PR #1, opened 2016-05-31
Apply the fixes that seemed to disappear from my local branch

Martin2112:
Mutter mutter, git .... bah humbug etc.


Martin2112:
TBR. Going to merge it because it doesn't compile without these. I thought they were in the import but Git disagreed.

PR #2, opened 2016-05-31
Add the storage mocks, which somehow also went missing.

Martin2112:


Martin2112:
TBR. Also merging because it's required for the tests to run. Sigh.

PR #4, opened 2016-06-02
Get rid of Gob encoding for SignedTimestamp

Martin2112:
It's a proto now so we don't need Gob encoder for it. Will sort out NodeID in a later PR.

### DIFF
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -137,14 +137,10 @@ func expandPlaceholderSql(sql string, num int) string {
 }
 
 func decodeSignedTimestamp(signedEntryTimestampBytes []byte) (trillian.SignedEntryTimestamp, error) {
-	// TODO: This will probably switch to proto serialization later
 	var signedEntryTimestamp trillian.SignedEntryTimestamp
-	buffer := bytes.NewBuffer(signedEntryTimestampBytes)
-	dec := gob.NewDecoder(buffer)
-	err := dec.Decode(&signedEntryTimestamp)
 
-	if err != nil {
-		glog.Warningf("Failed to encode payload: %s", err)
+	if err := proto.Unmarshal(signedEntryTimestampBytes, &signedEntryTimestamp); err != nil {
+		glog.Warningf("Failed to decode SignedTimestamp: %s", err)
 		return trillian.SignedEntryTimestamp{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


storage/mysql/tree_storage.go
```diff
@@ -137,33 +137,27 @@ func expandPlaceholderSql(sql string, num int) string {
 }
 
 func decodeSignedTimestamp(signedEntryTimestampBytes []byte) (trillian.SignedEntryTimestamp, error) {
-	// TODO: This will probably switch to proto serialization later
 	var signedEntryTimestamp trillian.SignedEntryTimestamp
-	buffer := bytes.NewBuffer(signedEntryTimestampBytes)
-	dec := gob.NewDecoder(buffer)
-	err := dec.Decode(&signedEntryTimestamp)
+
+	err := proto.Unmarshal(signedEntryTimestampBytes, &signedEntryTimestamp)
```
AlCutter:
```
if err := proto.blahblah; err != nil {
  // glog
  return ...
}
```

storage/mysql/tree_storage.go
```diff
@@ -137,33 +137,27 @@ func expandPlaceholderSql(sql string, num int) string {
...
 
 	if err != nil {
-		glog.Warningf("Failed to encode payload: %s", err)
+		glog.Warningf("Failed to decode payload: %s", err)
```
AlCutter: /payload/SignedTimestamp/ ?

storage/mysql/tree_storage.go
```diff
@@ -137,33 +137,27 @@ func expandPlaceholderSql(sql string, num int) string {
...
 func encodeSignedTimestamp(signedEntryTimestamp trillian.SignedEntryTimestamp) ([]byte, error) {
-	// TODO: This will probably switch to proto serialization later but we're avoiding those
-	// dependencies for the moment
-	var signedTimestampBuffer bytes.Buffer
-	enc := gob.NewEncoder(&signedTimestampBuffer)
-	err := enc.Encode(signedEntryTimestamp)
+	marshalled, err := proto.Marshal(&signedEntryTimestamp)
 
 	if err != nil {
 		glog.Warningf("Failed to encode payload: %s", err)
```
AlCutter: same /payload/SignedTimestamp/ sub here too?

storage/mysql/tree_storage.go

storage/mysql/tree_storage.go
Martin2112: Yeah it's because the original table column was payload, that still needs to be cleaned up. Done.

storage/mysql/tree_storage.go

PR #5, opened 2016-06-02
Get rid of gob serialization for node IDs and use protos

Martin2112:
I added a storage proto file as these are for internal use only.
Possibly could have used protos directly but as the IDs have associated
logic they'd probably have to be wrapped so this didn't seem too bad.


storage/mysql/tree_storage.go
```diff
@@ -166,31 +166,30 @@ func encodeSignedTimestamp(signedEntryTimestamp trillian.SignedEntryTimestamp) (
...
-func decodeNodeID(nodeIDBytes []byte, nodeID *storage.NodeID) error {
-	// TODO: This will probably switch to proto serialization later
-	buffer := bytes.NewBuffer(nodeIDBytes)
-	dec := gob.NewDecoder(buffer)
+// Node IDs are stored using proto serialization
+func decodeNodeID(nodeIDBytes []byte) (*storage.NodeID, error) {
+	var nodeIdProto storage.NodeIDProto
 
-	if err := dec.Decode(nodeID); err != nil {
+	err := proto.Unmarshal(nodeIDBytes, &nodeIdProto)
```
AlCutter:
```
if err := ...; err != nil {
...
}
```

PR #6, opened 2016-06-02
Fix TODO, inject hasher into sequencer

Martin2112:


log/sequencer.go
```diff
@@ -18,27 +18,26 @@ import (
 // There is no strong ordering guarantee but in general entries will be processed
 // in order of submission to the log.
 type Sequencer struct {
+	hasher trillian.Hasher
```
AlCutter: gofmt should line up the types for you I think

PR #8, opened 2016-06-02
Add a test that checks compact merkle tree state loading 

Martin2112:
Using data from the reference tree tests that correct nodes are accessed and resulting root hash is correct.

Not ready for review until branch it's based on is merged.

### DIFF
--- a/merkle/compact_merkle_tree_test.go
+++ b/merkle/compact_merkle_tree_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/google/trillian"
 )
 
-// This data came from the C++ CT tests
+// This data came from the C++ CT tests.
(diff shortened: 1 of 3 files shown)


merkle/compact_merkle_tree_test.go
```diff
@@ -3,11 +3,21 @@ package merkle
...
 	"encoding/hex"
+	"errors"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/google/trillian"
 )
 
+// This data came from the C++ CT tests
```
AlCutter: full stop :)

Martin2112: Done.

Martin2112:
I think this is now moot as the changes got mixed into a previous PR.

PR #14, opened 2016-06-13
Add Travis support

AlCutter:


Martin2112:
Might be best if you merge this and I think the followup I'm doing should fix the sql tests.

PR #23, opened 2016-06-15
Add outline log GRPC proto, service implementation and server skeleton

Martin2112:
It's still a bit WIP but I didn't think there was much point implementing operations until we decide how the pieces are going to be hooked up.

I'm starting to think that clients shouldn't use logids directly. We should give them an opaque token so they can't easily trash a different tree. Haven't implemented this yet though.


server/TrillianLogServer.go
```diff
@@ -0,0 +1,76 @@
+package server
+
+import (
+	"errors"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/storage"
+	"golang.org/x/net/context"
+	"sync"
```
AlCutter: this up to the previous block

server/TrillianLogServer.go
```diff
@@ -0,0 +1,76 @@
...
+	ErrNotImplemented = errors.New("Not yet implemented")
+)
+
+// LogStorageProviderFunc decouples the server from storage implementations
+type LogStorageProviderFunc func(int64) *storage.LogStorage
+
+// TrillianLogServer implements the RPC API defined in the proto
+type TrillianLogServer struct {
+	storageProvider LogStorageProviderFunc
+	// Must hold this lock before access to the storage map
```
AlCutter: /access to/accessing/?

server/TrillianLogServer.go
```diff
@@ -0,0 +1,76 @@
...
+	storageMapGuard sync.Mutex
+	// Map from storage ID to storage impl for that log
```
AlCutter: is `storage ID` TreeID elsewhere?

server/TrillianLogServer.go
```diff
@@ -0,0 +1,76 @@
...
+	storageMap map[int64]*storage.LogStorage
+}
+
+// NewTrillianLogServer creates a new RPC server backed by LogStorage
```
AlCutter: comment out of date wrt to LogStorage?

server/TrillianLogServer.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var (
+	// TODO: Delete when implementation done
+	// ErrNotImplemented is returned when an operation is not supported yet
+	ErrNotImplemented = errors.New("Not yet implemented")
+)
+
+// LogStorageProviderFunc decouples the server from storage implementations
+type LogStorageProviderFunc func(int64) *storage.LogStorage
```
AlCutter: Could the impl of this func reasonably fail and need to return an error (unknown ID, DB unavailable, etc.)?

storage/tools/common.go
```diff
@@ -6,18 +6,30 @@ import (
 	"github.com/google/trillian"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/mysql"
+	"github.com/google/trillian/server"
```
AlCutter: alpha ordering

storage/tools/common.go
```diff
@@ -6,18 +6,30 @@ import (
...
 )
 
 var logIdFlag = flag.String("logid", "logId", "The log id to use")
 var treeIdFlag = flag.Int64("treeid", 3, "The tree id to use")
 var storageTypeFlag = flag.String("storage_type", "mysql", "Which type of storage to use")
 var mysqlUriFlag = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test",
 	"uri to use with mysql storage")
+var logPortFlag = flag.Int("port", 8090, "Port to serve log requests on")
```
AlCutter: rename to `serverPort` or similar since this is a shared bunch of code?

storage/tools/common.go
```diff
@@ -6,18 +6,30 @@ import (
...
 
 func GetLogIdFromFlagsOrDie() trillian.LogID {
 	return trillian.LogID{[]byte(*logIdFlag), *treeIdFlag}
 }
 
+// TODO: This needs to be tidied up
```
AlCutter: Missing public function docs in here for most things

storage/tools/common.go
```diff
@@ -6,18 +6,30 @@ import (
...
+func GetLogStorageProviderFromFlags() server.LogStorageProviderFunc {
+	return func(x int64) *storage.LogStorage {
+		// TODO: We need to sort out exactly what a log id is. We really only need a tree id
+		// but having the log id as well could be a handy cross check as they can be guessed
+		storage := GetStorageFromFlagsOrDie(trillian.LogID{[]byte("This needs fixing"), x})
```
AlCutter: could just return it directly?

storage/tools/log_client/main.go
```diff
@@ -0,0 +1,42 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
```
AlCutter: Do you need this for the client?

server/TrillianLogServer.go
Martin2112: Done.

server/TrillianLogServer.go
Martin2112: Done but the other was also proper English like what she is written.

server/TrillianLogServer.go
Martin2112: Yes I meant treeID but developed a storage fixation.

server/TrillianLogServer.go
Martin2112: Done.

server/TrillianLogServer.go
Martin2112: Done.

storage/tools/common.go
Martin2112: Done.

storage/tools/common.go
Martin2112: Done. But I think this code won't be around too long.

storage/tools/common.go
Martin2112: Done.

storage/tools/common.go
Martin2112: Not sure if that works, was getting "x is pointer to interface, not interface" errors.

storage/tools/log_client/main.go
Martin2112: No. Removed.

storage/tools/common.go
```diff
@@ -39,12 +43,25 @@ func GetStorageFromFlagsOrDie(treeId trillian.LogID) storage.LogStorage {
...
 
-	panic(fmt.Sprintf("Unknown storage type: %s", *storageTypeFlag))
+	return nil, fmt.Errorf("Unknown storage type: %s", *storageTypeFlag)
 }
 
-func GetLogServingPort() int {
-	return *logPortFlag
-}
\ No newline at end of file
+// GetStorageFromFlags returns a configured storage instance, errors are fatal and if func init() {
```
AlCutter: docs vs. API name mismatch :)

AlCutter: "if func init() { returns" ?

Martin2112: Managed to trigger IDE template expansion without realizing it. Fixed.

storage/tools/common.go
```diff
@@ -39,12 +43,25 @@ func GetStorageFromFlagsOrDie(treeId trillian.LogID) storage.LogStorage {
...
+	storage, err := GetStorageFromFlags(treeId)
+
+	if err != nil {
+		panic(err)
+	}
+
+	return storage
+}
+
+// Gets the port number to be used when serving log data
```
AlCutter: nit: I _think_ golint will complain about this, the Hitch Hikers^W^W style guide has this to say on the subject:

> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

Martin2112: Done. Funny how Go ignores the existing conventions of every other language and then insists you follow theirs exactly. Grumble, mutter etc.

pphaneuf: Same convention in Python. :wink:

Martin2112: I rest my case ...

PR #24, opened 2016-06-15
Add a tree revision lookup API to NodeReader

Martin2112:
Add a tree revision lookup API to NodeReader + tests

### DIFF
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -34,7 +34,7 @@ const selectLatestSignedRootSql string = `SELECT TreeHeadTimestamp,TreeSize,Root
 		 ORDER BY TreeHeadTimestamp DESC LIMIT 1`
 const insertTreeHeadSql string = `INSERT INTO TreeHead(TreeId,TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature)
 		 VALUES(?,?,?,?,?,?)`
-const selectTreeRevisionAtSizeSql string = "SELECT TreeRevision FROM TreeHead WHERE TreeId=? AND TreeSize=?"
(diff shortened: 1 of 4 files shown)


storage/mysql/tree_storage.go
```diff
@@ -34,6 +34,7 @@ const selectLatestSignedRootSql string = `SELECT TreeHeadTimestamp,TreeSize,Root
 		 ORDER BY TreeHeadTimestamp DESC LIMIT 1`
 const insertTreeHeadSql string = `INSERT INTO TreeHead(TreeId,TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature)
 		 VALUES(?,?,?,?,?,?)`
+const selectTreeRevisionAtSizeSql string = "SELECT TreeRevision FROM TreeHead WHERE TreeId=? AND TreeSize=?"
```
AlCutter: There could be multiple result rows returned by this if nothing is added to the tree between multiple signings, I'd expect they should all have the same `TreeRevision` though.

storage/mysql/tree_storage.go
```diff
@@ -711,6 +712,20 @@ func (t *tx) removeSequencedLeaves(leaves []trillian.LogLeaf) error {
 	return nil
 }
 
+// TODO: This only works for sizes where there is a stored tree head. This is deliberate atm
+// as serving proofs at intermediate tree sizes is complicated and will be implemented later.
+func (t *tx) GetTreeRevisionAtSize(treeSize int64) (int64, error) {
```
AlCutter: API godoc missing on the new funcs in here

storage/mysql/tree_storage.go
Martin2112: Changed it to take the highest one, though they should be the same. Added a test for this.

storage/mysql/tree_storage.go
Martin2112: Cut and pasted, looks like an opportunity for it to get out of date though. Inheriting docs was probably another NIH thing.

PR #28, opened 2016-06-16
Fix up smt InclusionProof

AlCutter:


merkle/sparse_merkle_tree.go
```diff
@@ -86,28 +86,33 @@ func (s SparseMerkleTreeReader) InclusionProof(rev int64, key trillian.Key) ([]t
 	if err != nil {
 		return nil, err
 	}
+	// We're builing a full proof from a combination of whichever nodes we got
```
Martin2112: typo

merkle/sparse_merkle_tree_test.go
```diff
@@ -281,3 +285,82 @@ func TestRootAtRevisionCatchesNonRootNode(t *testing.T) {
 		t.Fatalf("Attempt to retrieve root with corrupt node did not result in error: %v", err)
 	}
 }
+
+func TestInclusionProofForNullEntryInEmptyTree(t *testing.T) {
+	const rev = 100
+	r, tx := getSparseMerkleTreeReaderWithMockTX(rev)
+	tx.On("GetMerkleNodes", int64(rev), mock.AnythingOfType("[]storage.NodeID")).Return([]storage.Node{}, nil)
```
Martin2112: I thought you could only use AnythingOfType in asserts. Are you sure the mocks are all wired up correctly?

merkle/sparse_merkle_tree.go
AlCutter: fidxe

merkle/sparse_merkle_tree_test.go
AlCutter: Pretty sure the tests would break if they weren't firing (and there'd be a panic from the mock for the  unexpected call)...

Martin2112: Odd. It didn't work for me and the doc says it's for use in Diff and Assert, but hey ho pip and dandy.

PR #29, opened 2016-06-16
Implement GetLeavesByIndex in server

Martin2112:
Also fix up some errors with pointer types, rename server go file to correct style, add tests and stuff, add options to client for what leaves get requested.


server/trillian_log_server.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (t *TrillianLogServer) GetLatestSignedLogRoot(ctx context.Context, req *trillian.GetLatestSignedLogRootRequest) (*trillian.GetLatestSignedLogRootResponse, error) {
+	return nil, ErrNotImplemented
+}
+
+func (t *TrillianLogServer) GetSequencedLeafCount(ctx context.Context, req *trillian.GetSequencedLeafCountRequest) (*trillian.GetSequencedLeafCountResponse, error) {
+	return nil, ErrNotImplemented
+}
+
+func (t *TrillianLogServer) GetLeavesByIndex(ctx context.Context, req *trillian.GetLeavesByIndexRequest) (*trillian.GetLeavesByIndexResponse, error) {
```
AlCutter: godoc missing on the exported funcs

server/trillian_log_server.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if !validateLeafIndices(req.LeafIndex) {
+		return &trillian.GetLeavesByIndexResponse{Status: buildStatusWithDesc(trillian.TrillianApiStatusCode_ERROR, "Invalid -ve leaf index in request")}, nil
+	}
+
+	// Read only transaction, failure to commit not propagated to client
```
AlCutter: Don't think you can do that - if the commit fails then the data read under the TX is invalid...

server/trillian_log_server.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (t *TrillianLogServer) GetLeavesByIndex(ctx context.Context, req *trillian.GetLeavesByIndexRequest) (*trillian.GetLeavesByIndexResponse, error) {
+	tx, err := t.prepareStorageTx(*req.LogId)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if !validateLeafIndices(req.LeafIndex) {
```
AlCutter: worth fetching the latest root and ensuring the req'd leaves are within the tree size?

server/trillian_log_server.go
Martin2112: That would appear to be a flaw in the cunning plan. I'll fix it in next push.

server/trillian_log_server.go
Martin2112: Wasn't sure about the overhead of that if someone is trying to read the whole log in batches. Not clear to me where this state should be kept and if it should be cached. Can I todo it for now?

server/trillian_log_server.go
Martin2112: Will fix.

server/trillian_log_server.go
```diff
@@ -75,26 +75,30 @@ func (t *TrillianLogServer) GetSequencedLeafCount(ctx context.Context, req *tril
...
 
 	if err != nil {
+		tx.Rollback()
 		return nil, err
 	}
 
 	leafProtos := leavesToProtos(leaves)
+
+	if err := t.commitAndLog(tx, "GetLeavesByIndex"); err != nil {
+		tx.Rollback()
```
AlCutter: Is it ok to call `Rollback()` if you already called `Commit()` and it failed?

Martin2112: It's doubtful but this is an application level tx object. There could be stuff that isn't cleaned up that we own, though there currently isn't.

Martin2112: Actually we could define the API so the TX impl must clean up if the commit fails then not worry about it. Is that better?

AlCutter: Maybe, dunno yet :)

Martin2112:
Implemented the change so it no longer calls rollback if commit fails.

PR #30, opened 2016-06-17
Implement QueueLeaves operation in server

Martin2112:
+ a bugfix + tests.

### DIFF
--- a/server/trillian_log_server.go
+++ b/server/trillian_log_server.go
@@ -56,7 +56,30 @@ func (t *TrillianLogServer) getStorageForLog(logId int64) (storage.LogStorage, e
 
 // QueueLeaves submits a batch of leaves to the log for later integration into the underlying tree.
 func (t *TrillianLogServer) QueueLeaves(ctx context.Context, req *trillian.QueueLeavesRequest) (*trillian.QueueLeavesResponse, error) {
-	return nil, ErrNotImplemented
+	leaves := protosToLeaves(req.Leaves)
+
+	if len(leaves) == 0 {
+		return &trillian.QueueLeavesResponse{Status: buildStatusWithDesc(trillian.TrillianApiStatusCode_ERROR, "Must queue at least one leaf")}, nil
+	}
+
+	tx, err := t.prepareStorageTx(*req.LogId)
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = tx.QueueLeaves(leaves)
+
+	if err != nil {
+		tx.Rollback()
+		return nil, err
+	}
+
+	if err := t.commitAndLog(tx, "QueueLeaves"); err != nil {
+		return nil, err
+	}
+
+	return &trillian.QueueLeavesResponse{Status: buildStatus(trillian.TrillianApiStatusCode_OK)}, nil
 }
 
 // GetInclusionProof obtains the proof of inclusion in the tree for a leaf that has been sequenced.
@@ -101,6 +124,7 @@ func (t *TrillianLogServer) GetLeavesByIndex(ctx context.Context, req *trillian.
 	leaves, err := tx.GetLeavesByIndex(req.LeafIndex)
 
 	if err != nil {
+		tx.Rollback()
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


server/trillian_log_server.go
```diff
@@ -56,7 +56,30 @@ func (t *TrillianLogServer) getStorageForLog(logId int64) (storage.LogStorage, e
...
+	tx, err := t.prepareStorageTx(*req.LogId)
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = tx.QueueLeaves(leaves)
+
+	if err != nil {
+		tx.Rollback()
```
AlCutter: An idea for removing the sprinkling of `Rollback()` calls (and avoid accidentally missing some out) might be to name the return values and then do something like this:
```go
func rollbackTXOnError(tx ..., err error) {
  if err != nil && tx != nil {
    tx.Rollback()
  }
}

func (t *TrillianLogServer) QueueLeaves(...) (resp *trilli..., err error) {
   ...
   tx, err := t.prepareStorageTx(*req.LogId)
   defer rollbackTXOnError(tx, err)
   ...
}
```

Then, if you call return anywhere in the function with a non-nil error, the tx will be rolled-back.

server/trillian_log_server.go
```diff
@@ -157,6 +181,20 @@ func (t *TrillianLogServer) commitAndLog(tx storage.LogTX, op string) error {
...
 }
 
+func protoToLeaf(proto *trillian.LeafProto) trillian.LogLeaf {
+	return trillian.LogLeaf{Leaf: trillian.Leaf{LeafHash: proto.LeafHash, LeafValue: proto.LeafData, ExtraData: proto.ExtraData}}
+}
+
+func protosToLeaves(protos []*trillian.LeafProto) []trillian.LogLeaf {
+	leaves := make([]trillian.LogLeaf, 0, len(protos))
+
+	for _, proto := range protos {
```
AlCutter: since you know the size of this list in advance, you could `make()` it at the right size (i.e. `make(..., len(protos), len(protos))`), and then use the index from the `range()` to just set the array element directly rather than appending.)

trillian_api.proto
```diff
@@ -33,6 +33,8 @@ message QueueLeavesRequest {
     repeated LeafProto leaves = 2;
 }
 
+// TODO: This will eventually contain the signed timestamps and stuff that we return for
```
AlCutter: Can you pop your name on the TODO?

server/trillian_log_server.go
Martin2112: I've allocated the capacity at the start so the appends won't do allocation. I think I prefer this to leaving holes in the result. I think our APIs are inconsistent on this so we should maybe decide.

trillian_api.proto
Martin2112: Done on both occurrences and rebuilt protos.

server/trillian_log_server.go
Martin2112: I'll have a look at it but I think there are cases where the storage tx succeeds but we return an error to the client for some other reason. Could get round this by having an application error base struct but not sure it's going to be cleaner.

AlCutter: No worries, your call.

server/trillian_log_server.go
AlCutter: I might be missing something, but I don't see how it could end up leaving holes in the result?
But anyway, yeah; I doubt there'd be that much of a difference in performance between the two since as you say you're preallocating, so feel free to leave it as is.

server/trillian_log_server.go
Martin2112: I think I'd like to defer this until the best approach is clearer. I don't like the explicit rollbacks either but the alternative needs thought.

server/trillian_log_server.go
Martin2112: In this case it doesn't matter as all elements are returned but if some were filtered out you'd have a slice with len(x) but the number of non-empty items in it would be x - y.

PR #31, opened 2016-06-17
Add description of Verifiable Logged Map to readme

heavenlyhash:
Make more explanation of the project available in the readme.

Content is ported almost unmodified from the PDF.

Focused mostly on the combined map structure.  I'm not sure if that's a valid elision, but my figuring was if brevity is important, it's the most interesting thing to highlight.


Martin2112:
Thanks for the PR. We do want to add more documentation but we also have a lot of code to write.

PR #33, opened 2016-06-21
Improve database tests, try to make them clean up locks

Martin2112:
Try to clean up transactions in database tests. And in some of the more complex cases make the test fail if the tx is left open. Add a bit more structure to some of the bigger tests.

### DIFF
--- a/storage/mysql/tree_storage_test.go
+++ b/storage/mysql/tree_storage_test.go
@@ -33,7 +33,8 @@ type logIDAndTest struct {
 }
 
 // Tests that access the db should each use a distinct log ID to prevent lock contention when
-// run in parallel or race conditions / unexpected interactions.
+// run in parallel or race conditions / unexpected interactions. Tests that pass should hold
+// no locks afterwards.
 
 var signedTimestamp = trillian.SignedEntryTimestamp{
 	TimestampNanos: proto.Int64(1234567890), LogId: createLogID("sign").logID.LogID, Signature: &trillian.DigitallySigned{Signature: []byte("notempty")}}
@@ -128,7 +129,7 @@ func TestOpenStateRollback(t *testing.T) {
 	}
 
 	assert.True(t, tx.Open(), "Transaction should be open on creation")
-	err = tx.Commit()
+	err = tx.Rollback()
 	assert.Nil(t, err, "Failed to commit: %v", err)
 	assert.False(t, tx.Open(), "Transaction should be closed after rollback")
 }
@@ -158,6 +159,8 @@ func TestNodeRoundTrip(t *testing.T) {
 
 	{
 		tx, err := s.Begin()
+		defer tx.Commit()
+		
 		if err != nil {
 			t.Fatalf("Failed to Begin: %s", err)
 		}
@@ -208,6 +211,7 @@ func TestQueueDuplicateLeafFails(t *testing.T) {
 	defer db.Close()
 	s := prepareTestStorage(logID, t)
 	tx := beginTx(s, t)
+	defer tx.Commit()
 
 	leaves := createTestLeaves(5, 10)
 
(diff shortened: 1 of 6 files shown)


Martin2112:
Motivation: We're going to need to get a list of logs to be sequenced (PR in progress). In tests this means putting the db in a clean state, which we can't do if locks are held.

storage/mysql/tree_storage_test.go
```diff
@@ -242,115 +283,142 @@ func TestDequeueLeaves(t *testing.T) {
...
-	tx := beginTx(s, t)
-
-	leaves := createTestLeaves(leavesToInsert, 20)
 
-	if err := tx.QueueLeaves(leaves); err != nil {
-		t.Fatalf("Failed to queue leaves: %v", err)
-	}
+	{
+		tx := beginTx(s, t)
+		defer failIfTXStillOpen(t, "TestDequeLeaves", tx)
```
AlCutter: Weird indentation here

storage/mysql/tree_storage_test.go
```diff
@@ -242,115 +283,142 @@ func TestDequeueLeaves(t *testing.T) {
...
 	leavesToDequeue1 := 3
 	leavesToDequeue2 := 2
-	leaves := createTestLeaves(leavesToInsert, 20)
 
-	if err := tx.QueueLeaves(leaves); err != nil {
-		t.Fatalf("Failed to queue leaves: %v", err)
-	}
+	{
+		tx := beginTx(s, t)
+		defer failIfTXStillOpen(t, "TestDequeueLeavesTwoBatches", tx)
```
AlCutter: indentation here too

storage/mysql/tree_storage_test.go
```diff
@@ -242,115 +283,142 @@ func TestDequeueLeaves(t *testing.T) {
...
 
-	ensureAllLeafHashesDistinct(leaves3, t)
+		if err != nil {
+			t.Fatalf("Failed to dequeue leaves: %v", err)
+		}
 
-	// Plus the union of the leaf batches should all have distinct hashes
-	leaves2 = append(leaves2, leaves3...)
-	ensureAllLeafHashesDistinct(leaves2, t)
+		if len(leaves3) != leavesToDequeue2 {
```
AlCutter: indentation

storage/mysql/tree_storage_test.go
```diff
@@ -485,15 +557,18 @@ func TestLatestSignedLogRoot(t *testing.T) {
...
-	root2, err := tx.LatestSignedLogRoot()
+	{
+		tx2 := beginTx(s, t)
+		defer tx2.Rollback()
+		root2, err := tx2.LatestSignedLogRoot()
 
-	if err != nil {
-		t.Fatalf("Failed to read back new log root: %v", err)
-	}
+		if err != nil {
```
AlCutter: indentation

storage/tree_storage.go
```diff
@@ -19,6 +19,11 @@ type TreeTX interface {
 
 	// Rollback aborts any performed operations. No updates must be applied to the underlying storage.
 	Rollback() error
+
+	// Open indicates if this transaction is open. An open transaction is one for which
+	// Commit() or Rollback() has never been called. Implementations must do all clean up
+	// in these methods so transactions are assumed closed regardless of the reported success.
+	Open() bool
```
AlCutter: Did you consider something like `IsOpen()`?  Could be confusing if it's taken as a verb.

storage/mysql/tree_storage_test.go
Martin2112: Strange. It looks OK in the editor and gofmt doesn't alter it. Same for other cases.

storage/tree_storage.go
Martin2112: I'd much prefer IsOpen but Get in Getters isn't idiomatic apparently. I'm going to make it IsOpen anyway.

PR #35, opened 2016-06-21
Factor out tree storage

AlCutter:
Builds on #34 .
There might be a better abstraction to be had in here, but I'm trying to keep the delta minimal until we have a more solid understanding of what the map storage needs to look like.


storage/mysql/tree_storage.go
```diff
@@ -60,22 +60,27 @@ const selectNodesSql string = `SELECT x.NodeId, x.MaxRevision, Node.NodeHash
...
-	allowDuplicates bool
-	hashSizeBytes   int
+// mySQLTreeStorage is shared between the mySQLLog- and (forthcoming) mySQLMap-
+// Storage implementations, and contains functionality which is common to both,
+type mySQLTreeStorage struct {
+	treeID        int64
+	db            *sql.DB
+	hashSizeBytes int
 
 	// Must hold the mutex before manipulating the statement map. Sharing a lock because
```
Martin2112: I assume you took the comments across as well? The diff looks weird like they're still in the original file.

Martin2112: Sorry I get it now I've had coffee. Forget that previous comment.

storage/mysql/tree_storage.go
```diff
@@ -183,91 +188,56 @@ func encodeNodeID(n storage.NodeID) ([]byte, error) {
 	return marshalledBytes, nil
 }
 
-func (m *mySQLLogStorage) getDeleteUnsequencedStmt(num int) (*sql.Stmt, error) {
+func (m *mySQLTreeStorage) getStmt(statement string, num int) (*sql.Stmt, error) {
```
Martin2112: We can possibly pull out all the statement building as a single unit. We'll need to make it more generic to support other databases anyway. No need to do now.

storage/mysql/tree_storage.go
```diff
@@ -183,91 +188,56 @@ func encodeNodeID(n storage.NodeID) ([]byte, error) {
...
 }
 
-func (m *mySQLLogStorage) GetMerkleNodes(treeRevision int64, nodeIDs []storage.NodeID) ([]storage.Node, error) {
-	t, err := m.Begin()
-
+func (m *mySQLTreeStorage) beginTreeTx() (treeTX, error) {
+	t, err := m.db.Begin()
 	if err != nil {
-		return nil, err
+		glog.Warningf("Could not start TX: %s", err)
```
Martin2112: mention tree tx in log message.

storage/mysql/tree_storage.go
AlCutter: Added a `TODO` to track that.

storage/mysql/tree_storage.go
```diff
@@ -183,91 +188,56 @@ func encodeNodeID(n storage.NodeID) ([]byte, error) {
...
-func (m *mySQLLogStorage) getDeleteUnsequencedStmt(num int) (*sql.Stmt, error) {
+func (m *mySQLTreeStorage) getStmt(statement string, num int) (*sql.Stmt, error) {
 	m.statementMutex.Lock()
 	defer m.statementMutex.Unlock()
 
-	if m.getDeleteUnsequenced[num] != nil {
-		return m.getDeleteUnsequenced[num], nil
+	if m.statements[statement] != nil {
+		if m.statements[statement][num] != nil {
+			return m.statements[statement][num], nil
```
AlCutter: Added `TODO` here too.

storage/mysql/tree_storage.go
AlCutter: Done.

PR #37, opened 2016-06-22
Add a Mysql MapStorage outline.

AlCutter:


storage/mysql/tree_storage.go
```diff
@@ -803,3 +874,19 @@ func (t *treeTX) Rollback() error {
 
 	return err
 }
+
+func (m *mapTX) Set(key []byte, value trillian.MapLeaf) error {
+	return errors.New("unimplemented")
```
Martin2112: Maybe we should have a set of common errors somewhere in the project.

PR #10, opened 2016-06-03
Implement the read only functionality for log storage

Martin2112:
And add some tests to show that it works

PR #40, opened 2016-06-22
Recreated Read Only Log PR against master

Martin2112:
It's #10 basically, without the merge conflicts. Will close that one.

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -13,7 +13,8 @@ import (
 	"github.com/google/trillian/storage"
 )
 
-const getLogPropertiesSql string = "SELECT AllowsDuplicateLeaves FROM Trees WHERE TreeId=?"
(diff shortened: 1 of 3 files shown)


storage/mysql/log_storage.go
```diff
@@ -13,7 +13,8 @@ import (
 	"github.com/google/trillian/storage"
 )
 
-const getLogPropertiesSql string = "SELECT AllowsDuplicateLeaves FROM Trees WHERE TreeId=?"
+const getTreePropertiesSql string = "SELECT AllowsDuplicateLeaves FROM Trees WHERE TreeId=?"
```
AlCutter: These three lines should probably be `...SQL` but I think there are a bunch of other consts like that, so maybe can fix those all together in a separate PR?

AlCutter:
Another, complimentary, idea might be to require the creation of a `...-readonly` DB user which does have any update/insert/delete rights granted, and have this code log in as that user in the read-only case.

Martin2112:
That's worth doing when we we finalize the schema. We should probably start opening issues to track these.

PR #41, opened 2016-06-22
Add a strawman Map API outline

AlCutter:


trillian_api.proto
```diff
@@ -113,3 +113,47 @@ service TrillianLog {
...
+}
+
+message GetMapLeavesResponse {
+  optional TrillianApiStatus status = 1;
+  repeated KeyValue key_value = 2;
+  optional SignedMapRoot map_root = 3;
+}
+
+message SetMapLeavesRequest {
+  repeated KeyValue key_value = 1;
```
Martin2112: Don't you need tree IDs?

AlCutter: Pah, IDs schmeIDs.

*cough*
Done.

PR #42, opened 2016-06-23
Implement get leaves by hash in RPC server

Martin2112:
And add some tests for the new code. Fix a previous typo in t.Fatalf as a free bonus.


Martin2112:
Why did travis not run on this? Too early in the morning?

PR #43, opened 2016-06-23
For sequencing add log api to get a list of configured logs

Martin2112:
Allow option to limit the results to logs that have unsequenced leaves to be processed to avoid then having to do status queries on each log.


storage/log_storage.go
```diff
@@ -82,3 +83,11 @@ type LogRootWriter interface {
 	// StoreSignedLogRoot stores a freshly created SignedLogRoot.
 	StoreSignedLogRoot(root trillian.SignedLogRoot) error
 }
+
+// LogMetadata provides access to information about the logs in storage
+type LogMetadata interface {
+	// GetActiveLogs returns the IDs of all the logs that are configured in storage
+	// if filterPendingWorkOnly is true then the list of logs returned includes only those
+	// with pending leaves that need to be integrated into the log.
+	GetActiveLogIDs(filterPendingWorkOnly bool) ([]trillian.LogID, error)
```
AlCutter: I wonder if this wouldn't just be better as two distinct API calls - they each seem to stand on their own as necessary functionality.

Martin2112: Can do.

Martin2112:
Done.

storage/mysql/log_storage.go
```diff
@@ -558,11 +558,12 @@ func (t* logTX) getActiveLogIDsInternal(sql string) ([]trillian.LogID, error) {
...
-	if (filterPendingWorkOnly) {
-		return t.getActiveLogIDsInternal(selectActiveLogsWithUnsequencedSql)
-	}
-
+// GetActiveLogIDs returns a list of the IDs of all configured logs
+func (t *logTX) GetActiveLogIDs() ([]trillian.LogID, error) {
 	return t.getActiveLogIDsInternal(selectActiveLogsSql)
 }
+
+// GetActiveLogIDsWithPendingWork returns a list of the IDs of all configured logs
```
AlCutter: "... with pending work"?

storage/mysql/storage_test.go
```diff
@@ -819,7 +819,7 @@ func TestGetActiveLogIDsFiltered(t *testing.T) {
 	s := prepareTestStorage(logID, t)
 	tx := beginTx(s, t)
 
-	logIDs, err := tx.GetActiveLogIDs(true)
+	logIDs, err := tx.GetActiveLogIDsWithPendingWork()
```
AlCutter: Probably worth renaming these tests too?

storage/mysql/log_storage.go
Martin2112: Ooops following line went missing. Reinstated.

storage/mysql/storage_test.go
Martin2112: Done, and fixed incorrect test setup in both.

PR #44, opened 2016-06-23
Implement map head storage

AlCutter:


storage/mysql/storage_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var allTables = []string{"Unsequenced", "TreeHead", "SequencedLeafData", "LeafData", "Node", "TreeControl", "Trees"}
+var allTables = []string{"Unsequenced", "TreeHead", "SequencedLeafData", "LeafData", "Node", "TreeControl", "Trees", "MapLeaf", "MapHead"}
```
Martin2112: Possibly a constraint error in this order but the CASCADE might take care of it.

PR #47, opened 2016-06-27
Move rpc server to log package

Martin2112:
No other changes.


Martin2112:
OK. I'll do this later when this packaging error won't happen.

PR #48, opened 2016-06-28
Add log server inc. sequencing and RPC server.

Martin2112:
Plus tests and sequencing bug fixes where struct members were not being correctly set. Update affected tests.

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -116,22 +116,24 @@ func (s Sequencer) SequenceBatch(limit int) (int, error) {
 		return 0, err
 	}
 
-	var mt *merkle.CompactMerkleTree
+	var merkleTree *merkle.CompactMerkleTree
 
 	if currentRoot.TreeSize == 0 {
 		// This should be an empty tree
 		if currentRoot.TreeRevision > 0 {
-			panic(fmt.Errorf("MT has zero size but non zero revision: %v", *mt))
+			// TODO(Martin2112): Remove panic and return error when we implement proper multi
+			// tenant scheduling.
+			panic(fmt.Errorf("MT has zero size but non zero revision: %v", *merkleTree))
 		}
-		mt, err = merkle.NewCompactMerkleTree(s.hasher), nil
+		merkleTree = merkle.NewCompactMerkleTree(s.hasher)
 	} else {
 		// Initialize the compact tree state to match the latest root in the database
-		mt, err = s.buildMerkleTreeFromStorageAtRoot(currentRoot, tx)
-	}
+		merkleTree, err = s.buildMerkleTreeFromStorageAtRoot(currentRoot, tx)
 
-	if err != nil {
-		tx.Rollback()
-		return 0, err
+		if err != nil {
+			tx.Rollback()
+			return 0, err
+		}
 	}
 
 	// We've done all the reads, can now do the updates.
@@ -141,7 +143,7 @@ func (s Sequencer) SequenceBatch(limit int) (int, error) {
 	newVersion := currentRoot.TreeRevision + int64(1)
 
 	// Assign leaf sequence numbers and collate node updates
-	nodeMap, sequenceNumbers := s.sequenceLeaves(mt, leaves)
+	nodeMap, sequenceNumbers := s.sequenceLeaves(merkleTree, leaves)
 
 	if len(sequenceNumbers) != len(leaves) {
 		panic(fmt.Sprintf("Sequencer returned %d sequence numbers for %d leaves", len(sequenceNumbers),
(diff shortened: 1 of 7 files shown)


log/sequencer.go
```diff
@@ -112,22 +113,32 @@ func (s Sequencer) SequenceBatch(limit int) error {
...
 	}
 
-	// Initialize the compact tree state to match the latest root in the database
-	mt, err := s.buildMerkleTreeFromStorageAtRoot(currentRoot, tx)
+	var mt *merkle.CompactMerkleTree
+
+	if currentRoot.TreeSize == 0 {
+		// This should be an empty tree
+		if currentRoot.TreeRevision > 0 {
+			panic(fmt.Errorf("MT has zero size but non zero revision: %v", *mt))
```
AlCutter: Should probably just return an error here - one day this'll get rolled into a multi-tenant sequencer and one corrupt log will stop progress on all the others.

AlCutter: /MT/merkle tree/ for clarity

log/sequencer.go
```diff
@@ -112,22 +113,32 @@ func (s Sequencer) SequenceBatch(limit int) error {
...
+		}
+		mt, err = merkle.NewCompactMerkleTree(s.hasher), nil
+	} else {
+		// Initialize the compact tree state to match the latest root in the database
+		mt, err = s.buildMerkleTreeFromStorageAtRoot(currentRoot, tx)
+	}
 
 	if err != nil {
```
AlCutter: This error catching is really only for the return of `buildMerkleTreeFromStorageAtRoot()` right? maybe move it into that `else{}` block?

server/log/main.go
```diff
@@ -0,0 +1,123 @@
...
+	glog.Infof("Creating RPC server starting on port: %d", *serverPortFlag)
+	// TODO(Martin2112): More flexible listen address configuration
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *serverPortFlag))
+
+	if err != nil {
+		glog.Errorf("Failed to listen on the server port: %d, because: %v", *serverPortFlag, err)
+		os.Exit(1)
+	}
+
+	// Start the sequencing loop which, will run until we terminate the process
```
AlCutter: spurious comma

server/sequencer_manager.go
```diff
@@ -0,0 +1,139 @@
...
+		default:
+		}
+
+		// TODO(Martin2112): Probably want to make the sequencer objects longer lived to
+		// avoid the cost of initializing their state each time but this works for now
+		storage, err := s.storageProvider(logID.TreeID)
+
+		// TODO(Martin2112): Honour the sequencing enabled in log parameters, needs an API change
+		// so deferring it
+		if err == nil {
```
AlCutter: indent error flow rather than business-as-usual.
(That might mean either having a separate `time.Sleep` and a `continue`, but that's not necessarily a bad thing - you may want to have a back-off strategy here for failures at some point?)

server/sequencer_manager.go
```diff
@@ -0,0 +1,139 @@
...
+	glog.Infof("Log sequencer starting")
+
+	// Outer sequencing loop, runs until terminated
+	for {
+		// TODO(Martin2112) using log ID zero because we don't have an id for metadata ops
+		// this API could improved
+		provider, err := s.storageProvider(0)
+
+		// If we get an error, we can't do anything but wait until the next run through
+		if err == nil {
```
AlCutter: re-jiggle to indent error flow instead

storage/mock_storage.go
```diff
@@ -159,6 +161,7 @@ func (t *MockLogTX) LatestSignedLogRoot() (trillian.SignedLogRoot, error) {
 
 // StoreSignedLogRoot is a mock
 func (t *MockLogTX) StoreSignedLogRoot(root trillian.SignedLogRoot) error {
+	fmt.Printf("**** %s\n", proto.CompactTextString(&root))
```
AlCutter: want this in there still?

log/sequencer.go
Martin2112: Yeah for the moment I wanted to halt the server because this is a "can't happen" case. I'll todo it to downgrade.

log/sequencer.go
Martin2112: That's somehow ended up weird, moved it around.

server/log/main.go

server/sequencer_manager.go
Martin2112: Done. Haven't used this sort of style since I wrote assembler code.

storage/mock_storage.go
Martin2112: Dinged.

PR #51, opened 2016-06-30
Proto3

AlCutter:


Martin2112:
LGTM, with a due sense of ill-defined unease.

PR #54, opened 2016-06-30
Again, with the proto generation.

AlCutter:
Previous attempts didn't work when there were cross-package proto imports.


Martin2112:
Turn it off and on again?

PR #55, opened 2016-07-01
Move HStar2 impl out to its own file.

AlCutter:
Dinged the `hi, lo` stuff too since we can represent that with the `[]values` slice directly.


merkle/hstar2.go
```diff
@@ -0,0 +1,124 @@
...
+			return nil, err
+		}
+		h := s.hasher.HashChildren(emptyRoot, emptyRoot)
+		if len(s.hStarEmptyCache) != n {
+			return nil, fmt.Errorf("cache wrong size - expected %d, but cache contains %d entries", n, len(s.hStarEmptyCache))
+		}
+		s.hStarEmptyCache = append(s.hStarEmptyCache, h)
+	}
+	if n >= len(s.hStarEmptyCache) {
+		return nil, fmt.Errorf("cache wrong size - expected %d or more, but cache contains %d entries", n, len(s.hStarEmptyCache))
```
Martin2112: I might be misreading it but the error message and if condition don't seem to match up.

merkle/hstar2_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"github.com/google/trillian"
+)
+
+// This root was calculated with the C++/Python sparse merkle tree code in the
+// github.com/google/certificate-transparency repo.
+const sparseEmptyRootHashB64 = "xmifEIEqCYCXbZUz2Dh1KCFmFZVn7DUVVxbBQTr1PWo="
+
+func mustDecode(b64 string) trillian.Hash {
```
Martin2112: Maybe TODO to collect these test helpers together.

merkle/hstar2.go
AlCutter: I think it was technically correct, but not very clearly put - I've changed it to hopefully be a bit clearer.

merkle/hstar2_test.go
AlCutter: Done.

PR #59, opened 2016-07-04
Inject a TreeHasher rather than just a Hasher

Martin2112:
Addresses TODO. Moves dependency on RFC6962 hasher further up in the
system.

### DIFF
--- a/log/sequencer_test.go
+++ b/log/sequencer_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/trillian/util"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/assert"
+	"github.com/google/trillian/merkle"
 )
 
 // These can be shared between tests as they're never modified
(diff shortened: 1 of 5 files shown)


log/sequencer_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/google/trillian/util"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/assert"
+	"github.com/google/trillian/merkle"
```
AlCutter: import ordering is wonky

server/sequencer_manager.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/google/trillian"
 	"github.com/google/trillian/log"
 	"github.com/google/trillian/util"
+	"github.com/google/trillian/merkle"
```
AlCutter: import ordering

Martin2112: Done.

log/sequencer_test.go
Martin2112: Done.

PR #60, opened 2016-07-05
Add signer. Make sequencer sign the roots it creates too.

Martin2112:
Add tests for signer changes and sequencing. The struct embedded fake inheritance stuff isn't right but it's already broken and I'll fix it in a followup. This PR is too big already.


log/sequencer.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/util"
+	"github.com/google/trillian/crypto"
```
AlCutter: import ordering

log/sequencer.go
```diff
@@ -82,6 +84,41 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []trillia
...
+	signer, err := s.keyManager.Signer()
+
+	if err != nil {
+		glog.Warningf("key manager failed to create crypto.Signer: %v", err)
+		return trillian.DigitallySigned{}, err
+	}
+
+	// TODO(Martin2112): Signature algorithm shouldn't be fixed here
+	trillianSigner := crypto.NewTrillianSigner(s.hasher, trillian.SignatureAlgorithm_ECDSA, signer)
+	signature, err := trillianSigner.Sign(merkleTree.CurrentRoot())
```
Martin2112: I'm bringing in object hash, in a separate PR then will fix this up to use it.

Martin2112:
I think this is ready for review again. Sorry it got so big but I thought it was better to review the two merged branches than the intermediate state that didn't work properly.

crypto/trillian_signer_test.go
```diff
@@ -16,17 +16,19 @@ import (
...
+func (m mockTrillianSigner) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) ([]byte, error) {
 	args := m.Called(rand, msg, opts)
 
 	return args.Get(0).([]byte), args.Error(1)
 }
 
-func (m mockSigner) Public() crypto.PublicKey {
+// Public is a mock
+func (m mockTrillianSigner) Public() crypto.PublicKey {
 	args := m.Called("Public")
```
AlCutter: `m.Called()`? (elsewhere too)

log/sequencer.go
```diff
@@ -82,9 +84,47 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []trillia
...
+	var merkleTree *merkle.CompactMerkleTree
+
+	if currentRoot.TreeSize == 0 {
+		// This should be an empty tree
+		if currentRoot.TreeRevision > 0 {
+			return nil, fmt.Errorf("MT has zero size but non zero revision: %v", *merkleTree)
+		}
+		return merkle.NewCompactMerkleTree(s.hasher), nil
+	} else {
+		// Initialize the compact tree state to match the latest root in the database
```
AlCutter: I guess you don't strictly need the `else{}` block here

log/sequencer.go
```diff
@@ -82,9 +84,47 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []trillia
...
+		return s.buildMerkleTreeFromStorageAtRoot(currentRoot, tx)
+	}
+}
+
+func (s Sequencer) signRoot(root trillian.SignedLogRoot) (trillian.DigitallySigned, error) {
```
AlCutter: Was there a reason not to take a `*trillian.SignedLogRoot` here, and set its signature member directly (iff the signing works as expected)?

log/sequencer.go
```diff
@@ -210,3 +246,60 @@ func (s Sequencer) SequenceBatch(limit int) (int, error) {
...
+		return err
+	}
+
+	// Build the updated root, ready for signing
+	newLogRoot := trillian.SignedLogRoot{
+		RootHash:       merkleTree.CurrentRoot(),
+		TimestampNanos: s.timeSource.Now().UnixNano(),
+		TreeSize:       merkleTree.Size(),
+		LogId:          currentRoot.LogId,
+		TreeRevision:   currentRoot.TreeRevision + 1,
```
AlCutter: I think it'd make sense for the sequencer process to store the signed tree root, rather than having a separate signer process - having the two things do similar work and having both re-calculate the `TreeRevision` scares me a bit.

server/log/main.go
```diff
@@ -55,6 +64,27 @@ func checkDatabaseAccessible(dbUri string) error {
 	return err
 }
 
+func loadPrivateKey(keyFile, keyPassword string) (crypto.KeyManager, error) {
```
AlCutter: This func could go into `crypto/`, might be useful elsewhere?

server/log/main.go
```diff
@@ -55,6 +64,27 @@ func checkDatabaseAccessible(dbUri string) error {
...
+	if len(keyFile) == 0 || len(keyPassword) == 0 {
```
AlCutter: what if someone has an unencrypted key? :)

server/log/main.go
```diff
@@ -55,6 +64,27 @@ func checkDatabaseAccessible(dbUri string) error {
...
+		return nil, errors.New("private key file and password must be specified")
+	}
+
+	pemData, err := ioutil.ReadFile(keyFile)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to read data from key file: %s", keyFile)
```
AlCutter: include the err in the string so they can figure out why it failed?

server/log/main.go
```diff
@@ -89,6 +119,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Load up our private key, exit if this fails to work
+	keyManager, err := loadPrivateKey(*privateKeyFile, *privateKeyPassword)
+
+	if err != nil {
+		glog.Errorf("Failed to load server key: %v", err)
+		os.Exit(1)
```
AlCutter: add a TODO to remove this during the multi-tenant work?
Also, is there not a `glog.Fatalf()` ?

server/log/main.go
```diff
@@ -101,9 +139,17 @@ func main() {
 
 	// Start the sequencing loop, which will run until we terminate the process
 	// TODO(Martin2112): Should respect read only mode and the flags in tree control etc
-	sequencerManager := server.NewSequencerManager(done, simpleMySqlStorageProvider, *batchSizeFlag, *sleepBetweenLogsFlag, *sleepBetweenRunsFlag)
+	// TODO(Martin2112): Plug in Key manager and load key, this is in another branch atm
+	// this is OK as we haven't added code to create a signer task yet
+	sequencerManager := server.NewLogOperationManager(done, simpleMySqlStorageProvider, *batchSizeFlag, *sequencerSleepBetweenRunsFlag, util.SystemTimeSource{}, server.NewSequencerManager(keyManager))
 	go sequencerManager.OperationLoop()
 
+	// Start the signer loop, which will run continuously
```
AlCutter: as we just talked about - I think the sequencer should created signed log roots, and then perhaps the *application* "signer" would choose which of those to re-format/re-sign and publish (e.g. a CT STH might be limited to one per hour, regardless of how many the trillian sequencer generates.)

server/log_operation_manager.go
```diff
@@ -8,87 +8,106 @@ import (
...
+// consistency checking or cleanup.
+type LogOperation interface {
+	Name() string
+	ExecutePass([]trillian.LogID, LogOperationManagerContext) bool
+}
+
+// LogOperationManagerContext bundles up the values so testing can be made easier
+type LogOperationManagerContext struct {
 	// done is a channel that provides an exit signal
 	done chan struct{}
```
AlCutter: Would https://godoc.org/golang.org/x/net/context be a good fit here?

server/log_operation_manager.go
```diff
@@ -8,87 +8,106 @@ import (
...
-	runLimit int
+	// oneShot is for use by tests only, it exits after one pass
+	oneShot bool
 	// timeSource allows us to mock this in tests
 	timeSource util.TimeSource
 }
 
-func (l logOperationManager) runOperationPass([]trillian.LogID) bool {
-	return false
+// logOperationManager controls scheduling activities for logs. At the moment it's very simple
```
AlCutter: s/logOperationManager/LogOperationManager/

server/log_operation_manager.go
```diff
@@ -8,87 +8,106 @@ import (
...
+		return false
+	}
 
-		if err != nil {
-			glog.Warningf("Failed to get log list for run: %v", err)
-			tx.Rollback()
-			continue
-		}
+	if err := tx.Commit(); err != nil {
+		glog.Warningf("Failed to commit getting logs, continuing anyway: %v", err)
```
AlCutter: seems to return false rather than continue?

crypto/trillian_signer_test.go
Martin2112: Done. Not sure what you mean by elsewhere? This is the only mock in this file.

log/sequencer.go
Martin2112: Sorry, I keep forgetting that this is Go, not some kind of structured programming language.

log/sequencer.go
Martin2112: IIRC It seemed to work out slightly neater this way but there wasn't much in it. It also means the signer can't modify the root in unexpected ways.

log/sequencer.go
Martin2112: Yes, this is very much V1 code. They can't both commit the same tree version so it should be safe until it gets redone.

server/log/main.go

server/log/main.go
Martin2112: They can load it with pem.Decode() or do we really want to allow that case here?

server/log/main.go

server/log/main.go
Martin2112: TODO added. Wasn't sure of exit code of glog.Fatalf but it does appear to exit(1).

server/log/main.go
Martin2112: Yes, this isn't the final version of this code. I can't do everything at once :-)

server/log_operation_manager.go
Martin2112: I hoped so but it's really for request scoped things and it seemed to be a bit of a hack for long lived operations.

server/log_operation_manager.go
Martin2112: Done.

server/log_operation_manager.go
Martin2112: This was inside a loop before it got refactored. Returning false does mean continue to run subsequent passes but it's not clear now so I've removed it.

log/sequencer.go
AlCutter: ok - can you add a big firey TODO? :)

PR #65, opened 2016-07-08
Summon the awesome power of objecthash to the project

Martin2112:
Add code to signer to serialize and sign log roots and update tests.


crypto/signer.go
```diff
@@ -43,3 +56,42 @@ func (s Signer) Sign(data []byte) (trillian.DigitallySigned, error) {
 		HashAlgorithm:      s.hasher.HashAlgorithm(),
 		Signature:          sig}, nil
 }
+
+// LogRootToJson returns the JSON representation of a signed log root that should be fed to
+// objecthash when verifying a signed root.
+// TODO(Martin2112): Decouple code to verify roots from Signer as you need a private
+// key to create these objects.
+func (s Signer) LogRootToJson(root trillian.SignedLogRoot) (string, error) {
```
AlCutter: Do you need to do this round-trip via JSON?
If this returned a `map[string]interface{}` with appropriate keys set (which is what I think you'll get with the JSON round-trip) then you could call `objecthash.ObjectHash()` directly.

Martin2112:
Possibly not. We talked about going via json at the meeting on Wed. I'll have another look at it on Monday.

Martin2112:
OK. I've ripped out the JSON roundtrip and now serialize a map directly.

crypto/signer.go
```diff
@@ -4,12 +4,12 @@ import (
 	"crypto"
 	"crypto/rand"
 	"fmt"
+	"encoding/base64"
```
AlCutter: ordering :)

crypto/signer.go
```diff
@@ -3,9 +3,13 @@ package crypto
 import (
 	"crypto"
 	"crypto/rand"
+	"encoding/base64"
 	"fmt"
+	"strconv"
 
+	"github.com/benlaurie/objecthash/go/objecthash"
 	"github.com/google/trillian"
+	"github.com/golang/glog"
```
AlCutter: ordering

crypto/signer.go
```diff
@@ -43,3 +47,32 @@ func (s Signer) Sign(data []byte) (trillian.DigitallySigned, error) {
 		HashAlgorithm:      s.hasher.HashAlgorithm(),
 		Signature:          sig}, nil
 }
+
+func (s Signer) hashRoot(root trillian.SignedLogRoot) []byte {
+	rootMap := make(map[string]interface{})
+
+	// Pull out the fields we want to hash. Caution: use string format for
+	// int64 values as they can overflow otherwise
```
AlCutter: maybe mention the JSON reasoning behind this?

crypto/signer.go
```diff
@@ -43,3 +47,32 @@ func (s Signer) Sign(data []byte) (trillian.DigitallySigned, error) {
...
+	rootMap["RootHash"] = base64.StdEncoding.EncodeToString(root.RootHash)
```
AlCutter: worth pulling these keys out as public constants somewhere?

crypto/signer.go
```diff
@@ -43,3 +47,32 @@ func (s Signer) Sign(data []byte) (trillian.DigitallySigned, error) {
...
+func (s Signer) SignLogRoot(root *trillian.SignedLogRoot) error {
+	objectHash := s.hashRoot(*root)
+	signature, err := s.Sign(objectHash[:])
+
+	if err != nil {
+		glog.Warningf("Signer failed to sign root: %v", err)
+		return err
+	}
+
+	(*root).Signature = &signature
```
AlCutter: `root.Signature = ...` should work

crypto/signer.go
Martin2112: Done.

crypto/signer.go
Martin2112: Done.

crypto/signer.go
Martin2112: Done.

crypto/signer.go
Martin2112: Done.

PR #66, opened 2016-07-12
Add a wrapped CertPool with extra features needed for get-roots

Martin2112:
plus tests plus add a copy of ca-certs to testdata.


crypto/pem_cert_pool.go
```diff
@@ -0,0 +1,70 @@
+package crypto
+
+import (
+	"encoding/pem"
+
+	"github.com/golang/glog"
+	"github.com/google/certificate-transparency/go/x509"
+	"crypto/sha1"
```
AlCutter: ordering

crypto/pem_cert_pool.go
```diff
@@ -0,0 +1,70 @@
+package crypto
+
+import (
+	"encoding/pem"
+
+	"github.com/golang/glog"
+	"github.com/google/certificate-transparency/go/x509"
```
AlCutter: I'm wondering if we should just use the golang `x509` package for now?

crypto/pem_cert_pool.go
```diff
@@ -0,0 +1,70 @@
...
+	"crypto/sha1"
+)
+
+// PEMCertPool is a wrapper / extension to x509.CertPool. It allows us to access the
+// raw certs, which we need to serve get-roots request and has stricter handling on loading
+// certs into the pool. CertPool ignores errors if at least one cert loads correctly but
+// PEMCertPool requires all certs to load.
+type PEMCertPool struct {
+	// maps from sha-1 to certificate, used for dup detection
+	fingerprintToCertMap map[[sha1.Size]byte]x509.Certificate
```
AlCutter: Why SHA-1, out of interest? are you worried about space?

AlCutter:
Wondering whether this stuff should live under the `ct_fe` dir, it's not really Trillian stuff per-se, right?
(perhaps the `ct_fe` should move under an `examples` package too at some point?)

crypto/pem_cert_pool.go
Martin2112: Done.

crypto/pem_cert_pool.go
Martin2112: Need to decide then as I've got several PRs that are pulling in this and the other CT specific packages.

crypto/pem_cert_pool.go
Martin2112: No real reason, just deduping certs. I've switched it to sha256.

Martin2112:
I left this in crypto as there's nothing CT specific about it. If the only user turns out to be the CT frontend then we can move it over.

crypto/pem_cert_pool.go
Martin2112: BTW not actually sure we can use the golang x509 as one of the changes was to make TBSCertificate visible.

AlCutter: ok, but *that* definitely is CT specific though, right?

Martin2112: Yeah it's a fair cop. But you can't mix the CT and golang x509 / asn1 / pkix packages easily. I don't want to move stuff until I've landed the next few PRs that modify the same code.

AlCutter:
I kinda assumed that clients of trillian will live in their own separate repos and the `ct_fe` stuff is really just an example of how to build something which uses trillian under the covers (hence wondering whether it might be better under `examples/ct_fe` or perhaps `examples/ct/frontend`), so I'd expect that this would only really get used, in this repo, for CT (it's not CT specific, but it is certificate specific...)

I don't feel super strongly about it, more just thinking aloud.  If you have a stronger preference to leave it here for now, then I'm cool with that, but maybe pop a TODO in saying the same?

Martin2112:
OK I'd kind of assumed we'd ship at least CT frontend with it as a working example. I don't really mind but I'd like to move it later to avoid hassle. I'll todo it.

crypto/pem_cert_pool.go
```diff
@@ -0,0 +1,70 @@
...
+// Skips over non certificate blocks in the data. Returns true if all certificates in the
+// data were parsed and added to the pool successfully and at least one certificate was found.
+func (p *PEMCertPool) AppendCertsFromPEM(pemCerts []byte) (ok bool) {
+	for len(pemCerts) > 0 {
+		var block *pem.Block
+		block, pemCerts = pem.Decode(pemCerts)
+		if block == nil {
+			break
+		}
+		if block.Type != "CERTIFICATE" || len(block.Headers) != 0 {
```
AlCutter: could make `"CERTIFICATE"` a constant?

Martin2112: Done.

PR #67, opened 2016-07-13
Add testonly package for code only for use by tests

Martin2112:
Refactor to move a couple of decoding helpers into testonly getting rid of dups.

### DIFF
--- a/testonly/doc.go
+++ b/testonly/doc.go
@@ -1,0 +1,9 @@
+/*
+Package testonly contains code and data that should only be used by tests.
+Production code MUST NOT depend on anything in this package. This will be enforced
+by tools where possible.
+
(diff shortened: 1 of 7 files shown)


testonly/doc.go
```diff
@@ -0,0 +1,9 @@
+/*
+Package testonly contains code and data that should only be used by tests.
+Production code MUST NOT depend on anything in this package. This will be enforced
+by tools where possible.
+
+As an example PEM encoded test certificates and helper functions to decode them are
```
AlCutter: Maybe... :)

AlCutter:
LGTM modulo wondering about PEM bits

Martin2112:
The alternative is to keep them in files and try to access them via relative paths, which I think will be more messy.

PR #68, opened 2016-07-13
Make NodeID store path data as BigEndian

AlCutter:


storage/types.go
```diff
@@ -83,7 +83,7 @@ func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits
 // index
 func NewNodeIDForTreeCoords(depth int64, index int64, maxLenBits int) NodeID {
 	r := NewEmptyNodeID(maxLenBits)
-	for i := 0; index > 0; i++ {
+	for i := len(r.Path) - 1; index > 0; i-- {
```
Martin2112: Mildly suspicious that you reversed this but the test is still index > 0

storage/types.go
```diff
@@ -93,16 +93,18 @@ func NewNodeIDForTreeCoords(depth int64, index int64, maxLenBits int) NodeID {
 
 // SetBit sets the ith bit to true if b is non-zero, and false otherwise.
 func (n *NodeID) SetBit(i int, b uint) {
+	bIndex := (n.PathLenBits - i - 1) / 8
 	if b == 0 {
-		n.Path[i/8] &= ^(1 << uint(i%8))
+		n.Path[bIndex] &= ^(1 << uint(i%8))
```
Martin2112: worth having lookup tables for these common bit shifts?

storage/types.go
```diff
@@ -93,16 +93,18 @@ func NewNodeIDForTreeCoords(depth int64, index int64, maxLenBits int) NodeID {
...
 	} else {
-		n.Path[i/8] |= (1 << uint(i%8))
+		n.Path[bIndex] |= (1 << uint(i%8))
 	}
 }
 
 // Bit returns 1 if the ith bit is true, and false otherwise.
 func (n *NodeID) Bit(i int) uint {
-	return uint((n.Path[i/8] >> uint(i%8)) & 0x01)
+	bIndex := (n.PathLenBits - i - 1) / 8
```
Martin2112: Doc comment should probably mention the endianicity, endianity ... which way round it is,

storage/types_test.go
```diff
@@ -14,12 +14,12 @@ func TestZerosNewNodeIDWithPrefix(t *testing.T) {
...
 func TestNewNodeIDWithPrefix(t *testing.T) {
 	n := NewNodeIDWithPrefix(0x12345678, 32, 32, 64)
-	if got, want := n.Path, []byte{0x00, 0x00, 0x00, 0x00, 0x78, 0x56, 0x34, 0x12}; !bytes.Equal(got, want) {
+	if got, want := n.Path, []byte{0x12, 0x34, 0x56, 0x78, 0x00, 0x00, 0x00, 0x00}; !bytes.Equal(got, want) {
 		t.Fatalf("Expected Path of %v, but got %v", want, got)
 	}
 
 	n = NewNodeIDWithPrefix(0x345678, 24, 24, 24)
-	if got, want := n.Path, []byte{0x78, 0x56, 0x34}; !bytes.Equal(got, want) {
+	if got, want := n.Path, []byte{0x34, 0x56, 0x78}; !bytes.Equal(got, want) {
```
Martin2112: Does this need tests not a multiple of 8 to prove it works?

storage/types.go
AlCutter: that's just because `index` is modified in the body of the loop, and that hasn't changed.

storage/types.go
AlCutter: Possibly - I'll add a `TODO` to investigate that later on.

storage/types.go
AlCutter: Added.

storage/types_test.go
AlCutter: Changed one test to use a non-multiple prefix length.

Martin2112:
LGTM (subject to Travis)

PR #69, opened 2016-07-13
Hook up SMT writer with storage "naively"

AlCutter:


merkle/hstar2.go
```diff
@@ -109,24 +109,28 @@ func (s *HStar2) hStarEmpty(n int) (trillian.Hash, error) {
 	return s.hStarEmptyCache[n], nil
 }
 
+var (
```
AlCutter: Yeah, although there'd need to be a set of consts for each tree-hashing scheme too, obviously.
I figure that just caching on first use is good enough for now.

PR #70, opened 2016-07-14
Move ensureErrorContains to testonly package

Martin2112:


testonly/errors.go
```diff
@@ -0,0 +1,19 @@
+package testonly
+
+import (
+	"strings"
+	"testing"
+)
+
+// EnsureErrorContains checks that an error contains a specific substring and fails a
+// test with a fatal if it does not or the error was nil.
```
AlCutter: It only does `Fatal()` if the err was nil, otherwise it calls `Error()`

Martin2112: Should have both been fatal. Fixed.

PR #72, opened 2016-07-14
Add testing for expected SMT writes too

AlCutter:


merkle/sparse_merkle_tree_test.go
```diff
@@ -287,46 +288,58 @@ func testSparseTreeFetches(t *testing.T, vec sparseTestVector) {
...
-			if _, ok := e[pathNode]; ok {
-				// we're modifying both children of a node because two keys are
-				// intersecting, since both will be recalculated neither will be read
-				// from storage so we remove the previously set expectation for this
-				// node's sibling, and skip adding one for this node:
-				delete(e, pathNode)
-				continue
+
+	{
+		readMutex.Lock()
```
Martin2112: Looks like you could just defer the unlock? Though it's hard to see via the web view.

merkle/sparse_merkle_tree_test.go
```diff
@@ -287,46 +288,58 @@ func testSparseTreeFetches(t *testing.T, vec sparseTestVector) {
...
+
+		// calculate the set of expected node reads.
```
Martin2112: This expectation setting is starting to look like a lot of work. In some ways it might be better if we had a fake storage but as previously discussed that's a maintenance overhead too.

merkle/sparse_merkle_tree_test.go
AlCutter: I'd hoped that too, but sadly the `defer` *really* means when the containing func returns, rather than when we leave the scope.

merkle/sparse_merkle_tree_test.go
AlCutter: Yeah, but actually I'm glad we went this way - it's tedious but I know for realz that this sets the expected nodes and only those ones - this actually helped me catch a bug.

PR #75, opened 2016-07-21
SMT Subtrees can use distinct transactions

AlCutter:


merkle/sparse_merkle_tree.go
```diff
@@ -97,14 +99,18 @@ type subtreeWriter struct {
 
 // getOrCreateChildSubtree returns, or creates and returns, a subtree worker
```
Martin2112: Bit pedantic but it's not actually a worker.

merkle/sparse_merkle_tree.go
```diff
@@ -265,7 +278,11 @@ func leafQueueSize(depths []int) int {
 }
 
 // newLocalSubtreeWriter creates a new local go-routing based subtree worker.
```
Martin2112: typo: go-routine

merkle/sparse_merkle_tree.go
AlCutter: Fixed

merkle/sparse_merkle_tree.go
AlCutter: Fixed

PR #76, opened 2016-07-21
Start on tree stratification

AlCutter:


storage/mysql/tree_storage.go
```diff
@@ -14,23 +14,23 @@ import (
...
 		 VALUES(?,?,?,?,?,?)`
 const selectTreeRevisionAtSizeSql string = "SELECT TreeRevision FROM TreeHead WHERE TreeId=? AND TreeSize=? ORDER BY TreeRevision DESC LIMIT 1"
 const selectActiveLogsSql string = "select TreeId, KeyId from Trees where TreeType='LOG'"
 const selectActiveLogsWithUnsequencedSql string = "SELECT DISTINCT t.TreeId, t.KeyId from Trees t INNER JOIN Unsequenced u WHERE TreeType='LOG' AND t.TreeId=u.TreeId"
 
-const selectNodesSql string = `SELECT x.NodeId, x.MaxRevision, Node.NodeHash
-				 FROM (SELECT n.NodeId, max(n.NodeRevision) AS MaxRevision
-							 FROM Node n
-							 WHERE n.NodeId IN (` + placeholderSql + `) AND
+const selectSubtreeSql string = `SELECT x.SubtreeId, x.MaxRevision, Subtree.Nodes
```
Martin2112: I think this query will perform similarly to the get nodes one. Don't worry about it now, we know we need to do optimization.

AlCutter: Yeah, that'll be fun...

PR #77, opened 2016-07-21
Add a subtree cache

AlCutter:


storage/subtree_cache.go
```diff
@@ -0,0 +1,168 @@
...
+	strataDepth = 8
+)
+
+// NewSubtreeCache returns a newly intialised cache ready for use.
+// TODO(al): consider supporting different sized subtrees - for now everything's subtrees of 8 levels.
+func NewSubtreeCache() SubtreeCache {
+	return SubtreeCache{
+		subtrees:      make(map[string]*SubtreeProto),
+		dirtyPrefixes: make(map[string]bool),
+		mutex:         new(sync.RWMutex),
```
Martin2112: Probably worth documenting what the mutex protects.

storage/subtree_cache.go
```diff
@@ -0,0 +1,168 @@
...
+	prefixSplit := (id.PrefixLenBits - 1) / strataDepth
+	suffixEnd := (id.PrefixLenBits-1)/8 + 1
+	s := Suffix{
+		bits: byte((id.PrefixLenBits-1)%strataDepth) + 1,
+		path: make([]byte, suffixEnd-prefixSplit),
+	}
+	// TODO(al): is all this copying actually necessary?
+	copy(s.path, id.Path[prefixSplit:suffixEnd])
+	if id.PrefixLenBits%8 != 0 {
+		suffixMask := (byte(0x1<<uint((id.PrefixLenBits%8))) - 1) << uint(8-id.PrefixLenBits%8)
```
Martin2112: I think we should be using look up tables for some of the bit flipping but don't worry about that now.

storage/subtree_cache.go
```diff
@@ -0,0 +1,168 @@
...
+// GetNodeHash retrieves the previously written hash and corresponding tree
+// revision for the given node ID.
+func (s *SubtreeCache) GetNodeHash(id NodeID, getSubtree GetSubtreeFunc) (trillian.Hash, int64, error) {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+
+	px, sx := splitNodeID(id)
+	prefixKey := string(px)
+	c := s.subtrees[prefixKey]
+	if c == nil {
```
Martin2112: Is c == nil a cache miss?

storage/subtree_cache.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for k, v := range s.subtrees {
+		if s.dirtyPrefixes[k] {
+			bk := []byte(k)
+			if !bytes.Equal(bk, v.Prefix) {
+				return fmt.Errorf("inconsistent cache: prefix key is %v, but cached object claims %v", bk, v.Prefix)
+			}
+			if len(v.Nodes) > 0 {
+				if err := setSubtree(v); err != nil {
+					return err
```
Martin2112: Presumably storage is inconsistent if we get here? What's the recovery strategy?

storage/subtree_cache.go
AlCutter: Good call, done.

storage/subtree_cache.go
AlCutter: Ok :)

storage/subtree_cache.go
AlCutter: Yes and no :) ...

You *should* be writing in the "future" here (i.e. at `LatestTreeRoot.treeRevision+1`), so partially succeeding updates haven't screwed the currently serving tree, but you have a problem if you want the tree to ever advance.

One way to deal with this is to ensure that any future attempts to write at this n+1th revision contain the same node updates as the set which were attempted in the current update, so e.g. we could have the Log Sequencer/Signer commit to a particular batches of sequenced entries for each tree revision and they will be retried (in order) on any failures.

Similarly for the Log-based Map, the mapping function can batch updates and commit to those, with the same in-order retry constraints.

You could imagine the Log Seq/Signer and Mapper not actually interacting with the trees directly at all, and just writing out work batches into a queue which some internal mechanism reads and applies (with retry) to the underlying trees.  A lighter-weight alternative might be something where the first attempt to write to the future revision stores some hash over the whole batch (before attempting to apply the batch) and any future retries must have the same hash or it's an error - this might make more sense it it turns out it's easy for the Seq/Signer Mapper to exactly replay previous failed updates for example.

Martin2112: If only there was some sort of technical document where you could write that sort of explanation ....

AlCutter: Coming RSN :)

PR #79, opened 2016-07-22
Use subtree cache

AlCutter:


merkle/sparse_merkle_tree.go
```diff
@@ -100,14 +100,15 @@ type subtreeWriter struct {
 // getOrCreateChildSubtree returns, or creates and returns, a subtree for the
 // specified childPrefix.
 func (s *subtreeWriter) getOrCreateChildSubtree(childPrefix []byte) (Subtree, error) {
-	childPrefixStr := string(childPrefix)
+	cp := append(make([]byte, 0, len(childPrefix)), childPrefix...)
```
Martin2112: Is that copy needed?

merkle/sparse_merkle_tree.go
```diff
@@ -176,8 +177,14 @@ func (s *subtreeWriter) RootHash() (trillian.Hash, error) {
 func nodeIDFromAddress(size int, prefix []byte, index *big.Int, depth int) storage.NodeID {
 	ib := index.Bytes()
 	t := make(trillian.Hash, size)
+	depthBytes := (depth-1)/8 + 1
```
Martin2112: I'm naturally suspicious and would be checking for depth <= 0

merkle/sparse_merkle_tree.go
```diff
@@ -198,11 +206,17 @@ func (s *subtreeWriter) buildSubtree() {
 			return
 		}
 		leaves = append(leaves, HStar2LeafHash{Index: new(big.Int).SetBytes(ih.index), LeafHash: ih.hash})
+		nodesToStore = append(nodesToStore,
+			storage.Node{
+				NodeID:       storage.NewNodeIDFromHash(append(append(make([]byte, 0, s.treeHasher.Size()), s.prefix...), ih.index...)),
```
Martin2112: Looks a bit ugly. Bytes.Join any good?

merkle/sparse_merkle_tree.go
```diff
@@ -215,14 +229,19 @@ func (s *subtreeWriter) buildSubtree() {
 				return nil, nil
 			}
 			if expected, got := nodeID, nodes[0].NodeID; !expected.Equivalent(got) {
-				return nil, fmt.Errorf("expected node ID %s from storage, but got %s", expected, got)
+				return nil, fmt.Errorf("expected node ID %s from storage, but got %s", expected.String(), got.String())
```
Martin2112: Do you actually have to call .String explicitly?

merkle/sparse_merkle_tree.go
AlCutter: To be honest, I don't know - I added a bunch of copies trying to debug issues - I suspect that most (possibly all) of them will turn out to be unnecessary, but I'll go back and figure that out for sure once I'm back.  I'll pop another TODO in here to mark the copy.

merkle/sparse_merkle_tree.go
```diff
@@ -285,15 +304,15 @@ func newLocalSubtreeWriter(rev int64, prefix []byte, depths []int, newTX newTXFu
 	}
 	tree := subtreeWriter{
 		treeRevision: rev,
-		prefix:       prefix,
+		prefix:       append(make([]byte, 0, len(prefix)), prefix...),
```
Martin2112: maybe a copyBytes() would clear some of these up a bit?

merkle/sparse_merkle_tree.go
AlCutter: Done.

merkle/sparse_merkle_tree.go
AlCutter: Nice, thanks!

merkle/sparse_merkle_tree.go
AlCutter: Yes, the formatter doesn't know to call it on your behalf (though I think you can (somehow) do that - big.Int seems to do that.)

merkle/sparse_merkle_tree.go
AlCutter: I'll punt on that for now, but if the copies turn out to be required I'll add it then - I've added a TODO to that effect.

PR #80, opened 2016-07-22
Add a toy to exercise the SMT writing code.

AlCutter:
Exports KeyHasher


merkle/map_hasher.go
```diff
@@ -8,15 +8,15 @@ import (
 // "null" hashes for the unused sections of a SparseMerkleTree.
 type MapHasher struct {
 	TreeHasher
-	keyHasher  keyHashFunc
+	HashKey    keyHashFunc
```
Martin2112: Not sure about this rename. I'd assume HashKey to be data, not a function.

AlCutter: It was in the verb sense to match the existing `HashLeaf` etc. funcs.

PR #82, opened 2016-07-25
Implement the CT add-chain handler and add some tests for it

Martin2112:
Note to reviewer: implements add-chain, validates certs, queues the leaf on the backend and formats the response. Doesn't handle deduping yet as the backend needs more work to enable this.


RJPercival:
Ok, glad I didn't get around to reviewing it yet then! Let me know when you'd like me to take a look at it.

examples/ct/ct_handlers.go
```diff
@@ -19,21 +30,28 @@ const (
...
+	contentTypeHeader      string     = "Content-Type"
+	contentTypeJSON        string     = "application/json"
+	jsonMapKeyCertificates string     = "certificates"
+	logVerboseLevel        glog.Level = 2
 )
 
 // CTRequestHandlers provides HTTP handler functions for CT V1 as defined in RFC 6962
 // and functionality to translate CT client requests into forms that can be served by a
 // log backend RPC service.
 type CTRequestHandlers struct {
```
RJPercival: Could you please document each of these members?

examples/ct/ct_handlers.go
```diff
@@ -44,23 +62,40 @@ type addChainRequest struct {
 	Chain []string
 }
 
+type addChainResponse struct {
+	SctVersion int    `json:sct_version`
+	ID         string `json:id`
+	Timestamp  string `json:timestamp`
```
RJPercival: Why is Timestamp a string, rather than a uint?

examples/ct/ct_handlers.go
```diff
@@ -44,23 +62,40 @@ type addChainRequest struct {
 	Chain []string
 }
 
+type addChainResponse struct {
```
RJPercival: Adding a comment linking to the relevant part of the RFC would be helpful.

examples/ct/ct_handlers.go
```diff
@@ -44,23 +62,40 @@ type addChainRequest struct {
...
+	Extensions string `json:extensions`
+	Signature  string `json:signature`
+}
+
 func parseBodyAsJSONChain(w http.ResponseWriter, r *http.Request) (addChainRequest, error) {
 	body, err := ioutil.ReadAll(r.Body)
 
 	if err != nil {
-		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		if glog.V(logVerboseLevel) {
```
RJPercival: Can be written more succinctly as `glog.V(logVerboseLevel).Info("Failed to read request body")`.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
Martin2112: JSON works with floats, this value is outside the range that can be represented. Or is there a trick to this somewhere?

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
Martin2112: OK, that should now be fixed.

examples/ct/ct_handlers.go
RJPercival: It was the CTRequestHandlers members I wanted documented, not the constants above.

examples/ct/ct_handlers.go
RJPercival: It looks like there are more places where this can be done, e.g. this line.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
```diff
@@ -19,48 +29,76 @@ const (
...
+			glog.Info("Failed to read request body")
+		}
+		sendHttpError(w, http.StatusBadRequest, err)
 		return addChainRequest{}, err
 	}
 
 	var req addChainRequest
 	if err := json.Unmarshal(body, &req); err != nil {
-		http.Error(w, http.StatusText(http.StatusBadRequest)+": "+err.Error(), http.StatusBadRequest)
+		glog.V(logVerboseLevel).Info("Failed to read request body")
```
RJPercival: Could you make this log message distinct from the one on line 83 please? "Failed to decode request body" would be fine. Is there a reason you don't log `err`?

examples/ct/ct_handlers.go
RJPercival: Thanks, that's great.

examples/ct/ct_handlers.go
Martin2112: The docs say it's less efficient to do it this way but are unclear on how much.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
```diff
@@ -89,19 +133,105 @@ func enforceMethod(w http.ResponseWriter, r *http.Request, method string) bool {
...
+
+		var logEntryBuffer bytes.Buffer
+		logEntry := NewCTLogEntry(merkleTreeLeaf, validPath)
+		if err := logEntry.Serialize(&logEntryBuffer); err != nil {
+			glog.Warningf("Failed to serialize log entry: %v", err)
+			sendHttpError(w, http.StatusInternalServerError, err)
+			return
+		}
+
+		leafHash := sha256.Sum256(leafBuffer.Bytes())
```
RJPercival: Is this different to a leaf hash as per RFC6962? Ordinarily, there'd be a prefixed 0 byte.

examples/ct/ct_handlers.go
RJPercival: Only if the arguments are expensive to evaluate, which they aren't in any of these cases.

examples/ct/ct_handlers.go
Martin2112: This is the leaf input to the tree hasher, the prefix is applied by the hasher. It's possible that the CT specific and generic parts don't quite connect up, which should become apparent fairly soon.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
RJPercival: I'm a bit unclear on that. The prefix can't possibly be applied after hashing `leafBuffer`. If it's the case that `leafHash` isn't a leaf hash in the RFC6962 sense, then a comment saying so at the least is advisable.

examples/ct/ct_handlers.go
```diff
@@ -89,19 +129,105 @@ func enforceMethod(w http.ResponseWriter, r *http.Request, method string) bool {
...
+		leafProto := trillian.LeafProto{LeafHash: leafHash[:], LeafData: leafBuffer.Bytes(), ExtraData: logEntryBuffer.Bytes()}
+
+		request := trillian.QueueLeavesRequest{LogId: c.logID, Leaves: []*trillian.LeafProto{&leafProto}}
+
+		ctx, _ := context.WithDeadline(context.Background(), getRPCDeadlineTime(c))
+
+		response, err := c.rpcClient.QueueLeaves(ctx, &request)
+
+		if err != nil || !rpcStatusOK(response.GetStatus()) {
+			// Request failed on backend, doesn't account for bad request etc. yet
```
RJPercival: Better to put a TODO here for accounting for bad requests, etc. so that it stands out.

examples/ct/ct_handlers.go
Martin2112: The leaf hash is a consistency check on leaf buffer. Anything sent to the backend is outside of RFC 6962. I'll add a comment.

examples/ct/ct_handlers.go
```diff
@@ -19,48 +29,78 @@ const (
...
+		glog.V(logVerboseLevel).Infof("Failed to read request body: %v", err)
+		sendHttpError(w, http.StatusBadRequest, err)
 		return addChainRequest{}, err
 	}
 
 	var req addChainRequest
 	if err := json.Unmarshal(body, &req); err != nil {
-		http.Error(w, http.StatusText(http.StatusBadRequest)+": "+err.Error(), http.StatusBadRequest)
+		glog.V(logVerboseLevel).Infof("Failed to parse request body: %v", err)
+		sendHttpError(w, http.StatusBadRequest, fmt.Errorf("Unmarshal failed with %v on %s", err, body))
```
RJPercival: I believe error strings are supposed to start lowercase.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go

Martin2112:
I think this is ready to merge. Let me know.

examples/ct/ct_handlers.go
```diff
@@ -202,3 +340,68 @@ func (c CTRequestHandlers) RegisterCTHandlers() {
 	http.HandleFunc(pathFor("get-roots"), wrappedGetRootsHandler(c.trustedRoots, c.rpcClient))
 	http.HandleFunc(pathFor("get-entry-and-proof"), wrappedGetEntryAndProofHandler(c.rpcClient))
 }
+
+// Generates a custom error page to give more information on why something didn't work
+// TODO(Martin2112): Not sure if we want to expose any detail or not
```
RJPercival: If `err != nil`, is `isPrecert` still meaningful? The following logging implies it is.

examples/ct/ct_handlers.go
```diff
@@ -202,3 +340,68 @@ func (c CTRequestHandlers) RegisterCTHandlers() {
...
+
+	if err != nil {
+		// We rejected it because the cert failed checks or we could not find a path to a root etc.
+		// Lots of possible causes for errors
+		glog.Warningf("Chain failed to verify: %v", req)
+		sendHttpError(w, http.StatusBadRequest, err)
+		return nil
+	}
+
+	isPrecert, err := IsPrecertificate(validPath[0])
```
RJPercival: If `err != nil`, is `isPrecert` still meaningful? The following logging implies it is.

examples/ct/ct_handlers.go
```diff
@@ -202,3 +340,68 @@ func (c CTRequestHandlers) RegisterCTHandlers() {
...
+	return validPath
+}
+
+// marshalLogIDAndSignatureForResponse is used by add-chain and add-pre-chain. It formats the
+// signature and log id ready to send to the client.
+func marshalLogIDAndSignatureForResponse(sct ct.SignedCertificateTimestamp, km crypto.KeyManager) ([sha256.Size]byte, string, error) {
+	logID, err := GetCTLogID(km)
+
+	if err != nil {
+		return [32]byte{}, "", fmt.Errorf("failed to marshal logID: %v %v", logID, err)
```
RJPercival: Does it make sense to log `logID` when `err != nil`?

examples/ct/ct_handlers.go
```diff
@@ -202,3 +340,68 @@ func (c CTRequestHandlers) RegisterCTHandlers() {
...
+		sendHttpError(w, http.StatusBadRequest, err)
+		return nil
+	}
+
+	return validPath
+}
+
+// marshalLogIDAndSignatureForResponse is used by add-chain and add-pre-chain. It formats the
+// signature and log id ready to send to the client.
+func marshalLogIDAndSignatureForResponse(sct ct.SignedCertificateTimestamp, km crypto.KeyManager) ([sha256.Size]byte, string, error) {
```
RJPercival: It might be better to pass `sct` and `km` as pointers, to avoid copying, though this probably has ramifications on other function signatures so feel free to defer/ignore this.

examples/ct/ct_handlers.go
Martin2112: I've split it up, was trying to cover too much in one statement.

examples/ct/ct_handlers.go
Martin2112: No.

examples/ct/ct_handlers.go
Martin2112: Leaving this for now as I've got another PR that needs this one.

examples/ct/ct_handlers_test.go
```diff
@@ -185,14 +218,14 @@ func TestPostHandlersAcceptNonEmptyCertChain(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		// For now they return not implemented as the handler is a stub
-		assert.Equal(t, http.StatusNotImplemented, resp.StatusCode, "Wrong status code for non empty chain in body")
+		// TODO(Martin2112): Remove not implemented from test when all the handlers have been written
+		assert.True(t, resp.StatusCode == http.StatusNotImplemented || resp.StatusCode == http.StatusBadRequest, "Wrong status code for GET to GET handler")
```
RJPercival: It could be helpful to include the value of `resp.StatusCode` in the failure message. I expect once the "not implemented" part of that test is removed, this would just turn into `assert.Equal` though, which would achieve this too.

examples/ct/ct_handlers_test.go
```diff
@@ -222,13 +255,195 @@ func TestGetRoots(t *testing.T) {
...
+// rejected
+func TestAddChainPrecert(t *testing.T) {
+	km := new(crypto.MockKeyManager)
+	client := new(trillian.MockTrillianLogClient)
+
+	roots := loadCertsIntoPoolOrDie(t, []string{testonly.CACertPEM})
+	reqHandlers := CTRequestHandlers{0x42, roots, client, km, time.Millisecond * 500, fakeTimeSource}
+
+	precert, err := fixchain.CertificateFromPEM(testonly.PrecertPEMValid)
+	_, ok := err.(x509.NonFatalErrors)
```
RJPercival: This might be clearer if you delete line 287 and...

examples/ct/ct_handlers_test.go
```diff
@@ -222,13 +255,195 @@ func TestGetRoots(t *testing.T) {
...
+
+	if err != nil && !ok {
+		t.Fatal(err)
```
RJPercival: ...change line 290 to be `assert.IsType(t, x509.NonFatalErrors, err)`

examples/ct/ct_handlers_test.go
```diff
@@ -222,13 +255,195 @@ func TestGetRoots(t *testing.T) {
...
+	}
+
+	var buffer bytes.Buffer
+	// It's tempting to avoid creating and flushing the intermediate writer but it doesn't work
+	writer := bufio.NewWriter(&buffer)
+	err := json.NewEncoder(writer).Encode(&chain)
+	writer.Flush()
+
+	if err != nil {
+		t.Fatalf("failed to create test json: %v", chain)
```
RJPercival: Could use `assert.NoError` here instead.

examples/ct/ct_handlers_test.go
```diff
@@ -222,13 +255,195 @@ func TestGetRoots(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to sign test SCT or Merkle Leaf: %v", err)
+	}
+
+	var b bytes.Buffer
+	if err := WriteMerkleTreeLeaf(&b, merkleLeaf); err != nil {
+		t.Fatalf("failed to serialize leaf: %v", err)
+	}
+
+	leafHash := sha256.Sum256(b.Bytes())
```
RJPercival: Could do with a comment saying that it's not an RFC6962 leaf hash, to avoid confusion.

examples/ct/ct_handlers_test.go
Martin2112: Done.

examples/ct/ct_handlers_test.go
Martin2112: As discussed I don't require that there is a non fatal error for the test to succeed. If the CT x509 is modified to accept the precert extension without complaint then I don't want this test to fail.

examples/ct/ct_handlers_test.go
Martin2112: Done, plus a couple of other places.

examples/ct/ct_handlers_test.go
Martin2112: Done.

examples/ct/ct_handlers_test.go
RJPercival: My proposal wouldn't violate that. Lines 287-290 would become:
```go
if err != nil {
  assert.IsType(t, x509.NonFatalErrors, err)
}
```
I think that clearly expresses the intent that if there is an error, it must be a non-fatal X509 error.

examples/ct/ct_handlers_test.go
```diff
@@ -434,16 +431,13 @@ func makeAddChainRequest(t *testing.T, reqHandlers CTRequestHandlers, body io.Re
 	handler := wrappedAddChainHandler(reqHandlers)
 
 	req, err := http.NewRequest("POST", "http://example.com/ct/v1/add-chain", body)
-	if err != nil {
-		t.Fatalf("test request setup failed: %v", err)
-	}
+
```
RJPercival: Nit: There is an abundance of vertical whitespace, but that doesn't bother me much.

examples/ct/ct_handlers_test.go
Martin2112: OK we were at cross purposes. Though you have to write x509.NonFatalErrors{}

examples/ct/ct_handlers_test.go
```diff
@@ -284,12 +284,9 @@ func TestAddChainPrecert(t *testing.T) {
 	reqHandlers := CTRequestHandlers{0x42, roots, client, km, time.Millisecond * 500, fakeTimeSource}
 
 	precert, err := fixchain.CertificateFromPEM(testonly.PrecertPEMValid)
-	_, ok := err.(x509.NonFatalErrors)
-
-	if err != nil && !ok {
-		t.Fatal(err)
+	if err != nil {
+		assert.IsType(t, x509.NonFatalErrors{}, err, "boom")
```
Martin2112: Cut and paste strikes again. Fixed.

PR #83, opened 2016-07-25
Add a doc comment to CT package

Martin2112:
Make it clear that only code in examples/ct should reference the CT repository.


examples/ct/doc.go
```diff
@@ -0,0 +1,9 @@
+/*
+Package ct contains a usage example by providing an implementation of an RFC6962 compatible CT
+log server using a Trillian log server as backend storage via its GRPC API.
+
+IMPORTANT: Only code rooted within this part of the tree should refer to the CT
+Github repository. Other parts of the system must not assume that the data they're
```
phad: maybe include github URL for CT?

Martin2112: Done.

PR #85, opened 2016-07-26
Add leaf index to backend API proto

Martin2112:
And update tests. Allows frontends to do additional checks on the data they get before passing it to clients.


server/trillian_log_server.go
```diff
@@ -222,7 +222,7 @@ func (t *TrillianLogServer) commitAndLog(tx storage.LogTX, op string) error {
 }
 
 func protoToLeaf(proto *trillian.LeafProto) trillian.LogLeaf {
-	return trillian.LogLeaf{Leaf: trillian.Leaf{LeafHash: proto.LeafHash, LeafValue: proto.LeafData, ExtraData: proto.ExtraData}}
+	return trillian.LogLeaf{SequenceNumber: proto.LeafIndex, Leaf: trillian.Leaf{LeafHash: proto.LeafHash, LeafValue: proto.LeafData, ExtraData: proto.ExtraData}}
```
phad: will (or does?) trillian.Leaf have a SequenceNumber property?

phad: looks like it does - should you be copying SequenceNumber across from proto to LogLeaf too?

server/trillian_log_server_test.go
```diff
@@ -182,7 +182,7 @@ func TestGetLeavesByIndexMultiple(t *testing.T) {
 func TestQueueLeavesStorageError(t *testing.T) {
 	test := newParameterizedTest("QueueLeaves",
 		func(t *storage.MockLogTX) {
-			t.On("QueueLeaves", []trillian.LogLeaf{leaf0}).Return(errors.New("STORAGE"))
+			t.On("QueueLeaves", []trillian.LogLeaf{leaf1}).Return(errors.New("STORAGE"))
```
phad: did you change from leaf0 to leaf1 because leaf0 has no SequenceNumber? does it need to stay that way?

server/trillian_log_server.go
Martin2112: Leaf doesn't have a sequence number as it's shared with the map implementation. LogLeaf does and I thought I did copy it both ways. I'll have another look at it.

server/trillian_log_server_test.go
Martin2112: Yes leaf0 has sequence number zero (implicitly) and I now pass one with sequence number 1, which failed the mock expectation.

server/trillian_log_server.go
Martin2112: It does get copied by protoToLeaf() and leafToProto(). Is that what you meant?

server/trillian_log_server_test.go
phad: I don't see that myself.

server/trillian_log_server.go
phad: Ignore, I was confusing trillian.Leaf{} for trillian.LogLeaf{}

PR #86, opened 2016-07-27
Implement Serialization for MerkleTreeLeaf and CTLogEntry

Martin2112:
Missing piece to be able to finish the add-chain handler. CTLogEntry is internal to server and will be used as in the C++ code to store the CT specific data when the leaf is passed to the backend.

### DIFF
--- a/examples/ct/structures.go
+++ b/examples/ct/structures.go
@@ -4,6 +4,10 @@ package ct
 
 import (
 	"crypto/sha256"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
 	"time"
 
 	"github.com/google/certificate-transparency/go"
@@ -13,6 +17,15 @@ import (
 )
 
 const millisPerNano int64 = 1000
+
+// CTLogEntry holds the data we send to the backend with the leaf. There is a LogEntry type in
+// the CT code but it is a superset of what we need.
+type CTLogEntry struct {
+	// The leaf structure that was built from the client submission
+	leaf ct.MerkleTreeLeaf
+	// The complete chain for the certificate or precertificate as raw bytes
+	chain []ct.ASN1Cert
+}
 
 // GetCTKeyID takes the key manager for a log and returns the LogID. (see RFC 6962 S3.2)
 // In CT V1 the log id is a hash of the public key.
@@ -28,7 +41,7 @@ func GetCTLogID(km crypto.KeyManager) ([sha256.Size]byte, error) {
 
 func serializeAndSignSCT(km crypto.KeyManager, leaf ct.MerkleTreeLeaf, sctInput ct.SignedCertificateTimestamp, t time.Time) (ct.MerkleTreeLeaf, ct.SignedCertificateTimestamp, error) {
 	// Serialize SCT signature input to get the bytes that need to be signed
-	res, err := ct.SerializeSCTSignatureInput(sctInput, ct.LogEntry{Leaf:leaf})
+	res, err := ct.SerializeSCTSignatureInput(sctInput, ct.LogEntry{Leaf: leaf})
 
 	if err != nil {
 		return ct.MerkleTreeLeaf{}, ct.SignedCertificateTimestamp{}, err
@@ -81,8 +94,8 @@ func SignV1SCTForCertificate(km crypto.KeyManager, cert *x509.Certificate, t tim
 	sctInput := getSCTForSignatureInput(t)
 
 	// Build up a MerkleTreeLeaf for the cert
-	timestampedEntry := ct.TimestampedEntry{Timestamp:sctInput.Timestamp, EntryType:ct.X509LogEntryType, X509Entry:cert.Raw}
-	leaf := ct.MerkleTreeLeaf{Version: ct.V1, LeafType:ct.TimestampedEntryLeafType, TimestampedEntry:timestampedEntry}
+	timestampedEntry := ct.TimestampedEntry{Timestamp: sctInput.Timestamp, EntryType: ct.X509LogEntryType, X509Entry: cert.Raw}
+	leaf := ct.MerkleTreeLeaf{Version: ct.V1, LeafType: ct.TimestampedEntryLeafType, TimestampedEntry: timestampedEntry}
 
 	return serializeAndSignSCT(km, leaf, sctInput, t)
 }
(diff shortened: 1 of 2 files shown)


examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
 	return ct.SignedCertificateTimestamp{
-		SCTVersion:ct.V1,
+		SCTVersion: ct.V1,
 		Timestamp:  uint64(t.UnixNano() / millisPerNano), // spec uses millisecond timestamps
 		Extensions: ct.CTExtensions{}}
-}
\ No newline at end of file
+}
+
+func NewLogEntry(leaf ct.MerkleTreeLeaf, certChain []*x509.Certificate) *CTLogEntry {
```
phad: Can you leave a TODO in here about verifying that the chain is well-formed?
By which I mean that each cert chains to the next.

Is there a plan to implement a root set restriction at some point?  That would need to be validated here too.

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+	chain := []ct.ASN1Cert{}
+
+	for _, cert := range certChain {
+		chain = append(chain, cert.Raw)
+	}
+
+	logEntry := new(CTLogEntry)
```
phad: You might more concisely write:

return &CTLogEntry{Leaf: leaf, Chain: chain}

replacing these 6 lines.

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+
+	logEntry.Leaf = leaf
+	logEntry.Chain = chain
+
+	return logEntry
+}
+
+// WriteTimestampedEntry writes out a TimestampedEntry structure in the binary format defined
+// by RFC 6962. The CT go code includes a deserializer but not a serializer so we might as
```
phad: By 'the CT go code' you mean the github.com/google/certificate-transparency/go  code?
Do you plan to move this serialization code into that repo at some point?

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+// well make this available.
+func WriteTimestampedEntry(w io.Writer, t ct.TimestampedEntry) error {
+	if err := binary.Write(w, binary.BigEndian, &t.Timestamp); err != nil {
```
phad: are you concerned at all about the fact that some varying amount of data might have been written to 'w' if the function goes on to return err below?

examples/ct/structures.go
phad: Is that why this func is exported?

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+		chain = append(chain, certBytes)
+	}
+
+	c.Chain = chain
+
+	return nil
+}
+
+// These came from the CT go code. Currently don't want to push changes upstream to make
+// them visible but this is an option for the future.
```
phad: I suppose if the exported WriteFoo() funcs above were pushed into the CT repo instead, then your duplications of writeBar() here would disappear.

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+	if err := WriteTimestampedEntry(w, l.TimestampedEntry); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Serialize writes out a CTLogEntry in binary form. This is not an RFC 6962 data structure
+// and is only used internally by the log.
+func (c CTLogEntry) Serialize(w io.Writer) error {
```
phad: If this is exported is there much point in the WriteFoo() funcs above being exported?

examples/ct/structures.go
Martin2112: Validation is done by the request handler before this gets created. The rest of the code works with x509.Certificate objects. This is only so we can serialize the extra data that the backend needs to store so we can implement CT on the generic storage.

examples/ct/structures.go
Martin2112: Unfortunately I can't as I need to extract the raw bytes from the cert objects.

examples/ct/structures.go
```diff
@@ -99,15 +112,208 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
...
+
+		chain = append(chain, certBytes)
+	}
+
+	c.Chain = chain
+
+	return nil
+}
+
+// These came from the CT go code. Currently don't want to push changes upstream to make
```
phad: Can you list exactly which funcs were duplicated from the CT repo.  Or add a similar end marker.  I'm assuming it's everything from here to the end of the file.

examples/ct/structures.go
Martin2112: I don't know yet. It might make more sense to push this code back to CT but it's server side rather than client so arguably it doesn't quite fit with what exists.

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
 
+func TestSerializeMerkleTreeLeafBadVersion(t *testing.T) {
+	var leaf ct.MerkleTreeLeaf
+
+	leaf.Version = 199 // Out of any expected valid range
+
+	var w io.Writer
+	err := WriteMerkleTreeLeaf(w, leaf)
+	assert.Error(t, err, "incorrectly serialized leaf with bad version")
+}
```
phad: So it would be _nice_ to assert that w wasn't written to because of errors ... dependent on my comment above.
I guess it depends how much this is example code versus something more robust.  (or whether you think a partially written-to stream matters)

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
+	leaf.Version = ct.V1
+	leaf.LeafType = 212
+
+	var w io.Writer
+	err := WriteMerkleTreeLeaf(w, leaf)
+	assert.Error(t, err, "incorrectly serialized leaf with bad leaf type")
+}
+
+func TestSerializeMerkleTreeLeafCert(t *testing.T) {
+	ts := ct.TimestampedEntry{Timestamp: 12345, EntryType: ct.X509LogEntryType, X509Entry: ct.ASN1Cert([]byte{0x10, 0x11, 0x12, 0x13, 0x20, 0x21, 0x22, 0x23}), Extensions: ct.CTExtensions{}}
```
phad: I think this struct would bear being split across lines, one field per line, for readabilitity's sake.

examples/ct/structures.go
Martin2112: I used io.Writer so it mirrors the existing Go serialization code. Any error will cause the frontend request to fail and the buffer won't be used. These are part of the frontend ->  backend flow and never get written to clients directly.

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
+	leaf := ct.MerkleTreeLeaf{LeafType: ct.TimestampedEntryLeafType, Version: ct.V1, TimestampedEntry: ts}
+
+	var buff bytes.Buffer
```
phad: cool.   I assume this grows dynamically as data is written into it.
This is what you'd need to use if you did my comments about all-or-nothing writing above.

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
+
+	if err != nil {
+		t.Fatalf("failed to read leaf: %v", err)
+	}
+
+	assert.Equal(t, leaf, *leaf2, "leaf mismatch after serialization roundtrip (cert)")
+}
+
+func TestSerializeMerkleTreePrecert(t *testing.T) {
+	ts := ct.TimestampedEntry{Timestamp: 12345, EntryType: ct.PrecertLogEntryType, PrecertEntry: ct.PreCert{IssuerKeyHash: [sha256.Size]byte{0x55, 0x56, 0x57, 0x58, 0x59}, TBSCertificate: ct.ASN1Cert([]byte{0x10, 0x11, 0x12, 0x13, 0x20, 0x21, 0x22, 0x23})}, Extensions: ct.CTExtensions{}}
```
phad: split across lines please

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
+
+	if err != nil {
+		t.Fatalf("failed to read leaf: %v", err)
+	}
+
+	assert.Equal(t, leaf, *leaf2, "leaf mismatch after serialization roundtrip (precert)")
+}
+
+func TestSerializeCTLogEntry(t *testing.T) {
+	ts := ct.TimestampedEntry{Timestamp: 12345, EntryType: ct.X509LogEntryType, X509Entry: ct.ASN1Cert([]byte{0x10, 0x11, 0x12, 0x13, 0x20, 0x21, 0x22, 0x23}), Extensions: ct.CTExtensions{}}
```
phad: and here

examples/ct/structures_test.go
```diff
@@ -147,6 +149,104 @@ func TestSignV1SCTForPrecertificate(t *testing.T) {
...
+	leaf := ct.MerkleTreeLeaf{LeafType: ct.TimestampedEntryLeafType, Version: ct.V1, TimestampedEntry: ts}
+
+	for chainLength := 1; chainLength < 10; chainLength++ {
```
phad: ooh!  10 certs ought to be enough for anyone!

examples/ct/structures.go
Martin2112: Yes, that's part of the argument for doing it. I might do it and refactor this once I've merged it and got the add-chain handler working. I need this to finish the handler.

examples/ct/structures_test.go
```diff
@@ -161,3 +261,20 @@ func setupMockKeyManager(toSign []byte) crypto.KeyManager {
 
 	return mockKeyManager
 }
+
+// Creates a dummy cert chain
+func createCertChain(numCerts int) []ct.ASN1Cert {
```
phad: hah, good job you didn't impl sig checking then :)

examples/ct/structures.go
Martin2112: I think I'm becoming more convinced that I should put the Write() stuff back into CT.

examples/ct/structures.go
Martin2112: I've expanded the comment to hopefully clarify this.

examples/ct/structures.go
Martin2112: Sorry, was looking at the wrong bit of code. Done.

examples/ct/structures.go
Martin2112: Done.

examples/ct/structures_test.go
Martin2112: I've added the extra checks.

examples/ct/structures_test.go
Martin2112: Done.

examples/ct/structures_test.go
Martin2112: Done.

examples/ct/structures_test.go
Martin2112: Done.

examples/ct/structures_test.go
Martin2112: I just wanted to be sure to test different variable length reads.

examples/ct/structures_test.go
Martin2112: Yes, this is all just a container for bytes that gets pushed to the backend and then retrieved to serve get-entries etc.

PR #87, opened 2016-07-29
Implement the CT add-pre-chain-handler

Martin2112:
This is mostly a refactoring and moving code around to break out helpers to make the add chain logic clearer and then a dedup of the code so that add-pre-chain shares everything. Most of the actual new code is tests.

### DIFF
--- a/examples/ct/ct_handlers.go
+++ b/examples/ct/ct_handlers.go
@@ -128,138 +128,91 @@ func enforceMethod(w http.ResponseWriter, r *http.Request, method string) bool {
 	return true
 }
 
-// All the handlers are wrapped so they have access to the RPC client
+// addChainInternal is called by add-chain and add-pre-chain as the logic involved in
+// processing these requests is almost identical
+func addChainInternal(w http.ResponseWriter, r *http.Request, c CTRequestHandlers, isPrecert bool) {
+	if !enforceMethod(w, r, httpMethodPost) {
+		// HTTP status code was already set
+		return
+	}
+
+	addChainRequest, err := parseBodyAsJSONChain(w, r)
+
+	if err != nil {
+		// HTTP status code was already set
+		return
+	}
+
+	// We already checked that the chain is not empty so can move on to verification
+	validPath := verifyAddChain(addChainRequest, w, *c.trustedRoots, isPrecert)
+
+	if validPath == nil {
+		// Chain rejected by verify. HTTP status code was already set
+		return
+	}
+
+	// Build up the SCT and MerkleTreeLeaf. The SCT will be returned to the client and
+	// the leaf will become part of the data sent to the backend.
+	var merkleTreeLeaf ct.MerkleTreeLeaf
+	var sct ct.SignedCertificateTimestamp
+
+	if isPrecert {
+		merkleTreeLeaf, sct, err = SignV1SCTForPrecertificate(c.logKeyManager, validPath[0], c.timeSource.Now())
+	} else {
+		merkleTreeLeaf, sct, err = SignV1SCTForCertificate(c.logKeyManager, validPath[0], c.timeSource.Now())
+	}
+
+	if err != nil {
+		glog.Warningf("Failed to create / serialize SCT or Merkle leaf: %v %v", sct, err)
+		sendHttpError(w, http.StatusInternalServerError, err)
+		return
+	}
+
+	// Inputs validated, pass the request on to the back end after hashing and serializing
+	// the data for the request
+	leafProto, err := buildLeafProtoForAddChain(merkleTreeLeaf, validPath)
+
+	if err != nil {
+		// Failure reason already logged
+		sendHttpError(w, http.StatusInternalServerError, err)
+		return
+	}
+
+	request := trillian.QueueLeavesRequest{LogId: c.logID, Leaves: []*trillian.LeafProto{&leafProto}}
+
+	ctx, _ := context.WithDeadline(context.Background(), getRPCDeadlineTime(c))
+
+	response, err := c.rpcClient.QueueLeaves(ctx, &request)
+
+	if err != nil || !rpcStatusOK(response.GetStatus()) {
+		// TODO(Martin2112): Possibly cases where the request we sent to the backend is invalid
+		// which isn't really an internal server error.
+		// Request failed on backend
+		sendHttpError(w, http.StatusInternalServerError, err)
+		return
+	}
+
+	// Success. We can now build and marshal the JSON response and write it out
+	err = marshalAndWriteAddChainResponse(sct, c.logKeyManager, w)
+
+	if err != nil {
+		// reason is logged and http status is already set
+		// TODO(Martin2112): Record failure for monitoring when it's implemented
+	}
+}
+
+// All the handlers are wrapped so they have access to the RPC client and other context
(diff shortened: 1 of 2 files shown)


examples/ct/ct_handlers_test.go
```diff
@@ -357,6 +358,48 @@ func TestAddChain(t *testing.T) {
...
+// Submit a chain with a valid path but using a cert instead of a precert. Should be rejected.
+func TestAddPrecertChainCert(t *testing.T) {
+	km := new(crypto.MockKeyManager)
+	client := new(trillian.MockTrillianLogClient)
+
+	roots := loadCertsIntoPoolOrDie(t, []string{testonly.CACertPEM})
+	reqHandlers := CTRequestHandlers{0x42, roots, client, km, time.Millisecond * 500, fakeTimeSource}
+
+	cert, err := fixchain.CertificateFromPEM(testonly.TestCertPEM)
+	_, ok := err.(x509.NonFatalErrors)
```
phad: this is a type assertion, right?
so I believe you are checking that CertificateFromPEM's err is nil or an x509.NonFatalError, and will t.Fatal for anything else?

examples/ct/ct_handlers_test.go
```diff
@@ -25,6 +24,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"golang.org/x/net/context"
+	"crypto/sha256"
```
phad: this import move seems spurious, crypto/sha256 normally comes  in the system imports at the top

examples/ct/ct_handlers_test.go
```diff
@@ -358,12 +372,42 @@ func TestAddChain(t *testing.T) {
...
+	reqHandlers := CTRequestHandlers{0x42, roots, client, km, time.Millisecond * 500, fakeTimeSource}
+
+	cert, err := fixchain.CertificateFromPEM(testonly.PrecertPEMValid)
+	_, ok := err.(x509.NonFatalErrors)
+
+	if err != nil && !ok {
+		t.Fatal(err)
+	}
+
+	pool := NewPEMCertPool()
```
phad: Does this PEMCertPool struct represent a certificate chain?

phad: Why the 'pool' name?  Are you memoizing certs once seen and storing a set of unique certs, or something?

examples/ct/ct_handlers_test.go
```diff
@@ -392,12 +436,89 @@ func TestAddPrecertChainCert(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	leaves := leafProtosForCert(t, km, pool.RawCertificates(), merkleLeaf)
+
+	client.On("QueueLeaves", mock.MatchedBy(deadlineMatcher), &trillian.QueueLeavesRequest{LogId: 0x42, Leaves: leaves}, mock.Anything /* []grpc.CallOption */).Return(&trillian.QueueLeavesResponse{Status: &trillian.TrillianApiStatus{StatusCode: trillian.TrillianApiStatusCode_OK}}, nil)
+
+	recorder := makeAddPrechainRequest(t, reqHandlers, chain)
+
+	assert.Equal(t, http.StatusOK, recorder.Code, "expected HTTP OK for valid add-chain: %v", recorder.Body)
```
phad: message: for valid 'add-pre-chain'

examples/ct/ct_handlers.go
```diff
@@ -494,3 +420,64 @@ func marshalLogIDAndSignatureForResponse(sct ct.SignedCertificateTimestamp, km c
...
+	w.Header().Set(contentTypeHeader, contentTypeJSON)
+	jsonData, err := json.Marshal(&resp)
+
+	if err != nil {
+		glog.Warningf("Failed to marshal add-chain resp: %v", resp)
+		sendHttpError(w, http.StatusInternalServerError, err)
+		return err
+	}
+
+	_, err = w.Write(jsonData)
```
phad: What is it being thrown away in the first part of the return?

examples/ct/ct_handlers_test.go
Martin2112: Yes, somewhat annoyingly the CT Go code returns NonFatalError when you parse a cert that contains our precert extension.

examples/ct/ct_handlers_test.go
Martin2112: Sometimes IntelliJ inserts them in the wrong place and they don't get fixed up. I've moved it back.

examples/ct/ct_handlers_test.go
Martin2112: It's a container for a bunch of certs. It does store them in the order added so can be used to represent a chain in tests but there's no requirement that the certs added have an relationship. It's a specialized version of CertPool, from the CT Go code so I retained the name.

examples/ct/ct_handlers_test.go
Martin2112: Done.

examples/ct/ct_handlers.go
Martin2112: The number of bytes written. The docs say writers must return an error if they write less than the data supplied.

examples/ct/ct_handlers_test.go
phad: I see. So, t.Fatal below occurs if (a) there was an err and (b) it wasn't a x509.NonFatalErrors.

So, since you parsed TestCertPEM (which is not a pre-cert) you definitely don't expect an x509.NonFatalErrors err to be returned.

However if TestCertPEM turned out to be a pre-cert the test would continue.

I think what I'm saying is that you'd want t.Fatal called if it turned out that TestCertPEM _was_ a precert.

In other words, the !ok part is superfluous.

examples/ct/ct_handlers_test.go
phad: OK, that's fine, still curious about the name though.  Does it implement pooling behaviour (re-use of objects for example)?

Martin2112: It originated in golang crypto/x509 but they didn't document why they picked it. I don't think it's ideal because it doesn't do what you suggest.

examples/ct/ct_handlers_test.go
Martin2112: You're right, well spotted. Probably cut and pasted from the precert test. Fixed.

Martin2112:
I think I've addressed all your comments now but it's really hard to be sure with this review UI.

phad:
LGTM for this PR.

I think you might consider tightening up checking of err returned from loading precerts in all the tests.

If I've understood correctly, you expect the X509.NonFatalError return whenever you load a precert, but the tests seem willing to continue running if a nil err is returned.

So code that says
err = LoadCertificateFromPEM(kPreCert)
_, ok := err.(X509.NonFatalErrors)
if err != nil && !ok {  t.Fatal () }

could be tightened up to:
err = LoadCertificateFromPEM(kPreCert)
_, ok := err.(X509.NonFatalErrors)
if err == nil || !ok {  t.Fatal () }

OTOH you might take the view that LoadCertificateFromPEM(kPreCert) ought to just return nil anyway.

Martin2112:
Thanks. As discussed I don't think that error should be returned by CT for something we expect to happen. I'll put a TODO in this file to revisit the tests if this behaviour is fixed upstream.

PR #88, opened 2016-08-02
Add some test helpers to example CT package

Martin2112:
These will be used in testing upcoming handler code.


Martin2112:
Seems that I was wrong about needing these. I'm going to close for now and reopen if necessary.

PR #89, opened 2016-08-04
Add the ct-get-entries handler

Martin2112:
Implements the handler for get-entries and adds tests for the input range validation, backend RPC failures, backend out of spec responses like dropped leaves from range and corrupt merkle leaves.

### DIFF
--- a/examples/ct/ct_handlers.go
+++ b/examples/ct/ct_handlers.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/golang/glog"
@@ -37,6 +38,12 @@ const (
 	jsonMapKeyCertificates string = "certificates"
 	// Logging level for debug verbose logs
 	logVerboseLevel glog.Level = 2
+	// Max number of entries we allow in a get-entries request
+	maxGetEntriesAllowed int64 = 50
+	// The name of the get-entries start parameter
+	getEntriesParamStart = "start"
+	// The name of the get-entries end parameter
+	getEntriesParamEnd = "end"
 )
 
 // CTRequestHandlers provides HTTP handler functions for CT V1 as defined in RFC 6962
@@ -79,6 +86,17 @@ type addChainResponse struct {
 	Timestamp  uint64 `json:timestamp`
 	Extensions string `json:extensions`
 	Signature  string `json:signature`
+}
+
+// getEntriesEntry is a struct that represents one element in a get-entries response
+type getEntriesEntry struct {
+	LeafInput []byte `json:leaf_input`
+	ExtraData []byte `json:extra_data`
+}
+
+// getEntriesResponse is a struct for marshalling get-entries respsonses. See RFC6962 Section 4.6
+type getEntriesResponse struct {
+	Entries []getEntriesEntry `json:entries`
 }
 
 func parseBodyAsJSONChain(w http.ResponseWriter, r *http.Request) (addChainRequest, error) {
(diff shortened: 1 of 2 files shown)


Martin2112:
Welcome back from your holiday Al!

examples/ct/ct_handlers.go
```diff
@@ -436,3 +522,86 @@ func marshalAndWriteAddChainResponse(sct ct.SignedCertificateTimestamp, km crypt
...
+	endIndex, err := strconv.ParseInt(r.FormValue(getEntriesParamEnd), 10, 64)
+
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return validateStartAndEnd(startIndex, endIndex, maxAllowedRange)
+}
+
+// validateStartAndEnd applies validation to the range params for get-entries. Returns nil if
```
AlCutter: Comment needs updating about the return values?

examples/ct/ct_handlers.go
```diff
@@ -436,3 +522,86 @@ func marshalAndWriteAddChainResponse(sct ct.SignedCertificateTimestamp, km crypt
...
+		return 0, 0, fmt.Errorf("start (%d) and end (%d) parameters must be >= 0", start, end)
+	}
+
+	if start > end {
+		return 0, 0, fmt.Errorf("start (%d) and end (%d) is not a valid range", start, end)
+	}
+
+	numEntries := end - start + 1
+
+	if numEntries > maxRange {
```
AlCutter: You probably know this already, but the RFC lets the server just truncate the requested range to one it's happy to serve. Don't know whether you want to support that, but might be worth popping a comment in if not?

examples/ct/ct_handlers.go
```diff
@@ -436,3 +522,86 @@ func marshalAndWriteAddChainResponse(sct ct.SignedCertificateTimestamp, km crypt
...
+
+// marshalGetEntriesResponse does the conversion from the backend response to the one we need for
+// an RFC compliant JSON response to the client.
+func marshalGetEntriesResponse(rpcResponse *trillian.GetLeavesByIndexResponse) (getEntriesResponse, error) {
+	jsonResponse := getEntriesResponse{}
+
+	for _, leaf := range rpcResponse.Leaves {
+		// We're only deserializing it to ensure it's valid, don't need the result
+		if _, err := ct.ReadMerkleTreeLeaf(bytes.NewBuffer(leaf.LeafData)); err != nil {
+			return getEntriesResponse{}, errors.New("Failed to deserialize merkle leaf from backend")
```
AlCutter: /Failed/failed/

examples/ct/ct_handlers.go
```diff
@@ -436,3 +522,86 @@ func marshalAndWriteAddChainResponse(sct ct.SignedCertificateTimestamp, km crypt
...
+	return nil
+}
+
+// marshalGetEntriesResponse does the conversion from the backend response to the one we need for
+// an RFC compliant JSON response to the client.
+func marshalGetEntriesResponse(rpcResponse *trillian.GetLeavesByIndexResponse) (getEntriesResponse, error) {
+	jsonResponse := getEntriesResponse{}
+
+	for _, leaf := range rpcResponse.Leaves {
+		// We're only deserializing it to ensure it's valid, don't need the result
```
AlCutter: It might be more useful to just return the data we got - if a particular leaf entry genuinely was stored in an unparseable fashion then nobody will be able to verify the tree's root hash if we refuse to return that data too. (modulo "it's a toy/example FE" etc. etc.)

examples/ct/ct_handlers.go
```diff
@@ -436,3 +522,86 @@ func marshalAndWriteAddChainResponse(sct ct.SignedCertificateTimestamp, km crypt
...
+		if _, err := ct.ReadMerkleTreeLeaf(bytes.NewBuffer(leaf.LeafData)); err != nil {
+			return getEntriesResponse{}, errors.New("Failed to deserialize merkle leaf from backend")
+		}
+
+		entry := getEntriesEntry{LeafInput: leaf.LeafData, ExtraData: leaf.ExtraData}
```
AlCutter: could in-line this below?

examples/ct/ct_handlers.go
Martin2112: The FE only rejects ranges that don't make sense like (-10, -8) or (20, 5) or are larger than we're willing to serve. I'll let the backend decide how to handle ranges that exceed the tree size. There are comments about this in the handler.

examples/ct/ct_handlers.go
Martin2112: Hmmm. You could read the RFC as not allowing to return something that's not a valid merkle leaf structure, which is what I implemented. I can do either, not sure what the right option is.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
Martin2112: I think I want to leave this check in for now so it can be hooked up with monitoring later as this probably indicates data corruption. I can make it return the bad data if you think that's better.

examples/ct/ct_handlers.go

examples/ct/ct_handlers.go
Martin2112: I've changed it to pass the data through anyway and updated the tests. Added a TODO to hook up monitoring for this.

PR #90, opened 2016-08-05
Implements the get-sth handler for CT example

Martin2112:
Adds the get-sth handler and adds a set of tests for several possible backend failures including receiving an invalid tree size or root hash.

### DIFF
--- a/examples/ct/ct_handlers.go
+++ b/examples/ct/ct_handlers.go
@@ -100,7 +100,7 @@ type getEntriesResponse struct {
 }
 
 // getSthResponse is a struct for marshalling get-sth responses. See RFC 6962 Section 4.3
-type getSthResponse struct {
+type getSTHResponse struct {
 	TreeSize        int64  `json:"tree_size"`
 	TimestampMillis int64  `json:"timestamp"`
 	RootHash        []byte `json:"sha256_root_hash"`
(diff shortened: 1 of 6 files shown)


examples/ct/ct_handlers.go
```diff
@@ -81,6 +81,14 @@ type addChainResponse struct {
 	Signature  string `json:signature`
 }
 
+// getSthResponse is a struct for marshalling get-sth responses. See RFC 6962 Section 4.3
+type getSthResponse struct {
```
AlCutter: getSTHResponse

examples/ct/structures.go
```diff
@@ -121,6 +121,38 @@ func SignV1SCTForPrecertificate(km crypto.KeyManager, cert *x509.Certificate, t
 	return serializeAndSignSCT(km, leaf, sctInput, t)
 }
 
+// SignV1TreeHead signs a tree head for CT. The input STH should have been built from a
+// backend response and already checked for validity.
+func SignV1TreeHead(km crypto.KeyManager, sth *ct.SignedTreeHead) error {
```
AlCutter: Should this stuff go into a different file?

AlCutter: Also, does it need to be exported?

Martin2112:
Rebased and conflicts fixed. Haven't done rework yet.

examples/ct/ct_handlers.go

examples/ct/structures.go
AlCutter: Just thinking that since this is a toy example it might be actively dangerous for someone to reuse bits of this to build a 'real' log?

Martin2112: True but I'm hoping to move it beyond toy status at some point. I think the serialization code is pretty much equivalent to the client side, and does have tests.

How about I pull this out into serialize.go, unexport it for now and we can revisit later?

AlCutter: SGTM

Martin2112: Done.

PR #91, opened 2016-08-05
Remove separate signer. Add code to sequencer to trigger signing

Martin2112:
at intervals if no new work has arrived. Probably needs more tests later but deferring until we've built more infrastructure like supporting read only logs and turning off sequencing / signing. This could be refactored later but I'm expecting we'll need log cleanup / checking tasks so leaving the infrastructure to support that.


log/sequencer_test.go
```diff
@@ -18,6 +18,16 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+// Long duration to prevent root signing kicking in for tests where we're only testing
+// sequencing
+const tenYears time.Duration = time.Hour * 24 * 365 * 10
+
+// Func that says root never expires, so we can be sure tests will only sign / sequence when  we
```
AlCutter: spurious extra space between "when" and "we"

log/sequencer.go
```diff
@@ -24,6 +23,10 @@ type Sequencer struct {
 	keyManager crypto.KeyManager
 }
 
+// CurrentRootExpiredFunc examines a signed log root and decides if it has expired with respect
```
AlCutter: I'd imagined a slightly different approach to this (not saying we should do that instead necessarily, but I thought I'd mention it); I was thinking that there would be Trillian STHs, which are produced with every single batch update to the tree (i.e. there's *always* a correct up-to-date signed root for the current state of the tree), and then there are application (i.e. CT) STHs, which are produced by an application "signer" which essentially just grabs the latest Trillian STH at whatever frequency is required, verifies it and then uses the tree hash/size/etc. to create a new application STH at the current time.

Martin2112: Well if we're into the realms of imagination ...

I think this is a good idea but I'd like to defer until I've got a 1st cut at serving proofs done. Otherwise we were going to have to have public / private tree heads to cope with cases where you're not allowed to issue STHs too frequently.

Martin2112: Opened #100.

log/sequencer_test.go
Martin2112: Done.

Issue #93, opened 2016-08-05
Implement backend inclusion proofs


Martin2112:
Fetching correct node set and stuff. Note: for this milestone initially only at tree sizes where we have an STH because the intermediate recalculations are complex. Needs to support by hash and by index, which are fairly similar operations.

Martin2112:
Not closed until get proof by index submitted.

PR #101, opened 2016-08-18
Add inclusion path nodeID calculations

Martin2112:
Based on prototype version but not using big.Int values. Part of #93.

### DIFF
--- a/storage/merkle_path.go
+++ b/storage/merkle_path.go
@@ -30,9 +30,9 @@ func CalcInclusionProofNodeAddresses(treeSize, index int64, maxBitLen int) ([]No
 		} else if sibling == lastNodeAtLevel {
 			// The tree may skip levels because it's not completely filled in. These nodes
 			// don't exist
-			drop := depth - subtreeDepth(treeSize, depth - 1)
+			drop := depth - subtreeDepth(treeSize, depth-1)
 			sibling = sibling << uint(drop)
-			proof = append(proof, NewNodeIDForTreeCoords(int64(depth - drop), sibling, maxBitLen))
+			proof = append(proof, NewNodeIDForTreeCoords(int64(depth-drop), sibling, maxBitLen))
 		}
 
 		node = node >> 1
--- a/storage/merkle_path.go
+++ b/storage/merkle_path.go
@@ -1,0 +1,67 @@
+package storage
+
+// CalcInclusionProofNodeAddresses returns the tree node IDs needed to
+// build an inclusion proof for a specified leaf and tree size. The maxBitLen parameter
+// is copied into all the returned nodeIDs.
+func CalcInclusionProofNodeAddresses(treeSize, index int64, maxBitLen int) []NodeID {
+	var proof []NodeID
+
+	sizeLessOne := treeSize - 1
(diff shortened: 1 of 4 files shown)


Martin2112:
Ping! other PRs depend on this one.

storage/merkle_path.go
```diff
@@ -0,0 +1,73 @@
+package storage
```
AlCutter: Would this fit better in `merkle/`?

storage/merkle_path.go
```diff
@@ -0,0 +1,73 @@
...
+
+// CalcInclusionProofNodeAddresses returns the tree node IDs needed to
+// build an inclusion proof for a specified leaf and tree size. The maxBitLen parameter
+// is copied into all the returned nodeIDs.
+func CalcInclusionProofNodeAddresses(treeSize, index int64, maxBitLen int) ([]NodeID, error) {
+	if index >= treeSize || index < 0 || treeSize < 1 || maxBitLen < 0 {
+		return []NodeID{}, fmt.Errorf("invalid params ts: %d index: %d, bitlen:%d", treeSize, index, maxBitLen)
+	}
+
+	var proof []NodeID
```
AlCutter: You could `make()` with preallocated capacity of the worst-case.

storage/merkle_path.go
Martin2112: NodeID belongs to storage though. I thought of splitting it to return abstract coords and then converting them to NodeID in storage but not sure it's worth it?

storage/merkle_path_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+type calcPathTestData struct {
+	treeSize     int64
+	leafIndex    int64
+	expectedPath []NodeID
+}
+
+// Expected paths built by examination of the example 7 leaf tree in RFC 6962. When comparing
+// with the document remember that our storage node layers are always populated from the bottom up.
```
AlCutter: Could you generate a bunch of test vectors up to some fairly large sizes from known-good implementations too? 7 leaves is a bit on the low-side - I know Ben got a *happy* surprise when working on something with more leaves than that :)

AlCutter: The fact that our storage works bottom-up shouldn't impact the proofs that are generated though, right?

Martin2112: You should get the same proofs, only the internal node coordinates are different.

I've got a todo in the consistency proof calculations to add larger tests. I'm planning to do that later as I've got a lot of stuff piling up in different branches.

storage/merkle_path.go
Martin2112: Done.

storage/merkle_path.go
AlCutter: Yeah, but this is mostly doing merkle related calculations and then representing the output of that as a bunch of `storage.NodeID`s.  The SMT writer code (in also in `merkle/`) has a dep on `storage.NodeID` too, since it needs to calculate NodeIDs to look them up.

Martin2112: Done.

PR #102, opened 2016-08-18
Update RPC API for inclusion proofs.

Martin2112:
Part of #93.


Martin2112:
I had to regenerate the mock client and it's caused a bit of noise.

PR #103, opened 2016-08-18
Bye bye testify assert

AlCutter:


Martin2112:
It does look less readable though.

PR #104, opened 2016-08-18
Bye bye testify mock 1

AlCutter:


crypto/mock_key_manager.go
```diff
@@ -1,48 +1,63 @@
...
+
 package crypto
 
 import (
-	"crypto"
-	"io"
-
-	"github.com/stretchr/testify/mock"
+	crypto "crypto"
+	gomock "github.com/golang/mock/gomock"
```
Martin2112: Missing blank line?

crypto/mock_key_manager.go
```diff
@@ -1,48 +1,63 @@
...
+}
+
+func (_m *MockKeyManager) EXPECT() *_MockKeyManagerRecorder {
+	return _m.recorder
 }
 
-// Mock for crypto.Signer
-type MockSigner struct {
-	mock.Mock
+func (_m *MockKeyManager) Signer() (crypto.Signer, error) {
```
Martin2112: Can we generate with doc comments? Tools are going to complain.

examples/ct/ct_handlers_test.go
```diff
@@ -168,20 +177,23 @@ func TestGetHandlersRejectPost(t *testing.T) {
...
+		}
 	}
-
-	client.AssertExpectations(t)
 }
 
 func TestPostHandlersRejectEmptyJson(t *testing.T) {
-	client := new(trillian.MockTrillianLogClient)
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
```
Martin2112: Is this doing the equivalent checks of AssertExpectations? Is there anything else we need to add?

examples/ct/ct_handlers_test.go
```diff
@@ -144,22 +147,28 @@ func TestPostHandlersOnlyAcceptPost(t *testing.T) {
 
 		// TODO(Martin2112): Remove this test when there are no more handlers to be implemented and
 		// rely on the handlers own tests
-		assert.Equal(t, http.StatusMethodNotAllowed, resp.StatusCode, "Wrong status code for GET to POST handler")
+		if expected, got := http.StatusMethodNotAllowed, resp.StatusCode; expected != got {
```
Martin2112: We could pull this block out as a test helper method ensureStatusCodeIs() or whatever.

examples/ct/ct_handlers_test.go
```diff
@@ -388,30 +435,41 @@ func TestAddChain(t *testing.T) {
...
-	assert.NoError(t, err, "failed to unmarshal json: %v", recorder.Body.Bytes())
+	if err = json.NewDecoder(recorder.Body).Decode(&resp); err != nil {
+		t.Fatalf("failed to unmarshal json: %v, body: %v", err, recorder.Body.Bytes())
+	}
 
-	assert.Equal(t, ct.V1, ct.Version(resp.SctVersion))
-	assert.Equal(t, ctMockLogID, resp.ID)
-	assert.Equal(t, uint64(1469185273000000), resp.Timestamp)
-	assert.Equal(t, "BAEABnNpZ25lZA==", resp.Signature)
+	if got, want := ct.Version(resp.SctVersion), ct.V1; got != want {
```
Martin2112: Compare the readability of these two blocks!

examples/ct/ct_handlers_test.go
```diff
@@ -972,11 +1135,23 @@ func leafProtosForCert(t *testing.T, km crypto.KeyManager, certs []*x509.Certifi
...
+type dlMatcher struct {
+}
+
+func deadlineMatcher() gomock.Matcher {
+	return dlMatcher{}
 }
 
-func deadlineMatcher(other context.Context) bool {
-	deadlineTime, ok := other.Deadline()
+func (d dlMatcher) Matches(x interface{}) bool {
```
Martin2112: Might be possible to get rid of this matcher or improve it with the new mock framework. OTOH I doubt it's worth spending a lot of time on it.

Martin2112:
I think this looks basically OK. I have some questions and suggestions.

crypto/mock_key_manager.go
AlCutter: Yep :)

crypto/mock_key_manager.go
AlCutter: Nope - mockgen doesn't seem to support doing that :/

examples/ct/ct_handlers_test.go
AlCutter: Yup, `.Finish()` does all of that, and it seems to be stricter in some cases too - doing an `EXPECT()...blah().Returns()` will match once only - you need to explicitly add and `.AnyTimes()` (or whatever) otherwise.

examples/ct/ct_handlers_test.go
AlCutter: Good call - I'll do that in a follow up later on, along with standardising all the `got, expected` orderings.

examples/ct/ct_handlers_test.go
AlCutter: Quite.

examples/ct/ct_handlers_test.go
AlCutter: I'll leave it for now on the "like-for-like(-ish)" argument, and we can sort it out in round two?

PR #107, opened 2016-08-22
Fix subtree lookup SQL.

AlCutter:


Martin2112:
Not sure if the code is passing a negated version or not. If so I think that can be changed too.

AlCutter:
Doesn't appear to be doing so, nope.

PR #108, opened 2016-08-22
Fork log server framework as map server outline

AlCutter:
Some of this code might look familiar :)
Should probably see if we can factor it out once the dust has settled.


server/vmap/trillian_map_server.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	// TODO: Delete when implementation done
+	// ErrNotImplemented is returned when an operation is not supported yet
+	ErrNotImplemented = errors.New("Not yet implemented")
+)
+
+// TODO: There is no access control in the server yet and clients could easily modify
+// any tree.
+
+// LogStorageProviderFunc decouples the server from storage implementations
```
Martin2112: Comment mismatch.

server/vmap/trillian_map_server/main.go
```diff
@@ -0,0 +1,127 @@
...
+var mysqlUriFlag = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test",
+	"uri to use with mysql storage")
+var serverPortFlag = flag.Int("port", 8091, "Port to serve log requests on")
+
+// TODO(Martin2112): Single private key doesn't really work for multi tenant and we can't use
+// an HSM interface in this way. Deferring these issues for later.
+var privateKeyFile = flag.String("private_key_file", "", "File containing a PEM encoded private key")
+var privateKeyPassword = flag.String("private_key_password", "", "Password for server private key")
+
+// TODO(Martin2112): Needs a more realistic provider of log storage with some caching
```
Martin2112: Comment still mentions log.

server/vmap/trillian_map_server.go
AlCutter: D'oh, updated.

server/vmap/trillian_map_server/main.go
AlCutter: gnash, updated too.

Martin2112:
OK for now, happy to leave the code sharing until later as there's more important stuff to do.

PR #112, opened 2016-08-24
Use "go generate" to auto build the mocks

AlCutter:


.travis.yml
```diff
@@ -8,6 +8,7 @@ install:
   - go get -d -v -t ./...
 
 script:
+  - go generate github.com/google/trillian/...
```
pphaneuf: Drive-by question: is this the sort of thing that could be used to run protoc?

AlCutter: Yes, and that's coming up next :)

pphaneuf: Nice!

Martin2112:
LGTM if Travis is happy.

PR #113, opened 2016-08-24
Enable auto-generation of protos too.

AlCutter:


Martin2112:
LGTM if travis ever finishes.

PR #114, opened 2016-08-25
Add support for Set/Get map leaves in mysql storage.

AlCutter:


storage/mysql/map_storage.go
```diff
@@ -67,12 +74,54 @@ type mapTX struct {
...
+	if err != nil {
+		return err
+	}
+	defer stmt.Close()
+
+	_, err = stmt.Exec(m.ms.mapID.TreeID, []byte(keyHash), m.writeRevision, flatValue)
+	if err != nil {
+		return err
+	}
+	return nil
```
Martin2112: Could just return err unless you're planning to log it.

storage/mysql/storage_test.go
```diff
@@ -889,6 +889,145 @@ func TestMapRootUpdate(t *testing.T) {
...
+func TestMapSetSameKeyInSameRevisionFails(t *testing.T) {
+	cleanTestDB()
+
+	mapID := createMapID("TestMapSetSameKeyInSameRevisionFails")
+	db := prepareTestMapDB(mapID, t)
+	defer db.Close()
+	s := prepareTestMapStorage(mapID, t)
+
+	keyHash := []byte("A Key Hash")
+	value := trillian.MapLeaf{
```
Martin2112: Could pull this value out as it's used in a couple of places.

storage/mysql/storage_test.go
```diff
@@ -889,6 +889,145 @@ func TestMapRootUpdate(t *testing.T) {
...
+
+	readRev := int64(1)
+
+	{
+		tx := beginMapTx(s, t)
+
+		if err := tx.Set(keyHash, value); err != nil {
+			t.Fatalf("Failed to set %v to %v: %v", keyHash, value, err)
+		}
+		tx.Commit()
```
Martin2112: I thought commit() could fail?

Martin2112:
Looks basically OK, couple of suggestions.

storage/mysql/map_storage.go
AlCutter: done

storage/mysql/storage_test.go
AlCutter: Done

storage/mysql/storage_test.go
AlCutter: Add checks for these new ones, but there are a *lot* of unchecked Commit() called in here, I've dropped a `TODO` in there to go back and sort those out too.

PR #115, opened 2016-08-25
Implements the handler for get-proof-by-hash in CT frontend

### LINKED ISSUE #94: Implement handler for get-proof-by-hash
In the example CT frontend including all the RFC 6962 gubbins.

Martin2112:
Includes tests. Closes #94.


examples/ct/ct_handlers.go
```diff
@@ -321,13 +330,68 @@ func wrappedGetSTHConsistencyHandler(rpcClient trillian.TrillianLogClient) http.
...
+		ctx, _ := context.WithDeadline(context.Background(), getRPCDeadlineTime(c))
+		response, err := c.rpcClient.GetInclusionProofByHash(ctx, &rpcRequest)
+
+		if err != nil || !rpcStatusOK(response.GetStatus()) {
+			sendHttpError(w, http.StatusInternalServerError, fmt.Errorf("RPC failed, possible extra info: %v", err))
+			return
+		}
+
+		// Per RFC 6962 section 4.5 we can't handle a case with duplicate leaves because the API is
+		// defined to return a single proof so it's an error if we don't get exactly one proof
```
AlCutter: In the 6962 domain the right thing to do would be to return the proof to the earlier entry.

Martin2112:
The tree_size param wasn't being handled. That should be fixed now.

examples/ct/ct_handlers.go
Martin2112: I'm also concerned that we could get a large number of results from storage in weird edge cases.

Martin2112: I've clarified the comment and added a TODO to make it do the right thing. Will probably tweak the api for this when I've caught up with everything else.

PR #119, opened 2016-08-26
Implement get inclusion proof by index

Martin2112:
Not directly part of CT api though can be used to implement get-entry-and-proof. Part of #93.


server/trillian_log_server.go
```diff
@@ -89,8 +89,53 @@ func (t *TrillianLogServer) QueueLeaves(ctx context.Context, req *trillian.Queue
 }
 
 // GetInclusionProof obtains the proof of inclusion in the tree for a leaf that has been sequenced.
+// Similar to the get proof by hash handler but one less step as we don't need to look up the index
 func (t *TrillianLogServer) GetInclusionProof(ctx context.Context, req *trillian.GetInclusionProofRequest) (*trillian.GetInclusionProofResponse, error) {
-	return nil, ErrNotImplemented
+	// Reject obviously invalid tree sizes and leaf indices
```
AlCutter: Seems like there's a lot of common code between this and the `...ByHash` funcs, could it be factored out?

Martin2112: Possibly but I need to untangle pending branches first.

Martin2112: Most of the duplication is in the tests. Apart from parameter validation they call the same function to retrieve the proofs.

Martin2112:
I've switched tests to use the parameterized version where appropriate. I think this is better now.

PR #121, opened 2016-08-26
Simple map server + example client

AlCutter:


examples/vmap/trillian_map_client/main.go
```diff
@@ -0,0 +1,63 @@
+package main
+
+import (
+	"flag"
+	"log"
```
Martin2112: glog? I've been using that everywhere.

server/vmap/trillian_map_server.go
```diff
@@ -54,14 +56,153 @@ func (t *TrillianMapServer) getStorageForMap(mapId int64) (storage.MapStorage, e
...
+		return nil, err
+	}
+	rootHash, err := smtWriter.CalculateRoot()
+
+	newRoot := trillian.SignedMapRoot{
+		TimestampNanos: time.Now().UnixNano(),
+		RootHash:       rootHash,
+		MapId:          s.MapID().MapID,
+		MapRevision:    tx.WriteRevision(),
+		Signature:      &trillian.DigitallySigned{},
```
Martin2112: Should signature be empty? TODO?

examples/vmap/trillian_map_client/main.go
AlCutter: Done.

server/vmap/trillian_map_server.go
AlCutter: Yeah, it's early days :)
Added a TODO.

PR #128, opened 2016-08-30
Add order by sequence to get leaves by hash and proof by hash

### LINKED ISSUE #130: Make the CT handler work if the backend sends multiple proofs
Should use the one with the lowest sequence number for the leaf index.

Martin2112:
Update CT handler to request this option and affected tests. Closes #130.

This is not a complete solution for the case where there are 1000s of leaves with the same hash but provides what we need to ensure the CT handler returns the correct result.


storage/mysql/log_storage.go
```diff
@@ -41,6 +41,12 @@ const selectLeavesByHashSql string = `SELECT l.LeafHash,l.TheData,s.SequenceNumb
 		     FROM LeafData l,SequencedLeafData s
 		     WHERE l.LeafHash = s.LeafHash
 		     AND l.LeafHash IN (` + placeholderSql + `) AND l.TreeId = ? AND s.TreeId = l.TreeId`
+// Same as above except with leaves ordered by sequence so we only incur this cost when necessary
+const selectLeavesByHashOrderedBySequenceSql string = `SELECT l.LeafHash,l.TheData,s.SequenceNumber,s.SignedEntryTimestamp
```
AlCutter: Could this be:

 ```go
const selectLeavesByHashOrderedBySequenceSQL string =  selectLeavesByHashSQL + " ORDER BY s.SequenceNumber"
```
?

PR #131, opened 2016-08-31
Add a Go implementation of a Merkle Tree that supports proofs

Martin2112:
Heavily based on the C++ code so has some caveats about leaf indexing and lazy evaluation. Intended for use as a cross check on proofs etc. in tests to better exercise the new code at larger sizes.

Part of #132.

### DIFF
--- a/merkle/merkle_tree.go
+++ b/merkle/merkle_tree.go
@@ -1,0 +1,511 @@
+package merkle
+
+// This is a fairly direct port of the C++ Merkle Tree. It has the same API and should have
+// similar performance.
+//
+// Note: this implementation evaluates the root lazily in the same way as the C++ code so
+// some methods that appear to be accessors can cause mutations to update the structure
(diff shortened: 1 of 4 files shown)


merkle/merkle_tree.go
```diff
@@ -0,0 +1,519 @@
...
+// TreeEntryDescriptor wraps a node and is used to describe tree paths, which are useful to have
+// access to when testing the code and examining how it works
+type TreeEntryDescriptor struct {
+	Value  TreeEntry
+	XCoord int // The horizontal node coordinate
+	YCoord int // The vertical node coordinate
+}
+
+// MerkleTree holds a Merkle Tree as a 2D node array
+type MerkleTree struct {
```
AlCutter: Maybe call it `InMemoryMerkleTree` so nobody uses it by mistake?

Martin2112: Done.

PR #138, opened 2016-09-01
Don't store internal subtree nodes.

AlCutter:
Instead, store just the subtree leaves and re-calculate the internal nodes on-the-fly as the subtrees are loaded from storage.

This reduces the amount of disk-space required for trees by about 50%, and seems to speed up operations by about 20%.


storage/cache/subtree_cache.go
```diff
@@ -170,3 +192,53 @@ func (s *SubtreeCache) Flush(setSubtree SetSubtreeFunc) error {
 	}
 	return nil
 }
+
+// makeSuffix creates a suffix key for indexing into the subtree's Leaves and
+// InternalNodes maps.
+func makeSuffix(depth int, index *big.Int) ([]byte, error) {
```
Martin2112: makeSuffixKey?

storage/cache/subtree_cache.go
```diff
@@ -170,3 +192,53 @@ func (s *SubtreeCache) Flush(setSubtree SetSubtreeFunc) error {
...
+	return sfx, nil
+}
+
+// populateSubtreeNodes re-creates the subtree's InternalNodes from the subtree
+// Leaves map.
+func (s *SubtreeCache) populateSubtreeNodes(st *storage.SubtreeProto) error {
+	st.InternalNodes = make(map[string][]byte)
+	rootID := storage.NewNodeIDFromHash(st.Prefix)
+	fullTreeDepth := s.treeHasher.Size() * 8
+	leaves := make([]merkle.HStar2LeafHash, 0, len(st.Leaves))
```
Martin2112: This use of HStar2 isn't map specific is it?

storage/cache/subtree_cache.go
```diff
@@ -170,3 +192,53 @@ func (s *SubtreeCache) Flush(setSubtree SetSubtreeFunc) error {
 	}
 	return nil
 }
+
+// makeSuffix creates a suffix key for indexing into the subtree's Leaves and
+// InternalNodes maps.
+func makeSuffix(depth int, index *big.Int) ([]byte, error) {
+	// TODO(al): only supports 8 bit subtree sizes currently
+	sfx := make([]byte, 2)
+	if depth >= 8 {
```
Martin2112: could tidy up the error message to "... depths up to 8 ..." as it currently reads like it must be 8.

storage/cache/subtree_cache.go
AlCutter: Done

storage/cache/subtree_cache.go
AlCutter: Rats, yes; I've made it so you can plug in different functions to reconstitute appropriately and make funcs for Logs & Maps.

storage/cache/subtree_cache.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	return nil
+}
+
+// makeSuffix creates a suffix key for indexing into the subtree's Leaves and
+// InternalNodes maps.
+func makeSuffixKey(depth int, index int64) (string, error) {
+	// TODO(al): only supports 8 bit subtree sizes currently
+	if depth >= 8 || depth < 0 {
+		return "", fmt.Errorf("found depth of %d, but we only support positive depths of 8 currently", depth)
```
Martin2112: error message still reads like 8 is only supported value. Suggest replacing "of" with "up to"

Martin2112:
It looks more or less OK but I don't yet have enough integration tests for the log to prove it works properly.

storage/cache/subtree_cache.go
AlCutter: done

storage/cache/subtree_cache.go
AlCutter: done

merkle/compact_merkle_tree.go
```diff
@@ -224,3 +224,10 @@ func (c *CompactMerkleTree) AddLeafHash(leafHash trillian.Hash, f setNodeFunc) (
 func (c CompactMerkleTree) Size() int64 {
 	return c.size
 }
+
+func (c CompactMerkleTree) Depth() int {
+	if c.size == 0 {
+		return 0
+	}
+	return bitLen(c.size - 1)
+}
```
taknira: So we're using the more traditional meaning of 'Depth' here then? :P

storage/cache/subtree_cache.go
```diff
@@ -27,6 +29,8 @@ type SubtreeCache struct {
 	dirtyPrefixes map[string]bool
 	// mutex guards access to the maps above.
 	mutex *sync.RWMutex
+
+	populateSubtree storage.PopulateSubtreeFunc
 }
 
 // Suffix represents the tail of of a NodeID, indexing into the Subtree which
```
taknira: In looking up what a Suffix is, I noticed a duplicacted 'of' here.  s/of of/of

storage/cache/subtree_cache.go
```diff
@@ -51,20 +55,24 @@ const (
...
+		populateSubtree: populateSubtree,
 	}
 }
 
 // splitNodeID breaks a NodeID out into its prefix and suffix parts.
 // unless ID is 0 bits long, Suffix must always contain at least one bit.
 func splitNodeID(id storage.NodeID) ([]byte, Suffix) {
 	if id.PrefixLenBits == 0 {
-		return []byte{}, Suffix{bits: 0, path: []byte{}}
+		return []byte{}, Suffix{bits: 0, path: []byte{0}}
```
Martin2112: Slightly surprised that an empty path isn't empty but you obviously changed this for a reason.

storage/cache/subtree_cache.go
```diff
@@ -126,7 +139,13 @@ func (s *SubtreeCache) getNodeHashUnderLock(id storage.NodeID, getSubtree GetSub
 
 	// finally look for the particular node within the subtree so we can return
 	// the hash & revision.
-	nh := c.Nodes[sx.serialize()]
+	var nh trillian.Hash
+	// Look up the hash in the appropriate map
+	if sx.bits == 8 {
```
Martin2112: Would be good to add a note on why this works and why the magic number 8.

merkle/compact_merkle_tree.go
AlCutter: `NumLevelsInTheCompactMerkleTree()` seemed a bit on the verbose side :)

storage/cache/subtree_cache.go
AlCutter: done done

storage/cache/subtree_cache.go
AlCutter: done

Martin2112:
Did you push the fix?

Martin2112:
Let's go with this and the soon to be written larger scale tests should help to verify it.

PR #140, opened 2016-09-01
Handle no existing TreeHead

AlCutter:


Martin2112:
Sorry, was probably going to write a dummy tree head when a log is created by the tool that doesn't exist yet.

Martin2112:
Still LGTM.

PR #145, opened 2016-09-02
Fix CompactMerkleTree

AlCutter:


merkle/compact_merkle_tree.go
```diff
@@ -11,12 +11,12 @@ import (
...
 type RootHashMismatchError struct {
-	ExpectedHash []byte
-	ActualHash   []byte
+	ExpectedHash trillian.Hash
+	ActualHash   trillian.Hash
 }
 
 func (r RootHashMismatchError) Error() string {
-	return fmt.Sprintf("root hash mismatch got: %v expected: %v", r.ExpectedHash, r.ActualHash)
+	return fmt.Sprintf("root hash mismatch got: %v expected: %v", r.ActualHash, r.ExpectedHash)
```
pphaneuf: This thing is used exactly once and is even exported (despite never being actually given out in a usable way, since it's hidden by the `error` interface)?!? Why not just use `fmt.Errorf` in that one place?

AlCutter: See `compact_merkle_tree_test.go:131` for how you'd use it.

pphaneuf:
Looks good enough to me. :wink:

PR #146, opened 2016-09-06
Log tree grows upwards

AlCutter:
Fixes a mismatch between the storage layer and the Log merkle calculations.


storage/types.go
```diff
@@ -79,16 +80,39 @@ func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits
...
-	for i := len(r.Path) - 1; index > 0; i-- {
+// index.
+// This method is used exclusively by the Log, and since the Log model grows upwards from the
+// leaves we modify the provided coords accordingly.
+//
+// index is the horizontal index into the tree at level depth, so the returned
+// NodeID will be zero padded on the right by depth places.
+func NewNodeIDForTreeCoords(depth int64, index int64, maxTreeDepth int) (NodeID, error) {
+	bl := bitLen(index)
+	if depth < 0 || bl > int(maxTreeDepth-int(depth)) {
```
taknira: Have I understood correctly:   Does depth start at 0 from the root of the tree, and index start from 0 at each level of the tree, and go along to 2^depth -1?  If so, I'm not convinced by the bl > int(maxTreeDepth - int(depth)) condition here - suppose the tree had 2^64 leaves, and you're trying to get the NodeID for the very last leaf - i.e. at index 2^64-1 on at depth 64 (double check the numbers because they could be wrong :P)  This would give bl = 64 which is > 64 - 64 = 0... would that not be a valid depth/index combo?

New to this code so please bear with me if I'm not understanding correctly!

taknira: Also, there doesn't appear to be any tests for causing this error :P

storage/types.go
```diff
@@ -79,16 +80,39 @@ func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits
...
+		return NodeID{}, fmt.Errorf("depth/index combination out of range: depth=%d index=%d", depth, index)
+	}
+	// this node is effectively a prefix of the subtree undernead (for non-leaf
```
taknira: s/this/This
s/undernead/underneath

storage/types.go
```diff
@@ -79,16 +80,39 @@ func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits
...
+}
+
 // NewNodeIDForTreeCoords creates a new NodeID for a Tree node with a specified depth and
-// index
-func NewNodeIDForTreeCoords(depth int64, index int64, maxLenBits int) NodeID {
-	r := NewEmptyNodeID(maxLenBits)
-	for i := len(r.Path) - 1; index > 0; i-- {
+// index.
+// This method is used exclusively by the Log, and since the Log model grows upwards from the
+// leaves we modify the provided coords accordingly.
```
taknira: Add a comma after 'leaves' - I spent longer than I care to admit parsing this before realising it's not the leaves we're modifying...  If you want to be super picky about grammar you could also add a comma after 'and'.

testonly/nodes.go
```diff
@@ -0,0 +1,13 @@
+package testonly
+
+import (
+	"github.com/google/trillian/storage"
+)
+
+func MustCreateNodeIDForTreeCoords(depth, index int64, maxBitLen int) storage.NodeID {
```
taknira: Small consistency thing - in definition of NewNodeIDForTreeCoords the third param is called 'maxTreeDepth', maybe use that here too?

taknira:
Mostly comment tweaks, and one understanding check.

storage/types.go
AlCutter: Yeah, this is a bit confusing :)
So: 
- there can't be `2^64` leaves in here because we use `int64` (mainly so that we *could* easily detect arithmetic truncation - though I suspect we need to add a bunch more checks for that), so we'd have a bad day once we got to `2^63-1`.
- in the 'Trillian log' world, depth does start at 0 from the leaves, so `index` at `depth=63` can only be 1 bit (because it's either the left or right child of the root), whereas a leaf at `depth=0, index=0`, would really have a `nodeID` of `0x0000000000000000` (assuming a 64 level tree)

the check you mention is trying to enforce this 2nd point - I've added a couple more test cases to hopefully illustrate a bit better.

I've also bumped `log.maxTreeDepth` to 63 :)

storage/types.go
AlCutter: Done

storage/types.go
AlCutter: Eats shoots & leafdata.
Done.

testonly/nodes.go
AlCutter: I changed `NewNodeIDForTreeCoords` to use `maxBitLen`.

taknira: Lies.  You changed it to maxPathBits :P

storage/types.go
taknira: In that case, clarify what the parameter depth represents in the function comment?  Pretty please?

AlCutter: Done!

testonly/nodes.go
AlCutter: *cough*, changed this one to match

storage/types.go
taknira: Ah you've done it.  Thanks :)

PR #148, opened 2016-09-08
Don't store individual node revision info in the subtree

AlCutter:
At the moment storing the node revisions at a more granular level than the subtree doesn't really help with anything and just consumes storage space, so stop storing them for now.


mock_log_client.go
```diff
@@ -4,9 +4,9 @@
 package trillian
 
 import (
-	"github.com/golang/mock/gomock"
-	"golang.org/x/net/context"
-	"google.golang.org/grpc"
+	gomock "github.com/golang/mock/gomock"
+	context "golang.org/x/net/context"
+	grpc "google.golang.org/grpc"
```
taknira: Why have you renamed these imports?  The go readability review comments doc has a section on imports that states you should 'Avoid renaming imports except to avoid a name collision'.

mock_log_client.go
```diff
@@ -46,6 +46,22 @@ func (_mr *_MockTrillianLogClientRecorder) GetConsistencyProof(arg0, arg1 interf
 	return _mr.mock.ctrl.RecordCall(_mr.mock, "GetConsistencyProof", _s...)
 }
 
+func (_m *MockTrillianLogClient) GetEntryAndProof(_param0 context.Context, _param1 *GetEntryAndProofRequest, _param2 ...grpc.CallOption) (*GetEntryAndProofResponse, error) {
```
taknira: Out of interest, why did you rename the parameters from ctx, in and opts?

taknira:
LGTM except for the random renaming changes in mock_log_client.go, which seem unrelated to the purpose of this PR?

mock_log_client.go
AlCutter: These mocks are auto generated by `mockgen` (I don't know why it decided to do this, other than perhaps being a newer version of the tool?)

mock_log_client.go
AlCutter: It wisnae me!

PR #151, opened 2016-09-12
Log sequencer writes leaf hashes

AlCutter:


Martin2112:
Need to fix whatever test should have broken with this change.

Issue #135, opened 2016-08-31
Basic CT handler monitoring


Martin2112:
Including at least:
- operation counts
- HTTP status count by operation
- backend returning invalid data

Martin2112:
Closing as we now export metrics from CTFE.

PR #153, opened 2016-09-13
Added an HTTP request count / status monitor via expvar

Martin2112:
Exploring an option for HTTP monitoring. Wraps the handlers so they'll publish to expvar. Part of #135.


Martin2112:
Going to close this for the moment as I don't think the requirements have been defined well enough yet.

PR #165, opened 2016-09-15
Fix log server connection handling

Martin2112:
Pull the storage caching up into the server so both the RPC and sequencing modules can share the same connections. Prevent sequencing creating endless connections.

### DIFF
--- a/server/log/main.go
+++ b/server/log/main.go
@@ -53,14 +53,14 @@ func getStorageForLog(logId int64) (storage.LogStorage, error) {
 	if !ok {
 		glog.Infof("Creating new storage for log: %d", logId)
 
-		s, err := simpleMySqlStorageProvider(logId)
+		var err error
+		s, err = simpleMySqlStorageProvider(logId)
 
 		if err != nil {
 			return s, err
 		}
 
 		storageMap[logId] = s
-		return s, nil
 	}
 
 	return s, nil
--- a/server/log/main.go
+++ b/server/log/main.go
@@ -60,6 +60,7 @@ func getStorageForLog(logId int64) (storage.LogStorage, error) {
 		}
 
 		storageMap[logId] = s
+		return s, nil
 	}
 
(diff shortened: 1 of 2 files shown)


server/log/main.go
```diff
@@ -32,12 +33,38 @@ var batchSizeFlag = flag.Int("batch_size", 50, "Max number of leaves to process
...
 }
 
+// TODO(Martin2112): Could pull this out as a wrapper so it can be used elsewhere
+func getStorageForLog(logId int64) (storage.LogStorage, error) {
+	storageMapGuard.Lock()
+	defer storageMapGuard.Unlock()
+
+	s, ok := storageMap[logId]
+
+	if ok {
```
AlCutter: Flip the sense of this so error/no-so-good flow is indented?

server/log/main.go
```diff
@@ -90,6 +117,8 @@ func main() {
 
 	glog.Info("**** Log Server Starting ****")
 
+	storageMap = make(map[int64]storage.LogStorage)
```
AlCutter: Could move this up to the var line at top?
```go
var storageMap = make(map[int64]storage.LogStorage)
```

server/log/main.go
```diff
@@ -32,12 +33,39 @@ var batchSizeFlag = flag.Int("batch_size", 50, "Max number of leaves to process
...
+		glog.Infof("Creating new storage for log: %d", logId)
+
+		s, err := simpleMySqlStorageProvider(logId)
+
+		if err != nil {
+			return s, err
+		}
+
+		storageMap[logId] = s
+		return s, nil
```
AlCutter: Can ding this return and fall-through, though you might need to have an explicit
```go
var err error
```
and change L51 to use `=` rather than `:=`

Martin2112: Done. Curse those damn idiomats.

PR #170, opened 2016-09-16
Bulk test client for log / RPC server 

Martin2112:
Client to queue leaves and read back / verify log data. Cross checks STH against a separate tree.


Martin2112:
Will be deleting this branch as the code is being migrated to an integration test.

PR #171, opened 2016-09-20
Log server signs leaves before passing them to storage

Martin2112:
Plus necessary fixups and test updates.  Subject to changing our minds on design but I wanted to at least get it working.

### DIFF
--- a/server/log/main.go
+++ b/server/log/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


Martin2112:
OK the strategic design committee (aka Al and me) think this signature is not adding any value so I'm going to abandon this PR and remove it from LogLeaf instead.

PR #172, opened 2016-09-21
Combine subtree writes into a single statement

AlCutter:
This allows the storage implementation to optimise multiple writes where
possible.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -17,7 +17,7 @@ import (
 type GetSubtreeFunc func(id storage.NodeID) (*storage.SubtreeProto, error)
 
 // SetSubtreeFunc describes a function which can store a Subtree into storage.
-type SetSubtreeFunc func(s *storage.SubtreeProto) error
+type SetSubtreesFunc func(s []*storage.SubtreeProto) error
 
 // SubtreeCache provides a caching access to Subtree storage.
 type SubtreeCache struct {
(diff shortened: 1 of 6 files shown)


storage/cache/subtree_cache.go
```diff
@@ -189,10 +189,11 @@ func (s *SubtreeCache) SetNodeHash(id storage.NodeID, h trillian.Hash, getSubtre
 }
 
 // Flush causes the cache to write all dirty Subtrees back to storage.
-func (s *SubtreeCache) Flush(setSubtree SetSubtreeFunc) error {
+func (s *SubtreeCache) Flush(setSubtrees SetSubtreesFunc) error {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
 
+	treesToWrite := make([]*storage.SubtreeProto, 0, len(s.dirtyPrefixes))
```
Martin2112: Seems possible that we might have a large number of dirty prefixes? I don't know if this is a real problem

storage/mysql/log_storage.go
```diff
@@ -95,19 +95,19 @@ func NewLogStorage(id trillian.LogID, dbURL string) (storage.LogStorage, error)
...
 
 func (m *mySQLLogStorage) getLeavesByIndexStmt(num int) (*sql.Stmt, error) {
-	return m.getStmt(selectLeavesByIndexSql, num)
+	return m.getStmt(selectLeavesByIndexSql, num, "?", "?")
 }
 
 func (m *mySQLLogStorage) getLeavesByHashStmt(num int, orderBySequence bool) (*sql.Stmt, error) {
 	if orderBySequence {
-		return m.getStmt(selectLeavesByHashOrderedBySequenceSQL, num)
+		return m.getStmt(selectLeavesByHashOrderedBySequenceSQL, num, "?", "?")
```
Martin2112: Might be better to make the code that handles expanding placeholders more general. I think this will do for the moment though.

storage/cache/subtree_cache.go
AlCutter: It's possible, I've added a `TODO` about batching the writes in the mysql impl.

storage/mysql/log_storage.go
AlCutter: Ack

PR #173, opened 2016-09-21
Mapper doesn't write for empty domains

AlCutter:


AlCutter:
Added a quick test for the tweak.

PR #176, opened 2016-09-30
Move test hashes used in compact_merkle_tree_test.go into testonly/

phad:
This is so they can be re-used in other tests.

### DIFF
--- a/testonly/compact_merkle_tree.go
+++ b/testonly/compact_merkle_tree.go
@@ -4,14 +4,21 @@ import (
 	"github.com/google/trillian"
 )
 
-func GetInputs() []trillian.Hash {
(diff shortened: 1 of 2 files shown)


testonly/compact_merkle_tree.go
```diff
@@ -0,0 +1,30 @@
+package testonly
+
+import (
+	"github.com/google/trillian"
+)
+
+func GetInputs() []trillian.Hash {
```
Martin2112: Might be better to give these and the functions below more specific names now for when they're used by other packages.

phad: Yeah, good call.  I'll add some documentation too.

PR #178, opened 2016-09-30
Add map integration test

AlCutter:


integration/map_integration_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func getClient() (*grpc.ClientConn, trillian.TrillianMapClient, error) {
+	conn, err := grpc.Dial(*server, grpc.WithInsecure())
+	if err != nil {
+		return nil, nil, err
+	}
+	return conn, trillian.NewTrillianMapClient(conn), nil
+}
+
+func TestMapIntegration1(t *testing.T) {
```
Martin2112: Don't think you need the "1" here.

.travis.yml
```diff
@@ -12,6 +12,7 @@ script:
   - go generate -run="genmock" -x github.com/google/trillian/...
   - go build -v ./...
   - go test -v ./...
+  - ./integration/map_integration_test.sh
```
Martin2112: It would be good to have a base for log / map Integration tests in Go. Maybe for later.

integration/map_integration_test.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Starting Map sever on port ${PORT}"
+
+go build ./server/vmap/trillian_map_server/
+./trillian_map_server --private_key_password=towel --private_key_file=${TESTDATA}/trillian-map-server-key.pem --port ${PORT} &
+sleep 2
+
+echo "Starting integration test"
+
+cd ${INTEGRATION_DIR}
+go test -tags=integration ./ --map_id ${TEST_TREE_ID} --server="localhost:${PORT}"
```
Martin2112: If the test hangs presumably the shell script does too.  Possibly worth adding a --timeout here.

.travis.yml
AlCutter: I've added an `integration_test.sh` which currently just sets some common ENV and calls the `map_integration_test.sh` script.

integration/map_integration_test.go
AlCutter: It was for "future expansion" :)
I've dinged it until I need it.

integration/map_integration_test.sh
AlCutter: good call, added

integration/map_integration_test.sh
```diff
@@ -0,0 +1,31 @@
...
+popd > /dev/null
+sleep 2
+
+echo "Starting Map integration test"
+
+cd ${INTEGRATION_DIR}
+go test -tags=integration --timeout=5m ./ --map_id ${TEST_TREE_ID} --server="localhost:${PORT}"
+
+echo "Stopping Map server on port ${PORT}"
+echo "--- PASS: Map Integration Test"
```
Martin2112: Having "PASS" there looks weird. Shouldn't it check the status of the go test?

AlCutter: The plan was that the script was `set -e` so it'd bail if there was a problem and this line wouldn't execute.
Perhaps that's a bit obscure though; I'll add an explicit check.

PR #179, opened 2016-09-30
Require Go 1.7.1

AlCutter:


README.md
```diff
@@ -7,6 +7,10 @@ white paper, which in turn could be thought of as an extension and
 generalisation of the ideas which underpin
 [Certificate Transparency](https://certificate-transparency.org).
 
+## Requirements
+
+You must have Go 1.7.1 installed.
```
phad: Is it worth explaining why?

AlCutter: Something has changed between `1.6.2` and `1.7.1` that affects the mysql driver somehow.
There's an integration test in another PR which works fine with Go 1.7.1, but fails with Go 1.6.2.

PR #182, opened 2016-10-03
Fix ordering of got,want in failure outputs for some tests.

phad:
Also switch from Fatalf to Errorf to allow testing to continue.


Martin2112:
Al was keen to use Fatalf() instead of Errorf(),  maybe check with him?

AlCutter:
For a lot of the tests in here (and particularly anything around the crypto/merkle functionality) if one of early tests fail then you pretty much know for sure that the rest will, so I'm not sure you get much (in this particular test) by swithing to `Errorf`, but I'm not super bothered either way :)

PR #183, opened 2016-10-04
Hetrogeneous tree stratification

AlCutter:
Allows the storage to use differently sized subtrees depending on their depth in the tree.
Mostly useful for the sparse map, where it's quite unlikely that entries below 80-bits of prefix will have (m)any neighbours in their subtrees.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -19,13 +19,14 @@ type GetSubtreeFunc func(id storage.NodeID) (*storage.SubtreeProto, error)
 // SetSubtreeFunc describes a function which can store a Subtree into storage.
 type SetSubtreesFunc func(s []*storage.SubtreeProto) error
 
-// strataInfo represents a single statum across the tree.
+// stratumInfo represents a single stratum across the tree.
 // It it used inside the SubtreeCache to determine which Subtree prefix should
 // be used for a given NodeID.
-type strataInfo struct {
+// Currently, the strata must have depths which are multiples of 8.
+type stratumInfo struct {
 	// prefixBytes is the number of prefix bytes above this stratum.
 	prefixBytes int
-	// depth is the depth of this stratum.
+	// depth is the number of levels in this stratum.
 	depth int
 }
 
@@ -33,7 +34,7 @@ type strataInfo struct {
 type SubtreeCache struct {
 	// prefixLengths contains the strata prefix sizes for each multiple-of-8 tree
 	// size.
-	strataInfo []strataInfo
+	stratumInfo []stratumInfo
 	// subtrees contains the Subtree data read from storage, and is updated by
 	// calls to SetNodeHash.
 	subtrees map[string]*storage.SubtreeProto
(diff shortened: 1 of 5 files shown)


storage/cache/subtree_cache.go
```diff
@@ -19,8 +19,21 @@ type GetSubtreeFunc func(id storage.NodeID) (*storage.SubtreeProto, error)
 // SetSubtreeFunc describes a function which can store a Subtree into storage.
 type SetSubtreesFunc func(s []*storage.SubtreeProto) error
 
+// strataInfo represents a single statum across the tree.
```
Martin2112: typo and possibly should be called stratumInfo.

storage/cache/subtree_cache.go
```diff
@@ -19,8 +19,21 @@ type GetSubtreeFunc func(id storage.NodeID) (*storage.SubtreeProto, error)
...
+// It it used inside the SubtreeCache to determine which Subtree prefix should
+// be used for a given NodeID.
+type strataInfo struct {
+	// prefixBytes is the number of prefix bytes above this stratum.
+	prefixBytes int
+	// depth is the depth of this stratum.
```
Martin2112: is this from the root or from the leaves?

storage/cache/subtree_cache.go
```diff
@@ -19,8 +19,21 @@ type GetSubtreeFunc func(id storage.NodeID) (*storage.SubtreeProto, error)
 // SetSubtreeFunc describes a function which can store a Subtree into storage.
 type SetSubtreesFunc func(s []*storage.SubtreeProto) error
 
+// strataInfo represents a single statum across the tree.
+// It it used inside the SubtreeCache to determine which Subtree prefix should
+// be used for a given NodeID.
+type strataInfo struct {
+	// prefixBytes is the number of prefix bytes above this stratum.
```
Martin2112: using bytes incurs the multiple of 8 limitation? If so it's worth commenting.

storage/cache/subtree_cache.go
```diff
@@ -39,52 +52,75 @@ type Suffix struct {
...
 // TODO(al): consider supporting different sized subtrees - for now everything's subtrees of 8 levels.
-func NewSubtreeCache(populateSubtree storage.PopulateSubtreeFunc) SubtreeCache {
+func NewSubtreeCache(strata []int, populateSubtree storage.PopulateSubtreeFunc) SubtreeCache {
+	// TODO(al): pass this in
+	maxTreeDepth := 256
+	// Precalculate strata information based on the passed in strata depths:
+	sInfo := make([]strataInfo, 0, maxTreeDepth/8)
+	t := 0
+	for _, s := range strata {
+		if s%8 != 0 {
```
Martin2112: this looks like it can go wrong if the caller passes in duplicate / inconsistent strata entries

storage/cache/subtree_cache.go
```diff
@@ -141,13 +177,14 @@ func (s *SubtreeCache) getNodeHashUnderLock(id storage.NodeID, getSubtree GetSub
...
 		}
+		sInfo := s.strataInfoForPrefixLength(subID.PrefixLenBits)
 		if c == nil {
 			// storage didn't have one for us, so we'll store an empty proto here
 			// incase we try to update it later on (we won't flush it back to
 			// storage unless it's been written to.)
 			c = &storage.SubtreeProto{
 				Prefix:        px,
-				Depth:         strataDepth,
+				Depth:         int32(sInfo.depth),
```
Martin2112: would it help to define this as int32 in the struct?

storage/cache/subtree_cache.go
AlCutter: done

storage/cache/subtree_cache.go
AlCutter: Neither, it's the number of levels in the stratum - I've updated the comment since it wasn't clear.

storage/cache/subtree_cache.go
AlCutter: Not hugely - then it'll need more casts elsewhere :/

storage/cache/subtree_cache.go
Martin2112: I meant that you could pass in duplicate entries with the same depth and different prefixes. I think this goes wrong but might not be worth defending against.

storage/cache/subtree_cache.go
AlCutter: Added a note about that

PR #185, opened 2016-10-04
Add map inclusion verifier

AlCutter:


merkle/map_verifier.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/storage"
+)
+
+// VerifyMapInclusionProof verifies that the passed in expectedRoot can be
+// reconstructed correctly given the other parameters.
+//
+// The process is essentially the same as the inclusion proof checking for
+// append-only logs, but with added support for "null" nodes as well as extra
```
Martin2112: this last bit reads strangely and should probably be a separate sentence. You could even spoil the ending by revealing what the assumptions are if you like!

merkle/map_verifier.go
```diff
@@ -0,0 +1,61 @@
...
+// assumptions can be made due to the fixed tree size.
+//
+// Returns nil on a successful verification, and an error otherwise.
+func VerifyMapInclusionProof(keyHash trillian.Hash, leafHash trillian.Hash, expectedRoot trillian.Hash, proof []trillian.Hash, h MapHasher) error {
+	hBits := h.Size() * 8
```
Martin2112: The magic number 8 is somewhat proliferating.

merkle/map_verifier_test.go
```diff
@@ -0,0 +1,316 @@
...
+			testonly.MustDecodeBase64("YmUpJx/UagsoBYv6PnFRaVYw3x6kAx3N3OOSyiXsGtg="),
+			testonly.MustDecodeBase64("CtC2GCsc3/zFn1DNkoUThUnn7k+DMotaNXvmceKIL4Y="),
+		},
+		testonly.MustDecodeBase64("U6ANU1en3BSbbnWqhV2nTGtQ+scBlaZf9kRPEEDZsHM="),
+	}, {
+		[]byte("key-1-848"),
+		[]byte("value-1-848"),
+		[]trillian.Hash{
+			// 246 x nil
+			nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil,
```
Martin2112: not sure if it would be better to generate these than having masses of 'nils'.  You could have e.g. a list of (index, base64).

AlCutter: I thought about that, but I figured it slightly complicates the testing code in order to re-constitute the proofs so I decided against it and shoved these vectors down under the rug^W^W^Wto the bottom of the file to keep it pretty.
I can change it if you feel strongly about it?

Martin2112: I don't think it really matters in test code.

PR #186, opened 2016-10-05
Fix bug that GetSequencedLeafCount was returning a count of the whole table

Martin2112:
Make it tree specific as API intended, add a test, which seemed to be missing.


storage/mysql/log_storage.go
```diff
@@ -27,7 +27,7 @@ const insertUnsequencedEntrySql string = `INSERT INTO Unsequenced(TreeId,LeafHas
      VALUES(?,?,?,?)`
 const insertSequencedLeafSql string = `INSERT INTO SequencedLeafData(TreeId,LeafHash,SequenceNumber)
 		 VALUES(?,?,?)`
-const selectSequencedLeafCountSql string = "SELECT COUNT(*) FROM SequencedLeafData"
+const selectSequencedLeafCountSql string = "SELECT COUNT(*) FROM SequencedLeafData WHERE TreeId=?"
```
AlCutter: heehee, whoops :)

PR #189, opened 2016-10-07
Fix nodeIDFromAddress for zero-prefixed indices.

AlCutter:


merkle/sparse_merkle_tree_test.go
```diff
@@ -571,6 +571,11 @@ func DISABLEDTestSparseMerkleTreeWriterBigBatch(t *testing.T) {
 	maybeProfileMemory(t)
 }
 
+func bigIntFromString(dec string) *big.Int {
```
Martin2112: this might as well panic on bad input like our other test decode helpers

merkle/sparse_merkle_tree.go
```diff
@@ -192,7 +191,7 @@ func nodeIDFromAddress(size int, prefix []byte, index *big.Int, depth int) stora
 	if depth > len(ib)*8 {
 		copy(t[len(prefix)+depthBytes-len(ib):], ib)
 	} else {
-		copy(t[len(prefix):], ib)
+		copy(t[size-len(ib):], ib)
```
Martin2112: Is there an easy explanation of what these two branches mean / do that could be comments?

merkle/sparse_merkle_tree_test.go
AlCutter: done

PR #193, opened 2016-10-07
Map integration test does inclusion verification

AlCutter:


Martin2112:
LGTM subject to the fix in the other branch fixing the build.

PR #203, opened 2016-11-02
Fix further hash confusion between raw and merkle hash

Martin2112:
The LeafData must have the raw SHA256 hash as we use it for deduping. The sequenced data needs both raw and merkle hashes as it must both link to the LeafData and be queried by Merkle Hash by clients. I've renamed database columns to make it clearer which is which.


storage/mysql/log_storage.go
```diff
@@ -18,32 +18,34 @@ import (
 
 const getTreePropertiesSQL string = "SELECT AllowsDuplicateLeaves FROM Trees WHERE TreeId=?"
 const getTreeParametersSQL string = "SELECT ReadOnlyRequests From TreeControl WHERE TreeID=?"
-const selectQueuedLeavesSQL string = `SELECT LeafHash,Payload
+const selectQueuedLeavesSQL string = `SELECT LeafRawHash,Payload
```
AlCutter: `DataHash`?

storage/mysql/log_storage.go
```diff
@@ -18,32 +18,34 @@ import (
...
-     VALUES(?,?,?,?)`
-const insertSequencedLeafSQL string = `INSERT INTO SequencedLeafData(TreeId,LeafHash,SequenceNumber)
+		 ORDER BY QueueTimestamp DESC,LeafRawHash ASC LIMIT ?`
+const insertUnsequencedLeafSQL string = `INSERT INTO LeafData(TreeId,LeafRawHash,TheData)
+		 VALUES(?,?,?) ON DUPLICATE KEY UPDATE LeafRawHash=LeafRawHash`
+const insertUnsequencedLeafSQLNoDuplicates string = `INSERT INTO LeafData(TreeId,LeafRawHash,TheData)
 		 VALUES(?,?,?)`
+const insertUnsequencedEntrySQL string = `INSERT INTO Unsequenced(TreeId,LeafRawHash,MessageId,Payload)
+     VALUES(?,?,?,?)`
+const insertSequencedLeafSQL string = `INSERT INTO SequencedLeafData(TreeId,LeafRawHash,LeafHash,SequenceNumber)
```
AlCutter: Would renaming `LeafHash` to `MerkleLeafHash` help with clarity too?

storage/mysql/log_storage.go
```diff
@@ -274,12 +276,23 @@ func (t *logTX) QueueLeaves(leaves []trillian.LogLeaf) error {
...
+		insertSQL = insertUnsequencedLeafSQLNoDuplicates
+	}
+
 	for _, leaf := range leaves {
 		// Create the unsequenced leaf data entry. We don't use INSERT IGNORE because this
 		// can suppress errors unrelated to key collisions. We don't use REPLACE because
 		// if there's ever a hash collision it will do the wrong thing and it also
 		// causes a DELETE / INSERT, which is undesirable.
-		_, err := t.tx.Exec(insertUnsequencedLeafSQL, t.ls.logID.TreeID,
+		_, err := t.tx.Exec(insertSQL, t.ls.logID.TreeID,
```
AlCutter: Might be a thing for another PR, but should this be using a prepared stmt?

storage/mysql/log_storage.go
```diff
@@ -474,7 +487,11 @@ func (t *logTX) UpdateSequencedLeaves(leaves []trillian.LogLeaf) error {
 			return errors.New("Sequenced leaf has incorrect hash size")
 		}
 
-		_, err := t.tx.Exec(insertSequencedLeafSQL, t.ls.logID.TreeID, []byte(leaf.LeafHash),
+		// Recompute the raw leaf hash, could be passed around in the leaf structure but
```
AlCutter: Would it make sense to add an extra field to hold it?

storage/mysql/log_storage.go
Martin2112: Not sure which is best. We use "leaf" a lot in the code but not sure we're always consistent.

storage/mysql/log_storage.go
Martin2112: It might. I can do that.

storage/mysql/log_storage.go
Martin2112: If we're going to do it for potential efficiency gains we should probably look at the other transaction level prepared statements. I don't think that belongs in this PR.

storage/mysql/log_storage.go
AlCutter: or `TheDataHash`, just to make it clear that it's a hash of `TheData` (similar relative naming could happen in the `LeafData` struct.

storage/mysql/log_storage.go
Martin2112: It does but we're again getting into slightly weird territory because the hash is part of the Leaf type, not LogLeaf.

Martin2112: Working on this now. I'm going to take the hit now and rename everything in Leaf / Logleaf end to end so it's always consistent.

storage/mysql/log_storage.go
AlCutter: I think `Leaf`is only used by the Log...
Anyway, don't worry about it for now as I think we're going to have to move those around a bit.
Pop a TODO in?

Martin2112:
OK I think this is all the renaming except for splitting off LeafValueHash, which can be done separately.

storage/mysql/log_storage.go
Martin2112: TODO added.

Martin2112:
PTAL.

PR #205, opened 2016-11-04
Add a Log integration test

Martin2112:
Queues leaves, reads them back and verifies the data is correct. Then compares the root hash with a separate compact tree, and a second tree implementation.


integration/log_integration_test.go
```diff
@@ -0,0 +1,333 @@
...
+			delete(leafDataPresenceMap, string(response.Leaves[l].LeafValue))
+
+			hash := hasher.HashLeaf(response.Leaves[l].LeafValue)
+
+			if !bytes.Equal(hash[:], response.Leaves[l].MerkleLeafHash) {
+				return nil, fmt.Errorf("leaf hash mismatch expected: %s, got: %s", base64.StdEncoding.EncodeToString(hash[:]), base64.StdEncoding.EncodeToString(response.Leaves[l].MerkleLeafHash))
+			}
+		}
+
+		currentLeaf += int64(params.readBatchSize)
```
AlCutter: should this be `len(response.Leaves)` in case the log decides to return fewer entries than requested?

integration/log_integration_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// By this point we expect to have seen all the leaves so there should be nothing in the map
+	if len(leafDataPresenceMap) != 0 {
+		return nil, fmt.Errorf("missing leaves from data read back: %v", leafDataPresenceMap)
+	}
+
+	return leafMap, nil
+}
+
+func checkLogSTHConsistency(logID trillian.LogID, tree *merkle.InMemoryMerkleTree, client trillian.TrillianLogClient, params testParameters) error {
```
AlCutter: I'd rename this method, if I were skimming I'd assume it's fetching and checking a consistency proof.

integration/log_integration_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// We use the leafMap as we need to use the same order for the memory tree to get the same hash.
+	for l := params.startLeaf; l < params.leafCount; l++ {
+		compactTree.AddLeaf(leafMap[l].LeafValue, func(depth int, index int64, hash trillian.Hash) {})
+		merkleTree.AddLeaf(leafMap[l].LeafValue)
+	}
+
+	// If the two reference results disagree there's no point in continuing the checks. This is a
+	// "can't happen" situation.
+	if bytes.Compare(compactTree.CurrentRoot(), merkleTree.CurrentRoot().Hash()) != 0 {
```
AlCutter: `!bytes.Equal(...)`

integration/log_integration_test.go
```diff
@@ -0,0 +1,333 @@
...
+		glog.Infof("Queueing %d leaves to log server ...", params.leafCount)
+		if err := queueLeaves(treeID, client, params); err != nil {
+			t.Fatalf("Failed to queue leaves: %v", err)
+		}
+	}
+
+	// Step 2 - Wait for queue to drain when server sequences, give up if it doesn't happen (optional)
+	if *awaitSequencingFlag {
+		glog.Infof("Waiting for log to sequence ...")
+		if err = waitForSequencing(treeID, client, params); err != nil {
```
Martin2112: Done.

integration/log_integration_test.go
Martin2112: Yes, done. I was assuming this wouldn't happen.

integration/log_integration_test.go
Martin2112: Done.

Martin2112:
PTYAL. I'll address the leaf value hash thing in a separare PR.

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+func checkLogRootHashMatches(logID int64, tree *merkle.InMemoryMerkleTree, client trillian.TrillianLogClient, params testParameters) error {
+	// Check the STH against the hash we got from our tree
+	resp, err := getLatestSignedLogRoot(client, logID)
+
+	if err != nil {
+		return err
+	}
+
+	// Hash must not be empty and must match the one we built ourselves
+	if len(resp.SignedLogRoot.RootHash) == 0 || !bytes.Equal(tree.CurrentRoot().Hash(), resp.SignedLogRoot.RootHash) {
```
AlCutter: Maybe break this out into two separate checks, one for length and the other for equality?

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+		return fmt.Errorf("root hash mismatch expected: %s, got: %s", base64.StdEncoding.EncodeToString(tree.CurrentRoot().Hash()), base64.StdEncoding.EncodeToString(resp.SignedLogRoot.RootHash))
```
AlCutter: could do this to cut the size down a bit:

```go
if got, want := base64.StdEncoding.EncodeToString(resp.SignedLogRoot.RootHash), base64.StdEncoding.EncodeToString(tree.CurrentRoot().Hash()); got != want {
return fmt,Errorf(...)
}
```

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	return nil
+}
+
+func makeQueueLeavesRequest(logID int64, leaves []trillian.LogLeaf) trillian.QueueLeavesRequest {
+	leafProtos := make([]*trillian.LogLeaf, 0, len(leaves))
+
+	for l := 0; l < len(leaves); l++ {
```
AlCutter: this could be a `range()` loop, I think?

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// We use the leafMap as we need to use the same order for the memory tree to get the same hash.
+	for l := params.startLeaf; l < params.leafCount; l++ {
+		compactTree.AddLeaf(leafMap[l].LeafValue, func(depth int, index int64, hash []byte) {})
+		merkleTree.AddLeaf(leafMap[l].LeafValue)
+	}
+
+	// If the two reference results disagree there's no point in continuing the checks. This is a
+	// "can't happen" situation.
+	if !bytes.Equal(compactTree.CurrentRoot(), merkleTree.CurrentRoot().Hash()) {
```
AlCutter: `if got, want := ...; !bytes.Equal(got, want) { ... }`

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+			if !ok {
+				return nil, fmt.Errorf("leaf data duplicated for leaf: %v", response.Leaves[l])
+			}
+
+			delete(leafDataPresenceMap, string(response.Leaves[l].LeafValue))
+
+			hash := hasher.HashLeaf(response.Leaves[l].LeafValue)
+
+			if !bytes.Equal(hash[:], response.Leaves[l].MerkleLeafHash) {
```
AlCutter: `if got, want := ...` and coule make the b64 encoding part of the got/want values

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+			req := makeQueueLeavesRequest(treeID, leaves)
+			ctx, cancelFunc := getRPCDeadlineContext()
+			response, err := client.QueueLeaves(ctx, &req)
+			cancelFunc()
+
+			if err != nil {
+				return err
+			}
+
+			if response.Status == nil || response.Status.StatusCode != trillian.TrillianApiStatusCode_OK {
```
AlCutter: `if got := response.Status; got != nil || got.StatusCode... {}`

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+		leafDataPresenceMap[fmt.Sprintf("Leaf %d", l + params.startLeaf)] = true
+	}
+
+	for currentLeaf < params.leafCount {
+		hasher := merkle.NewRFC6962TreeHasher(crypto.NewSHA256())
+
+		// We have to allow for the last batch potentially being a short one
+		numLeaves := params.leafCount - currentLeaf
+
+		if numLeaves > int64(params.readBatchSize) {
```
AlCutter: You could add a `bs := int64(params.readBatchSize)` up the top to save having to cast twice here, or, alternatively, make readBatchSizeFlag an int64?

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+		req := makeGetLeavesByIndexRequest(logID, currentLeaf + params.startLeaf, numLeaves)
+		ctx, cancelFunc := getRPCDeadlineContext()
+		response, err := client.GetLeavesByIndex(ctx, req)
+		cancelFunc()
+
+		if err != nil {
+			return nil, err
+		}
+
+		if response.Status == nil || response.Status.StatusCode != trillian.TrillianApiStatusCode_OK {
```
AlCutter: `if got := response.Status; got != nil || ... {}`

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+			return nil, fmt.Errorf("read leaves failed: %s %d", response.Status.Description, response.Status.StatusCode)
+		}
+
+		// Check we got the right leaf count
+		if int64(len(response.Leaves)) == 0 {
```
AlCutter: Do you need that int64 cast?
Also, should this be something like:
```go
if got, want := len(response.Leaves), numLeaves; got != want { ... }`
```
(based off the comment) ?

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+			return nil, fmt.Errorf("expected %d leaves log returned none", numLeaves)
+		}
+
+		// Check the leaf contents make sense. Can't rely on exact ordering as queue timestamps will be
+		// close between batches and identical within batches.
+		for l := 0; l < len(response.Leaves); l++ {
+			// Check for duplicate leaf index in response data - should not happen
+			if _, ok := leafMap[response.Leaves[l].LeafIndex]; ok {
```
AlCutter: `response.Leaves[l].LeafIndex` is used quite a lot in this loop, could assign it to a variable up here to reduce code, or better, use a `range()` style loop to do it for you.

integration/log_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+				return nil, fmt.Errorf("got duplicate leaf sequence number: %d", response.Leaves[l].LeafIndex)
+			}
+
+			leafMap[response.Leaves[l].LeafIndex] = response.Leaves[l]
+
+			// Test for having seen duplicate leaf data - it should all be distinct
+			_, ok := leafDataPresenceMap[string(response.Leaves[l].LeafValue)]
```
AlCutter: could pull the `string(response.Leaves[l].LeafValue)` out as a var since it's used below too?

integration/log_integration_test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Start the log server, and set an exit trap to ensure we kill it once we're done:
+pushd ${TRILLIAN_ROOT} > /dev/null
+go build ./server/trillian_log_server/
+./trillian_log_server --private_key_password=towel --private_key_file=${TESTDATA}/trillian-server-key.pem --port ${PORT} --signer_sleep_between_runs="1s" --sequencer_sleep_between_runs="1s" --batch_size=100 &
+trap "kill -INT %1" EXIT
+popd > /dev/null
+sleep 2
+
+
```
AlCutter: extra blank line

integration/log_integration_test.go
Martin2112: Done.

integration/log_integration_test.go
Martin2112: Done. Something something ... convention ... hobgoblins.

integration/log_integration_test.go
Martin2112: Weird cast removed. No, the log can return fewer leaves than requested. Returning none is definitely wrong.

integration/log_integration_test.go
Martin2112: OK I pulled out leaf to address both this and the previous grumble.

integration/log_integration_test.go
Martin2112: Done, not convinced it's better.

integration/log_integration_test.go
Martin2112: I think I can remove the length check and rely on the tree never producing a null hash. It's a different implementation to the one used by the log.

integration/log_integration_test.go
Martin2112: Done but again unconvinced.

integration/log_integration_test.go
Martin2112: It's a fair cop.

integration/log_integration_test.go
AlCutter: `got, ref` ?

PR #207, opened 2016-11-04
Kill off trillian.Leaf and trillian.LogLeaf, and use the proto structs instead.

AlCutter:


server/trillian_log_server.go
```diff
@@ -444,30 +438,22 @@ func (t *TrillianLogServer) commitAndLog(tx storage.LogTX, op string) error {
...
 }
 
-// TODO: Fill in the log leaf specific fields when we've implemented signed timestamps
-func leafToProto(leaf trillian.LogLeaf) *trillian.LeafProto {
-	return &trillian.LeafProto{LeafIndex: leaf.SequenceNumber, MerkleLeafHash: leaf.MerkleLeafHash, LeafValue: leaf.LeafValue, ExtraData: leaf.ExtraData}
-}
-
-func leavesToProtos(leaves []trillian.LogLeaf) []*trillian.LeafProto {
-	protos := make([]*trillian.LeafProto, 0, len(leaves))
+func pointerify(leaves []trillian.LogLeaf) []*trillian.LogLeaf {
```
Martin2112: I'd suggest renaming these but I can't think of anything better.

PR #226, opened 2016-11-16
Fix that leaves were being dequeued in reverse order. Ooops.

Martin2112:
And add a test that checks for the correct ordering.

### DIFF
--- a/storage/mysql/storage_test.go
+++ b/storage/mysql/storage_test.go
@@ -603,7 +603,7 @@ func TestDequeueLeavesTimeOrdering(t *testing.T) {
 		ensureAllLeafHashesDistinct(dequeue1, t)
 
 		// Ensure this is the second batch queued by comparing leaf data.
-		if !leafInRange(dequeue1[0], batchSize, batchSize + batchSize - 1) || !leafInRange(dequeue1[1], batchSize, batchSize + batchSize - 1) {
+		if !leafInRange(dequeue1[0], batchSize, batchSize+batchSize-1) || !leafInRange(dequeue1[1], batchSize, batchSize+batchSize-1) {
 			t.Fatalf("Got leaf from wrong batch (1st dequeue): (%s %s)", string(dequeue1[0].LeafValue), string(dequeue1[1].LeafValue))
 		}
 
@@ -625,7 +625,7 @@ func TestDequeueLeavesTimeOrdering(t *testing.T) {
 		ensureAllLeafHashesDistinct(dequeue2, t)
 
 		// Ensure this is the first batch by comparing leaf data.
-		if !leafInRange(dequeue2[0], 0, batchSize - 1) || !leafInRange(dequeue2[1], 0, batchSize - 1) {
+		if !leafInRange(dequeue2[0], 0, batchSize-1) || !leafInRange(dequeue2[1], 0, batchSize-1) {
 			t.Fatalf("Got leaf from wrong batch (2nd dequeue): (%s %s)", string(dequeue2[0].LeafValue), string(dequeue2[1].LeafValue))
 		}
 
@@ -1463,7 +1463,7 @@ func createTestLeaves(n, startSeq int64) []trillian.LogLeaf {
 	hasher := crypto.NewSHA256()
 
 	for l := int64(0); l < n; l++ {
-		lv := fmt.Sprintf("Leaf %d", l + startSeq)
+		lv := fmt.Sprintf("Leaf %d", l+startSeq)
 		leaf := trillian.LogLeaf{
 			LeafValueHash:  hasher.Digest([]byte(lv)),
 			MerkleLeafHash: hasher.Digest([]byte(lv)),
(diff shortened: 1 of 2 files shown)


storage/mysql/storage_test.go
```diff
@@ -552,6 +552,82 @@ func TestDequeueLeavesGuardInterval(t *testing.T) {
...
+
+	logID := createLogID("TestDequeueLeavesTimeOrdering")
+	db := prepareTestLogDB(logID, t)
+	defer db.Close()
+	s := prepareTestLogStorage(logID, t)
+	batchSize := 2
+
+	{
+		tx := beginLogTx(s, t)
+		defer failIfTXStillOpen(t, "TestDequeueLeavesTimeOrdering", tx)
```
AlCutter: `defer`ed funcs only run at the return of the enclosing function, not when the enclosing scope goes away, does that change anything here?

storage/mysql/storage_test.go
```diff
@@ -552,6 +552,82 @@ func TestDequeueLeavesGuardInterval(t *testing.T) {
...
+
+		leaves := createTestLeaves(int64(batchSize), 0)
+		leaves2 := createTestLeaves(int64(batchSize), int64(batchSize))
+
+		if err := tx.QueueLeaves(leaves, fakeQueueTime); err != nil {
+			t.Fatalf("Failed to queue leaves: %v", err)
```
AlCutter: `"QueueLeaves(..., ...) = %v", err`

storage/mysql/storage_test.go
```diff
@@ -552,6 +552,82 @@ func TestDequeueLeavesGuardInterval(t *testing.T) {
...
+		}
+
+		// These are one second earlier so should be dequeued first
+		if err := tx.QueueLeaves(leaves2, fakeQueueTime.Add(-time.Second)); err != nil {
+			t.Fatalf("Failed to queue leaves2: %v", err)
```
AlCutter: `"QueueLeaves(..., ...) = %v", err`

storage/mysql/storage_test.go
```diff
@@ -552,6 +552,82 @@ func TestDequeueLeavesGuardInterval(t *testing.T) {
...
+	}
+
+	{
+		// Now try to dequeue two leaves and we should get the second batch
+		tx2 := beginLogTx(s, t)
+		defer failIfTXStillOpen(t, "TestDequeueLeavesTimeOrdering", tx2)
+		dequeue1, err := tx2.DequeueLeaves(batchSize, fakeQueueTime)
+
+		if err != nil {
+			t.Fatalf("Failed to dequeue leaves (1st): %v", err)
```
AlCutter: error message style (and elsewhere)

storage/mysql/storage_test.go
```diff
@@ -552,6 +552,82 @@ func TestDequeueLeavesGuardInterval(t *testing.T) {
...
+
+		if got, want := len(dequeue1), batchSize; got != want {
+			t.Fatalf("Dequeue count mismatch (1st) got: %d, want: %d", got, want)
+		}
+
+		ensureAllLeafHashesDistinct(dequeue1, t)
+
+		// Ensure this is the second batch queued by comparing leaf data.
+		if !leafInRange(dequeue1[0], batchSize, batchSize + batchSize - 1) || !leafInRange(dequeue1[1], batchSize, batchSize + batchSize - 1) {
+			t.Fatalf("Got leaf from wrong batch (1st dequeue): (%s %s) %v", string(dequeue1[0].LeafValue), string(dequeue1[1].LeafValue), dequeue1)
```
AlCutter: Possibly don't need that last `%v` and corresponding `dequeue1`? (or don't need the individual elements)

storage/mysql/storage_test.go
Martin2112: No, they should all get committed by the end of the test. But I missed the defer for tx3 so I added that one.

storage/mysql/storage_test.go

storage/mysql/storage_test.go

storage/mysql/storage_test.go
Martin2112: They're not error types though.

storage/mysql/storage_test.go
Martin2112: It was originally just %v but the output was hard to read so I pulled out the mismatched data.

storage/mysql/storage_test.go
AlCutter: Sorry, I meant:
   change to "DequeueLeaves(...) = %v ..." style messages (and elsewhere)
but I was too lazy to type it all out again :)

storage/mysql/storage_test.go
AlCutter: `%+v` or `%#v` might be better?

Martin2112: I'll make it just print the mismatched strings. The rest of the leaf contents don't really matter in this context.

Martin2112:
Is this OK yet?

AlCutter:
Mostly, could you run gofmt on it and see what (if anything) it does on the `a+b` vs `a + b` front?

Martin2112:
Ran gofmt again from command line and it did change a+b. Speculative theory is that the intellij go fmt is interacting with the "format changed lines only" editor setting.

PR #240, opened 2016-11-18
Raise timeout for servers starting in integration tests

Martin2112:
and share setting between tests

### DIFF
--- a/integration/log_integration_test.sh
+++ b/integration/log_integration_test.sh
@@ -19,7 +19,7 @@ popd > /dev/null
 
 # Set an exit trap to ensure we kill the RPC server once we're done.
 trap "kill -INT ${RPC_SERVER_PID}" EXIT
-sleep ${STARTUP_WAIT_SECONDS}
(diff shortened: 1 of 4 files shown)


integration/log_integration_test.sh
```diff
@@ -19,7 +19,7 @@ popd > /dev/null
 
 # Set an exit trap to ensure we kill the RPC server once we're done.
 trap "kill -INT ${RPC_SERVER_PID}" EXIT
-sleep 2
+sleep ${STARTUP_WAIT_SECONDS}
```
AlCutter: You could also do:
`wget --retry-connrefused --waitretry=1 -t $(STARTUP_WAIT_SECONDS} localhost:${RPC_PORT}`

So that it continues as soon as the port is opened.

Martin2112:
I'll try that, possibly with a short sleep afterwards in case the port opens before the server is actually ready.

Martin2112:
Done, was slightly fiddly but I think it works.

PR #242, opened 2016-11-18
Adds a GetNodes function to SubtreeCache

AlCutter:
This moves some code out from the mysql storage implementation up into the subtree cache so it can be shared with other storage implementations.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -132,9 +132,9 @@ func (s *SubtreeCache) splitNodeID(id storage.NodeID) ([]byte, Suffix) {
 	return a[:prefixSplit], sfx
 }
 
-// preload calculates the set of subtrees required to cover the passed in node
-// IDs, uses getSubtrees to retrieve them, and finally populates the cache
-// structures with the data.
+// preload calculates the set of subtrees required to know the hashes of the
+// passed in node IDs, uses getSubtrees to retrieve them, and finally populates
+// the cache structures with the data.
 func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc) error {
 	s.mutex.Lock()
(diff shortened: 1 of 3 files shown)


storage/cache/subtree_cache.go
```diff
@@ -129,8 +132,10 @@ func (s *SubtreeCache) splitNodeID(id storage.NodeID) ([]byte, Suffix) {
 	return a[:prefixSplit], sfx
 }
 
-// Preload populates the cache based on a specified set of NodeIDs.
-func (s *SubtreeCache) Preload(ids []storage.NodeID, getSubtrees func(id []storage.NodeID) ([]*storagepb.SubtreeProto, error)) error {
+// preload calculates the set of subtrees required to cover the passed in node
```
Martin2112: By "cover" does that mean minimum set of subtrees that includes the nodes? Extra doc would be useful.

storage/cache/subtree_cache.go
```diff
@@ -163,8 +168,45 @@ func (s *SubtreeCache) Preload(ids []storage.NodeID, getSubtrees func(id []stora
 	return nil
 }
 
-// GetNodeHash retrieves the previously written hash and corresponding tree
-// revision for the given node ID.
+// GetNodes fetches multiple hashes from storage, and returns them an
```
Martin2112: That reads oddly. Possibly missing a word, "returns them in ..." ?

storage/cache/subtree_cache.go
AlCutter: Re-worded it a bit, any better?

Martin2112: did you push the change?

AlCutter: I was fixing the other comment :)
all pushed now.

Martin2112:
OK, reads better.

PR #249, opened 2016-11-29
Rewrite the skip level code in merkle path calculations

Martin2112:
Now is explicitly based on the tree shape at the snapshot size and hopefully easier to read. Should be read in conjunction with ```compact_merkle_tree```, which is where node writes come from. Add a few
more tests, more should be written.

### DIFF
--- a/merkle/merkle_path.go
+++ b/merkle/merkle_path.go
@@ -3,8 +3,12 @@ package merkle
 import (
 	"fmt"
 
+	"github.com/golang/glog"
 	"github.com/google/trillian/storage"
 )
+
+// Verbosity level for logging of debug related items
+const vLevel = 2
 
 // CalcInclusionProofNodeAddresses returns the tree node IDs needed to
 // build an inclusion proof for a specified leaf and tree size. The maxBitLen parameter
@@ -37,10 +41,10 @@ func CalcInclusionProofNodeAddresses(treeSize, index int64, maxBitLen int) ([]st
 			proof = append(proof, n)
 		} else if sibling == lastNodeAtLevel {
 			// The tree may skip levels because it's not completely filled in. These nodes
-			// don't exist
-			drop := depth - subtreeDepth(treeSize, depth-1)
-			sibling = sibling << uint(drop)
-			n, err := storage.NewNodeIDForTreeCoords(int64(depth-drop), sibling, maxBitLen)
+			// don't exist in storage, the value we want is a copy of a node further down
+			// (multiple levels may be skipped).
+			l, sibling := skipMissingLevels(treeSize, lastNodeAtLevel, depth, node)
+			n, err := storage.NewNodeIDForTreeCoords(int64(l), sibling, maxBitLen)
 			if err != nil {
 				return nil, err
 			}
@@ -76,17 +80,21 @@ func CalcConsistencyProofNodeAddresses(previousTreeSize, treeSize int64, maxBitL
 func snapshotConsistency(snapshot1, snapshot2 int64, maxBitLen int) ([]storage.NodeID, error) {
 	proof := make([]storage.NodeID, 0, bitLen(snapshot2)+1)
 
+	glog.V(vLevel).Infof("snapshotConsistency: %d -> %d", snapshot1, snapshot2)
+
 	level := 0
 	node := snapshot1 - 1
 
 	// Compute the (compressed) path to the root of snapshot2.
 	// Everything left of 'node' is equal in both trees; no need to record.
 	for (node & 1) != 0 {
+		glog.V(vLevel).Infof("Move up: l:%d n:%d", level, node)
 		node >>= 1
 		level++
 	}
 
 	if node != 0 {
+		glog.V(vLevel).Infof("Not root snapshot1: %d", node)
 		// Not at the root of snapshot 1, record the node
 		n, err := storage.NewNodeIDForTreeCoords(int64(level), node, maxBitLen)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


merkle/merkle_path.go
```diff
@@ -152,19 +164,59 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
+	return bits&mask != 0
+}
+
+// skipMissingLevels moves down the tree a level towards the leaves until the node exists. This
+// must terminate successfully as we will eventually reach the leaves, which are always written
+// and are at level 0. Missing nodes are intermediate nodes with one child, hence their value
+// is the same as the node lower down the tree as there is nothing to hash it with.
+func skipMissingLevels(snapshot, lastNode int64, level int, node int64) (int, int64) {
+	sibling := node ^ 1
+	for level > 0 && (node&1) == 0 && sibling == lastNode && !lastNodeWritten(int64(level), snapshot) {
```
AlCutter: if `sibling == lastNode` then can `(node&1)` ever be non-zero?

merkle/merkle_path.go
```diff
@@ -37,10 +41,10 @@ func CalcInclusionProofNodeAddresses(treeSize, index int64, maxBitLen int) ([]st
 			proof = append(proof, n)
 		} else if sibling == lastNodeAtLevel {
 			// The tree may skip levels because it's not completely filled in. These nodes
```
Martin2112: I'm not sure that the distinction is that clear cut. The C++ is effectively retaining intermediate nodes with one child, which costs very little as there is only one version stored. Also the output of this calculation is in the correct form for our storage implementation. I'll try to add some text.

merkle/merkle_path.go
```diff
@@ -129,18 +139,20 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
-			if level > 0 {
-				drop = level - subtreeDepth(snapshot, level-1)
-			}
-			sibling = sibling << uint(drop)
-			n, err := storage.NewNodeIDForTreeCoords(int64(level-drop), sibling, maxBitLen)
+			// for missing levels in the tree. This can only occur on the rightmost tree nodes because
+			// this is the only area of the tree that is not fully populated.
+			glog.V(vLevel).Infof("Last: S:%d L:%d", sibling, level)
+
+			// Account for non existent nodes - these can only be the right most node at an
```
Martin2112: Done.

merkle/merkle_path.go
```diff
@@ -152,19 +164,59 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
+//   / \     / \         /
+//   a b     c d        e
+//   | |     | |        |
+//   d0 d1   d2 d3      d4
+//
+// Tree size = 5, last index = 4 in binary = 100, append 1 for leaves = 1001.
+// Reading down the RHS: present, not present, not present, present = 1001. So when
+// attempting to fetch the sibling of k the tree should be descended twice to fetch
+// 'e'.
+func lastNodeWritten(d, ts int64) bool {
```
Martin2112: Done. We still have level v. depth inconsistencies elsewhere I think.

merkle/merkle_path.go
```diff
@@ -152,19 +164,59 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
-		p2++
+	// Last index is size - 1, we shift one to avoid special case (leaves always present)
+	bits := uint64(ts - 1)
+	// Test the bit in the path for the requested level
+	mask := uint64(1) << uint64(d-1)
+
+	return bits&mask != 0
+}
+
+// skipMissingLevels moves down the tree a level towards the leaves until the node exists. This
```
Martin2112: Drat, those are really painful to draw. I'll have a go.

merkle/merkle_path.go
```diff
@@ -152,19 +164,59 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
+// must terminate successfully as we will eventually reach the leaves, which are always written
+// and are at level 0. Missing nodes are intermediate nodes with one child, hence their value
+// is the same as the node lower down the tree as there is nothing to hash it with.
+func skipMissingLevels(snapshot, lastNode int64, level int, node int64) (int, int64) {
```
Martin2112: I left it as node because the caller uses that terminology. I could change all of them if you think it's better.

merkle/merkle_path.go
Martin2112: Hmmm. I don't see how. That was notional protection against it triggering incorrectly but I think it was probably voodoo and the lastNodeWritten() test is already handling this. I'll take it out.

merkle/merkle_path.go
Martin2112: I've tried to reword it.

merkle/merkle_path.go
```diff
@@ -169,38 +170,80 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
 // examining the bits of the last valid leaf index in a tree of the specified size. Zero bits
 // indicate nodes that are not stored at that tree size.
 //
-// Example:
+// Examples, all using a tree of size 5 leaves:
+//
+// As depicted in RFC 6962, nodes "float" upwards.
+//
+//						hash2
```
AlCutter: I think you might have some tabs in here :)

merkle/merkle_path.go
```diff
@@ -169,38 +170,80 @@ func pathFromNodeToRootAtSnapshot(node int64, level int, snapshot int64, maxBitL
...
+//			 / \           |
+//			/   \          e
+//		 /     \         |
+//		g       h       d4
+//	 / \     / \
+//	 a b     c d
+//	 | |     | |
+//	 d0 d1   d2 d3
+//
+// In C++ reference implementation, intermediate nodes are stored, leaves are at level 0.
```
AlCutter: In the C++

merkle/merkle_path.go
AlCutter: ok, we should probably go and make it all consistent at some point soon; this stuff is hard enough to understand as it is :)

AlCutter: my brains keeps reading this as a func which checks that something's been written ok, rather than determining whether a node _should_ exist in the given place.
I'm not sure what an alternative name might be though...

merkle/merkle_path.go
Martin2112: Done.

merkle/merkle_path.go
Martin2112: What about lastNodePresent()? Other ideas I have are quite long.

merkle/merkle_path.go
AlCutter: The docs are great, thanks!

merkle/merkle_path.go
AlCutter: Maybe we should create an issue to sweep through the codebase to make it consistent?

PR #261, opened 2016-12-08
Initial storage docs and diags

Martin2112:
Will add to this as we go but I thought we should at least document stuff like how the subtree storage works because it's not obvious.


docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+### NodeIDs
+
+Nodes in storage are uniquely identified by a `NodeID`. This combines a tree path with
+a revision number. The path is effectively used as a bitwise subtree prefix in the tree. 
+
+The same `NodeID` objects are used by both logs and maps but they are interpreted differently.
+There are API functions that create them for each case. Mixing node ID types in API calls will
+give incorrect results.
+
+### Subtree Compaction
```
AlCutter: /Compaction/Stratification/ to match up with the comments in the code

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+As an optimization the tree is not stored as a set of raw nodes but at as a collection of subtrees.
```
AlCutter: optimization,

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+Currently, subtrees must be be a multiple of 8 levels deep (referred to as `strataDepth` in the
+code). Only the bottom level nodes the "leaves" of the subtree are physically stored.
```
AlCutter: nodes (the "leaves") of the subtree

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+Subtrees are helpful for reads because it is likely that many of the nodes traversed in
+Merkle paths for proofs are part of the same subtree. The number of subtrees involved in a path
+through a large tree from the leaves to the root is also bounded. For writes the subtree update
+batches what would be many smaller writes into one larger but manageable one.
+
+We gain space efficiency by not storing intermediate nodes. This is a big saving as it avoids
+storing entire tree levels, which get very large as the tree grows. This is magnified further as we
+store many versions of the tree.
+
```
AlCutter: This saves effectively half the storage space, but we also lose out when (e.g. in maps) we're storing multiple subtree revisions (with the same prefix) just because one "leaf" differed...

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+This diagram shows a tree as it might actually be stored by our code using subtrees of depth 8.
+
+Each subtree does not include its "root" node, though this counts as part of the depth. There are
+additional subtrees below and to the right of the child subtree shown, they can't easily be shown
+in the diagram. Obviously, there could be less than 256 "leaf" nodes in the subtrees if they are not
+yet fully populated.
+
+![strata depth 8 tree](StratumDepth8.png, "Stratum Depth 8")
+
+As it's hard to visualize the structure at scale with stratum depth 8 some examples of smaller
```
AlCutter: Comma after 8, and maybe start a new sentence at "though" (-> Although)?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+Need stuff here.
+
+## Log Storage
+
+### The Log Tree
+
+The node tree built for a log is a representation of a Merkle Tree, which starts out empty and grows
+as leaves are added. A Merkle Tree of a specific size is a fixed and well-defined shape.
+                     
+Leaves are never removed and a completely filled in left subtree of the tree
```
AlCutter: /filled in/populated/ ?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+Log nodes are notionally addressed using a three dimensional coordinate tuple (level in tree, index
+in level, revision number).
+
+Level zero is always the leaf level and additional intermediate levels are added above this as the
+tree grows. Levels are only created when they are used.
+
+Index is the horizontal position of the node in the level, with zero being the leftmost node in
+each level.
+
+For example in a tree of size two the leaves are level 0 and index 0 and 1 and the root is
```
AlCutter: might be easier to parse if you expand it out: " are [level:0, index:0] and [level:0, index1], and the root is [level: 1, index 0]."

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+### Log Startup
+
+When log storage is intialized and its tree is not empty the existing state is loaded into a
+`compact_merkle_tree`. This can be done efficiently and only requires a few node accesses. As a
+crosscheck the root hash of the compact tree is compared against the current log root. If it does
+not match then log is corrupt and cannot be used.
+
+### Writing Leaves and Sequencing
+
+In the current RDBMS storage implementation log clients queue new leaves to the log, and the
```
AlCutter: /and the/and a/

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+When log storage is intialized and its tree is not empty the existing state is loaded into a
+`compact_merkle_tree`. This can be done efficiently and only requires a few node accesses. As a
+crosscheck the root hash of the compact tree is compared against the current log root. If it does
+not match then log is corrupt and cannot be used.
+
+### Writing Leaves and Sequencing
+
+In the current RDBMS storage implementation log clients queue new leaves to the log, and the 
+`LeafData` record is created. Further writes are coordinated by a single sequencer, which adds
```
AlCutter: /further writes/Writes to the merkle tree/ ?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+When leaves are added to the tree they are processed by a `merkle/compact_merkle_tree`, this causes a
+batched set of tree node updates to be applied. Each update is given its own revision number. The
+result is that a number of tree snapshots are directly available in storage. This contrasts with
+previous implementations where the tree is in RAM and only the most recent snapshot is directly
+available. Note that we may batch log updates so we don't necessarily have all intermediate tree
+snapshots directly available from storage.
+
+As an optimization intermediate nodes with only a left child are not stored. There is more detail
+on how this affects access to tree paths in the file `merkle/merkle_paths.go`. This differes from
```
AlCutter: /differes/differs/

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+Subtrees are keyed by the `NodeID` of the root (effectively a path prefix) and contain the
+intermediate nodes for that subtree, identified by their suffix. These are actually stored in a
+proto map where the key is the suffix path. 
+
+Subtrees are versioned to support the access model for nodes described above. Node addresses within
+the model are distinct because the path to a subtree must be unique.
+
+When node updates are applied they will affect one or more subtrees and caching is used to increase
+efficiency. After all updates have been done in-memory the cache is flushed to storage so each
```
phad: So you flush out all changes in the cached subtree in one write - does that imply that subtree's version number is revved exactly once?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+#### Subtree Diagrams
+
+This diagram shows a tree as it might actually be stored by our code using subtrees of depth 8.
+
+Each subtree does not include its "root" node, though this counts as part of the depth. There are
+additional subtrees below and to the right of the child subtree shown, they can't easily be shown
+in the diagram. Obviously, there could be less than 256 "leaf" nodes in the subtrees if they are not
+yet fully populated.
+
+![strata depth 8 tree](StratumDepth8.png, "Stratum Depth 8")
```
phad: Can you clarify as the diagram omits a lot of nodes; can you add some circles for the layer immediately above the leafs.  I think they're going to be yellow but it's not totally obvious.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+As it's hard to visualize the structure at scale with stratum depth 8 some examples of smaller
+depths might make things clearer, though these are not supported by the code the diagrams are
+much simpler.
+
+This diagram shows a tree with strata depth 2. It is a somewhat special case as all the levels are
```
phad: < latin pedant="on" >  shouldn't you say 'stratum' here as we're talking about a single subtree?

(I'm assuming that you consider a single subtree to occupy a single stratum (alongside its sibling subtrees) but maybe I've got that wrong?)

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+stored. Note that the root node is never stored and is always recalculated.
```
phad: So another surprise here for me now I read this and see the diagrams:  you say that StratumDepth2.png shows a subtree of stratum depth=2, yet there are only the leaf nodes in blue inside the wedge.  So you include the subtree root when counting depth.

So a general property of all these things is that, despite the root not being stored - you say it is always recalculated - it must correspond to one of the 'leaves' of the next stratum (I mean subtree) above and in those it will be stored.

i.e in all the diagrams the blue node at a sub-tree root 'overlays' a yellow node from the parent layer.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+The log stores two hashes per leaf, a raw SHA256 hash of the leaf data used for deduplication and
+the Merkle Leaf Hash of the data, which becomes part of the tree.
+
+### Log NodeIDs / Tree Coordinates
+
+Log nodes are notionally addressed using a three dimensional coordinate tuple (level in tree, index
+in level, revision number).
+
+Level zero is always the leaf level and additional intermediate levels are added above this as the
```
phad: There's a danger of confusion having a terminology that includes: 'level' and 'stratum'.  If I was correct above and 'sub-tree' == 'stratum' then clearly stratum != level, but I'm not totally sure.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+tree grows. Levels are only created when they are used.
+
+Index is the horizontal position of the node in the level, with zero being the leftmost node in
+each level.
+
+For example in a tree of size two the leaves are level 0 and index 0 and 1 and the root is
+level 1, index 0.
+
+The storage implementation must be able to provide access to nodes using this coordinate scheme but
+it is not required to store them this way. The current implementation collapses subtrees for
```
phad: 'collapses subtrees' -- are you referring to the same NodeID-addressed subtrees as you discuss in 'subtree compaction' above?  Can you use consistent verbs?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+increased write efficiency so nodes are not distinct database entities, this is hidden by the node
```
phad: , this is hidden  -> run-on sentence.  replace , with . or :

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+The log stores two hashes per leaf, a raw SHA256 hash of the leaf data used for deduplication and
+the Merkle Leaf Hash of the data, which becomes part of the tree.
+
+### Log NodeIDs / Tree Coordinates
+
+Log nodes are notionally addressed using a three dimensional coordinate tuple (level in tree, index
+in level, revision number).
+
+Level zero is always the leaf level and additional intermediate levels are added above this as the
+tree grows. Levels are only created when they are used.
```
phad: So would I be correct to conclude that as a tree grows, and new (higher level) intermediates are added, that existing nodes migrate from one stratum to new ones, to accommodate the extra level created?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+The storage implementation must be able to provide access to nodes using this coordinate scheme but
+it is not required to store them this way. The current implementation collapses subtrees for
+increased write efficiency so nodes are not distinct database entities, this is hidden by the node
+API.
+
+### Log Startup
+
+When log storage is intialized and its tree is not empty the existing state is loaded into a
+`compact_merkle_tree`. This can be done efficiently and only requires a few node accesses. As a
```
phad: 'a few node accesses' implies that you certainly don't need to read every leaf when calculating the new root hash using the compact merkle tree.

is that because it's possible to 'short circuit' that by adding the sub-tree roots at a particular instead, and have each 'contribute' the sum of all the descendant leaves?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+crosscheck the root hash of the compact tree is compared against the current log root. If it does
+not match then log is corrupt and cannot be used.
+
+### Writing Leaves and Sequencing
+
+In the current RDBMS storage implementation log clients queue new leaves to the log, and the 
+`LeafData` record is created. Further writes are coordinated by a single sequencer, which adds 
+leaves to the tree. The sequencer is responsible for ordering the leaves and creating the 
+`SequencedLeafData` row linking the leaf and its sequence number. If there are duplicate leaves in
+the log then they share `LeafData` rows. Leaves that have not been sequenced are not accessible
```
phad: is it even technically accurate to call these un-sequenced entities 'leaves' at all?  if they're not sequenced then don't have a position in the tree, so they can't be leaves of the tree yet :-P

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+`LeafData` record is created. Further writes are coordinated by a single sequencer, which adds 
+leaves to the tree. The sequencer is responsible for ordering the leaves and creating the 
+`SequencedLeafData` row linking the leaf and its sequence number. If there are duplicate leaves in
+the log then they share `LeafData` rows. Leaves that have not been sequenced are not accessible 
+via the log APIs.
+
+When leaves are added to the tree they are processed by a `merkle/compact_merkle_tree`, this causes a
+batched set of tree node updates to be applied. Each update is given its own revision number. The
+result is that a number of tree snapshots are directly available in storage. This contrasts with
+previous implementations where the tree is in RAM and only the most recent snapshot is directly
```
phad: by 'previous implementations' I take it you refer to previous CT impls and not earlier versions of Trillian.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+available. Note that we may batch log updates so we don't necessarily have all intermediate tree
+snapshots directly available from storage.
+
+As an optimization intermediate nodes with only a left child are not stored. There is more detail
+on how this affects access to tree paths in the file `merkle/merkle_paths.go`. This differes from
+the previous C++ in-memory tree implementation. In summary the code must handle cases where there
```
phad: you'll need some references here as Trillian has no C++ at all.  you're referring to a design in a different project in a different repo!

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+#### Serving Proofs
+
+API requests for proofs involve more work but both inclusion and consistency proofs follow the 
+same pattern.
+
+The node revision number for the tree size involved is obtained from storage and will be used in
+the subsequent node fetch.
+
+The tree path for the proof is calculated for a specified tree size using an algorithm based on the
+reference implementation of RFC 6962. The output of this is an ordered slice of `NodeIDs` that must
```
phad: Is the slice of NodeIDs that specify a particular tree revision?  Is that the same revision as you determine 'for the tree size involved' (above)?  In that case was I wrong to suppose that every sub-tree has it's own separate revision number?

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+concepts involved. The API for this is defined in `storage/tree_storage.go`. Related protos are in
+the `storage/storagepb` package.
+
+The model provides a versioned view of the tree. Revision numbers increase monotonically as the
+tree is mutated.
+
+### NodeIDs
+
+Nodes in storage are uniquely identified by a `NodeID`. This combines a tree path with
+a revision number. The path is effectively used as a bitwise subtree prefix in the tree.
```
AlCutter: yep, that's the fella.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+There are API functions that create them for each case. Mixing node ID types in API calls will
+give incorrect results.
+
+### Subtree Compaction
+
+As an optimization the tree is not stored as a set of raw nodes but at as a collection of subtrees.
+
+Currently, subtrees must be be a multiple of 8 levels deep (referred to as `strataDepth` in the
+code). Only the bottom level nodes the "leaves" of the subtree are physically stored.
+Intermediate subtree nodes are rehashed from the "leaves" when the subtree is loaded into memory.
```
Martin2112: More less. It's subtrees all the way down. The root of a subtree is never included in it though so they don't overlap. I hoped the diagram showed that. Let me know if not.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+Subtrees are keyed by the `NodeID` of the root (effectively a path prefix) and contain the
+intermediate nodes for that subtree, identified by their suffix. These are actually stored in a
+proto map where the key is the suffix path. 
+
+Subtrees are versioned to support the access model for nodes described above. Node addresses within
```
Martin2112: I think I am. Only affected subtrees are updated when nodes are written. For the log case once a subtree is full it's never updated again.

docs/storage/storage.md
Martin2112: Yes. The revision groups all the changes made in the transaction.

docs/storage/storage.md
Martin2112: Yes and no. The root is included in the depth but it's not stored so the subtrees don't overlap and a node always belongs to exactly one subtree.

docs/storage/storage.md
Martin2112: Yes, I'll read through this again. The level here is the actual level within the tree ignoring the subtree scheme.

docs/storage/storage.md
Martin2112: That's just making it too easy for people. Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: On the multiple of 8 comment - yes, I'll add that to the doc.

The code is initialized using a slice of subtree depths so you can mix them if you want.

Martin2112: It's also a trade-off, a larger depth means you save more storage but have to do more rehashing.

docs/storage/storage.md
Martin2112: Expanded the text to include this.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Tree Node Storage
+
+The node level of storage provides a fairly abstract tree model that is used to implement
+verifiable logs and maps. Most users will not need this level but it is important to know the
+concepts involved. The API for this is defined in `storage/tree_storage.go`. Related protos are in
+the `storage/storagepb` package.
+
+The model provides a versioned view of the tree. Revision numbers increase monotonically as the
+tree is mutated.
```
Martin2112: Added Al's comment to the doc.

docs/storage/storage.md
Martin2112: Added a note on this.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+### NodeIDs
+
+Nodes in storage are uniquely identified by a `NodeID`. This combines a tree path with
+a revision number. The path is effectively used as a bitwise subtree prefix in the tree. 
+
+The same `NodeID` objects are used by both logs and maps but they are interpreted differently.
+There are API functions that create them for each case. Mixing node ID types in API calls will
```
Martin2112: It could but we don't currently do that. Users shouldn't need to create node ids. The underlying tree storage is shared between logs and maps and just works with the path bits.

docs/storage/storage.md
```diff
@@ -0,0 +1,186 @@
...
+### Subtree Compaction
+
+As an optimization the tree is not stored as a set of raw nodes but at as a collection of subtrees.
+
+Currently, subtrees must be be a multiple of 8 levels deep (referred to as `strataDepth` in the
+code). Only the bottom level nodes the "leaves" of the subtree are physically stored.
+Intermediate subtree nodes are rehashed from the "leaves" when the subtree is loaded into memory.
+
+Subtrees are keyed by the `NodeID` of the root (effectively a path prefix) and contain the
+intermediate nodes for that subtree, identified by their suffix. These are actually stored in a
```
Martin2112: I think that got answered by a previous change?

docs/storage/storage.md
Martin2112: Added a note to make this explicit.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: Tried to clarify and added a note that level is global tree level, not involved in any subtree optimization.

docs/storage/storage.md
Martin2112: Not sure what you mean by "existing nodes" migrate. The existing tree revision is unaffected. New nodes are added at a new revision. Added a clarification.

docs/storage/storage.md
Martin2112: Yes, changed to "compacts".

docs/storage/storage.md
Martin2112: Done.

docs/storage/storage.md
Martin2112: It just needs to reinitialize a compact merkle tree. The intermediate nodes all exist in storage. I think it's log2 on the tree depth but can't find a statement to that effect in the Go or C++ code.

docs/storage/storage.md
Martin2112: It's arguable because they have LeafData rows to allow for deduping. I think you are technically correct and I've reworded it.

docs/storage/storage.md
Martin2112: I've put in a link to CT repo. This also glosses over the signer / frontend use of full / compact trees but I don't think that's important for readers of this doc.

docs/storage/storage.md
Martin2112: Fair point. Added link to CT code.

docs/storage/storage.md
Martin2112: It's latest version up to X, where X is the revision number of the tree size chosen.

Martin2112:
I think I addressed all the comments. Let me know if there are any I missed.

Martin2112:
Ping!

arielchapa (changes requested):
Ayuda

PR #266, opened 2016-12-13
Remove the code for signing log roots if the last one is too old.

Martin2112:
This is leftover from when we thought app layers wouldn't have their own storage. I don't think it's being used and probably doesn't do what we need anyway.


Martin2112:
Not so fast! This prevents a size zero root being created if none exist, which breaks the integration test. Needs more thought on how this should work. Should maybe pass a "no root" error back to the caller.

PR #269, opened 2016-12-14
Use ReadOnlyTX where possible in the LogRPCServer

AlCutter:


server/log_rpc_server_test.go
```diff
@@ -1549,19 +1549,24 @@ type makeRPCFunc func(*TrillianLogRPCServer) error
...
 	ctrl      *gomock.Controller
 	operation string
+	readonly  bool
 	prepareTx prepareMockTXFunc
 	makeRPC   makeRPCFunc
 }
 
-func newParameterizedTest(ctrl *gomock.Controller, operation string, prepareTx prepareMockTXFunc, makeRPC makeRPCFunc) *parameterizedTest {
-	return &parameterizedTest{ctrl, operation, prepareTx, makeRPC}
+func newParameterizedTest(ctrl *gomock.Controller, operation string, readonly bool, prepareTx prepareMockTXFunc, makeRPC makeRPCFunc) *parameterizedTest {
```
codingllama: Suggestion: use an enum instead of a boolean, so it's more readable.

type TransactionMode int
const (
    ReadOnly TransactionMode = iota
    ReadWrite
)

func newParameterizedTest(..., mode TransactionMode, ...)

PR #273, opened 2016-12-16
Return an empty list of node hashes when the compact merkle tree size is a power of 2

### LINKED ISSUE #200: merkle.CompactMerkleTree.Hashes() doesn't return [] for "perfect tree sizes"
Updates made to CompactMerkleTree in d1ee609 (PR #180) to return a copy of the internal Node state, have not properly covered the edge case where the tree is perfectly balanced (i.e. has 2^n leaf nodes).

In that situation the set of nodes should be empty:  the size & root hash alone describe the tree, as evidenced by the fact that when merkle.NewCompactMerkleTreeWithState() is called it will make 0 calls to the backing store via its getNodeFunc to retrieve hashes.

rolandshoemaker:
Dipping my toes into this code base with (what seems to me) a rather simple issue, feel free to tell me I've misunderstood the problem!

This change reuses the existing `isPerfectTree` method to check if the tree size is a power of two in `CompactMerkleTree.Hashes` and returns `[][]byte{}` if so. Also updates tests to reflect this new behavior.

Fixes #200.

### DIFF
--- a/merkle/compact_merkle_tree.go
+++ b/merkle/compact_merkle_tree.go
@@ -230,7 +230,7 @@ func (c CompactMerkleTree) Size() int64 {
 // Hashes returns a copy of the set of node hashes that comprise the compact representation of the tree.
 func (c CompactMerkleTree) Hashes() [][]byte {
 	if isPerfectTree(c.size) {
-		return [][]byte{}
+		return nil
 	}
 	n := make([][]byte, len(c.nodes))
 	copy(n, c.nodes)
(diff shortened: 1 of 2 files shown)


jsha:
Would `nil` be more idiomatic?

AlCutter:
Welcome :)
Looks like you could do with a rebase onto master to fix the Travis breakage…

merkle/compact_merkle_tree.go
```diff
@@ -229,6 +229,9 @@ func (c CompactMerkleTree) Size() int64 {
 
 // Hashes returns a copy of the set of node hashes that comprise the compact representation of the tree.
 func (c CompactMerkleTree) Hashes() [][]byte {
+	if isPerfectTree(c.size) {
+		return [][]byte{}
```
phad: +1  to the suggestion of returning nil here instead.

PR #286, opened 2017-01-05
Rename GetTreeRevisionAtSize to GetTreeRevisionIncludingSize + mysql impl

Martin2112:
And other noise caused by regenerating mocks. Update server to always request exact fetch so no behaviour changes yet. Update tests to expect true.

### DIFF
--- a/storage/mysql/log_storage_test.go
+++ b/storage/mysql/log_storage_test.go
@@ -670,20 +670,38 @@ func TestLatestSignedLogRoot(t *testing.T) {
 	}
 }
 
-func TestGetTreeRevisionAtNonExistentSizeError(t *testing.T) {
-	// Have to set all this up though we won't actually write anything
-	logID := createLogID("TestGetTreeRevisionAtSize")
+func TestGetTreeRevisionAtNonExistentSize(t *testing.T) {
+	// Should get an error regardless of whether we ask for an exact or inexact tree revision
+	logID := createLogID("TestGetTreeRevisionAtNonExistentSize")
 	db := prepareTestLogDB(logID, t)
 	defer db.Close()
 	s := prepareTestLogStorage(logID, t)
 	tx := beginLogTx(s, t)
 	defer commit(tx, t)
 
-	if revision, err := tx.GetTreeRevisionAtSize(0); err == nil {
+	if revision, err := tx.GetTreeRevisionAtSize(25, true); err == nil {
+		t.Fatalf("Incorrectly returned a tree revision for tree size 25: %d", revision)
+	}
+
+	if revision, err := tx.GetTreeRevisionAtSize(25, false); err == nil {
+		t.Fatalf("Incorrectly returned a tree revision for tree size 25 (inexact): %d", revision)
+	}
+}
+
+func TestGetTreeRevisionAtNonExistentSizeError(t *testing.T) {
+	// Have to set all this up though we won't actually write anything
+	logID := createLogID("TestGetTreeRevisionAtNonExistentSizeError")
+	db := prepareTestLogDB(logID, t)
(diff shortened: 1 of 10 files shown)


storage/tree_storage.go
```diff
@@ -34,7 +34,9 @@ type TreeTX interface {
 type NodeReader interface {
 	// GetTreeRevisionAtSize returns the max node version for a tree at a particular size.
 	// It is an error to request tree sizes larger than the currently published tree size.
```
codingllama: Remove this line?

storage/mysql/tree_storage.go
```diff
@@ -356,16 +357,17 @@ func checkResultOkAndRowCountIs(res sql.Result, err error, count int64) error {
...
+// For an inexact tree size this implementation always returns the next largest revision if an
+// exact one does not exist but it isn't required to do so.
+func (t *treeTX) GetTreeRevisionAtSize(treeSize int64, exact bool) (int64, error) {
 	// Negative size is not sensible and a zero sized tree has no nodes so no revisions
 	if treeSize <= 0 {
-		return 0, fmt.Errorf("Invalid tree size: %d", treeSize)
+		return 0, fmt.Errorf("invalid tree size: %d", treeSize)
 	}
 
+	stmt := map[bool]string{false: selectTreeRevisionAtSizeOrLargerSQL, true:selectTreeRevisionAtSizeSQL}[exact]
```
codingllama: nit: Use an if/else?

storage/mysql/log_storage_test.go
```diff
@@ -734,11 +752,25 @@ func TestGetTreeRevisionAtSize(t *testing.T) {
 		}
 
 		// But an intermediate value shouldn't work
-		treeRevision3, err := tx.GetTreeRevisionAtSize(21)
+		treeRevision3, err := tx.GetTreeRevisionAtSize(21, true)
 
 		if err == nil {
```
codingllama: Suggestion: use the if initialization statement for variables that don't need to live past a single check?

if treeRevision3, err := tx.GetTreeRevisionAtSize(21, true); err == nil {
    t.Fatalf("Unexpectedly returned revision for nonexistent tree size: %d", treeRevision3)
}

(Ditto for other occurrences.)

storage/tree_storage.go
Martin2112: No it's still true. There can't be any revisions larger than the tree size so it will still error if you request one.

storage/mysql/tree_storage.go
Martin2112: I really miss the '?' operator you get in proper languages.

storage/mysql/log_storage_test.go
Martin2112: Yes this is some of the oldest code in the repo and needs to be looked at again but I was trying to limit the scope for now.

storage/mysql/tree_storage.go
codingllama: That makes 2 of us. :)

storage/mysql/log_storage_test.go
Martin2112: Done.

storage/mysql/tree_storage.go

Martin2112:
OK @AlCutter was quibbling about the API change in this PR. It turns out that I need to make another change because I need the tree size corresponding to the version to do the snapshot recomputation. So I can either leave this as is or remove the parameter and return the version, size tuple leaving the caller to do the right thing.

Martin2112:
Assigned to Al to make sure we agree on this API.

Martin2112:
Going to have to merge this as I'm blocked otherwise. Silence is consent!

PR #288, opened 2017-01-06
Unify all the proof tests for rpc server that test invalid requests and make them table driven.

Martin2112:
Until it's been implemented. Update tests. Unify all the proof tests that test invalid requests and make them table driven.

### DIFF
--- a/server/log_rpc_server_test.go
+++ b/server/log_rpc_server_test.go
@@ -45,24 +45,28 @@ var getByHashRequest2 = trillian.GetLeavesByHashRequest{LogId: logID2, LeafHash:
 
 var getInclusionProofByHashRequestBadTreeSize = trillian.GetInclusionProofByHashRequest{LogId: logID1, TreeSize: -50, LeafHash: []byte("data")}
 var getInclusionProofByHashRequestBadHash = trillian.GetInclusionProofByHashRequest{LogId: logID1, TreeSize: 50, LeafHash: []byte{}}
+var getInclusionProofByHashRequestRehash = trillian.GetInclusionProofByHashRequest{LogId: logID1, TreeSize: 25, LeafHash: []byte("ahash"), AllowRehashing:true}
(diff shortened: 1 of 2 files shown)


server/log_rpc_server_test.go
```diff
@@ -657,35 +698,22 @@ func TestGetLeavesByLeafValueHash(t *testing.T) {
...
-}
+		// Request should fail validation before any storage operations
+		mockStorage := storage.NewMockLogStorage(ctrl)
+		registry := testonly.NewRegistryWithLogProvider(mockStorageProviderFunc(mockStorage))
+		server := NewTrillianLogRPCServer(registry, fakeTimeSource)
 
-func TestGetProofByHashBadHash(t *testing.T) {
-	ctrl := gomock.NewController(t)
-	defer ctrl.Finish()
+		_, err := server.GetInclusionProofByHash(context.Background(), test.hReq)
```
codingllama: Suggestion: Use if initialization?

(Same for others.)

Martin2112: Sadly it needs access to resp later and I was previously criticized for multiple else clauses in a similar situation.

Martin2112:
This has been reworked to remove the API changes as I believe this feature can be implemented safely without it. I thought it was still worth having the test improvements.

server/log_rpc_server_test.go
codingllama: I meant it basically as combining this and the if below, so it wouldn't need any more if/else clauses than we already have.

if _, err := server.GetInclusionProofByHash(context.Background(), test.hReq); err == nil || ... {
	t.Errorf("...")
}

(Just clarifying, please take it as a suggestion.)

Martin2112:
OK the last round of changes might have moved the code you commented. I'll have another look.

Martin2112:
That code is removed in #288 where I cleaned up all the failure mode proof tests.

PR #292, opened 2017-01-17
Use path from node to root for inclusion proofs

Martin2112:
Removes code duplication and prepares for rehashing support.


Martin2112:
I should probably have got @pphaneuf to review this.

PR #293, opened 2017-01-17
Add a fake node reader for improved proof tests coming up

Martin2112:
Can be initialized from a set of nodes or by using a compact Merkle tree to record the updates that would be made at multiple revisions. This is in storage/testonly rather than testonly to avoid import cycles.

### DIFF
--- a/storage/testonly/fake_node_reader.go
+++ b/storage/testonly/fake_node_reader.go
@@ -59,7 +59,7 @@ func (f FakeNodeReader) GetTreeRevisionAtSize(treeSize int64) (int64, error) {
 	return f.treeRevision, nil
 }
 
-// GetTreeRevisionAtSize implements the corresponding NodeReader API
+// GetMerkleNodes implements the corresponding NodeReader API
 func (f FakeNodeReader) GetMerkleNodes(treeRevision int64, NodeIDs []storage.NodeID) ([]storage.Node, error) {
 	if f.treeRevision > treeRevision {
 		return []storage.Node{}, fmt.Errorf("GetMerkleNodes() got treeRevision:%d, want up to: %d", treeRevision, f.treeRevision)
--- a/storage/testonly/fake_node_reader.go
+++ b/storage/testonly/fake_node_reader.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"encoding/hex"
+	"github.com/golang/glog"
 	"github.com/google/trillian/crypto"
 	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/storage"
@@ -40,8 +41,8 @@ func NewFakeNodeReader(mappings []NodeMapping, treeSize, treeRevision int64) *Fa
 		_, ok := nodeMap[mapping.NodeID.String()]
 
 		if ok {
-			// Duplicate mapping - bail out
-			return nil
+			// Duplicate mapping - the test data is invalid so don't continue
+			glog.Fatalf("NewFakeNodeReader duplicate mapping for: %s in:\n%v", mapping.NodeID.String(), mappings)
 		}
 
 		nodeMap[mapping.NodeID.String()] = mapping.Node
@@ -50,7 +51,7 @@ func NewFakeNodeReader(mappings []NodeMapping, treeSize, treeRevision int64) *Fa
 	return &FakeNodeReader{nodeMap: nodeMap, treeSize: treeSize, treeRevision: treeRevision}
 }
 
-// GetTreeRevisionAtSize implements the corresponding NodeReader API
+// GetTreeRevisionAtSize implements the corresponding NodeReader API.
 func (f FakeNodeReader) GetTreeRevisionAtSize(treeSize int64) (int64, error) {
 	if f.treeSize != treeSize {
 		return int64(0), fmt.Errorf("GetTreeRevisionAtSize() got treeSize:%d, want: %d", treeSize, f.treeSize)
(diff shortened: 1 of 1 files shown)


storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+
+	for _, batch := range batches {
+		nodeMap := make(map[string]storage.Node)
+		for _, leaf := range batch.Leaves {
+			// We're only interested in the side effects of adding leaves - the node updates
+			tree.AddLeaf([]byte(leaf), func(depth int, index int64, hash []byte) {
+				nID, err := storage.NewNodeIDForTreeCoords(int64(depth), index, 64)
+
+				if err != nil {
+					panic(fmt.Errorf("failed to create a nodeID for tree - should not happen d:%d i:%d",
```
codingllama: nit: In the grand scheme of things it doesn't matter much, but I would prefer a test failure to a panic. Maybe return an error, or pass in a *testing.T? (Not sure if the latter is idiomatic either.)

Feel free to push back.

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+						depth, index))
+				}
+
+				nodeMap[nID.String()] = storage.Node{NodeID: nID, NodeRevision: batch.TreeRevision, Hash: hash}
+			})
+		}
+
+		// Sanity check the tree root hash against the one we expect to see.
+		if got, want := hex.EncodeToString(tree.CurrentRoot()), batch.ExpectedRoot; got != want {
```
codingllama: FYI, you can do fmt.Sprintf("%x", myBytes) for the same effect.

(Not asking for changes or anything, just an FYI. I'm too lazy to import encoding/hex if fmt is already there.)

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+	readers []FakeNodeReader
+}
+
+// LeafBatch describes a set of leaves to be loaded into a MultiFakeNodeReader via a compact
+// merkle tree. As each batch is added to the tree a set of node updates are collected
+// and recorded in a FakeNodeReader for that revision.
+type LeafBatch struct {
+	TreeRevision int64
+	Leaves       []string
+	ExpectedRoot string
```
codingllama: nit: Mention ExpectedRoot's encoding, or type as []byte and use bytes.Compare() instead?

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+	return &MultiFakeNodeReader{readers: readers}
+}
+
+// NewMultiFakeNodeReaderFromLeaves uses a compact Merkle tree to set up the nodes at various
+// revisions. It collates all node updates from a batch of leaf data into one FakeNodeReader.
+// This has the advantage of not needing to manually create all the data structures but the
+// disadvantage is that a bug in the compact tree could be reflected in test using this
+// code. To help guard against this we check the tree root hash after each batch has been
+// processed.
+func NewMultiFakeNodeReaderFromLeaves(batches []LeafBatch) *MultiFakeNodeReader {
```
codingllama: Do you expect batches to be in ascending TreeRevision order? Please document.

Should we add guards against that?

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+		return int64(0), fmt.Errorf("GetTreeRevisionAtSize() got treeSize:%d, want: %d", treeSize, f.treeSize)
+	}
+
+	return f.treeRevision, nil
+}
+
+// GetMerkleNodes implements the corresponding NodeReader API
+func (f FakeNodeReader) GetMerkleNodes(treeRevision int64, NodeIDs []storage.NodeID) ([]storage.Node, error) {
+	if f.treeRevision > treeRevision {
+		return []storage.Node{}, fmt.Errorf("GetMerkleNodes() got treeRevision:%d, want up to: %d", treeRevision, f.treeRevision)
```
codingllama: Suggestion: Return nil instead of []storage.Node{}? Less characters, type-change proof.

(Same for other occurrences, if you take if, ofc.)

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+// must be distinct.
+func NewFakeNodeReader(mappings []NodeMapping, treeSize, treeRevision int64) *FakeNodeReader {
+	nodeMap := make(map[string]storage.Node)
+
+	for _, mapping := range mappings {
+		_, ok := nodeMap[mapping.NodeID.String()]
+
+		if ok {
+			// Duplicate mapping - bail out
+			return nil
```
Martin2112: I've made it a glog.Fatalf(). If it happens the test is broken and needs fixing.

storage/testonly/fake_node_reader.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		nodeMap[mapping.NodeID.String()] = mapping.Node
+	}
+
+	return &FakeNodeReader{nodeMap: nodeMap, treeSize: treeSize, treeRevision: treeRevision}
+}
+
+// GetTreeRevisionAtSize implements the corresponding NodeReader API
```
Martin2112: Done.

storage/testonly/fake_node_reader.go
Martin2112: I think we've been inconsistent on this in various places throughout the code. If you did this in C++ / Java the code would be littered with nil checks but it probably does make sense in Go. Let's start doing it.

storage/testonly/fake_node_reader.go
Martin2112: Doc added. We've found hex to be more user friendly in tests.

storage/testonly/fake_node_reader.go
Martin2112: Doc added. Added a check to reject invalid input. Thought about sorting the batches instread but there could still be dupes and the current Go sorting API is onerous.

storage/testonly/fake_node_reader.go
Martin2112: Apparently, passing around testing.T is not approved Go dogma. But we do it as the alternatives are not great. Sadly a rich unit testing framework is also apparently against Go dogma. I've made it another glog.Fatalf(). These cases mean the test input is broken.

storage/testonly/fake_node_reader.go
Martin2112: Fair point. We've always just imported hex.

storage/testonly/fake_node_reader.go
codingllama: glog.Fatalf sounds like a good compromise. LGTM.

storage/testonly/fake_node_reader.go
codingllama: SGTM.

PR #294, opened 2017-01-17
Pass args to reset script

gdbelvin:
Allow custom username / password in db setup script
by passing command line arguments to mysql

### DIFF
--- a/scripts/resetdb.sh
+++ b/scripts/resetdb.sh
@@ -3,6 +3,7 @@ echo "Completely wipe and reset database 'test'."
 read -p "Are you sure? " -n 1 -r
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
+    # A command line supplied -u will override the first argument.
     mysql -u root "$@" -e 'DROP DATABASE IF EXISTS test;'
     mysql -u root "$@" -e 'CREATE DATABASE test;'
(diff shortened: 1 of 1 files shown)


scripts/resetdb.sh
```diff
@@ -3,9 +3,9 @@ echo "Completely wipe and reset database 'test'."
 read -p "Are you sure? " -n 1 -r
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
-    mysql -u root -e 'DROP DATABASE IF EXISTS test;'
-    mysql -u root -e 'CREATE DATABASE test;'
-    mysql -u root -e "GRANT ALL ON test.* TO 'test'@'localhost' IDENTIFIED BY 'zaphod';"
-    mysql -u root -D test < storage/mysql/storage.sql
+    mysql -u root "$@" -e 'DROP DATABASE IF EXISTS test;'
```
codingllama: I can see how that would allow for a different password, but does that allow for a different user? Does a later "-u" overrides a previous one?

I would prefer something more explicit, like:
mysql_opts="$@"
if [[ ! "$mysql_opts" || "${mysql_opts##*-u *}" ]]; then
  mysql_opts="-u root $mysql_opts"
fi
mysql "$mysql_opts" ...

Or maybe just add a comment saying -u overrides?

PR #299, opened 2017-01-18
Document LogLeaf

gdbelvin:


trillian_api.proto
```diff
@@ -26,10 +26,16 @@ message TrillianApiStatus {
 }
 
 message LogLeaf {
+    // merkle_leaf_hash is over leaf data and optional extra_data.
     bytes merkle_leaf_hash = 1;
+    // leaf_value contains arbitray data.
```
codingllama: nit: typo on "arbitrary"

trillian_api.proto
```diff
@@ -26,10 +26,16 @@ message TrillianApiStatus {
...
     bytes leaf_value = 2;
+    // extra_data is optional.
```
codingllama: What distinguishes leaf_value and extra_data? Could you add some examples? Are both end-user/personality supplied, or is extra_data recorded for internal Trillian purposes?

PR #301, opened 2017-01-18
Initial stab at writing some map docs

Martin2112:
Very rough.


Martin2112:
Are you planning to write Map docs? Should I keep this rough summary on hold, merge it or abandon it?

PR #306, opened 2017-01-19
Fix GROUP BY for map on newer mysql versions

Martin2112:
And update the doc. I think this is correct because we don't expect to get multiple rows with the same KeyHash and LeafValue in a consistent map tree as the keys are distinct hence none are collapsed to a group in the result set and if I wave my hands and say this very loudly then people will probably believe me.


Martin2112:
"For every complex problem there is an answer that is clear, simple, and wrong." - H. L. Mencken

gdbelvin:
Love the quote

PR #307, opened 2017-01-19
Remove dependency on disabling ONLY_FULL_GROUP_BY

gdbelvin:
in http://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html
MYSQL stipulates that it can return _any_ row from a group by clause.
MySQL now produces warnings when queries do not return unique results.
This PR fixes the GetMapLeaf SQL query to return a unique result.

It also stops storing inverted revision numbers since a particular
database ordering is not longer required.

### DIFF
--- a/storage/mysql/map_storage.go
+++ b/storage/mysql/map_storage.go
@@ -29,7 +29,7 @@ const selectMapLeafSQL string = `
 	SELECT TreeId, KeyHash, MAX(MapRevision) as maxrev
 	FROM MapLeaf t0
 	WHERE t0.KeyHash IN (` + placeholderSQL + `) AND
-	      t0.TreeId = ? AND t0.MapRevision >= ?
+	      t0.TreeId = ? AND t0.MapRevision <= ?
 	GROUP BY t0.TreeId, t0.KeyHash
  ) t2
  ON t1.TreeId=t2.TreeId
--- a/storage/mysql/map_storage.go
+++ b/storage/mysql/map_storage.go
@@ -21,13 +21,20 @@ const selectLatestSignedMapRootSQL string = `SELECT MapHeadTimestamp, RootHash, 
 
 const insertMapLeafSQL string = `INSERT INTO MapLeaf(TreeId, KeyHash, MapRevision, LeafValue) VALUES (?, ?, ?, ?)`
 
-// Note that MapRevision is stored negated, hence the odd equality check below:
-const selectMapLeafSQL string = `SELECT KeyHash, MAX(MapRevision), LeafValue
-	 FROM MapLeaf
-	 WHERE KeyHash IN (` + placeholderSQL + `) AND
-	       TreeId = ? AND
-				 MapRevision >= ?
-	 GROUP BY KeyHash`
+const selectMapLeafSQL string = `
+ SELECT t1.KeyHash, t1.MapRevision, t1.LeafValue
+ FROM MapLeaf t1
+ INNER JOIN
+ (
+	SELECT TreeId, KeyHash, MAX(MapRevision) as maxrev
+	FROM MapLeaf t0
+	WHERE t0.KeyHash IN (` + placeholderSQL + `) AND
+	      t0.TreeId = ? AND t0.MapRevision >= ?
+	GROUP BY t0.TreeId, t0.KeyHash
+ ) t2
+ ON t1.TreeId=t2.TreeId
+ AND t1.KeyHash=t2.KeyHash
+ AND t1.MapRevision=t2.maxrev`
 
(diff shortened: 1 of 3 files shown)


storage/mysql/map_storage.go
```diff
@@ -21,13 +21,20 @@ const selectLatestSignedMapRootSQL string = `SELECT MapHeadTimestamp, RootHash,
...
-// Note that MapRevision is stored negated, hence the odd equality check below:
-const selectMapLeafSQL string = `SELECT KeyHash, MAX(MapRevision), LeafValue
-	 FROM MapLeaf
-	 WHERE KeyHash IN (` + placeholderSQL + `) AND
-	       TreeId = ? AND
-				 MapRevision >= ?
-	 GROUP BY KeyHash`
+const selectMapLeafSQL string = `
+ SELECT t1.KeyHash, t1.MapRevision, t1.LeafValue
+ FROM MapLeaf t1
```
Martin2112: This formatting is possibly incorrect? Did you run gofmt?

storage/mysql/map_storage_test.go
```diff
@@ -161,44 +160,43 @@ func TestMapSetGetMultipleRevisions(t *testing.T) {
...
-			ExtraData: []byte(fmt.Sprintf("Some Extra Data %d", i)),
-		}
-	}
-
-	for i := 0; i < numRevs; i++ {
+	for _, tc := range []struct {
+		rev  int64
+		leaf trillian.MapLeaf
+	}{
+		{0, trillian.MapLeaf{keyHash, []byte{0}, []byte{0}, []byte{0}}},
```
Martin2112: Might read neater if you kept the strings for the values.

storage/mysql/map_storage_test.go
```diff
@@ -161,44 +160,43 @@ func TestMapSetGetMultipleRevisions(t *testing.T) {
...
-		}
-		if got, want := &readValues[0], &values[i]; !proto.Equal(got, want) {
-			t.Fatalf("At rev %d read back %v, but expected %v", i, got, want)
-		}
-		if err := tx.Commit(); err != nil {
-			t.Fatalf("At rev %d failed to commit: %v", i, err)
+		// Read. Verify that we get the current tc.
+		for i := int64(0); i < tc.rev; i++ {
+			tx := beginMapTx(s, t)
+
```
Martin2112: possible spurious blank line

storage/mysql/map_storage_test.go
```diff
@@ -161,44 +160,43 @@ func TestMapSetGetMultipleRevisions(t *testing.T) {
...
+	for _, tc := range []struct {
+		rev  int64
+		leaf trillian.MapLeaf
+	}{
+		{0, trillian.MapLeaf{keyHash, []byte{0}, []byte{0}, []byte{0}}},
+		{1, trillian.MapLeaf{keyHash, []byte{1}, []byte{1}, []byte{1}}},
+		{2, trillian.MapLeaf{keyHash, []byte{2}, []byte{2}, []byte{2}}},
+		{3, trillian.MapLeaf{keyHash, []byte{3}, []byte{3}, []byte{3}}},
+	} {
+		// Write the current test case.
```
Martin2112: This isn't equivalent to previous test as it no longer does all writes before reading back. It's unlikely but it could possibly hide bugs. On some Merkle tests we do them twice, all at once and then like this incrementally. Not sure it's worth it here though.

gdbelvin:
Quick note: I was unable to determine why the tests on Travis are passing. 
If a modern version of MySQL is being installed and used, it should fail, but it's not, and I can't find any code that disables the ONLY_FULL_GROUP_BY option.

storage/mysql/tree_storage.go
```diff
@@ -60,11 +60,6 @@ func openDB(dbURL string) (*sql.DB, error) {
 		return nil, err
 	}
 
-	if _, err := db.Exec("SET sql_mode = 'STRICT_ALL_TABLES'"); err != nil {
```
Martin2112: Not sure why this was removed, doesn't immediately seem related. Is there some problem with it enabled now?

gdbelvin:
David found the version of mysql that Travis is using:
`mysql Ver 14.14 Distrib 5.5.41, for debian-linux-gnu (x86_64) using readline 6.2`

This is old enough that it doesn't enforce ONLY_FULL_GROUP_BY

Martin2112:
I don't know why the test used to work. With negated values it wrote an initial revision 0, which would always be >= any of the other later revisions (e.g. -3). Plus it appears that any of the matching leaf values could be returned, which is why we needed this change. Bit of a mystery.

Martin2112:
LGTM if Travis happy.

PR #313, opened 2017-01-23
Remove some dead code from tree storage

Martin2112:


Martin2112:
@daviddrysdale I don't think these are used by CT any more?

PR #315, opened 2017-01-23
Add Go Report Card and GoDoc to Readme

gdbelvin:


gdbelvin:
There's not really a way to get them to exclude the generated files...

PR #336, opened 2017-01-27
Move HashKey to the personality layer

### LINKED ISSUE #332: Map Index / HashKey should be computed by the personality
The current Map Hasher interface contains a HashKey function to turn a string into a sha256 index in the map. This index, however, should be computed by the personality, not the map. Key Transparency, for instance, computes the index as the output of a privately keyed signature function. 

If this sounds good, I'll convert the Map interfaces to accept a `index []byte` rather than `key []byte` or `HashedKey []byte`, and remove the `HashKey` function from the MapHasher.

gdbelvin:
The personality should be responsible for computing the index
of values in the map.

- Creates a personality layer hasher in testonly.
- Renames HashKey to index.

Closes #332


server/vmap/trillian_map_server/main.go
```diff
@@ -132,15 +115,9 @@ func main() {
 
 	// Bring up the RPC server and then block until we get a signal to stop
 	rpcServer := startRPCServer(lis, *serverPortFlag, registry)
-	go awaitSignal(rpcServer)
+	defer glog.Flush()
```
AlCutter: This will not work the way you want I fear... `glog.Fatalf` below will do *stuff* and then call `os.Exit`, but the [docs for os.Exit](https://golang.org/pkg/os/#Exit) have this to say: _"... deferred functions are *not* run"_.

storage/mysql/map_storage.go
```diff
@@ -124,21 +124,22 @@ func (m *mapTX) Set(keyHash []byte, value trillian.MapLeaf) error {
...
 	defer stmt.Close()
 
-	_, err = stmt.Exec(m.ms.mapID, []byte(keyHash), m.writeRevision, flatValue)
+	_, err = stmt.Exec(m.ms.mapID, keyHash, m.writeRevision, flatValue)
 	return err
 }
 
-func (m *mapTX) Get(revision int64, keyHashes [][]byte) ([]trillian.MapLeaf, error) {
-	stmt, err := m.ms.getStmt(selectMapLeafSQL, len(keyHashes), "?", "?")
+// TODO: API problem: there's nothing to say which map leaves correspond to which indexes.
```
AlCutter: `MapLeaf.Index` tells (or should) that.
As we spoke about earlier there's still some fiddling to do with the API though, either removing `IndexValue` or replacing `[]MapLeaf` with `map[_something_]=MapLeaf`... Maybe update the `TODO` slightly?

storage/mysql/map_storage.go
```diff
@@ -124,21 +124,22 @@ func (m *mapTX) Set(keyHash []byte, value trillian.MapLeaf) error {
...
 		return nil, err
 	}
 	stx := m.tx.Stmt(stmt)
 	defer stx.Close()
 
-	args := make([]interface{}, 0, len(keyHashes)+2)
-	for _, k := range keyHashes {
-		args = append(args, []byte(k[:]))
+	args := make([]interface{}, len(indexes), len(indexes)+2)
+	for i, index := range indexes {
```
AlCutter: No need to change this, ding the `i` and use `append` - there's unlikely to be much (any?) performance boost by doing it this way when the underlying array is pre-allocated, and even if there were it pales into insignificance with the call to the database you're about to make, and `append` is idiomatic

testonly/hasher.go
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+package testonly
+
+// This file implements the hashing functions that are part of a Trillian
+// personality.
+
+import "crypto/sha256"
+
+// HashKey converts an identity string into a map index using SHA256.
```
AlCutter: /an identity string/a map key/ ?

testonly/hasher.go
```diff
@@ -0,0 +1,40 @@
...
+func HashKey(key string) []byte {
+	h := sha256.New()
+	h.Write([]byte(key))
+	return h.Sum(nil)
+}
+
+// TransparentHash returns a key that can be visually inspected.
+// This supports testing where it was nice to see what the key was.
+func TransparentHash(key string) []byte {
+	size := sha256.New().Size()
```
AlCutter: `sha256.Size`

trillian_api.proto
```diff
@@ -204,8 +204,9 @@ service TrillianLog {
 
 // MapLeaf represents the data behind Map leaves.
 message MapLeaf {
-  // key_hash is the hash of the key for this leaf.
-  bytes key_hash = 1;
+  // index is the location of this leaf.
+  // All indexes for a given Map must contain a constant number of bits.
```
AlCutter: Worth popping a `TODO` in the map Set/Get RPC handlers to enforce this, maybe?

examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -120,11 +128,13 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
...
 		MapId:    m.mapID,
-		Key:      make([][]byte, 0, len(domains)),
+		Index:    make([][]byte, len(domains)),
 		Revision: -1,
 	}
-	for k := range domains {
-		getReq.Key = append(getReq.Key, []byte(k))
+	i := 0
+	for d := range domains {
+		getReq.Index[i] = HashDomain(d)
```
gdbelvin: Allocating the array directly is slightly more efficient than using append.
http://stackoverflow.com/questions/21362950/golang-getting-a-slice-of-keys-from-a-map

This would be a bit more elegant if domains was a slice where I could iterate over `i,d := range domains`, but it's a map, which means range iterates over `key, values` rather than `index, values`

storage/mysql/map_storage.go
gdbelvin: In my opinion, it's easier to read, and slightly faster. I've made this change in multiple places.  Can you point to where append is idiomatic?

http://stackoverflow.com/questions/21362950/golang-getting-a-slice-of-keys-from-a-map

trillian_api.proto
gdbelvin: done

storage/mysql/map_storage.go
AlCutter: It _might_ be marginally faster, although the benchmark I made (https://play.golang.org/p/FYh5arq6D9) is not clear cut on the matter: 4 runs:
```
BenchmarkDirect-12    	1000000000	         3.28 ns/op
BenchmarkAppend-12    	2000000000	         4.42 ns/op
BenchmarkDirect-12    	1000000000	         4.27 ns/op
BenchmarkAppend-12    	2000000000	         3.15 ns/op
BenchmarkDirect-12    	1000000000	         3.38 ns/op
BenchmarkAppend-12    	1000000000	         1.98 ns/op
BenchmarkDirect-12    	1000000000	         3.00 ns/op
BenchmarkAppend-12    	1000000000	         2.00 ns/op
```

But even if it were faster (and I suspect it could only ever be marginally faster since both approaches have to do bounds checking) you're adding complexity to pay for it; in this case an extra variable, and in other places where you've done the same thing you've added an extra explicit `var` line to create the variable, as well as an extra explicit `i++` because you're iterating over a map.   But, then, the next thing which happens is many orders of magnitude slower (because it goes to a remote DB server, and then possibly to disk, so your complexity down-payment is for nothing.

`append` is a fundamental part of the language - it's a built-in keyword, it doesn't get more idiomatic than that :)

AlCutter:
BTW, we tend to reply to all comments in reviews (even if it's just a "Done/Ack") to help the reviewer know what's been seen/addressed/skipped etc.  I see one "done", but many more unanswered comments, although I do see you've made quite a few changes in the latest commit.

AlCutter:
Also the build's bust :)

storage/mysql/map_storage.go
gdbelvin: Under the hood, I think they're doing something extremely similar, which explains why their performance is similar. For me this is more about code readability.

storage/mysql/map_storage.go
gdbelvin: done

gdbelvin:
Fixed the build and comments. PTAL

examples/ct/ctmapper/lookup/lookup.go
```diff
@@ -14,6 +15,13 @@ import (
 var mapServer = flag.String("map_server", "", "host:port for the map server")
 var mapID = flag.Int("map_id", -1, "Map ID to write to")
 
+// HashDomain converts a domain into a map index.
```
gdbelvin: This has been bumpted to `examples/ct/ctmapper/hasher.go`

examples/ct/ctmapper/mapper/mapper.go
gdbelvin: I've been convinced.

integration/map.go
```diff
@@ -42,117 +42,89 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
...
+	r, err := client.GetSignedMapRoot(ctx, &trillian.GetSignedMapRootRequest{MapId: mapID})
+	if err != nil {
+		return fmt.Errorf("failed to get map head: %v", err)
 	}
 
-	{
-		// Check values
-		getReq := trillian.GetMapLeavesRequest{
+	if got, want := r.MapRoot.MapRevision, int64(numBatches); got != want {
+		return fmt.Errorf("got SMH with revision %d, expected %d", got, want)
```
gdbelvin: done

integration/map.go
```diff
@@ -42,117 +42,89 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
...
-			getReq.Key = make([][]byte, 0, batchSize)
-			for y := 0; y < batchSize; y++ {
-				getReq.Key = append(getReq.Key, expectedKeys[keyOrder[i]])
-				i++
+		r, err := client.GetLeaves(ctx, getReq)
+		if err != nil {
+			return fmt.Errorf("failed to get values: %v", err)
+		}
+		if got, want := len(r.IndexValueInclusion), len(getReq.Index); got != want {
+			return fmt.Errorf("got %d values, expected %d", got, want)
```
gdbelvin: done

server/vmap/trillian_map_server/main.go
gdbelvin: Good point. I'll use `glog.Errorf` instead

storage/mysql/map_storage.go
gdbelvin: ok, I've been convinced. using append now.

testonly/hasher.go
gdbelvin: done

gdbelvin:
PTAL

examples/ct/ctmapper/mapper/mapper.go
AlCutter: Excellent :)

PR #338, opened 2017-01-27
Pass context down into storage layer calls.

AlCutter:


server/log_operation_manager.go
```diff
@@ -144,7 +143,8 @@ func (l LogOperationManager) OperationLoop() {
 		// Wait for the configured time before going for another pass
 		time.Sleep(l.context.sleepBetweenRuns)
 
-		quit := l.getLogsAndExecutePass()
+		// TODO(alcutter): want a child context with deadline here?
```
Martin2112: All this code is probably up for rework at some point.

storage/mysql/map_storage_test.go
```diff
@@ -15,8 +16,9 @@ func TestMapRootUpdate(t *testing.T) {
 	cleanTestDB(DB)
 	prepareTestMapDB(DB, mapID, t)
 	s := prepareTestMapStorage(DB, mapID, t)
-	tx := beginMapTx(s, t)
-	defer tx.Commit()
+	ctx := context.Background()
```
Martin2112: context.TODO() used elsewhere. Should they all be the same?

AlCutter: The TODOs are because we need to fix the problem that there's no context being passed in, I think a context.Background is fine here.

PR #339, opened 2017-01-27
Bump to latest proto release: 3.1.0

AlCutter:


Martin2112:
Not sure this is the underlying cause. Slightly worried if we submit this it will break for everyone else.  I suggest more experimentation first.

codingllama:
FYI: #404 bumps protoc to 3.2.0.

Martin2112:
An auspicious PR number.

Martin

Issue #347, opened 2017-01-31
API: Set single place for Index to be set and returned.


gdbelvin:
The Map API is intended to expose a key / value interface, yet the "value" aka. `MapLeaf` currently contains the `key` aka `Index`.  `Index` is also contained in `IndexValue`, `IndexInclusionProof` and several other messages, producing confusion about where and when to set Index. 

Proposal: 
Use proto3 map in Set / Get if we can find a scalar value for index. Perhaps a hex string? 
Or standadize on the IndexValue message. 

Blocking google/keytransparency#486

gdbelvin:
After team discussion we decided to include index in the `MapLeaf` itself.

PR #348, opened 2017-02-01
Remove TrillianStatus and TrillianStatusCode

### LINKED ISSUE #335: Use gRPC status to return errors
Consider removing TrillianStatusCode so that clients have a single place to check for errors.

gdbelvin:
Simplify error checking by having a single location to report errors.

Also use standardized error codes for better error processing.

Closes #335


AlCutter:
I like the idea of switching to standardised error codes, but the `QueueLeaves` call needs to return one status for each queued entry, i.e. a single leaf failing to queue for whatever reason shouldn't abort the whole batch.

gdbelvin:
I agree with you, however, the current api returns only one status for the whole batched `QueueLeaves` request.  Returning one status code per operation requires returning a `map` of some kind between indexes and status codes, bringing us back to #347. 

I suggest we merge this as it maintains the current granularity of error codes, and create a new PR for returning individual error codes once we have a design plan for that.

gdbelvin:
btw, for returning individual error codes, we can use github.com/googleapis/google/rpc/status.proto

AlCutter:
Ok, fair enough - I think we're probably going to get to the API changes in short order, but could you just pop a TODO in there next to the affected APIs if there's not already one?

gdbelvin:
Added TODOs

PR #352, opened 2017-02-01
Use an explicit logID in the sequencer

codingllama:
This solves a TODO and removes util.LogID.

### DIFF
--- a/server/sequencer_manager.go
+++ b/server/sequencer_manager.go
@@ -49,7 +49,7 @@ func (s SequencerManager) Name() string {
 
 // ExecutePass performs sequencing for the specified set of Logs.
 func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManagerContext) bool {
-	glog.V(1).Infof("beginning sequencing run for %v active log(s)", len(logIDs))
+	glog.V(1).Infof("Beginning sequencing run for %v active log(s)", len(logIDs))
 
 	successCount := 0
 	leavesAdded := 0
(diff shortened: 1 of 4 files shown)


codingllama:
Related to issue #324

server/sequencer_manager.go
```diff
@@ -49,7 +49,7 @@ func (s SequencerManager) Name() string {
 
 // ExecutePass performs sequencing for the specified set of Logs.
 func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManagerContext) bool {
-	glog.V(1).Infof("Beginning sequencing run for %d active log(s)", len(logIDs))
+	glog.V(1).Infof("beginning sequencing run for %v active log(s)", len(logIDs))
```
AlCutter: No need to lowercase "beginning" as this is a log message rather than an error.

server/sequencer_manager.go
```diff
@@ -75,18 +75,16 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
 			continue
 		}
 
 		successCount++
 		leavesAdded += leaves
 	}
 
-	glog.V(1).Infof("Sequencing run completed %d succeeded %d failed %d leaves integrated", successCount, len(logIDs)-successCount, leavesAdded)
-
+	glog.V(1).Infof("sequencing run completed %v succeeded %v failed %v leaves integrated", successCount, len(logIDs)-successCount, leavesAdded)
```
AlCutter: same here

server/sequencer_manager.go
codingllama: Done.

Thankfully I didn't do that *everywhere* in a bunch of different PRs. Imagine that... :)

server/sequencer_manager.go
codingllama: Done.

codingllama:
PTAL?

server/sequencer_manager.go
AlCutter: hehe, that's, er, lucky! :)

PR #353, opened 2017-02-02
Fix subtree cache edge case. For logs only write internal nodes of incompletely filled subtrees

Martin2112:
Because they cannot be reconstructed purely from the nodes of a single subtree - for explanation see comments in code. Add a test that fails without this fix because writes cross subtree boundaries. Remove the log proto repopulate known answer test as it's no longer applicable. Should not affect maps.


storage/cache/subtree_cache.go
```diff
@@ -46,8 +46,10 @@ type SubtreeCache struct {
 	dirtyPrefixes map[string]bool
 	// mutex guards access to the maps above.
 	mutex *sync.RWMutex
-
+	// used to rebuild internal nodes when subtrees are loaded
```
AlCutter: "populateSubtree is used..."

storage/cache/subtree_cache.go
```diff
@@ -46,8 +46,10 @@ type SubtreeCache struct {
...
 	populateSubtree storage.PopulateSubtreeFunc
+	// used for preparation work when subtrees are about to be written to storage
```
AlCutter: "prepareSubt..."

storage/cache/subtree_cache.go
```diff
@@ -46,8 +46,10 @@ type SubtreeCache struct {
...
+	prepareSubtreeWrite storage.PrepareSubtreeWriteFunc
```
AlCutter: I know it was me who caused this, but it might be worth renaming these two fields to have shorter names?

storage/cache/subtree_cache.go
```diff
@@ -400,11 +405,23 @@ func PopulateMapSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+// nodes when the subtree is fully populated. For an explanation of why see the comments
+// below for PrepareLogSubtreeWrite.
 func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtreeFunc {
 	return func(st *storagepb.SubtreeProto) error {
-		st.InternalNodes = make(map[string][]byte)
 		cmt := merkle.NewCompactMerkleTree(treeHasher)
+		if st.Depth < 1 {
+			return fmt.Errorf("populate log subtree with invalid depth: %d", st.Depth)
+		}
+		fullyPopulatedLeafCount := 1 << uint(st.Depth)
```
AlCutter: maxLeaves?

storage/cache/subtree_cache.go
```diff
@@ -424,7 +441,9 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
 					// TODO(al): Don't panic Mr. Mainwaring.
 					panic(err)
 				}
-				if depth > 0 {
+				// Don't put leaves into the internal map and only update if we're rebuilding internal
```
AlCutter: I'm not sure I follow the comment - isn't this block going to update the internal map?

storage/cache/subtree_cache.go
```diff
@@ -436,3 +455,39 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
 		return nil
 	}
 }
+
+// PrepareMapSubtreeWrite prepares a map subtree for writing. For maps the internal
+// nodes are never written to storage and are thus always cleared
```
AlCutter: "." :)

storage/cache/subtree_cache.go
```diff
@@ -436,3 +455,39 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+}
+
+// PrepareLogSubtreeWrite prepares a log subtree for writing. If the subtree is fully
+// populated the internal nodes are cleared. Otherwise they are written.
+//
+// To see why this is necessary consider the case where a tree has a single full subtree
+// and then an additional leaf is added.
+//
+// This causes an extra level to be added to the tree with an internal node that is a hash
+// of the root of the left full subtree and the new leaf. Thus the internal nodes cannot be
```
AlCutter: "(note that the leaves are stored at level 0, which will be in a lower stratum in this example)"

storage/cache/subtree_cache.go
```diff
@@ -436,3 +455,39 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+// subtree.
+//
+// Fully populated subtrees don't have this problem because by definition they can only
+// contain internal nodes built from their own contents.
+func PrepareLogSubtreeWrite() storage.PrepareSubtreeWriteFunc {
+	return func(st *storagepb.SubtreeProto) error {
+		if st.Depth < 1 {
+			return fmt.Errorf("prepare subtree for log write invalid depth: %d", st.Depth)
+		}
+		fullyPopulatedNodeCount := 1 << uint(st.Depth)
```
AlCutter: maxLeaves?

storage/cache/subtree_cache.go
```diff
@@ -400,11 +405,23 @@ func PopulateMapSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+		if st.Depth < 1 {
+			return fmt.Errorf("populate log subtree with invalid depth: %d", st.Depth)
+		}
+		fullyPopulatedLeafCount := 1 << uint(st.Depth)
+
+		// We're going to rebuild internal nodes if the subtree is fully populated so clear them now
+		if len(st.Leaves) == fullyPopulatedLeafCount {
+			st.InternalNodes = make(map[string][]byte)
+		}
+
```
AlCutter: Maybe add a check that in the other case st.InternalNodes is not nil and contains a valid number of nodes given the number of leaves?

storage/mysql/tree_storage.go
```diff
@@ -270,13 +270,13 @@ func (t *treeTX) storeSubtrees(subtrees []*storagepb.SubtreeProto) error {
 	// TODO(al): probably need to be able to batch this in the case where we have
 	// a really large number of subtrees to store.
 	args := make([]interface{}, 0, len(subtrees))
+	glog.Warningf("writing %d subtrees", len(subtrees))
```
AlCutter: Still want this in here? :)

Martin2112:
My additional checks were incorrect, hopefully fixed now. Going to address the review comments next.

storage/cache/subtree_cache.go
Martin2112: Done.

storage/cache/subtree_cache.go
Martin2112: Done.

storage/cache/subtree_cache.go
Martin2112: populate and prepare? I don't think it harms clarity because a.) there already isn't any and b.) they only get called once each.

storage/cache/subtree_cache.go
Martin2112: Done, and documented for anyone else lucky enough to work with this code.

storage/cache/subtree_cache.go
Martin2112: Not necessarily. We always have to go through the loop because we need the root hash from the CMT as we don't store it (there's a TODO). In the case where the internal leaves were saved with the subtree we don't update the map because they're correct when we started. I've tried a different explanation.

storage/cache/subtree_cache.go
Martin2112: Note added.

Martin2112: I might add a diagram to the storage doc later.

storage/mysql/tree_storage.go
Martin2112: No. I'll take it out. There's not enough logging to debug logic issues so maybe some should go back as log.V() later.

storage/cache/subtree_cache.go
```diff
@@ -433,6 +459,58 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+			// subtrees on disk so we can detect corruption.
+			return fmt.Errorf("log repop got: %d internal nodes, want: %d", got, want)
+		}
+
+		return nil
+	}
+}
+
+// PrepareMapSubtreeWrite prepares a map subtree for writing. For maps the internal
+// nodes are never written to storage and are thus always cleared
```
AlCutter: "." on the end of the comment

storage/cache/subtree_cache.go
```diff
@@ -433,6 +459,58 @@ func PopulateLogSubtreeNodes(treeHasher merkle.TreeHasher) storage.PopulateSubtr
...
+//
+// Fully populated subtrees don't have this problem because by definition they can only
+// contain internal nodes built from their own contents.
+func PrepareLogSubtreeWrite() storage.PrepareSubtreeWriteFunc {
+	return func(st *storagepb.SubtreeProto) error {
+		st.InternalNodeCount = uint32(len(st.InternalNodes))
+		if st.Depth < 1 {
+			return fmt.Errorf("prepare subtree for log write invalid depth: %d", st.Depth)
+		}
+		fullyPopulatedNodeCount := 1 << uint(st.Depth)
```
AlCutter: maxLeaves?

storage/mysql/tree_storage.go
```diff
@@ -150,7 +150,7 @@ func (m *mySQLTreeStorage) setSubtreeStmt(num int) (*sql.Stmt, error) {
 	return m.getStmt(insertSubtreeMultiSQL, num, "VALUES(?, ?, ?, ?)", "(?, ?, ?, ?)")
 }
 
-func (m *mySQLTreeStorage) beginTreeTx(ctx context.Context, treeID int64, hashSizeBytes int, strataDepths []int, populateSubtree storage.PopulateSubtreeFunc) (treeTX, error) {
+func (m *mySQLTreeStorage) beginTreeTx(ctx context.Context, treeID int64, hashSizeBytes int, strataDepths []int, populateSubtree storage.PopulateSubtreeFunc, prepareSubtreeWrite storage.PrepareSubtreeWriteFunc) (treeTX, error) {
```
AlCutter: use the shorter `populate` and `prepare` names for the args here too maybe?

storage/cache/subtree_cache.go
Martin2112: Done.

storage/cache/subtree_cache.go
Martin2112: Done.

storage/mysql/tree_storage.go

PR #354, opened 2017-02-02
AdminStorage interface and API protos

codingllama:
Adds the AdminStorage interface, read-only and read/write AdminTX and the
corresponding Tree proto to the API.


trillian.proto
```diff
@@ -22,12 +34,107 @@ enum SignatureAlgorithm {
...
+  // Optional.
+  string description = 9;
+
+  // Timestamp of tree creation.
+  // Readonly.
+  int64 create_time_millis_since_epoch = 10;
+
+  // Timestamp of last tree update.
+  // Readonly (automatically assigned on updates).
+  int64 update_time_millis_since_epoch = 11;
```
AlCutter: Would it be useful to have a `repeated Audit` field where the API appends details of any mutations made to this pb?

codingllama: Nice? Yes. Useful? Not so much, I think.

We only allow changes to display_name and description, which are pretty inconsequential ATM.

codingllama:
Thanks for the quick review!

I pushed a few other changes, PTAL?

trillian.proto
AlCutter: I was thinking of when state could be changed in particular.

codingllama: I wasn't planning on adding a history at this stage. I'd like to give it more thought if that's something we want to have.

PR #355, opened 2017-02-02
Add a larger test of building proof paths to the fetcher test

Martin2112:
Should also help to localize further subtree cache bugs because this test is independent of real storage / caching.

### DIFF
--- a/server/proof_fetcher_test.go
+++ b/server/proof_fetcher_test.go
@@ -153,12 +153,12 @@ func TestTree813FetchAll(t *testing.T) {
 			for i, f := range fetches {
 				t.Errorf("Fetch: %d => %s", i, f.NodeID.CoordString())
 			}
-			t.Fatalf("(%d, %d, %d): got proof len: %d, want: %d: %v\n%v", ts, ts, l, got, want, fetches, refProof)
+			t.Fatalf("(%d, %d): got proof len: %d, want: %d: %v\n%v", ts, l, got, want, fetches, refProof)
 		}
 
 		for i := 0; i < len(proof.ProofNode); i++ {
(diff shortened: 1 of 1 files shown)


server/proof_fetcher_test.go
```diff
@@ -126,6 +126,44 @@ func TestRehasher(t *testing.T) {
...
+		}
+
+		// We use +1 here because of the 1 based leaf indexing of this implementation
+		refProof := mt.PathToRootAtSnapshot(l + 1, ts)
+
+		if got, want := len(proof.ProofNode), len(refProof); got != want {
+			for i, f := range fetches {
+				t.Errorf("Fetch: %d => %s", i, f.NodeID.CoordString())
+			}
+			t.Fatalf("(%d, %d, %d): got proof len: %d, want: %d: %v\n%v", ts, ts, l, got, want, fetches, refProof)
```
AlCutter: logging ts twice?

server/proof_fetcher_test.go
```diff
@@ -126,6 +126,44 @@ func TestRehasher(t *testing.T) {
...
+		}
+
+		for i := 0; i < len(proof.ProofNode); i++ {
+			if got, want := hex.EncodeToString(proof.ProofNode[i].NodeHash), hex.EncodeToString(refProof[i].Value.Hash()); got != want {
+				t.Fatalf("(%d, %d, %d): %d got proof node: %s, want: %s l:%d fetches: %v", ts, ts, l, i, got, want, len(proof.ProofNode), fetches)
```
AlCutter: same here

server/proof_fetcher_test.go
Martin2112: Cough. Might have cut and pasted stuff. Fixed.

server/proof_fetcher_test.go
Martin2112: Also fixed.

PR #358, opened 2017-02-03
Rename /*MapTX/*MapTreeTX/ and add "ForTree" suffix to Begin/Snapshot

codingllama:
Analogous to the LogStorage change, this allows us to introduce a new kind of
MapTX that is not tied to any particular treeID.


codingllama:
Issue #324

PR #359, opened 2017-02-03
Add error returns to compact tree set node func etc.

Martin2112:
Allows some panics / TODOs to be removed and fixes that some errors could previously be ignored.

### DIFF
--- a/merkle/compact_merkle_tree.go
+++ b/merkle/compact_merkle_tree.go
@@ -176,7 +176,7 @@ func (c *CompactMerkleTree) AddLeaf(data []byte, f setNodeFunc) (int64, []byte, 
 	h := c.hasher.HashLeaf(data)
 	seq, err := c.AddLeafHash(h, f)
 	if err != nil {
-		return -1, nil, err
+		return 0, nil, err
 	}
 	return seq, h, err
 }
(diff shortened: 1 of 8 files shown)


merkle/compact_merkle_tree.go
```diff
@@ -158,39 +160,45 @@ func (c *CompactMerkleTree) recalculateRoot(f setNodeFunc) {
...
+	return nil
 }
 
 // AddLeaf calculates the leafhash of |data| and appends it to the tree.
 // |f| is a callback which will be called multiple times with the full MerkleTree coordinates of nodes whose hash should be updated.
-func (c *CompactMerkleTree) AddLeaf(data []byte, f setNodeFunc) (int64, []byte) {
+func (c *CompactMerkleTree) AddLeaf(data []byte, f setNodeFunc) (int64, []byte, error) {
 	h := c.hasher.HashLeaf(data)
-	return c.AddLeafHash(h, f), h
+	seq, err := c.AddLeafHash(h, f)
```
codingllama: nit: Return "default" values for seq and h if err != nil

if err != nil {
  return 0, nil, err
}

Martin2112: Maybe -1, nil, err as 0 is a valid sequence number.

Issue #331, opened 2017-01-26
Hashing Interface


gdbelvin:
Hasher needs to be an interface to support alternative hashing implementations. 
Here's a proposal that supports both logs and maps.  
Individual implementations are not required to incorporate all input fields into their hash. 

```
// TreeHasher provides hash functions for tree implementations.
type TreeHasher interface {
	HashLeaf(treeID, index []byte, depth int, dataHash []byte) Hash
	HashEmpty(treeID, index []byte, depth int) Hash
	HashInterior(left, right Hash) Hash
}
```

Steps:
- [x] Move HashKey to Personality
- [x] Create Hasher interface
- [x] Create test implementations
- [x] Migrate to new function interfaces.

codingllama:
I'd suggest renaming HashInterior() to HashNode().

Do we expect leaves to have depth = 0? Because, if so, that's height, not depth (https://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height)

Martin2112:
I'm not clear if this will turn out to be a special case for Key Transparency. If so then every other client will have stuff in the API that they don't use. Is there maybe a better way?

@codingllama  We've used 'depth' a lot in the code as well as 'level', which is what we probably should have stuck to.

codingllama:
Isn't the depth/height/level, in a way, a component of the index? I don't see too much problem in including it there.

Some smalls inconsistencies do bother me, though:
* "dataHash []byte" vs "left, right Hash" - Why not make all of them Hash?
* HashInterior has neither index nor depth - That makes the interface KT specific, IMO
* treeID should probably be an int64

We may also want a type to represent the (index, depth) pair.

@Martin2112 If we mean "height" when we say "depth", we should at least stop using the latter.

Martin2112:
It's effectively part of the NodeID in logs but not maps and is not part of the hash in logs or our existing maps. We should try to keep anything outside of personality code as generic as possible. I'm just questioning if we've done that here. It may not be possible to do it cleanly either.

Depth probably crept in from the RFC where trees grow from the top down. For obvious reasons we don't want to have to update as the tree grows so we work in the other direction. All I'm saying is that there may well be places where it's been used instead of node level, which could cause confusion. In some places e.g. the strata code depth is the correct term. I don't know whether Gary is using it in the correct sense here.

gdbelvin:
Incorporating the comments above:

```
// TreeHasher provides hash functions for tree implementations.
[...]

PR #364, opened 2017-02-07
Remove TreeHasher.Digest

gdbelvin:
Partial work towards #331

Future Hasher API does not have a Digest function.
This PR removes it in preparation for that.

The code base that mostly used Digest was the Map tests
which I converted to table driven tests for clarity and
ease of code maintenance.

### DIFF
--- a/merkle/map_verifier.go
+++ b/merkle/map_verifier.go
@@ -28,7 +28,7 @@ import (
 // append-only logs, but adds support for nil/"default" proof nodes.
 //
 // Returns nil on a successful verification, and an error otherwise.
-func VerifyMapInclusionProof(index []byte, leafHash []byte, expectedRoot []byte, proof [][]byte, h MapHasher) error {
+func VerifyMapInclusionProof(index, leafHash, expectedRoot []byte, proof [][]byte, h MapHasher) error {
 	hBits := h.Size() * 8
 
 	if got, want := len(proof), hBits; got != want {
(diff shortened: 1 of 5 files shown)


merkle/map_verifier.go
```diff
@@ -40,9 +40,6 @@ func VerifyMapInclusionProof(index []byte, leafHash []byte, expectedRoot []byte,
 	if got, want := len(leafHash)*8, hBits; got != want {
 		return fmt.Errorf("invalid leafHash length %d, expected %d", got, want)
 	}
-	if got, want := len(expectedRoot)*8, hBits; got != want {
```
gdbelvin: Removing this check allows tests to supply clean dummy values. 
It doesn't reduce the robustness of the checks because the `expectedRoot` is `bytes.Equal` compared with the computed roots.

merkle/tree_hasher.go
```diff
@@ -52,28 +52,24 @@ func NewRFC6962TreeHasher() TreeHasher {
 
 // HashEmpty returns the hash of an empty element for the tree
 func (t TreeHasher) HashEmpty() []byte {
-	return t.Digest([]byte{})
+	return t.New().Sum(nil)
 }
 
 // HashLeaf returns the Merkle tree leaf hash of the data passed in through leaf.
 // The data in leaf is prefixed by the LeafHashPrefix.
 func (t TreeHasher) HashLeaf(leaf []byte) []byte {
```
gdbelvin: Using the hasher interface is more efficient because `h.Write` can absorb inputs without first needing to copy all the inputs into a separate, contiguous input buffer.

merkle/map_verifier.go
Martin2112: Not sure what you mean about "clean" values.

gdbelvin: By "clean" I mean that tests can use `[]byte("fake root")` instead of `hasher.LeafHash([]byte("fake root")`  Requiring test values to be the output of a hash function doesn't make much sense, especially when we're trying to test for error cases.

PR #366, opened 2017-02-07
Adding ReadOnlyMapTX and DatabaseChecker interfaces

codingllama:
ReadOnlyMapTX allows us to create map snapshot transactions without a treeID.
DatabaseChecker allows us to check connectivity in more optimal manner (ie,
without listing all tree IDs, which is likelyt to become expensive at some point).

ReadOnlyMapTX and ReadOnlyLogTX both implement DatabaseChecker. Connectivity
checks have been modified to use the new method.

### DIFF
--- a/storage/map_storage.go
+++ b/storage/map_storage.go
@@ -20,6 +20,19 @@ import (
 	"github.com/google/trillian"
 )
 
+// ReadOnlyMapTX provides a read-only view into log data.
+// A ReadOnlyMapTX, unlike ReadOnlyMapTreeTX, is not tied to a particular tree.
+type ReadOnlyMapTX interface {
+	DatabaseChecker
+
+	// Commit ensures the data read by the TX is consistent in the database. Only after Commit the
+	// data read should be regarded as valid.
+	Commit() error
+
+	// Rollback discards the read-only TX.
+	Rollback() error
+}
+
 // ReadOnlyMapTreeTX provides a read-only view into the Map data.
 // A ReadOnlyMapTreeTX can only read from the tree specified in its creation.
 type ReadOnlyMapTreeTX interface {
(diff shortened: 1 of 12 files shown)


storage/mysql/log_storage.go
```diff
@@ -138,6 +137,10 @@ func (t *readOnlyLogTX) Rollback() error {
 	return t.tx.Rollback()
 }
 
+func (t *readOnlyLogTX) CheckDatabaseAccessible() error {
```
gdbelvin: /CheckDatabaseAccessible/IsConnected/

Prefer verbs in function names.

codingllama: Done.

codingllama:
PTAL?

storage/mysql/log_storage.go
AlCutter: "Check" is a perfectly cromulent verb :)

storage/map_storage.go
```diff
@@ -20,6 +20,19 @@ import (
 	"github.com/google/trillian"
 )
 
+// ReadOnlyMapTX provides a read-only view into log data.
+// A ReadOnlyMapTX, unlike ReadOnlyMapTreeTX, is not tied to a particular tree.
+type ReadOnlyMapTX interface {
+	DatabaseChecker
```
AlCutter: Would this be better as a func on the Storage itself?

storage/mysql/log_storage.go
codingllama: I've reverted back to "CheckDatabaseAccessible", on the grounds of:
1. We have a few methods named like that
2. It better matches the interface name
3. *IsConnected* is not entirely true, as the storage doesn't necessarily keep an active connection (plus it may suggest a different kind of usage)
4. Al doesn't seem to mind it

storage/map_storage.go
codingllama: Done.

I've moved DatabaseChecker out of TX and to Storage interfaces. I've elected to keep ReadOnlyMapTX, even though it does nothing currently, as I hope it'll be an incentive to adding methods to it when the time comes. (I really don't want another treeID refactor in my hands.)

codingllama:
PTAL?

codingllama:
Thanks Al, I'll go ahead and merge this.

PR #369, opened 2017-02-07
Recalculate intermediate hashes using current tree.

AlCutter:


merkle/merkle_path.go
```diff
@@ -38,14 +38,35 @@ func (n NodeFetch) Equivalent(other NodeFetch) bool {
 	return n.Rehash == other.Rehash && n.NodeID.Equivalent(other.NodeID)
 }
 
+// checkSnapshot performs a couple of simple sanity checks on sa and treeSize
```
Martin2112: sa? typo?

merkle/merkle_path.go
```diff
@@ -38,14 +38,35 @@ func (n NodeFetch) Equivalent(other NodeFetch) bool {
...
+// and returns an error if there's a problem.
+func checkSnapshot(ssDesc string, ss, treeSize int64) error {
+	if ss < 1 {
+		return fmt.Errorf("%s %d < 1", ssDesc, ss)
```
Martin2112: can you add something to the error message to show where it's from.

merkle/merkle_path.go
AlCutter: whoops, fixed.

merkle/merkle_path.go
AlCutter: Done at the next level up, how's that?

Martin2112: LGTM.

PR #371, opened 2017-02-08
Commit() and Rollback() were dropping errors

Martin2112:
In particular the cache flush could fail to write subtrees and the  TX would appear to succeed.


gdbelvin:
Consider using the go tool  https://github.com/kisielk/errcheck to ensure that we're catching all errors.

PR #372, opened 2017-02-09
MySQL AdminStorage implementation

codingllama:
Adds a new MySQL AdminStorage implementation, plus a generic AdminStorage
tester, which may be used to test any AdminStorage implementation (not just the
MySQL one).

A few schema changes have been made. A follow up PR will wire the new storage in
the current code and further add constraints to the schema, plus remove some
redundant columns. Notably, this PR enables strict mode on MySQL, without which
invalid enums generate warnings instead of errors (and get truncated to an
empty string).

A few general utilities have been added as well (tree ID generation, tree
validation), which may be used by other AdminStorage implementations.


storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+		&createDatetime,
+		&updateDatetime,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert all things!
+	if ts, ok := trillian.TreeState_value[treeState]; ok {
+		tree.TreeState = trillian.TreeState(ts)
```
AlCutter: nitty question - would it be better to flip the success/fail blocks around to closer match the normal "indent the error" idiom?

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+// define one.
+type row interface {
+	Scan(dest ...interface{}) error
+}
+
+func readTree(row row) (*trillian.Tree, error) {
+	tree := &trillian.Tree{}
+
+	// Enums and Datetimes need an extra conversion step
+	var treeState string
```
AlCutter: could do a var (...) block here, or even var treeType, hashStrategy, ... string

gdbelvin: This could be collapsed into one line:
`var treeState, treeType, ... string`

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+		&tree.Description,
+		&createDatetime,
+		&updateDatetime,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert all things!
+	if ts, ok := trillian.TreeState_value[treeState]; ok {
```
gdbelvin: Move into separate function?  This might push you over the cyclomatic complexity warning.

storage/mysql/admin_storage.go
AlCutter: Also, adding _another_ block of the intermediate vars would let you ditch the indented "success" bits, at the cost of having those vars in-scope for longer.  Might be a bit more readable, though.
Take it or leave it as you see fit.

storage/mysql/storage.sql
```diff
@@ -4,23 +4,40 @@
...
+-- instead of warning.
+-- https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict
+SET GLOBAL sql_mode = 'STRICT_ALL_TABLES';
 
 -- Tree parameters should not be changed after creation. Doing so can
 -- render the data in the tree unusable or inconsistent.
+-- TODO(codingllama): Add NOT NULL constraints to most columns
 CREATE TABLE IF NOT EXISTS Trees(
-  TreeId                INTEGER NOT NULL,
+  TreeId                BIGINT NOT NULL,
```
gdbelvin: What's the reason for the BIGINT change?

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+			break
+		}
+	}
+	if duplicatePolicy == "" {
+		return nil, fmt.Errorf("unexpected DuplicatePolicy value: %v", newTree.DuplicatePolicy)
+	}
+
+	_, err = insertTreeStmt.Exec(
+		newTree.TreeId,
+		1, /* KeyId */
```
AlCutter: Does this need a TODO?

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+		nowDatetime,                                                            /* UpdateTime */
+		newTree.HashAlgorithm.String(),                                         /* LeafHasherType */
+		newTree.HashAlgorithm.String(),                                         /* TreeHasherType */
+		newTree.DuplicatePolicy == trillian.DuplicatePolicy_DUPLICATES_ALLOWED, /* AllowsDuplicateLeaves */
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// MySQL silently truncates data when running on non-strict mode.
```
AlCutter: /on/in/

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+		defaultSignIntervalSeconds,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &newTree, nil
+}
+
+func toMillisSinceEpoch(t time.Time) int64 {
```
AlCutter: I'd consider storing nanos anyway, just because everything* else in trillian is nanos and it could sow some confusion.

storage/mysql/admin_storage_test.go
```diff
@@ -0,0 +1,67 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mysql
+
+import (
+	"testing"
+
+	"context"
```
AlCutter: "context" should be up above "testing"

storage/mysql/admin_storage_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/google/trillian/storage"
+	"github.com/google/trillian/storage/testonly"
+)
+
+var selectTreeControlByID = "SELECT ReadOnlyRequests, SigningEnabled, SequencingEnabled, SequenceIntervalSeconds, SignIntervalSeconds FROM TreeControl WHERE TreeId = ?"
```
AlCutter: could it be const?

storage/testonly/admin_storage_tester.go
```diff
@@ -0,0 +1,309 @@
...
+// MapTree is a valid, MAP-type trillian.Tree for tests.
+var MapTree *trillian.Tree = &trillian.Tree{
+	TreeState:          trillian.TreeState_ACTIVE,
+	TreeType:           trillian.TreeType_MAP,
+	HashStrategy:       trillian.HashStrategy_RFC_6962,
+	HashAlgorithm:      trillian.HashAlgorithm_SHA256,
+	SignatureAlgorithm: trillian.SignatureAlgorithm_ECDSA,
+	DuplicatePolicy:    trillian.DuplicatePolicy_DUPLICATES_ALLOWED,
+	DisplayName:        "Llamas Map",
+	Description:        "Key Transparency map for all your digital llama needs.",
```
AlCutter: hehe

storage/mysql/admin_storage.go
codingllama: gocyclo isn't complaining. I'd rather keep it here, as otherwise functions will multiply really quickly.

I tried a bunch of alternative ways of abstracting/simplifying this, but I couldn't find a way around the barrage of if/else/return err that Golang likes so much.

storage/testonly/admin_storage_tester.go
```diff
@@ -0,0 +1,309 @@
...
+	for _, v := range values {
+		m[v.TreeId] = v
+	}
+	return m
+}
+
+// TestAdminTXClose verifies the behavior of Close() with and without explicit Commit() / Rollback() calls.
+func (tester *AdminStorageTester) TestAdminTXClose(t *testing.T) {
+	tests := []struct {
+		commit       bool
```
AlCutter: worth adding a "desc string" member to the tests and adding it to the log msgs below (rather than index)?

storage/mysql/storage.sql
codingllama: We picked int64 for tree IDs, which don't fit on a MySQL integer.

Fun fact: non-strict MySQL silently truncates int64 to int32, which caused a bunch of my IDs to become int32_max_value. Good times.

codingllama:
Thanks for reviewing so quickly! I'll see what's up with Travis, address remaining comments and ping the thread when I'm done.

storage/testonly/admin_storage_tester.go
```diff
@@ -0,0 +1,309 @@
...
+		{rollback: true, wantRollback: true},
+		{wantRollback: true}, // Close() before Commit() or Rollback() will cause a rollback
+	}
+
+	s := tester.NewAdminStorage()
+	for i, test := range tests {
+		ctx := context.Background()
+		tx, err := s.Begin(ctx)
+		if err != nil {
+			t.Fatalf("%v: Begin() = (_, %v), want = (_, nil)", i, err)
```
AlCutter: prefer `Error.*` if possible (could this perhaps `Errorf` and then `continue` for example?)

storage/tree_id.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"crypto/rand"
+	"math"
+	"math/big"
+)
+
+// NewTreeID generates a random, positive, non-zero tree ID.
+func NewTreeID() (int64, error) {
+	id, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
```
AlCutter: could you just do `return rand.Int63()+1`?

storage/tree_validation.go
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"fmt"
+	"github.com/google/trillian"
```
AlCutter: separate sys & remote imports with a blank line

storage/mysql/admin_storage.go
codingllama: Done.

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mysql
+
+import (
+	"database/sql"
+	"fmt"
+	"github.com/google/trillian"
```
codingllama: Done.

Hmm, I thought gofmt did that. Ran goimports on everything.

storage/mysql/admin_storage.go
```diff
@@ -0,0 +1,394 @@
...
+	"time"
+)
+
+const (
+	defaultSequenceIntervalSeconds = 60
+	defaultSignIntervalSeconds     = 60
+)
+
+var (
+	selectTrees = `
```
codingllama: Done.

storage/mysql/admin_storage.go
codingllama: Done.

storage/mysql/admin_storage.go
codingllama: Inverting the ifs: could do. I decided against testing the negative, which I find a bit weird if there's an else block as well. I don't have a strong feel which is better (or "less worse") in the context of Go, so let know if strongly prefer one.

Adding another block: the intent behind the if/else chains is to limit the scope of the variables. Since pretty much everything here is an int32 being cast to an enum, and all enums have similar ranges (most between 0-2), I'm a bit paranoid of mismatching them (which could easily happen in a future copy/paste).

storage/mysql/admin_storage.go
codingllama: Added TODO to (possibly) kill it.

storage/mysql/admin_storage.go
codingllama: Done. Thanks.

storage/mysql/admin_storage.go
codingllama: Datetime stops at seconds and trillian.Tree has dates in millis_since_epoch. Nanos won't really help here.

(Plus we don't write the int64 value, we write the date-as-string one.)

storage/mysql/admin_storage_test.go
codingllama: Done.

storage/mysql/admin_storage_test.go
codingllama: Done.

storage/testonly/admin_storage_tester.go
codingllama: I had one at some point, but the index works fine for a small number of tests, so I removed it in favor of this.

storage/tree_validation.go
codingllama: Done.

storage/testonly/admin_storage_tester.go
codingllama: I'm using Fatal for test setup failures and Error for everything else. The idea is that failures on test setup are either really bad (like a broken tx.Begin) or trace back to bugs that should be handled by other tests, so the whole loop goes down with them.

storage/tree_id.go
codingllama: Discussed offline.

TL;DR: I don't want to use math/rand, pick some seed and risk IDs being predictable. It doesn't seem like it would be a problem anyway, but I think we get a better mileage from crypto/rand regardless.

storage/testonly/admin_storage_tester.go
codingllama: https://goo.gl/IdpVB7

codingllama:
PTAL?

PR #373, opened 2017-02-09
Parallelise sequencing runs

AlCutter:
Just a quick hack to make things happen a little quicker.

### DIFF
--- a/server/sequencer_manager.go
+++ b/server/sequencer_manager.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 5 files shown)


server/sequencer_manager.go
```diff
@@ -54,42 +55,55 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
+		wg.Add(1)
+		logID := logID
+		go func() {
+			defer wg.Done()
+
+			start := time.Now()
+
+			// TODO(Martin2112): Honor the sequencing enabled in log parameters, needs an API change
+			// so deferring it
+			storage, err := s.registry.GetLogStorage()
```
codingllama: nit: Move this out of the loop?

It seems we could move hasher and sequencer out too.

server/sequencer_manager.go
```diff
@@ -54,42 +55,55 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
-
-		leaves, err := sequencer.SequenceBatch(ctx, logID, logctx.batchSize)
-		if err != nil {
-			glog.Warningf("%v: Error trying to sequence batch for: %v", logID, err)
-			continue
-		}
-
-		successCount++
-		leavesAdded += leaves
+		wg.Add(1)
```
codingllama: Should we have an upper bound on the number of goroutines?

server/sequencer_manager.go
```diff
@@ -54,42 +55,55 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
+		}()
 	}
 
-	glog.V(1).Infof("Sequencing run completed %v succeeded %v failed %v leaves integrated", successCount, len(logIDs)-successCount, leavesAdded)
+	wg.Wait()
+
+	glog.V(1).Infof("Sequencing group run completed %v succeeded %v failed %v leaves integrated", successCount, len(logIDs)-successCount, leavesAdded)
+
+	select {
+	case <-logctx.ctx.Done():
```
codingllama: A couple comments here:

1. Neither the loop above nor goroutines check logctx.ctx.Done() anymore, which means they'll keep working regardless of cancellation
2. It seems strange that we basically return ctx.Done() as a result of the method. Should this be something like "return len(logIDs) == successCount" instead?
3. Borrowing from 2., could you document what the return value is meant to represent for this method?

server/sequencer_manager.go
```diff
@@ -54,42 +55,55 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
+			sequencer := log.NewSequencer(hasher, logctx.timeSource, storage, s.keyManager)
+			sequencer.SetGuardWindow(s.guardWindow)
+
+			leaves, err := sequencer.SequenceBatch(ctx, logID, logctx.batchSize)
+			if err != nil {
+				glog.Warningf("%v: Error trying to sequence batch for: %v", logID, err)
+				return
+			}
+			d := time.Now().Sub(start).Seconds()
+			glog.Infof("%v: sequenced %d leaves in %.2f seconds (%d qps)", logID, leaves, d, float64(leaves)/d)
```
codingllama: Would it be more meaningful to time the entire run?

server/sequencer_manager.go
```diff
@@ -54,42 +55,55 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
...
+
+			successCount++
```
codingllama: Protect writes to successCount and leavesAdded with a mutex or sync/atomic?

server/sequencer_manager.go
AlCutter: Done

server/sequencer_manager.go
AlCutter: Done, I left `hasher` and `sequencer` inside one depends on the other and, in theory at least, `hasher` could be different for each tree.

server/sequencer_manager.go
AlCutter: I've added timing for the whole run too; I think both are interesting.

server/sequencer_manager.go
AlCutter: done

server/sequencer_manager.go
AlCutter: I've ripped the whole return value out for now, it doesn't really make any sense to have it here any longer.

AlCutter:
I was going for _minimum viable line-count_ with my hack as I think this whole area really needs to be revisited now that we know more (it was perfectly fine to get us up and running though), but I've added the extra bits 'cos I'm nice like that :)

codingllama:
It is "minimum viable line-count" now :P

(Thanks for the changes.)

PR #377, opened 2017-02-14
Provide crypto.KeyManager through extension.Registry

RJPercival:
Allows crypto keys to be obtained from different sources, depending on the registry implementation.

### DIFF
--- a/server/sequencer_manager_test.go
+++ b/server/sequencer_manager_test.go
@@ -76,13 +76,10 @@ func TestSequencerManagerNothingToDo(t *testing.T) {
 	defer mockCtrl.Finish()
 
 	mockStorage := storage.NewMockLogStorage(mockCtrl)
-	mockKeyManager := crypto.NewMockKeyManager(mockCtrl)
-	mockKeyManager.EXPECT().SignatureAlgorithm().AnyTimes().Return(sigpb.DigitallySigned_ECDSA)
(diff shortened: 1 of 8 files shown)


server/sequencer_manager_test.go
```diff
@@ -169,8 +170,19 @@ func mockStorageProviderForSequencer(mockStorage storage.LogStorage) testonly.Ge
 	}
 }
 
-func registryForSequencer(mockStorage storage.LogStorage) extension.Registry {
-	return testonly.NewRegistryWithLogProvider(mockStorageProviderForSequencer(mockStorage))
+func mockKeyManagerProviderForSequencer(mockKM crypto.KeyManager) testonly.GetKeyManagerFunc {
+	return func(id int64) (crypto.KeyManager, error) {
+		if mockKM == nil {
+			return nil, fmt.Errorf("mock KeyManager is nil")
```
AlCutter: can just use errors.New(...) here since there are no args.

AlCutter:
Did you mean to move the `testonly/keys.go` under `crypto/`?
You'll probably need a `crypto/testonly` package if you want to do that (currently you have two packages in the `crypto/` dir)

server/sequencer_manager_test.go
RJPercival: Done.

RJPercival:
Travis is green.

RJPercival:
Rebased following merge of PR #387, which addressed the import cycle. PTAL.

PR #378, opened 2017-02-14
Delete some obsolete documentation

RJPercival:
While addressing issue #324, the treeID parameter in the extension.Registry interface and the cachedRegistry were removed. However, not all of the documentation referred to them was.


codingllama:
I've restarted the Travis build. Please make sure it's green before merging.

PR #379, opened 2017-02-14
Move DigitallySigned to its own proto package.

gdbelvin:
This allows external libraries to depend on it and enforces
better decoupling in the code base.


crypto/key_manager.go
```diff
@@ -25,7 +25,7 @@ import (
 	"io/ioutil"
 
 	"github.com/golang/glog"
-	"github.com/google/trillian"
+	spb "github.com/google/trillian/proto/signature"
```
gdbelvin: No strict conflicts.  I've had a habit of naming protobuf packages *pb for code clarity.

crypto/key_manager.go
```diff
@@ -38,9 +38,9 @@ type KeyManager interface {
 	// manager.
 	Signer() (crypto.Signer, error)
 	// SignatureAlgorithm returns the value that identifies the signature algorithm.
-	SignatureAlgorithm() trillian.SignatureAlgorithm
+	SignatureAlgorithm() spb.DigitallySigned_SignatureAlgorithm
 	// HashAlgorithm returns the type of hash that will be used for signing with this key.
-	HashAlgorithm() trillian.HashAlgorithm
+	HashAlgorithm() crypto.Hash
```
gdbelvin: This was only being used to pass the hash algorithm to the Signature library.

gdbelvin: `hashers.go` is no longer needed since the API now accepts `crypto.Hash` directly.

crypto/hashers.go
```diff
@@ -1,40 +0,0 @@
...
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-package crypto
-
-import (
-	"crypto"
-	_ "crypto/sha256" // Register the SHA256 algorithm
```
gdbelvin: sha256 is registered in the signature library where it is used.

crypto/signer.go
```diff
@@ -34,48 +35,58 @@ const (
...
-	h, ok := hashLookup[hashAlgo]
+func NewSigner(hashAlgo crypto.Hash, sigAlgo spb.DigitallySigned_SignatureAlgorithm, signer crypto.Signer) *Signer {
+	_, ok := reverseSignerHashLookup[hashAlgo]
 	if !ok {
 		// TODO(gbelvin): return error from Signer.
 		panic("unsupported hash algorithm")
 	}
 
-	return &Signer{h, signer, sigAlgo}
+	return &Signer{hashAlgo, signer, sigAlgo}
```
gdbelvin: Good call.

crypto/key_manager.go
Martin2112: I'm surprised it wasn't being used elsewhere and that nobody needs to use the maps.

crypto/key_manager.go
Martin2112: As we've done with all ours except the top level one. What did you think about moving it inside the relevant package and calling it somethingpb?

crypto/signer.go
gdbelvin: done

crypto/hashers.go
```diff
@@ -1,40 +0,0 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
```
gdbelvin: This delete is appropriate because I've moved the relevant hash enums into `DigitallySigned` 
The replacement lookup dictionaries are in `signature.go` 

`HashingStrategy` is entirely different.

gdbelvin:
@Martin2112 PTAL

crypto/key_manager.go

crypto/key_manager.go
gdbelvin: This is different than `HashingStrategy` which is used in maps. 
This hash function is only the one used for signing.

Martin2112:
In that case could we pass the hasher directly to NewSigner instead of
looking it up?

gdbelvin:
Yes, in this PR, we start passing the proper hasher directly to the Signer.

crypto/key_manager.go
```diff
@@ -38,9 +38,9 @@ type KeyManager interface {
 	// manager.
 	Signer() (crypto.Signer, error)
 	// SignatureAlgorithm returns the value that identifies the signature algorithm.
-	SignatureAlgorithm() trillian.SignatureAlgorithm
+	SignatureAlgorithm() sigpb.DigitallySigned_SignatureAlgorithm
 	// HashAlgorithm returns the type of hash that will be used for signing with this key.
-	HashAlgorithm() trillian.HashAlgorithm
+	HashAlgorithm() crypto.Hash
```
Martin2112: This doesn't return an algorithm any more so is potentially confusing. Can you TODO it for rework later.

crypto/signer.go
```diff
@@ -34,48 +35,62 @@ const (
...
+	reverseSignerHashLookup = map[crypto.Hash]sigpb.DigitallySigned_HashAlgorithm{
+		crypto.SHA256: sigpb.DigitallySigned_SHA256,
+	}
+)
+
 // Signer is responsible for signing log-related data and producing the appropriate
 // application specific signature objects.
 type Signer struct {
-	hasher       crypto.Hash
+	hash         crypto.Hash
```
Martin2112: Hasher was OK for this, either will do.

PR #386, opened 2017-02-16
Client: Verify consistency proofs

gdbelvin:
Add consistency proof checking between STR updates. 
- Add tests that inject errors into responses for testing error conditions. 
- Moves responsibility for computing MerkleLeafHash to the client.

Contributes to #297

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -109,12 +109,35 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 		LogId: c.LogID,
 	}
 	resp, err := c.client.GetLatestSignedLogRoot(ctx, req)
+	str := resp.SignedLogRoot
 	if err != nil {
 		return err
 	}
-	// TODO(gdbelvin): Verify SignedLogRoot
+	// TODO(gdbelvin): Verify SignedLogRoot Signature
 
-	c.STR = *resp.SignedLogRoot
+	// Implicitly trust the first STH we get.
+	if c.STR.TreeSize != 0 {
+		// Get consistency proof.
+		req := &trillian.GetConsistencyProofRequest{
+			LogId:          c.LogID,
+			FirstTreeSize:  c.STR.TreeSize,
+			SecondTreeSize: str.TreeSize,
+		}
+		proof, err := c.client.GetConsistencyProof(ctx, req)
+		if err != nil {
+			return err
+		}
+		// Verify consistency proof.
+		v := merkle.NewLogVerifier(c.hasher)
+		if err := v.VerifyConsistencyProof(
+			c.STR.TreeSize, str.TreeSize,
+			c.STR.RootHash, str.RootHash,
+			convertProof(proof.Proof)); err != nil {
+			return err
+		}
+	}
+
+	c.STR = *str
 	return nil
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -18,6 +18,7 @@ package client
 import (
 	"bytes"
 	"context"
+	"crypto"
 	"crypto/sha256"
 	"errors"
 	"time"
@@ -25,9 +26,12 @@ import (
 	"github.com/google/trillian"
 	"github.com/google/trillian/client/backoff"
 	"github.com/google/trillian/merkle"
+	"github.com/google/trillian/merkle/rfc6962"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
+
+var treeHasher = rfc6962.TreeHasher{Hash: crypto.SHA256}
 
 // LogClient represents a client for a given Trillian log instance.
 type LogClient struct {
@@ -61,7 +65,7 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
 	switch {
 	case grpc.Code(err) == codes.AlreadyExists:
 		// If the leaf already exists, don't wait for an update.
-		return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.STR.TreeSize)
+		return c.getInclusionProof(ctx, treeHasher.HashLeaf(leaf.LeafValue), c.STR.TreeSize)
 	case err != nil:
 		return err
 	default:
@@ -185,7 +189,6 @@ func buildLeaf(data []byte) *trillian.LogLeaf {
 	hash := sha256.Sum256(data)
 	leaf := &trillian.LogLeaf{
 		LeafValue:        data,
-		MerkleLeafHash:   hash[:],
 		LeafIdentityHash: hash[:],
 	}
 	return leaf
--- a/client/client.go
+++ b/client/client.go
@@ -117,7 +117,7 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 	}
 	// TODO(gdbelvin): Verify SignedLogRoot Signature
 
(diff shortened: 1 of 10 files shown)


gdbelvin:
Looks like the server is returning valid consistency proofs!

codingllama:
I mean, could we actually give it a bad proof and have a test that checks for failure? The code without verification would pass the tests much like this one.

gdbelvin:
Hmm... That would mean mocking out the server to produce incorrect results,
or intercepting valid results and mucking with them.
Would you prefer that I build a fuzzing framework to test this?

codingllama:
It doesn't need to be a fuzzing test, although that could be interesting in other areas. A mock TrillianLogClient is what I had in mind.

gdbelvin:
PTAL

client/client_test.go
```diff
@@ -43,16 +43,42 @@ func TestAddLeaf(t *testing.T) {
...
 
-	if err, want := client.AddLeaf(ctx, []byte("foo")), codes.DeadlineExceeded; grpc.Code(err) != want {
-		t.Errorf("AddLeaf(): %v, want, %v", err, want)
-	}
-	env.Sequencer.OperationLoop() // Sequence the new node.
-	glog.Infof("try AddLeaf again")
-	if err := client.AddLeaf(ctx, []byte("foo")); err != nil {
-		t.Errorf("Failed to add Leaf: %v", err)
+		if err, want := client.AddLeaf(ctx, []byte(test.desc)), codes.DeadlineExceeded; grpc.Code(err) != want {
+			t.Errorf("AddLeaf(%v): %v, want, %v", test.desc, err, want)
```
codingllama: Add a `continue` after t.Errorf? It doesn't like the current test could go on after this.

examples/ct/handlers.go
```diff
@@ -81,8 +82,11 @@ const (
 	GetEntryAndProofName  = EntrypointName("GetEntryAndProof")
 )
 
-// Entrypoints is a list of entrypoint names as exposed in statistics/logging.
-var Entrypoints = []EntrypointName{AddChainName, AddPreChainName, GetSTHName, GetSTHConsistencyName, GetProofByHashName, GetEntriesName, GetRootsName, GetEntryAndProofName}
+var (
+	// Entrypoints is a list of entrypoint names as exposed in statistics/logging.
+	Entrypoints   = []EntrypointName{AddChainName, AddPreChainName, GetSTHName, GetSTHConsistencyName, GetProofByHashName, GetEntriesName, GetRootsName, GetEntryAndProofName}
+	treeHasher, _ = merkle.Factory(merkle.RFC6962SHA256Type)
```
codingllama: nit: Move the merkle.Factory() call to a `getDefaultHasher` method and have it panic if `err != nil`? Otherwise, if it errors, we won't know until the system starts getting null pointers from treeHasher.

server/log_rpc_server.go
```diff
@@ -78,12 +78,6 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
 	}
 	leaves := depointerify(req.Leaves)
 
-	// TODO(al): Hasher must be selected based on log config.
-	th, _ := merkle.Factory(merkle.RFC6962SHA256Type)
```
codingllama: Why isn't this necessary anymore? Do we expect the personality to hash, or is it happening in another layer already?

gdbelvin: I think we should make the personality supply all the data in `LogLeaf`. 
This allows personalities to define their own leaf hashing algorithms without needing to change server code. What do you think?

codingllama: I don't have a strong opinion, but I don't think I completely understand what the implications might be either.

This is straying a bit out of the purpose of this PR, would you mind reverting and maybe putting a separate PR together for this change? I think this should be discussed with more people in the team.

client/client.go
```diff
@@ -115,6 +117,11 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 	}
 	// TODO(gdbelvin): Verify SignedLogRoot Signature
 
+	if str.TreeSize == c.STR.TreeSize && bytes.Equal(str.RootHash, str.RootHash) {
```
gdbelvin: yep. done

client/client_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+// QueueLeaves forwards requests.
+func (c *MockLogClient) QueueLeaves(ctx context.Context, in *trillian.QueueLeavesRequest, opts ...grpc.CallOption) (*trillian.QueueLeavesResponse, error) {
+	return c.c.QueueLeaves(ctx, in)
+}
+
+// GetInclusionProof forwards requests and modifies the response.
+func (c *MockLogClient) GetInclusionProof(ctx context.Context, in *trillian.GetInclusionProofRequest, opts ...grpc.CallOption) (*trillian.GetInclusionProofResponse, error) {
+	resp, err := c.c.GetInclusionProof(ctx, in)
+	if c.mGetInclusionProof {
+		resp.Proof.ProofNode[3].NodeHash[4] ^= 3
```
gdbelvin: done

client/client_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return resp, err
+}
+
+// GetInclusionProofByHash forwards requests.
+func (c *MockLogClient) GetInclusionProofByHash(ctx context.Context, in *trillian.GetInclusionProofByHashRequest, opts ...grpc.CallOption) (*trillian.GetInclusionProofByHashResponse, error) {
+	resp, err := c.c.GetInclusionProofByHash(ctx, in)
+	if c.mGetInclusionProof {
+		log.Print(resp)
```
gdbelvin: done

client/client_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+		resp.Proof[0].ProofNode[0].NodeHash[4] ^= 3
```
gdbelvin: done

client/client_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return resp, err
+}
+
+// GetConsistencyProof forwards requests and modifies responses.
+func (c *MockLogClient) GetConsistencyProof(ctx context.Context, in *trillian.GetConsistencyProofRequest, opts ...grpc.CallOption) (*trillian.GetConsistencyProofResponse, error) {
+	resp, err := c.c.GetConsistencyProof(ctx, in)
+	if c.mGetConsistencyProof {
+		log.Print(resp)
```
gdbelvin: done

client/client_test.go

examples/ct/handlers.go
gdbelvin: I'll just use a method that can't fail

server/log_rpc_server.go
gdbelvin: I'll create a separate PR

gdbelvin:
Done PTAL

examples/ct/handlers.go
codingllama: Even better.

server/log_rpc_server.go
codingllama: Thanks.

PR #387, opened 2017-02-17
Move testonly/extension.go to server/test_registry.go

RJPercival:
Avoids an import cycle that prevents trillian/{extension,storage} and soon trillian/crypto from importing  trillian/testonly. Since testRegistry is only used by the server package, moved it to there and simplified the
code accordingly.


server/test_registry.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/google/trillian/storage"
+)
+
+type GetLogStorageFunc func() (storage.LogStorage, error)
+type GetMapStorageFunc func() (storage.MapStorage, error)
+
+// testRegistry implements extension.Registry.
+// It delegates its implementation to its member funcs.
+// It is intended as a stub for testing.
+type testRegistry struct {
```
codingllama: I wonder if we could get away with just using a MockExtensionRegistry in the tests above.

Feel free to push back on this.

RJPercival: Done.

PR #388, opened 2017-02-17
Use pointers to return empty values on err

gdbelvin:
I believe it is canonical Go to return pointers to objects, especially when a function can error.

### DIFF
--- a/examples/ct/handlers.go
+++ b/examples/ct/handlers.go
@@ -310,7 +310,7 @@ func addChainInternal(ctx context.Context, c LogContext, w http.ResponseWriter, 
 
 	// As the Log server has successfully queued up the Merkle tree leaf, we can
 	// respond with an SCT.
-	err = marshalAndWriteAddChainResponse(*sct, c.logKeyManager, w)
+	err = marshalAndWriteAddChainResponse(sct, c.logKeyManager, w)
 	if err != nil {
 		// reason is logged and http status is already set
 		// TODO(Martin2112): Record failure for monitoring when it's implemented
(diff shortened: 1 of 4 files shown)


examples/ct/handlers.go
```diff
@@ -310,7 +310,7 @@ func addChainInternal(ctx context.Context, c LogContext, w http.ResponseWriter,
 
 	// As the Log server has successfully queued up the Merkle tree leaf, we can
 	// respond with an SCT.
-	err = marshalAndWriteAddChainResponse(sct, c.logKeyManager, w)
+	err = marshalAndWriteAddChainResponse(*sct, c.logKeyManager, w)
```
codingllama: nit: Change marshalAndWriteAddChainResponse to take a pointer?

examples/ct/handlers_test.go
```diff
@@ -1221,8 +1221,8 @@ func createJSONChain(t *testing.T, p PEMCertPool) io.Reader {
 	return bufio.NewReader(&buffer)
 }
 
-func logLeavesForCert(t *testing.T, km crypto.KeyManager, certs []*x509.Certificate, merkleLeaf ct.MerkleTreeLeaf, isPrecert bool) []*trillian.LogLeaf {
-	leafData, err := tls.Marshal(merkleLeaf)
+func logLeavesForCert(t *testing.T, km crypto.KeyManager, certs []*x509.Certificate, merkleLeaf *ct.MerkleTreeLeaf, isPrecert bool) []*trillian.LogLeaf {
```
codingllama: nit: There doesn't seem to be much benefit in this particular change, maybe revert?

examples/ct/handlers.go
gdbelvin: done

examples/ct/handlers_test.go
gdbelvin: Done

PR #389, opened 2017-02-17
Make KeyManager appropriate for private keys

### LINKED ISSUE #381: KeyManager: Perform all key verification during New* functions
The KeyManager interface currently supports returning `error` for several of the Get* functions. 
This dramatically increases code complexity for calling functions.  All these error params can be eliminated if the New* functions require a valid key to be loaded before returning a KeyManager object. Is there a strong reason to support starting Trillian without key material? 

```
type KeyManager interface {
	Signer() (crypto.Signer, error)
	SignatureAlgorithm() spb.DigitallySigned_SignatureAlgorithm
	HashAlgorithm() crypto.Hash
	GetPublicKey() (crypto.PublicKey, error)
	GetRawPublicKey() ([]byte, error)
}
```

Proposed interface:
```
New(...) (KeyManager, error)
type KeyManager interface {
	Signer() crypto.Signer
	SignatureAlgorithm() spb.DigitallySigned_SignatureAlgorithm
	HashAlgorithm() crypto.Hash
	GetPublicKey() crypto.PublicKey
	GetRawPublicKey() []byte
}
```

gdbelvin:
- Check key validity when creating the key manager.
- Separate public key functions into the verifier.
- Remove extraneous error conditions.

Closes #381


crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
+}
 
-	k.serverPublicKey = parsedKey
-	return nil
+// NewFromPrivateKey creates PrivateKeyManager using a private key.
+func NewFromPrivateKey(key crypto.PrivateKey) (PrivateKeyManager, error) {
+	var signer crypto.Signer
+	var sigAlgo sigpb.DigitallySigned_SignatureAlgorithm
+
+	switch key.(type) {
```
RJPercival: Why not `switch key := key.(type) {`? That would eliminate the need for type assertions on lines 79 and 82.

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
+	case *ecdsa.PrivateKey:
+		signer = key.(crypto.Signer)
+		sigAlgo = sigpb.DigitallySigned_ECDSA
+	case *rsa.PrivateKey:
+		signer = key.(crypto.Signer)
+		sigAlgo = sigpb.DigitallySigned_RSA
+	default:
+		return nil, errors.New("unsupported key type")
```
RJPercival: Might be helpful to include the type of the key in the error message, e.g. `fmt.Errorf("unsupported key type: %T", key)`

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
+	}
+
+	return &PEMKeyManager{
```
RJPercival: Could duplicate the return in each case statement and get rid of the signer and sigAlgo vars.

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
-// not been previously called successfully.
-func (k PEMKeyManager) GetPublicKey() (crypto.PublicKey, error) {
-	if k.serverPublicKey == nil {
-		return nil, errors.New("called GetPublicKey() but one is not loaded")
+	if key, err := x509.ParseECPrivateKey(key); err == nil {
+		return key, nil
 	}
-
-	return k.serverPublicKey, nil
+	return nil, errors.New("could not parse private key")
```
RJPercival: This discards all of the error messages from the parsing functions, which could potentially be helpful.

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
-	return nil, sigpb.DigitallySigned_ANONYMOUS, errors.New("could not parse private key")
+	key, err := parsePrivateKey(der)
+	if err != nil {
+		return nil, err
+	}
+	return NewFromPrivateKey(key)
 }
 
-// LoadPasswordProtectedPrivateKey initializes and returns a new KeyManager using a PEM encoded
+// NewFromPrivatePEMFile initializes and returns a new KeyManager using a PEM encoded
```
RJPercival: s/KeyManager/PrivateKeyManager/

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
+	return nil, errors.New("could not parse private key")
 }
 
-// GetRawPublicKey returns the DER encoded public key bytes as loaded from the file.
-// This is needed for some applications that exchange or embed key hashes in structures.
-// The result will be an error if a public key has not been loaded
-func (k PEMKeyManager) GetRawPublicKey() ([]byte, error) {
-	if k.rawPublicKey == nil {
-		return nil, errors.New("called GetRawPublicKey() but one is not loaded")
+// NewFromPrivatePEM returns key manager for a password protected PEM object.
```
RJPercival: The PEM need not necessarily be password protected.

crypto/key_manager.go
```diff
@@ -24,50 +24,27 @@ import (
...
-	// This is needed for some applications that exchange or embed key hashes in structures.
-	// It is an error to call this before a public key has been loaded
-	GetRawPublicKey() ([]byte, error)
+	// PublicKey returns the public key corresponding to the private key.
+	PublicKey() crypto.PublicKey
 }
 
 // PEMKeyManager is an instance of KeyManager that loads its key data from an encrypted
 // PEM file.
 type PEMKeyManager struct {
```
RJPercival: I don't think PEMKeyManager actually has anything to do with PEM any more. All of the PEM processing has been moved to free functions. Consider renaming it.

crypto/verifier.go
```diff
@@ -29,6 +31,24 @@ import (
...
+	if publicBlock == nil {
+		return nil, errors.New("could not decode PEM for public key")
+	}
+	if len(rest) > 0 {
+		return nil, errors.New("extra data found after PEM key decoded")
+	}
+
+	parsedKey, err := x509.ParsePKIXPublicKey(publicBlock.Bytes)
+	if err != nil {
+		return nil, errors.New("unable to parse public key")
```
RJPercival: Include `err` in returned error message.

crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
 }
 
-// GetRawPublicKey returns the DER encoded public key bytes as loaded from the file.
-// This is needed for some applications that exchange or embed key hashes in structures.
-// The result will be an error if a public key has not been loaded
-func (k PEMKeyManager) GetRawPublicKey() ([]byte, error) {
-	if k.rawPublicKey == nil {
-		return nil, errors.New("called GetRawPublicKey() but one is not loaded")
+// NewFromPrivatePEM returns key manager for a password protected PEM object.
+func NewFromPrivatePEM(pemBlock []byte, password string) (PrivateKeyManager, error) {
```
RJPercival: `NewFromPrivatePEM()` takes its PEM as `[]byte`, but `PublicKeyFromPEM()` takes its PEM as `string`. Choose one for consistency. I slightly prefer `string`, as PEM is ASCII and it allows passing `const string` vars directly to the func without casting.

examples/ct/handlers_test.go
```diff
@@ -88,25 +88,36 @@ LPd2G+er1/w5wxpM/hvZbWc0yoLyLd5aDIu73YJde28+dhKtjbMAp+IRaYhgIyYi
...
 	info.mockCtrl = gomock.NewController(t)
-	info.km = crypto.NewMockKeyManager(info.mockCtrl)
-	info.km.EXPECT().GetRawPublicKey().AnyTimes().Return([]byte("key"), nil)
+
+	info.km = crypto.NewMockPrivateKeyManager(info.mockCtrl)
+	pubkey, err := crypto.PublicKeyFromPEM(ctTesttubePublicKey)
+	if err != nil {
+		panic(err)
+	}
+	info.km.EXPECT().PublicKey().AnyTimes().Return(pubkey)
```
RJPercival: I've noticed that some of our mock expectations are completely unnecessary. By saying `AnyTimes()`, the method doesn't ever have to be called during the test. I'm not sure whether this is a good or bad thing. Unnecessary code is generally bad, but I suppose this makes the tests more resilient to changes. Fundamentally, we're not using these mocks as mocks a lot of the time, but as stubs instead, which is probably where this oddity comes from.

examples/ct/handlers_test.go
```diff
@@ -346,8 +357,8 @@ func TestAddChain(t *testing.T) {
 			if got, want := ct.Version(resp.SCTVersion), ct.V1; got != want {
 				t.Errorf("resp.SCTVersion=%v; want %v", got, want)
 			}
-			if got, want := hex.EncodeToString(resp.ID), ctMockLogID; got != want {
```
RJPercival: I suspect it was done this way to produce more readable error messages. I don't have a strong opinion on this though.

crypto/key_manager.go
gdbelvin: In general, fewer exit conditions (ie return statements) are better for code readability and maintenance. Do you have a strong preference on this?

examples/ct/serialize_test.go
```diff
@@ -68,7 +60,7 @@ func TestSignV1SCTForCertificate(t *testing.T) {
 	}
 
 	if !reflect.DeepEqual(got, expected) {
-		t.Fatalf("Mismatched SCT (cert), got %v, expected %v", got, expected)
+		t.Fatalf("Mismatched SCT (cert), got \n%v, expected \n%v", got, expected)
```
RJPercival: Using something like https://github.com/d4l3k/messagediff would give the nicest failure messages, but I don't expect you to add that in this PR.

examples/ct/structures_test.go
```diff
@@ -28,49 +27,33 @@ import (
...
-var fixedTime = time.Date(2017, 9, 7, 12, 15, 23, 0, time.UTC)
-
-// Public key for Google Testtube log, taken from CT Github repository
-const ctTesttubePublicKey string = `
+var (
+	fixedTime       = time.Date(2017, 9, 7, 12, 15, 23, 0, time.UTC)
+	ctTesttubeLogID = [32]byte{176, 204, 131, 229, 165, 249, 125, 107, 175, 124, 9, 204,
+		40, 73, 4, 135, 42, 199, 232, 139, 19, 44, 99, 80, 183, 198, 253, 38, 225, 108, 108, 119}
+	// Public key for Google Testtube log, taken from CT Github repository
+	ctTesttubePublicKey = `
```
RJPercival: Did you intend to change `ctTesttubePublicKey` from const to var?

examples/ct/structures_test.go
```diff
@@ -106,21 +89,25 @@ func TestSerializeLogEntry(t *testing.T) {
 }
 
 // Creates a mock key manager for use in interaction tests
-func setupMockKeyManager(ctrl *gomock.Controller, toSign []byte) *crypto.MockKeyManager {
-	mockKeyManager := setupMockKeyManagerForSth(ctrl, toSign)
-	mockKeyManager.EXPECT().GetRawPublicKey().AnyTimes().Return([]byte("key"), nil)
+func setupMockPrivateKeyManager(ctrl *gomock.Controller, toSign []byte) *crypto.MockPrivateKeyManager {
+	pubkey, err := crypto.PublicKeyFromPEM(ctTesttubePublicKey)
+	if err != nil {
+		panic(err)
```
RJPercival: Panicking, even in tests, is generally prohibited. Issues include that it causes tabular tests to abort early. Just return the error and fail the test.

crypto/key_manager.go
RJPercival: No, I'm fine with leaving this as if. Any comment that starts "could" or "might" is optional. I think fewer variables are also better for readability and maintenance, but it really doesn't make much difference here. I'll just point out that almost all of the examples of switches in Effective Go use my proposed pattern though: https://golang.org/doc/effective_go.html#switch

crypto/key_manager.go
```diff
@@ -24,50 +24,27 @@ import (
...
-	// manager.
-	Signer() (crypto.Signer, error)
+// PrivateKeyManager supports signing data with a private key that may be stored
+// in a secure location, which is not immediately available to this client.
+type PrivateKeyManager interface {
+	// Signer returns a crypto.Signer that can sign data using the private key.
+	Signer() crypto.Signer
 	// SignatureAlgorithm returns the value that identifies the signature algorithm.
 	SignatureAlgorithm() sigpb.DigitallySigned_SignatureAlgorithm
 	// HashAlgorithm returns the type of hash that will be used for signing with this key.
```
gdbelvin: Removed HashAlgorithm

crypto/key_manager.go
gdbelvin: Renamed to `LocalSigner`

crypto/key_manager.go

crypto/key_manager.go

crypto/key_manager.go
gdbelvin: Love the example. You've convinced me.

crypto/key_manager.go

crypto/key_manager.go
gdbelvin: It's not clear which error to return. Do you have a preference?

crypto/key_manager.go

crypto/verifier.go
gdbelvin: done

crypto/key_manager.go

examples/ct/handlers_test.go
gdbelvin: Yep. I agree, we're using them as stubs. 

I'd prefer to remove the mocks and properly sign things.

examples/ct/handlers_test.go
gdbelvin: In general, it's probably better to use `fmt.Errorf("%x", data)` when printing data for the screen.

examples/ct/structures_test.go
gdbelvin: Yes, purely for aesthetic purposes since it was the only const, and the associated vars - `ctTesttubeLogID` need to be a var because it is a slice.

examples/ct/structures_test.go
gdbelvin: done

gdbelvin:
Rebased and implemented the requested changes. PTAL

crypto/key_manager.go
```diff
@@ -36,38 +36,33 @@ type PrivateKeyManager interface {
 	SignatureAlgorithm() sigpb.DigitallySigned_SignatureAlgorithm
 }
 
-// PEMKeyManager is an instance of KeyManager that loads its key data from an encrypted
-// PEM file.
-type PEMKeyManager struct {
+// LocalSigner signs objects using in-memory key material.
+type LocalSigner struct {
```
RJPercival: I don't think you need to export LocalSigner.

crypto/key_manager.go
```diff
@@ -84,8 +79,8 @@ func parsePrivateKey(key []byte) (crypto.PrivateKey, error) {
 }
 
 // NewFromPrivatePEM returns key manager for a password protected PEM object.
```
RJPercival: The comment here needs changing to make it obvious that the PEM doesn't *need* to be password-protected.

RJPercival: For example, `NewFromPrivateKey()` says "The key **may** be protected by a password".

crypto/key_manager.go
RJPercival: You could return all 3, e.g. `fmt.Errorf("could not parse private key as PKCS1 (%v), PKCS8 (%v) or SEC1 (%v)", err1, err2, err3)`

RJPercival:
A couple of comments left to address, but then it'll LGTM. I recommend using the "squash and merge" option when you're finished, to keep the master branch clean of "nit" and "fix rebase" commits.

crypto/key_manager.go

crypto/key_manager.go

crypto/key_manager.go

gdbelvin:
Done. PTAL

PR #390, opened 2017-02-17
Client: Working signature verification

gdbelvin:
This adds signature verification to the client.

It also separates `HashLogRoot` from `crypto/signer.go` since it is not generic. 
It also adds a generic `SignObject`, `VerifyObject` functions.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -121,7 +121,7 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 	str := resp.SignedLogRoot
 
 	// Verify SignedLogRoot signature.
-	hash := crypto.HashTrillianSignedLogRoot(*str)
+	hash := crypto.HashLogRoot(*str)
 	if err := crypto.Verify(c.pubKey, hash, str.Signature); err != nil {
 		return err
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -18,20 +18,18 @@ package client
 import (
 	"bytes"
 	"context"
-	"crypto"
+	gocrypto "crypto"
 	"crypto/sha256"
 	"errors"
 	"time"
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/client/backoff"
+	"github.com/google/trillian/crypto"
 	"github.com/google/trillian/merkle"
-	"github.com/google/trillian/merkle/rfc6962"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
-
-var treeHasher = rfc6962.TreeHasher{Hash: crypto.SHA256}
 
 // LogClient represents a client for a given Trillian log instance.
 type LogClient struct {
@@ -40,15 +38,17 @@ type LogClient struct {
 	hasher   merkle.TreeHasher
 	STR      trillian.SignedLogRoot
 	MaxTries int
+	pubKey   gocrypto.PublicKey
 }
 
 // New returns a new LogClient.
-func New(logID int64, client trillian.TrillianLogClient, hasher merkle.TreeHasher) *LogClient {
+func New(logID int64, client trillian.TrillianLogClient, hasher merkle.TreeHasher, pubKey gocrypto.PublicKey) *LogClient {
 	return &LogClient{
 		LogID:    logID,
 		client:   client,
 		hasher:   hasher,
 		MaxTries: 3,
+		pubKey:   pubKey,
 	}
 }
 
@@ -60,12 +60,12 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
 		return err
 	}
 
-	leaf := buildLeaf(data)
+	leaf := c.buildLeaf(data)
 	err := c.queueLeaf(ctx, leaf)
 	switch {
 	case grpc.Code(err) == codes.AlreadyExists:
 		// If the leaf already exists, don't wait for an update.
-		return c.getInclusionProof(ctx, treeHasher.HashLeaf(leaf.LeafValue), c.STR.TreeSize)
+		return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.STR.TreeSize)
 	case err != nil:
 		return err
 	default:
@@ -115,12 +115,18 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 		LogId: c.LogID,
 	}
 	resp, err := c.client.GetLatestSignedLogRoot(ctx, req)
-	str := resp.SignedLogRoot
 	if err != nil {
 		return err
 	}
-	// TODO(gdbelvin): Verify SignedLogRoot Signature
+	str := resp.SignedLogRoot
 
+	// Verify SignedLogRoot signature.
+	hash := crypto.HashTrillianSignedLogRoot(*str)
+	if err := crypto.Verify(c.pubKey, hash, str.Signature); err != nil {
+		return err
+	}
+
+	// Verify Consistency proof.
 	if str.TreeSize == c.STR.TreeSize && bytes.Equal(str.RootHash, c.STR.RootHash) {
 		// Tree has not been updated.
 		return nil
@@ -147,7 +153,6 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 			return err
 		}
 	}
-
 	c.STR = *str
 	return nil
 }
@@ -185,10 +190,11 @@ func convertProof(proof *trillian.Proof) [][]byte {
 	return neighbors
 }
 
-func buildLeaf(data []byte) *trillian.LogLeaf {
+func (c *LogClient) buildLeaf(data []byte) *trillian.LogLeaf {
 	hash := sha256.Sum256(data)
 	leaf := &trillian.LogLeaf{
 		LeafValue:        data,
+		MerkleLeafHash:   c.hasher.HashLeaf(data),
 		LeafIdentityHash: hash[:],
 	}
 	return leaf
--- a/crypto/data_formats.go
+++ b/crypto/data_formats.go
@@ -1,0 +1,64 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crypto
+
+import (
+	"encoding/base64"
+	"strconv"
+
+	"github.com/benlaurie/objecthash/go/objecthash"
+	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/sigpb"
+)
+
+// This file contains struct specific mappings and data structures.
+// TODO: remove data-structure specific operations.
+
+// Constants used as map keys when building input for ObjectHash. They must not be changed
+// as this will change the output of hashRoot()
+const (
+	mapKeyRootHash       string = "RootHash"
+	mapKeyTimestampNanos string = "TimestampNanos"
+	mapKeyTreeSize       string = "TreeSize"
+)
+
+// HashTrillianSignedLogRoot hashes SignedLogRoot objects in a custom way.
+func HashTrillianSignedLogRoot(root trillian.SignedLogRoot) []byte {
+	rootMap := make(map[string]interface{})
+
+	// Pull out the fields we want to hash. Caution: use string format for int64 values as they
+	// can overflow when JSON encoded otherwise (it uses floats). We want to be sure that people
+	// using JSON to verify hashes can build the exact same input to ObjectHash.
+	rootMap[mapKeyRootHash] = base64.StdEncoding.EncodeToString(root.RootHash)
+	rootMap[mapKeyTimestampNanos] = strconv.FormatInt(root.TimestampNanos, 10)
+	rootMap[mapKeyTreeSize] = strconv.FormatInt(root.TreeSize, 10)
+
+	hash := objecthash.ObjectHash(rootMap)
+
+	return hash[:]
+}
+
+// SignLogRoot updates a log root to include a signature from the crypto signer this object
+// was created with. Signatures use objecthash on a fixed JSON format of the root.
+func (s Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
+	objectHash := HashTrillianSignedLogRoot(root)
+	signature, err := s.Sign(objectHash[:])
+
+	if err != nil {
+		return nil, err
+	}
+
+	return signature, nil
+}
--- a/crypto/data_formats.go
+++ b/crypto/data_formats.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/benlaurie/objecthash/go/objecthash"
 	"github.com/google/trillian"
-	"github.com/google/trillian/crypto/sigpb"
 )
 
 // This file contains struct specific mappings and data structures.
@@ -34,8 +33,9 @@ const (
 	mapKeyTreeSize       string = "TreeSize"
 )
 
-// HashLogRoot hashes SignedLogRoot objects in a custom way
-// using the map of properties listed above.
+// HashLogRoot hashes SignedLogRoot objects using ObjectHash with
+// "RootHash", "TimestampNanos", and "TreeSize", used as keys in
+// a map.
 func HashLogRoot(root trillian.SignedLogRoot) []byte {
 	// Pull out the fields we want to hash.
 	// Caution: use string format for int64 values as they can overflow when
@@ -48,9 +48,4 @@ func HashLogRoot(root trillian.SignedLogRoot) []byte {
 
 	hash := objecthash.ObjectHash(rootMap)
 	return hash[:]
-}
-
-// SignLogRoot returns a signature using objecthash on a fixed JSON format of the root.
-func (s *Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
-	return s.Sign(HashLogRoot(root))
 }
--- a/crypto/data_formats.go
+++ b/crypto/data_formats.go
@@ -24,7 +24,7 @@ import (
 )
 
 // This file contains struct specific mappings and data structures.
-// TODO: remove data-structure specific operations.
+// TODO(alcutter): remove data-structure specific operations.
 
 // Constants used as map keys when building input for ObjectHash. They must not be changed
 // as this will change the output of hashRoot()
@@ -46,7 +46,6 @@ func HashTrillianSignedLogRoot(root trillian.SignedLogRoot) []byte {
 	rootMap[mapKeyTreeSize] = strconv.FormatInt(root.TreeSize, 10)
 
 	hash := objecthash.ObjectHash(rootMap)
-
 	return hash[:]
 }
 
@@ -55,7 +54,6 @@ func HashTrillianSignedLogRoot(root trillian.SignedLogRoot) []byte {
 func (s Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
 	objectHash := HashTrillianSignedLogRoot(root)
 	signature, err := s.Sign(objectHash[:])
-
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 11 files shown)


crypto/key_manager.go
```diff
@@ -82,140 +59,84 @@ func (k PEMKeyManager) HashAlgorithm() crypto.Hash {
...
+
+	switch key.(type) {
+	case *ecdsa.PrivateKey:
+		signer = key.(crypto.Signer)
+		sigAlgo = sigpb.DigitallySigned_ECDSA
+	case *rsa.PrivateKey:
+		signer = key.(crypto.Signer)
+		sigAlgo = sigpb.DigitallySigned_RSA
+	default:
+		return nil, errors.New("unsupported key type")
```
codingllama: nit: Add the key type to the error message?

RJPercival:
Once you rebase this on master, we can take a look.

crypto/data_formats.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/base64"
+	"strconv"
+
+	"github.com/benlaurie/objecthash/go/objecthash"
+	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/sigpb"
+)
+
+// This file contains struct specific mappings and data structures.
+// TODO: remove data-structure specific operations.
```
gdbelvin: done

crypto/data_formats.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return hash[:]
+}
+
+// SignLogRoot updates a log root to include a signature from the crypto signer this object
+// was created with. Signatures use objecthash on a fixed JSON format of the root.
+func (s Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
+	objectHash := HashTrillianSignedLogRoot(root)
+	signature, err := s.Sign(objectHash[:])
+
```
gdbelvin: done

crypto/data_formats.go
```diff
@@ -0,0 +1,64 @@
...
+// Constants used as map keys when building input for ObjectHash. They must not be changed
+// as this will change the output of hashRoot()
+const (
+	mapKeyRootHash       string = "RootHash"
+	mapKeyTimestampNanos string = "TimestampNanos"
+	mapKeyTreeSize       string = "TreeSize"
+)
+
+// HashTrillianSignedLogRoot hashes SignedLogRoot objects in a custom way.
+func HashTrillianSignedLogRoot(root trillian.SignedLogRoot) []byte {
```
gdbelvin: done

gdbelvin:
Rebased on master and fixed comments.  PTAL.

crypto/data_formats.go
```diff
@@ -0,0 +1,62 @@
...
+	// using JSON to verify hashes can build the exact same input to ObjectHash.
+	rootMap[mapKeyRootHash] = base64.StdEncoding.EncodeToString(root.RootHash)
+	rootMap[mapKeyTimestampNanos] = strconv.FormatInt(root.TimestampNanos, 10)
+	rootMap[mapKeyTreeSize] = strconv.FormatInt(root.TreeSize, 10)
+
+	hash := objecthash.ObjectHash(rootMap)
+	return hash[:]
+}
+
+// SignLogRoot updates a log root to include a signature from the crypto signer this object
```
RJPercival: "SignLogRoot updates a log root" - I don't think it does. It just returns the signature, which is better than modifying the root IMO. Comment needs correcting to reflect what the func actually does.

crypto/data_formats.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Constants used as map keys when building input for ObjectHash. They must not be changed
+// as this will change the output of hashRoot()
+const (
+	mapKeyRootHash       string = "RootHash"
+	mapKeyTimestampNanos string = "TimestampNanos"
+	mapKeyTreeSize       string = "TreeSize"
+)
+
+// HashTrillianSignedLogRoot hashes SignedLogRoot objects in a custom way.
```
RJPercival: "in a custom way" - could the comment elaborate on that a little please?

crypto/data_formats_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/google/trillian"
+)
+
+const (
+	publicKeyFile      = "../testdata/log-rpc-server.pubkey.pem"
+	privateKeyFile     = "../testdata/log-rpc-server.privkey.pem"
```
RJPercival: These paths might not work inside of Google. Instead, could you import "github.com/google/trillian/testonly" and use testonly.Demo{Private,Public}Key?

crypto/data_formats_test.go
```diff
@@ -0,0 +1,67 @@
...
+				TreeSize:       2,
+			},
+		},
+	} {
+		before := test.root // Capture test value.
+		signature, err := signer.SignLogRoot(test.root)
+		if err != nil {
+			t.Errorf("Failed to sign log root: %v", err)
+		}
+		// Check root is not modified
```
RJPercival: `SignLogRoot` takes `test.root` by value, so couldn't modify it even if it wanted to. Presumably the implementation of `SignLogRoot` changed since this test was written. Either way, I don't think it's necessary to test that `SignLogRoot` doesn't do something it can't possibly do.

crypto/signer.go
```diff
@@ -64,54 +53,39 @@ func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.S
 }
 
 // Sign obtains a signature after first hashing the input data.
-func (s Signer) Sign(data []byte) (sigpb.DigitallySigned, error) {
+func (s Signer) Sign(data []byte) (*sigpb.DigitallySigned, error) {
```
RJPercival: I don't think you need to copy `s` when invoking this func.

crypto/signer.go
```diff
@@ -64,54 +53,39 @@ func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.S
...
 	h := s.hash.New()
 	h.Write(data)
 	digest := h.Sum(nil)
 
 	if len(digest) != s.hash.Size() {
-		return sigpb.DigitallySigned{}, fmt.Errorf("hasher returned unexpected digest length: %d, %d",
+		return nil, fmt.Errorf("hasher returned unexpected digest length: %d, %d",
```
RJPercival: Use the "got %d, want %d" format of error message to make it clear which of the two values is which.

crypto/signer.go
```diff
@@ -64,54 +53,39 @@ func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.S
...
+	j, err := json.Marshal(obj)
 	if err != nil {
-		glog.Warningf("Signer failed to sign root: %v", err)
-		return sigpb.DigitallySigned{}, err
+		return nil, err
 	}
-
-	return signature, nil
+	hash := objecthash.CommonJSONHash(string(j))
+	sig, err := s.Sign(hash[:])
```
RJPercival: Could just do `return s.Sign(hash[:])`

crypto/verifier.go
```diff
@@ -49,8 +61,23 @@ func PublicKeyFromPEM(pemEncodedKey string) (crypto.PublicKey, error) {
...
 
+// VerifyObject verifies the output of SignObject.
+func VerifyObject(pub crypto.PublicKey, obj interface{}, sig *sigpb.DigitallySigned) error {
+	j, err := json.Marshal(obj)
+	if err != nil {
+		return err
+	}
+	hash := objecthash.CommonJSONHash(string(j))
+
+	if err := Verify(pub, hash[:], sig); err != nil {
```
RJPercival: Could just `return Verify(pub, hash[:], sig)`

crypto/verifier.go
```diff
@@ -20,19 +20,31 @@ import (
...
+	pemData, err := ioutil.ReadFile(keyFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read: %s. %v", keyFile, err)
+	}
+	return PublicKeyFromPEM(pemData)
+}
+
 // PublicKeyFromPEM converts a PEM object into a crypto.PublicKey
-func PublicKeyFromPEM(pemEncodedKey string) (crypto.PublicKey, error) {
+func PublicKeyFromPEM(pemEncodedKey []byte) (crypto.PublicKey, error) {
```
RJPercival: Isn't this reversing a review comment I made on the last PR, about PEM parameters having a consistent type?

crypto/data_formats.go
gdbelvin: I added the TODO and assigned it to @AlCutter.
It's my opinion that we should remove everything in this file and make signing a generic thing rather than creating new functions for every object we want to sign.  But that should be a separate PR.  Just separating it here so it's easier to remove.

crypto/key_manager.go

crypto/data_formats.go
gdbelvin: Ie. Not with a generic object hash. -- ie. using the special map that is defined above.

crypto/data_formats.go
RJPercival: I think the TODO should have your name in it then. The name in a TODO isn't who it's assigned to, but who can provide the most context when someone wants to fix it. It sounds like that's you.

crypto/data_formats.go
RJPercival: Rephrasing the comment to say that would be fine :)

crypto/data_formats.go
gdbelvin: done

crypto/data_formats.go
gdbelvin: done

crypto/data_formats.go
```diff
@@ -0,0 +1,62 @@
...
+func HashTrillianSignedLogRoot(root trillian.SignedLogRoot) []byte {
```
gdbelvin: done

crypto/data_formats.go
```diff
@@ -0,0 +1,62 @@
...
+	rootMap := make(map[string]interface{})
```
gdbelvin: done

crypto/data_formats.go
gdbelvin: done

crypto/data_formats.go
```diff
@@ -0,0 +1,62 @@
...
+	rootMap[mapKeyTimestampNanos] = strconv.FormatInt(root.TimestampNanos, 10)
+	rootMap[mapKeyTreeSize] = strconv.FormatInt(root.TreeSize, 10)
+
+	hash := objecthash.ObjectHash(rootMap)
+	return hash[:]
+}
+
+// SignLogRoot updates a log root to include a signature from the crypto signer this object
+// was created with. Signatures use objecthash on a fixed JSON format of the root.
+func (s Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
```
gdbelvin: done

crypto/data_formats_test.go
gdbelvin: done

crypto/signer.go
gdbelvin: done

crypto/signer.go
gdbelvin: done.  Removing this unneeded test.

crypto/signer.go
gdbelvin: done

crypto/verifier.go
gdbelvin: yep.  Fixing now.

crypto/verifier.go
gdbelvin: done

crypto/data_formats_test.go
gdbelvin: done

gdbelvin:
PTAL

crypto/data_formats.go
```diff
@@ -0,0 +1,56 @@
...
+		mapKeyRootHash:       base64.StdEncoding.EncodeToString(root.RootHash),
+		mapKeyTimestampNanos: strconv.FormatInt(root.TimestampNanos, 10),
+		mapKeyTreeSize:       strconv.FormatInt(root.TreeSize, 10)}
+
+	hash := objecthash.ObjectHash(rootMap)
+	return hash[:]
+}
+
+// SignLogRoot returns a signature using objecthash on a fixed JSON format of the root.
+func (s *Signer) SignLogRoot(root trillian.SignedLogRoot) (*sigpb.DigitallySigned, error) {
```
codingllama: nit: Move this method to signer.go? I find it unusual for the method to be defined in a separate file from the type.

crypto/data_formats_test.go
```diff
@@ -0,0 +1,56 @@
...
+package crypto
+
+import (
+	"testing"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/testonly"
+)
+
+func TestSignLogRoot(t *testing.T) {
```
codingllama: Move test to signer_test.go?

crypto/data_formats_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+package crypto
+
+import (
+	"testing"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/testonly"
+)
+
```
codingllama: Add a test for HashLogRoot?

It may not be useful to check specific hash values, but we definitely want to make sure that RootHash, TimestampNanos and TreeSize are part of the computation.

crypto/verifier.go
```diff
@@ -20,17 +20,29 @@ import (
...
 	"math/big"
 
+	"github.com/benlaurie/objecthash/go/objecthash"
 	"github.com/google/trillian/crypto/sigpb"
 )
 
 // ErrVerify occurs whenever signature verification fails.
 var ErrVerify = errors.New("signature verification failed")
 
+// PublicKeyFromFile returns the public key contained in the keyFile.
```
codingllama: nit: Add ", in PEM format" ?

crypto/verifier.go
```diff
@@ -49,8 +61,19 @@ func PublicKeyFromPEM(pemEncodedKey string) (crypto.PublicKey, error) {
 	return parsedKey, nil
 }
 
+// VerifyObject verifies the output of SignObject.
```
codingllama: nit: "verifies the output of SignObject". I'm not sure what that means. Is SignObject == obj? Please rephrase. We should probably mention that obj is marshaled to JSON and hashed with CommonJSONHash before verification.

crypto/data_formats_test.go
codingllama: I don't think this got addressed.

client/client.go
```diff
@@ -115,12 +115,18 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
...
 	}
 	resp, err := c.client.GetLatestSignedLogRoot(ctx, req)
-	str := resp.SignedLogRoot
 	if err != nil {
 		return err
 	}
-	// TODO(gdbelvin): Verify SignedLogRoot Signature
+	str := resp.SignedLogRoot
 
+	// Verify SignedLogRoot signature.
```
codingllama: nit: Needless comment?

crypto/data_formats.go
```diff
@@ -0,0 +1,56 @@
...
+// Constants used as map keys when building input for ObjectHash. They must not be changed
+// as this will change the output of hashRoot()
+const (
+	mapKeyRootHash       string = "RootHash"
+	mapKeyTimestampNanos string = "TimestampNanos"
+	mapKeyTreeSize       string = "TreeSize"
+)
+
+// HashLogRoot hashes SignedLogRoot objects in a custom way
+// using the map of properties listed above.
```
codingllama: nit: Please document the contract of HashLogRoot more explicit. The properties above are not public, therefore I don't think they should be referenced by documentation.

crypto/data_formats.go
```diff
@@ -0,0 +1,56 @@
...
+	rootMap := map[string]string{
+		mapKeyRootHash:       base64.StdEncoding.EncodeToString(root.RootHash),
+		mapKeyTimestampNanos: strconv.FormatInt(root.TimestampNanos, 10),
+		mapKeyTreeSize:       strconv.FormatInt(root.TreeSize, 10)}
+
+	hash := objecthash.ObjectHash(rootMap)
+	return hash[:]
+}
+
+// SignLogRoot returns a signature using objecthash on a fixed JSON format of the root.
```
codingllama: Suggestion: "SignLogRoot returns a signature over the HashLogRoot() of root"

crypto/verifier.go
codingllama: Suggestion: "(...) verifies the output of Signer.SignObject"

crypto/data_formats.go
gdbelvin: I'll remove this function.

crypto/data_formats_test.go
gdbelvin: done

crypto/data_formats_test.go
gdbelvin: done

crypto/verifier.go
gdbelvin: done

client/client.go
gdbelvin: na, I think it adds some structure to the verifications this is doing.

crypto/data_formats.go
gdbelvin: removed this func

crypto/data_formats.go
gdbelvin: how's this:

```
// HashLogRoot hashes SignedLogRoot objects using ObjectHash with
// "RootHash", "TimestampNanos", and "TreeSize", used as keys in
// a map.
```

crypto/verifier.go
gdbelvin: done

crypto/signer_test.go
```diff
@@ -111,11 +111,45 @@ func TestSignLogRootSignerFails(t *testing.T) {
...
+	_, err := logSigner.Sign(HashLogRoot(root))
 
 	testonly.EnsureErrorContains(t, err, "signfail")
 }
 
 func createTestSigner(mock *MockSigner) *Signer {
 	return NewSigner(sigpb.DigitallySigned_RSA, mock)
 }
+
+func TestSignLogRoot(t *testing.T) {
```
codingllama: Remove the test too? The method doesn't exist anymore.

testonly/integration/clientserver.go
```diff
@@ -40,12 +40,15 @@ const (
 )
 
 var (
+	publicKeyFile      = "../testdata/log-rpc-server.pubkey.pem"
```
codingllama: I think @RJPercival added a comment here about using the constants of testonly/keys.go instead of reading files, but I can't find it anymore.  :/

Could you change it, please?

crypto/signer_test.go
gdbelvin: done

testonly/integration/clientserver.go
gdbelvin: done

gdbelvin:
PTAL

crypto/signer_test.go
```diff
@@ -51,30 +50,29 @@ func (i usesSHA256Hasher) String() string {
 }
 
 func TestSigner(t *testing.T) {
```
codingllama: Would you mind keeping the old test, in addition to the one you added? I think both add value in different areas.

gdbelvin: I'll add some of the old test back in.

crypto/signer_test.go
```diff
@@ -65,12 +66,18 @@ func TestSigner(t *testing.T) {
...
 	} {
-		signature, err := signer.Sign(test.message)
+		sig, err := signer.Sign(test.message)
+		if got, want := sig.HashAlgorithm, sigpb.DigitallySigned_SHA256; got != want {
+			t.Fatalf("Hash alg incorrect, got %v expected %d", got, want)
+		}
+		if got, want := sig.SignatureAlgorithm, sigpb.DigitallySigned_ECDSA; got != want {
+			t.Fatalf("Sig alg incorrect, got %v expected %v", got, want)
+		}
 		if err != nil {
```
codingllama: Move `if err != nil` to line 70, add a `continue` after the `t.Errorf("Failed to sign ...")` below. We shouldn't assume `sig != nil` if `err != nil`.

codingllama: Could you check that `len(sig.Signature) > 0`? With that I think I'm happy for you to drop the other test.

gdbelvin: done

crypto/signer_test.go
```diff
@@ -67,14 +67,17 @@ func TestSigner(t *testing.T) {
 		{message: []byte("message")},
 	} {
 		sig, err := signer.Sign(test.message)
+		if err != nil {
+			t.Errorf("Failed to sign log root: %v", err)
```
codingllama: nit: Add a continue here? If signing failed there isn't much point in progressing the test.

PR #397, opened 2017-02-22
Add Close() method to TX types, defer call it in all occurrences

codingllama:
I've tracked all TXs being opened and made sure they're being defer-closed after
creation. There's no more need to tx.Rollback() at every exit point.

A few deferred commits have been removed and made explicit as well (mostly on
the Map server implementation).

### DIFF
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -404,6 +404,13 @@ func (t *treeTX) Rollback() error {
 	return nil
 }
 
+func (t *treeTX) Close() error {
+	if !t.closed {
+		return t.Rollback()
+	}
+	return nil
+}
+
 func (t *treeTX) IsOpen() bool {
 	return !t.closed
 }
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -406,7 +406,11 @@ func (t *treeTX) Rollback() error {
 
 func (t *treeTX) Close() error {
 	if !t.closed {
-		return t.Rollback()
+		err := t.Rollback()
+		if err != nil {
+			glog.Warningf("Rollback error on Close(): %v", err)
+		}
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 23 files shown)


storage/mysql/tree_storage.go
```diff
@@ -404,6 +404,13 @@ func (t *treeTX) Rollback() error {
 	return nil
 }
 
+func (t *treeTX) Close() error {
+	if !t.closed {
+		return t.Rollback()
```
Martin2112: All of the deferred Close() calls are hiding errors if the rollback fails.

Martin2112:
As we could argue that if people want explicit rollback error handling they can still do this by manually calling Rollback() I think it's probably enough to make Close() log any errors.

storage/mysql/tree_storage.go
codingllama: As were most tx.Rollback() calls before it.

From a theoretical standpoint, I think Close() returning an error is the proper way to do it. I could add logging to Close, if you prefer, but I think errors arising from it are uninteresting: if Close() is taking action, that's because something else already went wrong, and that something else is the error we really care about.

Martin2112: Errors on rollback are unlikely with an RDBMS but if they happen then it probably means the connection is in a bad state somehow. Either due to a bug or the server going away. This is probably not the right level to detect the latter.

Returning the error looks like the right thing. I generally think all errors should be handled or logged, could be a low severity level and done via deferring a logging helper function. 

I think the errors are probably uninteresting to clients but I'm wary of masking bugs in our code.

Martin2112:
This is another one I'll have to merge into my experimental branches. Hopefully it won't cause too many conflicts though.

storage/mysql/tree_storage.go
codingllama: Added logging to Close() fns.

codingllama:
Rebased and fixed comments, PTAL?

(Sorry about the merge conflicts, hopefully it won't be too painful.)

PR #398, opened 2017-02-22
Make personality compute MerkleLeafHash

gdbelvin:
This
- Improves API readability. All fields of LogLeaf are provided by client.
- Improves API usability. Client's don't have to guess what hash algorithm was used to compute the MerkleTreeHash.
- Improves future flexibility. Server won't need to be modified to support new MerkleLeafHash algos.
- Improves server speed. Saves a DB lookup of the hash algo.

### DIFF
--- a/examples/ct/handlers.go
+++ b/examples/ct/handlers.go
@@ -16,6 +16,7 @@ package ct
 
 import (
 	"context"
+	gocrypto "crypto"
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
@@ -36,6 +37,7 @@ import (
 	"github.com/google/certificate-transparency/go/x509"
 	"github.com/google/trillian"
 	"github.com/google/trillian/crypto"
+	"github.com/google/trillian/merkle/rfc6962"
 	"github.com/google/trillian/util"
 )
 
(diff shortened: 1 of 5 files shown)


Martin2112:
This allows clients to send mismatched or invalid hashes because we can't do any checks on the value. I need to think a bit about whether this causes any problems over and above what a misbehaving client already can. Does @AlCutter have an opinion on this?

gdbelvin:
If I understand correctly, it's up to the personality to define the relationship between LeafValue and `MerkleLeafHash`, so technically any value for `MerkleLeafHash` should be "valid".

Martin2112:
But currently the server computes it so we know it matches the data that we were given. It might be OK, I just need to think about it.

codingllama:
I've been thinking about this, but I'm not entirely convinced it's a good idea.

A few reasons are:
* Clients may use the wrong algorithm or preimage protection. The former we *may* be able to tell by hash size, the latter we can't distinguish
* Trillian has to hash intermediate nodes and root anyway, so it doesn't eliminate the need to properly setup and match hashing algos/strategy. I'd be more willing if we could take away all hashing from Trillian, but IIUC that's not practical (we'd need a bidi connection to the personality during sequencing, plus added latency, etc).

As it stands this is spreading responsibility, but I think I'd rather concentrate it on Trillian instead.

gdbelvin:
Good discussion. Continuing for the sake of argument:

* Pre-image resistance:  The personality should be responsible for choosing algorithms that are suitable for it's security needs.  It should not be Trillian's responsibility if this is done improperly. 
* Spreading responsibility: Good point
   * Hashing internal nodes is simple enough that it could be standardized. 
   * Hashing empty nodes is more tricky, but that only effects maps.  Maps might not be able to export all of this responsibility.

Martin2112:
Hmmm. Internal node hashes use preimage as well. I don't think this can
necessarily be standardized because if the client chooses the preimage then
it could collide with the node preimage value chosen by the server, losing
second preimage resistance (see section 2.1 of the CT spec). If everybody
uses RFC 6962 preimage values then that's not a problem but that seems like
it might not be realistic.

Hashing empty means knowing the hash algorithm, so the server still needs
it.

Martin

PR #407, opened 2017-02-24
Pull out the statement cache from storage

Martin2112:
Make it agnostic to SQL dialects. Callers must now do all parameter expansions.


Martin2112:
Closing this for a rethink on if there's a better way.

PR #411, opened 2017-02-28
Pass PrivateKeyManager when creating a new Signer

RJPercival:
Simplifies the code a bit by removing a superfluous parameter.

### DIFF
--- a/crypto/signer.go
+++ b/crypto/signer.go
@@ -30,27 +30,26 @@ var sigpbHashLookup = map[crypto.Hash]sigpb.DigitallySigned_HashAlgorithm{
 // Signer is responsible for signing log-related data and producing the appropriate
 // application specific signature objects.
 type Signer struct {
-	hash crypto.Hash
-	key  PrivateKeyManager
+	hash         crypto.Hash
+	signer       crypto.Signer
+	sigAlgorithm sigpb.DigitallySigned_SignatureAlgorithm
 }
 
 // NewSigner creates a new Signer wrapping up a hasher and a signer. For the moment
 // we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced
 // here.
 func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.Signer) *Signer {
-	return NewSignerFromPrivateKeyManager(localSigner{
-		Signer:             signer,
-		signatureAlgorithm: sigAlgo,
-	})
+	return &Signer{
+		hash:         crypto.SHA256,
+		signer:       signer,
+		sigAlgorithm: sigAlgo,
+	}
 }
 
-// NewSignerFromPrivateKeyManager creates a new Signer wrapping up a hasher and a signer.
+// NewSignerFromPrivateKeyManager creates a new Signer wrapping up a hasher and a private key.
 // For the moment, we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced here.
 func NewSignerFromPrivateKeyManager(key PrivateKeyManager) *Signer {
-	return &Signer{
-		hash: crypto.SHA256,
-		key:  key,
-	}
+	return NewSigner(key.SignatureAlgorithm(), key)
 }
 
 // Sign obtains a signature after first hashing the input data.
@@ -59,13 +58,13 @@ func (s *Signer) Sign(data []byte) (*sigpb.DigitallySigned, error) {
 	h.Write(data)
 	digest := h.Sum(nil)
 
-	sig, err := s.key.Sign(rand.Reader, digest, s.hash)
+	sig, err := s.signer.Sign(rand.Reader, digest, s.hash)
 	if err != nil {
 		return nil, err
 	}
 
 	return &sigpb.DigitallySigned{
-		SignatureAlgorithm: s.key.SignatureAlgorithm(),
+		SignatureAlgorithm: s.sigAlgorithm,
 		HashAlgorithm:      sigpbHashLookup[s.hash],
 		Signature:          sig,
 	}, nil
--- a/crypto/signer.go
+++ b/crypto/signer.go
@@ -37,7 +37,16 @@ type Signer struct {
 // NewSigner creates a new Signer wrapping up a hasher and a signer. For the moment
 // we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced
 // here.
-func NewSigner(key PrivateKeyManager) *Signer {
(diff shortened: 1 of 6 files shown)


crypto/signer.go
```diff
@@ -24,30 +24,25 @@ import (
...
-	sigAlgorithm sigpb.DigitallySigned_SignatureAlgorithm
+	hash   crypto.Hash
+	signer PrivateKeyManager
 }
 
 // NewSigner creates a new Signer wrapping up a hasher and a signer. For the moment
 // we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced
 // here.
-func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.Signer) *Signer {
+func NewSigner(signer PrivateKeyManager) *Signer {
```
gdbelvin: So the question here is one of philosophy. 
Pro: This change makes it easier to create `Signer`s from `PrivateKeyManager`
Con: This now couples the Signer to KeyManager. Now I can't pass in my own generic private without invoking PrivateKeyManager.  
- This might make testing a bit harder since we'll need to create an extra object to create Signer.
- This means that this Signing library will be less useful to other projects that might not want to incorporate PrivateKeyManager.

What do you think?

crypto/verifier.go
```diff
@@ -31,8 +31,13 @@ import (
 	"github.com/google/trillian/crypto/sigpb"
 )
 
-// ErrVerify occurs whenever signature verification fails.
-var ErrVerify = errors.New("signature verification failed")
+var (
+	errVerify = errors.New("signature verification failed")
```
gdbelvin: You might want to make `errVerify` publicly exported so other modules can see what went wrong.

RJPercival: In other languages, identifying a type of error based on the error message would be frowned upon. Is this another of those cases where Go goes against convention? In most languages, you would basically use type assertions to identify a particular kind of error (and therefore have something like `type VerifyError struct { Error }`)?

crypto/signer.go
RJPercival: `PrivateKeyManager` exposes exactly what `Signer` needs (and no more), so it seems reasonable to use it here. The tests actually get a little simpler in general. I don't think it's onerous to ask callers to provide a `PrivateKeyManager`, but I could add a `NewFromSigner()` func to key_manager.go to make it extra easy to provide one if you have a `crypto.Signer` that isn't a `crypto.PrivateKey`.

crypto/verifier.go
gdbelvin: In Go, the convention is that you can test errors like so `if err == ErrVerify` which is quite nice. Now in this particular case, if you don't think external libraries should be doing anything special with this error condition, then it's perfectly fine to hide it for better data hiding. 

Thanks for this change. Keep it as `errVerify`

gdbelvin:
For the next step in this PR, please comment on the coupling between `Signer` and `PrivateKeyManger`

RJPercival:
I already have: https://github.com/google/trillian/pull/411#discussion_r103456359

crypto/verifier.go
RJPercival: Ok, from https://blog.golang.org/error-handling-and-go, I got the impression that type assertions are the way to go, but I'll bear this in mind for the future. However, in this case, ErrVerify is not always returned when signature verification fails, as only the verifyECDSA func returns it (verifyRSA doesn't).

crypto/signer.go
RJPercival: We should improve the names of some of the funcs in the crypto package come to think of it. There are a collection of "NewFrom" funcs in key_manager.go which should either be in a "keymanager" package or else be named "NewKeyManagerFrom" (or something along those lines).

gdbelvin: Good suggestions.  Please add
- In Signer: `NewFromKeyManager`.  This function should just call `New` with the appropriate expanded arguments. 
- `key_manager.go` should probably be moved to a separate package `crypto/keymanger`

crypto/verifier.go
gdbelvin: One can use type assertions to distinguish between types defined with `type newError ...`  
All errors from `error.New` or `fmt.Errorf` return the standard error type `stringError` which is better distinguished with an equality operator that simply compares the `Error()` strings.

gdbelvin:
Sorry for the delay. Consider these changes:

- In Signer: `NewFromKeyManager`. This function should just call `New` with the appropriate expanded arguments.
- `key_manager.go` should probably be moved to a separate package `crypto/keymanager`
- `Signer` may also optionally be moved to `crypto/signer`

crypto/signer.go
```diff
@@ -23,31 +23,33 @@ import (
...
-	sigAlgorithm sigpb.DigitallySigned_SignatureAlgorithm
+	hash crypto.Hash
+	key  PrivateKeyManager
 }
 
 // NewSigner creates a new Signer wrapping up a hasher and a signer. For the moment
 // we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced
 // here.
 func NewSigner(sigAlgo sigpb.DigitallySigned_SignatureAlgorithm, signer crypto.Signer) *Signer {
+	return NewSignerFromPrivateKeyManager(localSigner{
```
gdbelvin: Almost there. 

Consider making `NewSignerFromKeyManager` call `NewSigner` rather than the other way around.  Observe that the current state requires creating an extra intermediate object.

gdbelvin:
One last change and we'll be good to go

crypto/signer.go
RJPercival: That's true, but only for the rare (currently never) occasion when `NewSigner()` is called. In the common case of calling `NewSignerFromPrivateKeyManager()` directly, it results in a comparable saving (by not having to copy the `PrivateKeyManager`).

RJPercival: On the other hand, if `Signer` and `PrivateKeyManager` move into their own packages, `NewSigner()` will no longer have access to `localSigner`, so your way would be simpler.

RJPercival: Would you like me to do that package move as part of this PR?

gdbelvin: Keep the package move for a separate PR, but reverse the `New` functions in this one

gdbelvin:
Created #414 to track progress on our conversation. 
Ping me when this is ready for another round of review.

RJPercival:
Done. `NewSignerFromPrivateKeyManager()` will change back into `NewSigner()` with the next PR, which will remove the sigAlgo parameter from `NewSigner()`.

PR #418, opened 2017-03-07
Admin server skeleton

codingllama:
* RPCs return Unimplemented
* Added unit and integration tests
* AdminServer wired into Log and Map server binaries

### DIFF
--- a/server/admin/admin_server.go
+++ b/server/admin/admin_server.go
@@ -1,0 +1,53 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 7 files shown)


gdbelvin:
I see this is creating a 3rd binary.  Extra binaries are typically difficult to manage. 
Just wondering if this kind of API belongs in a long running back-end job such as the signer with a restrictive authorization policy?

codingllama:
@gdbelvin there's no 3rd binary, I'm plugging Admin into both Log and Map server binaries.

It seems reasonable to me that we'd perform Admin using those same endpoints. There are no outstanding requirements for Admin RPCs either (ie, the volume should be pretty much negligible).

codingllama:
Issue #310

gdbelvin:
Ah, my bad.  Carry on. 
Btw, might want to add a PR description.

codingllama:
Good call. Apparently, when there's >1 commit, Github doesn't use the description on the first.

server/admin/admin_server.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var errNotImplemented = grpc.Errorf(codes.Unimplemented, "not implemented")
+
+// adminServer is an implementation of trillian.TrillianAdminServer.
+type adminServer struct{}
+
+// New returns a trillian.TrillianAdminServer implementation.
+func New() trillian.TrillianAdminServer {
+	return &adminServer{}
```
AlCutter: Do you anticipate a need for work to be done in here?

codingllama: Yes. Here and pretty much everywhere else on the PR.

At the very least this will need to be `func New(extension.Registry)`.

PR #424, opened 2017-03-10
Remove PrivateKeyManager in favour of new crypto/keys package

### LINKED ISSUE #414: Replace PrivateKeyManager with crypto.Signer
After a conversation with @RJPercival, we decided that the current `PrivateKeyManager` interface could be completely replaced with the pure `crypto.Signer` interface. 

The current `PrivateKeyManager` interface only contains a `DigitallySigned_SignatureAlgorithm` but the signature algorithm can be derived from `Signer.Public()`.  

This change
- Removes the opportunity for algorithm mismatch and several error conditions. 
- Makes the `Signer` library generic.

RJPercival:
The `PrivateKeyManager` interface merely added the ability to get a `sigpb.DigitallySigned_SignatureAlgorithm` from a `crypto.Signer`. Instead, `keys.SignatureAlgorithm()` can be used on a `crypto.PublicKey` to discover the signature algorithm. A `crypto.Signer` exposes its `crypto.PublicKey` via its `Public()` method, meaning `keys.SignatureAlgorithm()` can be used to discover the signature algorithm of any `crypto.Signer`.

This simplifies the code, and consequently allows some mocks to be deleted as well.

The extension.Registry is modified to return a `trillian.crypto.Signer`, rather than a `PrivateKeyManager`.

Fixes #414.


crypto/signer.go
```diff
@@ -30,26 +31,23 @@ var sigpbHashLookup = map[crypto.Hash]sigpb.DigitallySigned_HashAlgorithm{
...
+		signer: signer,
 	}
 }
 
-// NewSignerFromPrivateKeyManager creates a new Signer wrapping up a hasher and a private key.
-// For the moment, we only support SHA256 hashing and either ECDSA or RSA signing but this is not enforced here.
-func NewSignerFromPrivateKeyManager(key PrivateKeyManager) *Signer {
-	return NewSigner(key.SignatureAlgorithm(), key)
+// Public returns the public key that can verify signatures produced by s.
+func (s *Signer) Public() crypto.PublicKey {
```
gdbelvin: This method might not be needed.

crypto/signer_test.go
```diff
@@ -50,13 +51,13 @@ func (i usesSHA256Hasher) String() string {
 }
 
 func TestSigner(t *testing.T) {
-	km, err := NewFromPrivatePEM(testonly.DemoPrivateKey, testonly.DemoPrivateKeyPass)
+	key, err := keys.NewFromPrivatePEM(testonly.DemoPrivateKey, testonly.DemoPrivateKeyPass)
```
gdbelvin: Please use a table driven test here

examples/ct/structures_test.go
```diff
@@ -87,23 +79,15 @@ func TestSerializeLogEntry(t *testing.T) {
...
-	mockKeyManager.EXPECT().SignatureAlgorithm().AnyTimes().Return(spb.DigitallySigned_ECDSA)
-	return mockKeyManager, nil
-}
 
-// As above but we don't expect the call for a public key as we don't need it for an STH
-func setupMockPrivateKeyManagerForSth(ctrl *gomock.Controller, toSign []byte) *crypto.MockPrivateKeyManager {
-	mockKeyManager := crypto.NewMockPrivateKeyManager(ctrl)
-	mockKeyManager.EXPECT().Sign(gomock.Any(), toSign, gomock.Any()).AnyTimes().Return([]byte("signed"), nil)
-	return mockKeyManager
+	return crypto.NewSigner(testonly.NewSignerWithFixedSig(key, fakeSig)), nil
```
gdbelvin: This is much better.  Thanks for this

server/sequencer_manager_test.go
```diff
@@ -95,20 +110,23 @@ func TestSequencerManagerSingleLogNoLeaves(t *testing.T) {
...
+		t.Fatalf("Failed to create test signer (%v)", err)
+	}
 
 	mockStorage.EXPECT().BeginForTree(gomock.Any(), logID).Return(mockTx, nil)
 	mockTx.EXPECT().Commit().Return(nil)
 	mockTx.EXPECT().Close().Return(nil)
 	mockTx.EXPECT().WriteRevision().AnyTimes().Return(writeRev)
 	mockTx.EXPECT().LatestSignedLogRoot().Return(testRoot0, nil)
 	mockTx.EXPECT().DequeueLeaves(50, fakeTime).Return([]*trillian.LogLeaf{}, nil)
-	mockKeyManager := crypto.NewMockPrivateKeyManager(mockCtrl)
```
gdbelvin: Thanks for removing the mocks!

crypto/signer.go
RJPercival: There are two reasons to have this method:
1) It keeps the methods exposed by `Signer` as close to the `crypto.Signer` interface as possible.
2) By exposing the `crypto.PublicKey`, the `keys.SignatureAlgorithm()` func can be used to determine the `sigpb.DigitallySigned_SignatureAlgorithm` used by the `Signer`.

gdbelvin: I see that now.  Good choice.

crypto/verifier_test.go
```diff
@@ -40,22 +41,21 @@ func TestSignVerify(t *testing.T) {
...
-			t.Errorf("LoadPrivateKey(_, %v)=%v, want nil", test.password, err)
+			t.Errorf("LoadPrivateKey(_, %q)=%v, want nil", test.password, err)
 			continue
 		}
-		signer := NewSignerFromPrivateKeyManager(km)
 
 		// Sign and Verify.
 		msg := []byte("foo")
-		signed, err := signer.Sign(msg)
+		signed, err := NewSigner(key).Sign(msg)
```
gdbelvin: So sharp 👍

gdbelvin:
This is looking _really_ sharp.   
- Moved `New*` operations to the `keys` package. 
- `Signer` now takes a single `crypto.Signer`. 
- Especially liked the new simplicity of `NewSigner(key).Sign()` 
- Updated to table driven tests so we can change these packages easily in the future.
- Replaced auto-mocks with precise Signers for error injection and testing. 

Well done.

PR #425, opened 2017-03-10
Add private key to trillian.Tree proto and MySQL storage

RJPercival:
This will soon be used to determine the private key used for signing.


storage/mysql/admin_storage.go
```diff
@@ -320,6 +331,11 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
 		return nil, fmt.Errorf("unexpected DuplicatePolicy value: %v", newTree.DuplicatePolicy)
 	}
 
+	privateKey, err := proto.Marshal(newTree.PrivateKey)
```
codingllama: Should we validate private key types we know of before marshaling? For example, for PEMKeyFile, we could check that the file exists, password works, so on.

There's some generic validation on storage/tree_validation.go. We should at least verify that the Any's TypeUrl and Value are not empty there, I think.

RJPercival: I've added some validation that attempts to unmarshal the `Any`, which at least checks that there's a proto in there. Proper validation of key will have to come later, once I've added the code that takes this `Any` and returns a `crypto.Signer`. I think that'd have to take the form of attempting to produce an STH. In the case of updating the private key, this STH could be verified using the old key to check that the key itself hasn't changed (only the provider of the key).

storage/tree_validation_test.go
```diff
@@ -78,6 +85,8 @@ func TestValidateTreeForCreation(t *testing.T) {
 		{tree: invalidDuplicatePolicy, wantErr: true},
 		{tree: invalidDisplayName, wantErr: true},
 		{tree: invalidDescription, wantErr: true},
+		{tree: unsupportedKey, wantErr: true},
+		{tree: invalidKey, wantErr: true},
```
codingllama: Could you add a key change test under TestValidateTreeForUpdate a well?

RJPercival: Done.

RJPercival:
Travis was failing because I used the British English spelling of "serialised", and we have a check for that sort of thing. Fixed now.

codingllama:
Please make sure Travis is happy; apart from that, feel free to merge.

RJPercival:
Travis is happy now - I had to add a serialized trillian.PrivateKey proto to scripts/createtree.sh to make the integration tests pass.

PR #433, opened 2017-03-15
Return an error if the signed tree root is not, in fact, signed

RJPercival:
This could happen if there is a problem generating a signature server-side. The server really shouldn't be sending a STR without a signature, but it currently does if a signing error occurs.


client/client.go
```diff
@@ -120,6 +120,10 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 	}
 	str := resp.SignedLogRoot
 
+	if str.Signature == nil {
```
gdbelvin: Please move this check to be inside `crypto.Verify`

RJPercival: Done.

gdbelvin:
Consider adding a nil test to the signature tests as well please.

RJPercival:
Done.

PR #434, opened 2017-03-15
Get private key from admin storage instead of command-line flag

RJPercival:
The private key of a tree is now discovered via the admin storage interface. This allows different trees to use different keys.

The --private_key_file and --private_key_password flags have been removed.

A new keys.Provider interface abstracts away how a private key / signer is acquired for a tree. The default implementation supports only loading PEM-encoded keys from disk.

### DIFF
--- a/crypto/keys/private.go
+++ b/crypto/keys/private.go
@@ -15,6 +15,7 @@
 package keys
 
 import (
+	"context"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rsa"
@@ -23,7 +24,15 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+
+	"github.com/google/trillian"
 )
+
+// Provider handles acquisition of signers for trees.
+type Provider interface {
+	// Signer returns a signer for the given tree.
+	Signer(context.Context, *trillian.Tree) (crypto.Signer, error)
+}
 
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
 // The key may be protected by a password.
--- a/crypto/keys/private.go
+++ b/crypto/keys/private.go
@@ -28,10 +28,13 @@ import (
 	"github.com/google/trillian"
 )
 
-// Provider handles acquisition of signers for trees.
-type Provider interface {
-	// Signer returns a signer for the given tree.
-	Signer(context.Context, *trillian.Tree) (crypto.Signer, error)
+// SignerFactory creates signers for Trillian trees.
+// A signers may be created by loading a private key, interfacing with a HSM,
+// or sending network requests to a remote key management service, to give a few
+// examples.
+type SignerFactory interface {
+	// NewSigner returns a signer for the given tree.
+	NewSigner(context.Context, *trillian.Tree) (crypto.Signer, error)
 }
 
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
--- a/extension/builtin/default_registry.go
+++ b/extension/builtin/default_registry.go
@@ -15,12 +15,16 @@
 package builtin
 
 import (
+	"context"
+	"crypto"
 	"database/sql"
 	"flag"
+	"fmt"
 
 	_ "github.com/go-sql-driver/mysql" // Load MySQL driver
 
-	"github.com/google/trillian/crypto"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/extension"
 	"github.com/google/trillian/storage"
@@ -30,16 +34,11 @@ import (
 var (
 	// MySQLURIFlag is the mysql db connection string.
 	MySQLURIFlag = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "uri to use with mysql storage")
-	// TODO(Martin2112): Single private key doesn't really work for multi tenant and we can't use
-	// an HSM interface in this way. Deferring these issues for later.
-	privateKeyFile     = flag.String("private_key_file", "", "File containing a PEM encoded private key")
-	privateKeyPassword = flag.String("private_key_password", "", "Password for server private key")
 )
 
 // Default implementation of extension.Registry.
 type defaultRegistry struct {
-	db     *sql.DB
-	signer *crypto.Signer
+	db *sql.DB
 }
 
 func (r *defaultRegistry) GetAdminStorage() storage.AdminStorage {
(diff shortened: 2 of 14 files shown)


RJPercival:
Travis won't be happy until #435 is fixed.

RJPercival:
#435 is now fixed, and Travis is happy.

crypto/keys/private.go
```diff
@@ -23,8 +24,16 @@ import (
...
 	"fmt"
 	"io/ioutil"
+
+	"github.com/google/trillian"
 )
 
+// Provider handles acquisition of signers for trees.
+type Provider interface {
+	// Signer returns a signer for the given tree.
+	Signer(context.Context, *trillian.Tree) (crypto.Signer, error)
```
gdbelvin: This looks like a factory method. Consider calling this `NewSigner`

crypto/keys/private.go
```diff
@@ -23,8 +24,16 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+
+	"github.com/google/trillian"
 )
 
+// Provider handles acquisition of signers for trees.
+type Provider interface {
```
gdbelvin: This looks like a factory interface. Consider calling this `SignerFactory`

extension/builtin/default_registry.go
```diff
@@ -54,27 +53,45 @@ func (r *defaultRegistry) GetMapStorage() (storage.MapStorage, error) {
 	return mysql.NewMapStorage(r.db), nil
 }
 
-func (r *defaultRegistry) GetSigner(treeID int64) (*crypto.Signer, error) {
-	return r.signer, nil
+func (r *defaultRegistry) GetKeyProvider() (keys.Provider, error) {
```
gdbelvin: Go style doesn't typically use "Get" as a function prefix. Consider `SignerFactory()`

Btw, this is the right abstraction to be making here. Good call.

extension/builtin/default_registry.go
```diff
@@ -54,27 +53,45 @@ func (r *defaultRegistry) GetMapStorage() (storage.MapStorage, error) {
...
 		return nil, err
 	}
-	key, err := keys.NewFromPrivatePEMFile(*privateKeyFile, *privateKeyPassword)
-	if err != nil {
-		return nil, err
+	return NewExtensionRegistry(db)
+}
+
+// keyProvider implements keys.Provider.
+type keyProvider struct{}
```
gdbelvin: Consider moving this to a new file in the same folder for better readability. A kind of one class per file rule of thumb. 

Consider renaming to `signerFactory`

crypto/keys/private.go
RJPercival: I was cautious about putting "New" at the beginning of this. It's arguably more of a "GetSigner" method, in that it goes and acquires a private key or surrogate to use as a signer. Go style doesn't really believe in "Get" prefixes though.

extension/builtin/default_registry.go
RJPercival: This is for consistency with the other methods. However, all of these methods cease to exist with PR #438.

extension/builtin/default_registry.go
RJPercival: This file ceases to exist in PR #438.

extension/builtin/default_registry.go
```diff
@@ -54,27 +53,45 @@ func (r *defaultRegistry) GetMapStorage() (storage.MapStorage, error) {
...
+
+// Signer returns a crypto.Signer for the given tree.
+func (p keyProvider) Signer(ctx context.Context, tree *trillian.Tree) (crypto.Signer, error) {
+	if tree.PrivateKey == nil {
+		return nil, fmt.Errorf("tree %d has no PrivateKey", tree.GetTreeId())
```
RJPercival: I believe you're supposed to use the "Get" methods to access proto fields, because they do more than just return the value (e.g. returning the default value if one exists). While that doesn't really apply here, it's a good convention to stick to.

server/sequencer_manager.go
```diff
@@ -96,13 +105,26 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
 					continue
 				}
 
-				signer, err := s.registry.GetSigner(logID)
+				snapshot, err := admin.Snapshot(ctx)
```
RJPercival: Done.

testonly/integration/clientserver.go
```diff
@@ -47,7 +46,11 @@ var (
 	sleepBetweenRuns = 100 * time.Millisecond
 	timeSource       = util.SystemTimeSource{}
 	// PublicKey returns the public key that verifies responses from this server.
-	PublicKey, _ = keys.NewFromPublicPEM(to.DemoPublicKey)
+	PublicKey, _ = keys.NewFromPublicPEMFile("../testdata/log-rpc-server.pubkey.pem")
+	privateKey   = &trillian.PEMKeyFile{
+		Path:     "../testdata/log-rpc-server.privkey.pem",
```
RJPercival: Fixed.

server/sequencer_manager_test.go
```diff
@@ -108,10 +124,12 @@ func TestSequencerManagerSingleLogNoLeaves(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
+	logID := stestonly.LogTree.GetTreeId()
```
RJPercival: That's fine, it just has to be a valid log ID.

crypto/keys/private.go
gdbelvin: `Generate` is also a factory method in a sense, since it creates new objects of a type that the provider will decide.  What do you think? 

Btw, having interfaces with optional implementations seems less than ideal.  If `Generate` will be optional, maybe there's a slightly different place to put it? Perhaps a utility method that's not part of the interface?

crypto/keys/private.go
gdbelvin: Yeah, I get that it feels a bit more like a "Get", but in another sense, it's a "New" because it creates a new object that meets the `crypto.Signer` interface. It just fills that object with info that it's fetching from elsewhere.  

It might be clearer for users to use something other than `Signer` though, since it's also the name of an interface and an object already.  No need to reinvent [buffalo buffalo](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo) :-)

extension/builtin/default_registry.go
gdbelvin: Fair enough.  No need to change it then.

extension/builtin/default_registry.go
gdbelvin: Sounds good

gdbelvin:
Suggest changing `Signer` to `NewSigner` to avoid ambiguity from the object and interface. 
Other than that, it's good to go

crypto/keys/private.go
RJPercival: At least a portion of the key generation will be provider-specific though, so it can't be just a utility method. PEMKeyProvider *could* implement it, but I'm cautious about doing so because of the security implications. I've got [a branch that implements it](https://github.com/RJPercival/trillian/blob/pemkey_generate/crypto/keys/pem_key_provider.go) if you'd like to see how it could be done.

I'll rename `keys.Provider` to `keys.SignerFactory`, as you make a good argument.

crypto/keys/private.go
RJPercival: `Signer()` renamed to `NewSigner()`.

crypto/keys/private.go
RJPercival: Done.

gdbelvin:
This is much clearer. Thanks!

PR #444, opened 2017-03-16
Fix go vet error on log_rpc_server_test

codingllama:


codingllama:
Somewhat related: #445.

I'm not sure how this slipped from Travis.

PR #454, opened 2017-03-21
Create Client Interface

gdbelvin:
Creating an interface for the client allows other clients to be
created (eg. a non-verifying client).
It also enables unit testing with mock clients.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -37,20 +37,30 @@ type LogClient struct {
 	LogID    int64
 	client   trillian.TrillianLogClient
 	hasher   merkle.TreeHasher
-	STR      trillian.SignedLogRoot
-	MaxTries int
+	str      trillian.SignedLogRoot
+	maxTries int
 	pubKey   gocrypto.PublicKey
 }
 
 // New returns a new LogClient.
-func New(logID int64, client trillian.TrillianLogClient, hasher merkle.TreeHasher, pubKey gocrypto.PublicKey) *LogClient {
+func New(logID int64, client trillian.TrillianLogClient, hasher merkle.TreeHasher, pubKey gocrypto.PublicKey) LogVerifier {
 	return &LogClient{
 		LogID:    logID,
 		client:   client,
 		hasher:   hasher,
-		MaxTries: 3,
+		maxTries: 3,
 		pubKey:   pubKey,
 	}
+}
+
+// STR returns the current trusted SignedLogRoot
+func (c *LogClient) STR() trillian.SignedLogRoot {
+	return c.str
+}
+
+// SetMaxTries sets the maximum number of requests to make before returning DeadlineExceeded.
+func (c *LogClient) SetMaxTries(tries int) {
+	c.maxTries = tries
 }
 
 // AddLeaf adds leaf to the append only log.
@@ -66,19 +76,19 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
 	switch {
 	case grpc.Code(err) == codes.AlreadyExists:
 		// If the leaf already exists, don't wait for an update.
-		return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.STR.TreeSize)
+		return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.str.TreeSize)
 	case err != nil:
 		return err
 	default:
 		err := grpc.Errorf(codes.NotFound, "Pre-loop condition")
-		for i := 0; grpc.Code(err) == codes.NotFound && i < c.MaxTries; i++ {
+		for i := 0; grpc.Code(err) == codes.NotFound && i < c.maxTries; i++ {
 			// Wait for TreeSize to update.
-			if err := c.waitForSTRUpdate(ctx, c.MaxTries); err != nil {
+			if err := c.waitForSTRUpdate(ctx, c.maxTries); err != nil {
 				return err
 			}
 
 			// Get proof by hash.
-			err = c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.STR.TreeSize)
+			err = c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.str.TreeSize)
 		}
 		return err
 	}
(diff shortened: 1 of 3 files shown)


client/interface.go
```diff
@@ -0,0 +1,29 @@
...
+
+package client
+
+import (
+	"context"
+
+	"github.com/google/trillian"
+)
+
+// LogVerifier is a client that verifies output from Trillian.
```
Martin2112: LogVerifier doesn't seem like a good name as it implies a log reader whereas the API defines a writer. Also AddLeaf does a lot more work than just adding a leaf but I can't think of a better name right now.

I think we can solve the 'STR' naming by using 'Root'. The interface name includes 'Log' already.

```
type VerifyingLogClient interface {
  AddLeaf(ctx context.Context, data []byte) error
￼  UpdateRoot(ctx context.Context) error
￼  Root() trillian.SignedLogRoot
￼}
```

There's a lot of extra overhead compared to a simple client that doesn't get proofs so we should be clear what the indended use is. The documentation should include a description of AddLeaf and how it blocks and is not guaranteed to get a proof result within the specified deadline.

client/client.go
```diff
@@ -37,22 +37,32 @@ type LogClient struct {
...
 		LogID:    logID,
 		client:   client,
 		hasher:   hasher,
-		MaxTries: 3,
+		maxTries: 3,
 		pubKey:   pubKey,
 	}
 }
 
+// STR returns the current trusted SignedLogRoot
```
Martin2112: 'trusted' is slightly incorrect as there is no consistency proof check from the first root read from the log.

If the log has a size > 1 you could ask for a consistency proof from 1 to current size and verify that it's correct rather than always skipping for the first root. I can't comment on that bit of code thanks to Github.

client/client.go
```diff
@@ -37,22 +37,32 @@ type LogClient struct {
 	LogID    int64
 	client   trillian.TrillianLogClient
 	hasher   merkle.TreeHasher
-	STR      trillian.SignedLogRoot
-	MaxTries int
+	str      trillian.SignedLogRoot
```
gdbelvin: Good suggestion.  Thanks!

client/client.go
gdbelvin: Thanks for pointing this out. I'll make the comment a bit clearer. 
Is it standard practice for clients to blindly trust the first root they see? 
Is there a standard way of starting a client off from an empty root or something like that?

client/client.go
```diff
@@ -37,22 +37,32 @@ type LogClient struct {
...
 	}
 }
 
+// STR returns the current trusted SignedLogRoot
+func (c *LogClient) STR() trillian.SignedLogRoot {
+	return c.str
+}
+
+// SetMaxTries sets the maximum number of requests to make before returning DeadlineExceeded.
+func (c *LogClient) SetMaxTries(tries int) {
```
gdbelvin: That's a great suggestion.  Thanks!

client/interface.go
gdbelvin: Added documentation.

Martin2112:
Submitting clients e.g. new certs to CT generally don't need to verify.
Monitors will pick it up later and they retain the previous state of the
tree between runs. Perhaps this client needs to allow a previous root to
specified on creation.

gdbelvin:
As a future feature, this client could be created with a previously known root, or even save and restore it's current state from disk, but that's outside the scope of this PR.

client/client.go
```diff
@@ -34,40 +34,35 @@ import (
...
-	c.maxTries = tries
+// Root returns the current trusted SignedLogRoot.
+// The client blindly trusts the first Root it sees.
+// All others must have consistency proofs between the last Root seen and the new one.
+func (c *LogClient) Root() trillian.SignedLogRoot {
+	return c.root
 }
 
 // AddLeaf adds leaf to the append only log.
 // Blocks until it gets a verifiable response.
```
Martin2112: mention the deadline.

client/client.go
```diff
@@ -76,52 +71,52 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
...
+				"%v. TreeSize: %v, want > %v. Tried %v times.",
+				err, c.root.TreeSize, startTreeSize, i+1)
 		}
 		time.Sleep(b.Duration())
 	}
 }
 
-// UpdateSTR retrieves the current SignedLogRoot and verifies it.
-func (c *LogClient) UpdateSTR(ctx context.Context) error {
+// UpdateRoot retrieves the current SignedLogRoot and verifies it.
```
Martin2112: Maybe explain what "verifies" means either here or in the definition of the verifying client.

client/client.go
```diff
@@ -138,17 +133,17 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
 	}
 
 	// Verify Consistency proof.
-	if str.TreeSize == c.str.TreeSize && bytes.Equal(str.RootHash, c.str.RootHash) {
+	if str.TreeSize == c.root.TreeSize && bytes.Equal(str.RootHash, c.root.RootHash) {
```
Martin2112: Might want to just check the size here if log screws up and issues 2 STHs with different hashes we don't want to block here.

client/client.go
```diff
@@ -138,17 +133,17 @@ func (c *LogClient) UpdateSTR(ctx context.Context) error {
...
 		// Tree has not been updated.
 		return nil
 	}
 
-	// Implicitly trust the first STH we get.
-	if c.str.TreeSize != 0 {
+	// Implicitly trust the first root we get.
```
Martin2112: TODO to check signature? Or is this already happening. Ignore if so.

client/client_test.go
```diff
@@ -68,23 +70,24 @@ func TestAddLeaf(t *testing.T) {
 		},
 	} {
 		client := New(logID, test.client, testonly.Hasher, env.PublicKey)
-		client.SetMaxTries(1)
-
-		if err, want := client.AddLeaf(ctx, []byte(test.desc)), codes.DeadlineExceeded; grpc.Code(err) != want {
-			t.Errorf("AddLeaf(%v): %v, want, %v", test.desc, err, want)
-			continue
+		{
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(timeout))
```
Martin2112: Is this liable to cause test flakes by use of absolute time? Can we use a mock context somehow if so?

client/client.go
gdbelvin: Updated the documentation.

client/client.go

client/client.go

client/client.go
gdbelvin: Signature check happens on L131

client/client_test.go
gdbelvin: In this particular case, it won't cause test flakiness because I'm only using the timeout for the parts of the test that I want to timeout.

PR #458, opened 2017-03-21
Client ListByIndex and GetByIndex

gdbelvin:
Adds ListByIndex and GetByIndex to the client using the existing server methods.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -21,6 +21,7 @@ import (
 	gocrypto "crypto"
 	"crypto/sha256"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/google/trillian"
@@ -86,6 +87,39 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
 		}
 		return err
 	}
+}
+
+// GetByIndex returns a single leaf at the requested index.
+func (c *LogClient) GetByIndex(ctx context.Context, index int64) (*trillian.LogLeaf, error) {
+	resp, err := c.client.GetLeavesByIndex(ctx, &trillian.GetLeavesByIndexRequest{
+		LogId:     c.LogID,
+		LeafIndex: []int64{index},
+	})
+	if err != nil {
+		return nil, err
+	}
+	if got, want := len(resp.Leaves), 1; got != want {
+		return nil, fmt.Errorf("len(leaves): %v, want %v", got, want)
+	}
+	return resp.Leaves[0], nil
+}
+
+// ListByIndex returns the requested leaves by index.
+func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*trillian.LogLeaf, error) {
+	indexes := make([]int64, count)
+	for i := range indexes {
+		indexes[i] = start + int64(i)
+	}
+
+	resp, err := c.client.GetLeavesByIndex(ctx,
+		&trillian.GetLeavesByIndexRequest{
+			LogId:     c.LogID,
+			LeafIndex: indexes,
+		})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Leaves, nil
 }
 
 // waitForRootUpdate repeatedly fetches the Root until the TreeSize changes
--- a/client/client.go
+++ b/client/client.go
@@ -22,6 +22,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/google/trillian"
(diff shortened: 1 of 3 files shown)


codingllama:
Hey, sorry for the delay on reviewing this. Would you mind rebasing first, the changelog seems to mix commits from a few different PRs.

gdbelvin:
This is now rebased and clean.  PTAL.

client/client.go
```diff
@@ -88,6 +89,39 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
...
+}
+
+// ListByIndex returns the requested leaves by index.
+func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*trillian.LogLeaf, error) {
+	indexes := make([]int64, count)
+	for i := range indexes {
+		indexes[i] = start + int64(i)
+	}
+
+	resp, err := c.client.GetLeavesByIndex(ctx,
```
codingllama: Just `return c.client.GetLeavesByIndex(...)` ?

client/client_test.go
```diff
@@ -33,6 +34,91 @@ func TestAddGetLeaf(t *testing.T) {
...
+	client := New(logID, cli, testonly.Hasher, env.PublicKey)
+	// Add a few test leaves.
+	leafData := [][]byte{
+		[]byte("A"),
+		[]byte("B"),
+		[]byte("C"),
+	}
+
+	// TODO(gdbelvin): Replace with batch API.
+	// TODO(gdbelvin): Replace with AddSequenedLeaves API.
```
codingllama: int: Typo on "Sequenced"

client/client_test.go
```diff
@@ -33,6 +34,91 @@ func TestAddGetLeaf(t *testing.T) {
...
+	for _, l := range leafData {
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(timeout))
+		defer cancel()
```
codingllama: Use a single context, outside the loop?

Note that the defers inside the loop will only run once TestGetByIndex exits. Moving the ctx out of the loop clears that up too.

client/client_test.go
```diff
@@ -33,6 +34,91 @@ func TestAddGetLeaf(t *testing.T) {
...
+		if err, want := client.AddLeaf(ctx, l), codes.DeadlineExceeded; grpc.Code(err) != want {
```
codingllama: I wish we could do this without the timeouts. Have you considered having a version of AddLeaf that doesn't block?

A more general question: some logs take a long time to integrate leaves (MMDs go up to 24h). Is the AddLeaf behavior desirable considering such scenarios? Should we dissociate adding from verifying? (I don't expect any changes related to this, it's just a question of mine.)

client/client_test.go
```diff
@@ -33,6 +34,91 @@ func TestAddGetLeaf(t *testing.T) {
...
+		env.Sequencer.OperationLoop() // Sequence the new leaves in-order.
+	}
+
+	for i, l := range leafData {
+		leaf, err := client.GetByIndex(context.Background(), int64(i))
+		if err != nil {
+			t.Errorf("Failed to GetByIndex(%v): %v", i, err)
+		}
+		if got, want := leaf.LeafValue, l; !bytes.Equal(got, want) {
+			t.Errorf("GetByIndex(%v) = %v, want %v", i, got, want)
```
codingllama: nit: Use %X for the binary values?

client/client_test.go
```diff
@@ -33,6 +34,91 @@ func TestAddGetLeaf(t *testing.T) {
...
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(timeout))
+		defer cancel()
+		if err, want := client.AddLeaf(ctx, l), codes.DeadlineExceeded; grpc.Code(err) != want {
+			t.Fatalf("AddLeaf(%v): %v, want, %v", l, err, want)
+		}
+		env.Sequencer.OperationLoop() // Sequence the new leaves in-order.
+	}
+
+	// Fetch leaves.
+	leaves, err := client.ListByIndex(context.Background(), 0, 3)
```
codingllama: Both tests are very similar. Should we maybe rename `TestGetByIndex` and just do this assertion there?

client/client.go
gdbelvin: `GetLeavesByIndex` returns `GetLeavesByIndexResponse`.  This API would like to return the `leaves` variable from within `GetLeavesByIndexResponse`.

client/client_test.go
gdbelvin: thanks

client/client_test.go
gdbelvin: A blocking AddLeaf is indeed bad, but if a client distrusts Trillian, how else should it be convinced that it's data was integrated? To not block, the client needs an SCT and the associated SCT verification routines. 

More concretely, when publishing a map head, the verifiable map personality's sequencer can't proceed until it can provide an inclusion proof for the map head back to clients.

client/client_test.go
gdbelvin: Thanks

client/client_test.go
gdbelvin: After looking at this, we figured out that ctx needs to be inside the loop because each AddLeaf needs to timeout in order to succeed. 

In a future PR, we can have a separate, non-blocking AddLeaf request.

client/client_test.go
gdbelvin: Not sure I follow. Could you elaborate?

gdbelvin:
Addressed nits.  PTAL

client/client_test.go
codingllama: (already discussed, commenting for future reference) Test setup is pretty much identical. Suggestions are to refactor setup to a 3rd function or maybe merge the tests.

client/client.go
codingllama: Right, my bad.

client/client_test.go
codingllama: (Discussed offline too, comments for completeness.)

Re how to verify: we should look into returning some sort of promise. I'm personally against calling it SCT, as I don't want Trillian promises to be mistaken for "real" SCTs, but we definitely want something to play the corresponding part in Trillian.

Re quick integration / blocking: that doesn't solve the problem; outages and delays do happen, and when those happen the client will get an error and have nothing to show for it. I think dissociating Add from Verify (plus having a promise to show for adds) is key for good reliability in face of errors.

PR #464, opened 2017-03-22
Separate proto files for each service

gdbelvin:
Cosmetic change to separate out the Trillian Map messages and
service definition for better readability and code isolation.


gdbelvin:
I would support having a separate file for each service since it makes things much easier to read. 
Do you think the supporting trillian.proto should be split as well?  

The other thing this does is make the dependencies between the different parts of the API explicit, which is quite nice.

codingllama:
I'd say we keep trillian.proto, but split api into 3 per-service files, but I'd rather avoid having 2 files per service for now.

AlCutter:
Sounds fine to me, I like the idea of the explicitly visible dependency.

gdbelvin:
How does this look? PTAL.

trillian_admin_api.proto
```diff
@@ -0,0 +1,92 @@
...
+
+// DeleteTree request.
+message DeleteTreeRequest {
+  // ID of the tree to delete.
+  int64 tree_id = 1;
+}
+
+// Trillian Administrative interface.
+// Allows creation and management of Trillian trees (both log and map trees).
+service TrillianAdmin {
```
codingllama: Suggestion: Move the service to the top? The service is the interesting part, requests/responses are pretty similar / standard.

I'd say the same for the other services.

gdbelvin: The service is indeed the most interesting part. I put it at the bottom because it depends on the messages above it, and someone told me once that that's how dependencies in proto files should be ordered. This seems to be consistent with the other proto files I've seen.

PR #466, opened 2017-03-23
Add a simple counter for monitoring.

pphaneuf:
This is a very modest beginnings towards production-grade monitoring facilities.

### DIFF
--- a/monitoring/metric/counter.go
+++ b/monitoring/metric/counter.go
@@ -1,0 +1,107 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metric
+
+import (
+	"context"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// A Counter is a metric that can only increase.
+type Counter interface {
+	Add(n int64)
+}
+type counter struct {
+	mu              sync.Mutex
+	name            string
+	value           int64
+	lastDumped      time.Time
+	lastDumpedValue int64
+}
+
+type safeMetrics struct {
+	mu sync.Mutex
+	m  map[string]*counter
+}
+
+var (
+	metrics = safeMetrics{
+		m: make(map[string]*counter),
+	}
+)
+
+func (m *counter) Add(n int64) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.value += n
+}
+
+// NewCounter defines a cumulative metric. The name should be unique
+// within a binary.
+func NewCounter(name string) Counter {
+	c := counter{name: name, lastDumped: time.Now()}
+	metrics.mu.Lock()
+	defer metrics.mu.Unlock()
+	if dup := metrics.m[c.name]; dup != nil {
+		glog.Fatal("duplicate metric name registered: ", c.name)
+	}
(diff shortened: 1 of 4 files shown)


monitoring/metric/counter.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		m := metrics.m[key]
+		m.mu.Lock()
+		defer m.mu.Unlock()
+		delta := m.value - m.lastDumpedValue
+		now := time.Now()
+		duration := now.Sub(m.lastDumped)
+		qps := float64(delta) / duration.Seconds()
```
Martin2112: Not sure about building QPS in here. But we don't necessarily want to explicity distinguish between rate and non rate counters. What do you think?

server/trillian_log_server/main.go
```diff
@@ -68,6 +70,9 @@ func main() {
 	glog.CopyStandardLogTo("WARNING")
 	glog.Info("**** Log RPC Server Starting ****")
 
+	// Enable dumping of metrics to the log at regular interval.
+	go metric.DumpToLog(context.Background(), 10*time.Second)
```
Martin2112: Can we have an expvar export as well? Followup is fine. And a flag to control log frequency would be handy for my testing.

monitoring/metric/counter.go
```diff
@@ -0,0 +1,99 @@
...
+	m  map[string]*counter
+}
+
+var (
+	metrics = safeMetrics{
+		m: make(map[string]*counter),
+	}
+)
+
+func (m *counter) Add(n int64) {
```
gdbelvin: Consider making an interface for this

monitoring/metric/counter.go
```diff
@@ -0,0 +1,99 @@
...
+	lastDumpedValue int64
+}
+
+type safeMetrics struct {
+	mu sync.Mutex
+	m  map[string]*counter
+}
+
+var (
+	metrics = safeMetrics{
```
pphaneuf: The model is that code everywhere can just create their metrics, without any knowledge of where they should go, just caring about updating them, and then, their values can be queried and published in different ways (the `DumpToLog` is the simplistic one I'm providing initially, but there would be an HTTP handler that would return them in a format suitable for Prometheus, for example).

monitoring/metric/counter.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func (m *counter) Add(n int64) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.value += n
+}
+
+func NewCounter(name string) Counter {
+	c := counter{name: name, lastDumped: time.Now()}
```
pphaneuf: The time aspect of this is somewhat related to having a very simplistic (as in, mostly inlined) "backend", dumping to the logs. Ideally, a more fleshed out backend would be keeping track of the time itself, possibly?

Keep an eye out for further developments... :smiley:

monitoring/metric/counter.go
pphaneuf: I see this as part of the "dumping to log backend", rather than the metric, but they're slightly intertwined for now. In the case of a Prometheus handler, we wouldn't do that at all, we'd just give out the current value, and let Prometheus calculate rates.

monitoring/metric/counter.go
```diff
@@ -0,0 +1,99 @@
...
+	glog.Info("dumping metrics:")
+	keys := make([]string, 0, len(metrics.m))
+	for k, _ := range metrics.m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		m := metrics.m[key]
+		m.mu.Lock()
+		defer m.mu.Unlock()
```
pphaneuf: Function scoping, not block scoping, damn it! :stuck_out_tongue: 

I'll unlock explicitly, then.

server/trillian_log_server/main.go
pphaneuf: I'd like to have expvars for them, yes, I'll build on this.

I'll add a flag now, though, good idea.

monitoring/metric/counter.go
Martin2112: OK let's try this for now.

pphaneuf:
This is related to working on #467, of course.

PR #469, opened 2017-03-23
Refactor log/map server entry points and add REST support

codingllama:
Shared logic between log and map servers has been refactored to server.Main.
REST support has been added, via grpc-gateway, using the same port as expvars.

The admin server implementation used by log and map servers is now REST-enabled,
provided that the HTTP/REST endpoint is not explicitly disabled.


codingllama:
Issue #310 

Example usage:
```bash
# tab1
./trillian_log_server -alsologtostderr

#tab2
tree_id=$(./createtree \
    -admin_server=localhost:8090 \
    -pem_key_path=testdata/log-rpc-server.privkey.pem \
    -pem_key_password=towel)
curl -H 'Content-Type: application/json' localhost:8091/v1beta1/trees/$tree_id
{"tree_id":"4442952201554314888","tree_state":"ACTIVE","tree_type":"LOG","hash_strategy":"RFC_6962","hash_algorithm":"SHA256","signature_algorithm":"RSA","duplicate_policy":"DUPLICATES_NOT_ALLOWED","create_time_millis_since_epoch":"1490295619000","update_time_millis_since_epoch":"1490295619000","private_key":{"@type":"type.googleapis.com/trillian.PEMKeyFile","path":"testdata/log-rpc-server.privkey.pem","password":"towel"}}
```

(createtree is from #465)

.travis.yml
```diff
@@ -35,6 +35,8 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/golang/protobuf/proto
   - go get -u github.com/golang/protobuf/protoc-gen-go
+  - go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
+  - go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
```
gdbelvin: Are we generating swagger definitions as well?

server/trillian_log_server/main.go
```diff
@@ -28,100 +28,69 @@ import (
...
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/reflection"
 )
 
 var (
 	mySQLURI            = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	serverPortFlag      = flag.Int("port", 8090, "Port to serve log RPC requests on")
-	exportRPCMetrics    = flag.Bool("export_metrics", true, "If true starts HTTP server and exports stats")
-	httpPortFlag        = flag.Int("http_port", 8091, "Port to serve HTTP metrics on")
+	httpPortFlag        = flag.Int("http_port", 8091, "Port to serve HTTP metrics and REST requests on (negative means disabled)")
```
gdbelvin: Consider renaming this flag to something like `metrics_port`. 
Now that we have http and grpc endpoints, `http_port` is overloaded.

.travis.yml
codingllama: Good catch, removed.

server/trillian_log_server/main.go
codingllama: `metrics_port` was a good name *before*, but now it doubles as metrics *and* REST endpoint, so I don't think the rename is appropriate.

We could do a bit more port-fu (like you mentioned to me before) and use a single port for everything. I'll propose that in a top-level comment.

codingllama:
Something I forgot to mention yesterday:

We could do a little more port-fu, like we do to [switch between grpc-gateway and expvar](https://github.com/google/trillian/pull/469/files#diff-d3d4b092ce8121dc129c8b29e3545206R80), and look for `Content-Type: application/grpc` as well. That would allow us to serve all traffic from a single port. I felt I had enough going on here already, so I didn't do it, but I'm happy to follow up in a subsequent PR.

server/trillian_log_server/main.go
```diff
@@ -28,100 +28,69 @@ import (
...
+	s := grpc.NewServer(grpc.UnaryInterceptor(stats.Interceptor()))
+	// No defer: server ownership is delegated to server.Main
 
-	// Set up the listener for the server
-	// TODO(Martin2112): More flexible listen address configuration
-	glog.Infof("Creating RPC server starting on port: %d", *serverPortFlag)
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *serverPortFlag))
-	if err != nil {
-		glog.Exitf("Failed to listen on the server port: %d, because: %v", *serverPortFlag, err)
+	var httpEndpoint = ""
```
AlCutter: `var httpEndpoint string` (or `httpEndpoint := ""`)

server/vmap/trillian_map_server/main.go
```diff
@@ -28,91 +27,63 @@ import (
...
+	s := grpc.NewServer()
+	// No defer: server ownership is delegated to server.Main
 
-	// Set up the listener for the server
-	glog.Infof("Creating RPC server starting on port: %d", *serverPortFlag)
-	// TODO(Martin2112): More flexible listen address configuration
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *serverPortFlag))
-	if err != nil {
-		glog.Exitf("Failed to listen on the server port: %d, because: %v", *serverPortFlag, err)
+	var httpEndpoint = ""
```
AlCutter: Same suggestion here

server/trillian_log_server/main.go
gdbelvin: Perhaps just update the flag description then to indicate both purposes.

server/trillian_log_server/main.go
codingllama: Done.

server/vmap/trillian_map_server/main.go
codingllama: Done.

server/trillian_log_server/main.go
codingllama: It says it already, but let me know if you'd like any changes.

codingllama:
Travis is failing because the `expvar.Handler` was only added in go1.8. I'm investigating a way around it, but I'll probably have a go (no pun intended) at the 1.8 update later today.

codingllama:
PTAL?

server/trillian_log_server/main.go
```diff
@@ -28,100 +28,69 @@ import (
...
 
-	// Set up the listener for the server
-	// TODO(Martin2112): More flexible listen address configuration
-	glog.Infof("Creating RPC server starting on port: %d", *serverPortFlag)
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *serverPortFlag))
-	if err != nil {
-		glog.Exitf("Failed to listen on the server port: %d, because: %v", *serverPortFlag, err)
+	httpEndpoint := ""
+	if *httpPortFlag >= 0 {
+		httpEndpoint = fmt.Sprintf("localhost:%v", *httpPortFlag)
```
AlCutter: You're probably going to want a way to specify the address to bind to too, I expect?

Operators may want to expose the RPC and HTTP servers on different addresses too, depending on the deployment specifics, would it make sense to dump the *Port flags and just have the bind address string passed in by the flags for this and the RPC endpoints?

codingllama: Yes, I'd like that too. Raised #477, will follow up in a separate PR.

PR #471, opened 2017-03-24
Redact private key from admin responses

codingllama:


server/admin/admin_server.go
```diff
@@ -64,7 +65,7 @@ func (s *Server) getTreeImpl(ctx context.Context, request *trillian.GetTreeReque
 	if err := tx.Commit(); err != nil {
 		return nil, err
 	}
-	return tree, nil
+	return redact(tree), nil
```
codingllama: Good call, added to `Tree` / `trillian.proto`.

PR #473, opened 2017-03-24
Make key paths work regardless of working directory

RJPercival:
`RelativeToPackage(path)` will convert `path` to an absolute path. This means the key files can be found regardless of the location of test.

### DIFF
--- a/storage/testonly/admin_storage_tester.go
+++ b/storage/testonly/admin_storage_tester.go
@@ -25,6 +25,7 @@ import (
 	"github.com/google/trillian"
 	spb "github.com/google/trillian/crypto/sigpb"
 	"github.com/google/trillian/storage"
+	testonly2 "github.com/google/trillian/testonly"
 	"github.com/kylelemons/godebug/pretty"
 )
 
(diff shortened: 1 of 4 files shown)


storage/testonly/admin_storage_tester.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/google/trillian"
 	spb "github.com/google/trillian/crypto/sigpb"
 	"github.com/google/trillian/storage"
+	testonly2 "github.com/google/trillian/testonly"
```
codingllama: nit: Could we find a better alias? `trilliantestonly`, `trltestonly`, `ttestonly`, something else?

testonly/integration/clientserver.go
```diff
@@ -30,7 +30,8 @@ import (
 	"github.com/google/trillian/extension"
 	"github.com/google/trillian/server"
 	"github.com/google/trillian/storage/mysql"
-	"github.com/google/trillian/storage/testonly"
+	strtestonly "github.com/google/trillian/storage/testonly"
```
codingllama: nit: `str` reminds me of `string`. Maybe `stestonly`?

storage/testonly/admin_storage_tester.go
RJPercival: Done.

testonly/integration/clientserver.go
RJPercival: Done.

PR #478, opened 2017-03-27
Run stringer on errors.Code

RJPercival:
This makes error codes easier to read:
`fmt.Sprint(codes.InvalidArgument) == "InvalidArgument"` instead of `"3"`


RJPercival:
Travis is happier now.

PR #480, opened 2017-03-27
Add support for key generation via registry.SignerFactory

RJPercival:
`admin.Server` will now check whether a tree's private key exists before attempting to create it. If it does not exist, but the `SignerFactory` implements `keys.Generator`, it will attempt to generate a new key. 

To support this, the `crypto/keys` package now returns `trillian.Error` where appropriate so that callers, such as `admin.Server`, may make decisions based on the type of error that occurred (e.g. distinguishing a key that does not exist from an invalid key).


crypto/keys/keys_test.go
```diff
@@ -138,21 +139,19 @@ func TestLoadPrivateKeyAndSign(t *testing.T) {
 		switch {
 		case test.keyPEM != "":
 			k, err = NewFromPrivatePEM(test.keyPEM, test.keyPass)
-			switch gotErr := err != nil; {
-			case gotErr != test.wantLoadErr:
-				t.Errorf("%s: NewFromPrivatePEM(_, _) = (%v, %v), want err? %t", test.name, k, err, test.wantLoadErr)
+			if got, want := errors.ErrorCode(err), test.wantLoadErr; got != want {
```
codingllama: Suggestion: Move the if outside the cases?

```go
var k crypto.Signer
var err error
switch {
case test.keyPEM != "":
  k, err = NewFromPrivatePEM(test.keyPEM, test.keyPass)
case test.keyPath != "":
  k, err = NewFromPrivatePEMFile(test.keyPath, test.keyPass)
}
if got, want := ... {
  t.Errorf("%v: NewFromPrivate() ...", ...)
}
```

There's a slight loss in the error message, but since we log the test name it should be fine. The readability / code duplication gain is significant, though.

crypto/keys/keys_test.go
```diff
@@ -138,21 +139,19 @@ func TestLoadPrivateKeyAndSign(t *testing.T) {
...
-			switch gotErr := err != nil; {
-			case gotErr != test.wantLoadErr:
-				t.Errorf("%s: NewFromPrivatePEMFile(_, _) = (%v, %v), want err? %t", test.name, k, err, test.wantLoadErr)
+			if got, want := errors.ErrorCode(err), test.wantLoadErr; got != want {
+				t.Errorf("%s: NewFromPrivatePEMFile(_, _) = (_, %v: %v), want err.Code == %v", test.name, got, err, want)
 				continue
-			case gotErr:
+			} else if got != errors.OK {
 				continue
 			}
```
codingllama: nit: I'd be OK removing the line below. The test doesn't need to check itself.

crypto/keys/pem_signer_factory.go
```diff
@@ -32,18 +32,18 @@ type PEMSignerFactory struct{}
 // NewSigner returns a crypto.Signer for the given tree.
 func (f PEMSignerFactory) NewSigner(ctx context.Context, tree *trillian.Tree) (crypto.Signer, error) {
 	if tree.GetPrivateKey() == nil {
-		return nil, fmt.Errorf("tree %d has no PrivateKey", tree.GetTreeId())
+		return nil, errors.Errorf(errors.InvalidArgument, "tree %d has no PrivateKey", tree.GetTreeId())
```
codingllama: nit: FailedPrecondition? Isn't that a system state failure?

crypto/keys/pem_signer_factory.go
```diff
@@ -32,18 +32,18 @@ type PEMSignerFactory struct{}
...
+		return nil, errors.Errorf(errors.InvalidArgument, "failed to unmarshal private key for tree %d: %v", tree.GetTreeId(), err)
 	}
 
 	switch privateKey := privateKey.Message.(type) {
 	case *trillian.PEMKeyFile:
 		return NewFromPrivatePEMFile(privateKey.GetPath(), privateKey.GetPassword())
 	}
 
-	return nil, fmt.Errorf("unsupported PrivateKey type for tree %d: %T", tree.GetTreeId(), privateKey.Message)
+	return nil, errors.Errorf(errors.Unimplemented, "unsupported PrivateKey type for tree %d: %T", tree.GetTreeId(), privateKey.Message)
```
codingllama: nit: InvalidArgument? Unimplemented tends to refer to the entire RPC.

crypto/keys/private.go
```diff
@@ -34,20 +34,32 @@ import (
...
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
 // The key may be protected by a password.
 func NewFromPrivatePEMFile(keyFile, keyPassword string) (crypto.Signer, error) {
 	pemData, err := ioutil.ReadFile(keyFile)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read private key from file %q: %v", keyFile, err)
+		if os.IsNotExist(err) {
+			return nil, errors.Errorf(errors.NotFound, "file does not exist: %q", keyFile)
+		}
+		return nil, errors.Errorf(errors.InvalidArgument, "failed to read private key from file %q: %v", keyFile, err)
```
codingllama: nit: FailedPrecondition.

crypto/keys/private.go
```diff
@@ -57,13 +69,19 @@ func NewFromPrivatePEMFile(keyFile, keyPassword string) (crypto.Signer, error) {
...
+		return nil, errors.New(errors.InvalidArgument, "failed to decode PEM block")
+	}
 	if len(rest) > 0 {
-		return nil, errors.New("extra data found after PEM decoding")
+		return nil, errors.New(errors.InvalidArgument, "extra data found after PEM decoding")
 	}
 
 	der := block.Bytes
 	if password != "" {
 		pwdDer, err := x509.DecryptPEMBlock(block, []byte(password))
```
codingllama: nit: pwdDER?

crypto/keys/testonly/signer_factory.go
```diff
@@ -0,0 +1,89 @@
...
+
+// KeyGenerator is a stub keys.SignerFactory and keys.Generator.
+type KeyGenerator struct {
+	SignerFactory
+	// generateErr is the error that will be returned by Generate().
+	generateErr error
+}
+
+// Generate simulates generation of a new private key for the given tree.
+func (s KeyGenerator) Generate(ctx context.Context, tree *trillian.Tree) error {
```
codingllama: nit: `g KeyGenerator` (instead of `s`)?

crypto/keys/testonly/signer_factory.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// KeyGenerator is a stub keys.SignerFactory and keys.Generator.
+type KeyGenerator struct {
+	SignerFactory
+	// generateErr is the error that will be returned by Generate().
+	generateErr error
+}
+
+// Generate simulates generation of a new private key for the given tree.
```
codingllama: Document that subsequent invocations will return an error, regardless of the supplied tree?

server/admin/admin_server.go
```diff
@@ -72,12 +74,28 @@ func (s *Server) getTreeImpl(ctx context.Context, request *trillian.GetTreeReque
...
-		return nil, errors.WrapError(err)
+		return nil, svrerrors.WrapError(err)
 	}
 	return tree, err
 }
 
 func (s *Server) createTreeImpl(ctx context.Context, request *trillian.CreateTreeRequest) (*trillian.Tree, error) {
+	// Test that the tree's private key exists. If it doesn't, generate it.
+	if _, err := s.registry.SignerFactory.NewSigner(ctx, request.GetTree()); err != nil {
+		if errors.ErrorCode(err) != errors.NotFound {
```
codingllama: I dislike the double-edged logic. I suggest the following:

* If a non-nil PrivateKey was supplied it must be valid or the RPC will fail
* If a nil PrivateKey was supplied we try to generate it (fail if we can't).

Generating a new key in case of errors may be surprising for callers. It'll also be hard to detect, given that we redact PKs from responses (it happens silently and without any feedback).

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
+type testFixture struct {
 	registry   extension.Registry
 	as         *storage.MockAdminStorage
 	tx         *storage.MockAdminTX
 	snapshotTX *storage.MockReadOnlyAdminTX
 	server     *Server
 }
 
-// setupAdminStorage configures storage mocks according to input parameters.
+type testOptions struct {
```
codingllama: Suggestion: maybe rename to setupOpts, so it's clearer where it's applied.

server/admin/admin_server_test.go
```diff
@@ -119,42 +124,55 @@ func TestAdminServer_GetTree(t *testing.T) {
 	invalidTree.TreeState = trillian.TreeState_HARD_DELETED
 
 	tests := []struct {
-		desc              string
-		getErr, commitErr bool
+		desc string
+		// getErr is the error that will be returned by the mock AdminTX.GetTree() method.
+		getErr error
```
codingllama: Maybe rename to getTreeErr and lose the comment?

server/admin/admin_server_test.go
```diff
@@ -119,42 +124,55 @@ func TestAdminServer_GetTree(t *testing.T) {
...
+		opts   testOptions
 	}{
 		{
 			desc: "success",
+			opts: testOptions{
+				shouldSnapshot: true,
```
codingllama: Could we phrase opts in a way that the defaults are the success path? I'm more interested in tests expressing what fails instead of what works.

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
 	tests := []struct {
-		desc                 string
-		req                  *trillian.CreateTreeRequest
-		createErr, commitErr bool
+		desc string
+		req  *trillian.CreateTreeRequest
+		// shouldCreate indicates whether AdminTX.CreateTree() is expected to be called.
+		shouldCreate bool
+		// createErr is the error that will be returned by the mock AdminTX.CreateTree() method.
+		createErr error
```
codingllama: As above: creatTreeErr and drop the comment?

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
+			req:          &trillian.CreateTreeRequest{Tree: testonly.LogTree},
+			shouldCreate: true,
+			opts: testOptions{
+				shouldBeginTx: true,
+				shouldCommit:  true,
+				commitErr:     errors.New("commit error"),
+			},
+		},
+		{
+			desc: "non-existent key and SignerFactory does not implement keys.Generator",
```
codingllama: Maybe `desc: "noKeyWithoutGenerator"`, so the error message is smaller.

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
 			req:  &trillian.CreateTreeRequest{Tree: testonly.LogTree},
+			opts: testOptions{
+				signerFactory: ktestonly.NewSignerFactoryWithErr(terrors.New(terrors.NotFound, "key not found")),
+			},
 		},
 		{
-			desc:      "invalidTree",
-			req:       &trillian.CreateTreeRequest{Tree: &invalidTree},
-			createErr: true,
+			desc:         "non-existent key but SignerFactory implements keys.Generator",
```
codingllama: `"noKeyWithGenerator"` (as above).

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
+			shouldCreate: true,
+			opts: testOptions{
+				signerFactory: ktestonly.NewKeyGenerator(),
+				shouldBeginTx: true,
+				shouldCommit:  true,
+			},
 		},
 		{
-			desc:      "commitError",
+			desc: "key generation failure",
```
codingllama: nit: "keyGenFailure", "invalidKey" (below), so on for others.

server/admin/admin_server_test.go
```diff
@@ -174,107 +192,176 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
 // Whether the snapshot/TX is expected to be committed (and if it should error doing so) is
 // controlled via shouldCommit and commitErr parameters.
-func setupAdminStorage(ctrl *gomock.Controller, snapshot, shouldCommit, commitErr bool) adminTestSetup {
+func setupServer(ctrl *gomock.Controller, opts testOptions) testFixture {
 	as := storage.NewMockAdminStorage(ctrl)
 
 	var snapshotTX *storage.MockReadOnlyAdminTX
-	var tx *storage.MockAdminTX
-	if snapshot {
+	if opts.shouldSnapshot {
```
codingllama: I like what you did here. Nice.

server/admin/admin_server.go
RJPercival: The issue with that logic is that the `tree.PrivateKey` field currently tells the `keys.Generator` how to generate the key. For example, if `PrivateKey` was a `PEMKeyFile`, it would tell the `Generator` what password to encrypt the key with. Other key types may similarly require some metadata in order to generate a key. I suppose that isn't absolutely necessary though, as the `Generator` could probably just be hard-coded with particular settings, at the cost of some flexibility. Your proposal certainly has clearer behaviour.

crypto/keys/keys_test.go
RJPercival: Which line is that?

codingllama: Sorry, not sure why I added this so far away. I meant line 158 (which Github definitely won't let me add comments on):

```go
default:
  t.Errorf("%s: No PEM or file path set in test definition", test.name)
  continue
```

crypto/keys/keys_test.go
```diff
@@ -100,13 +101,13 @@ func TestLoadPrivateKeyAndSign(t *testing.T) {
 		{
 			name:        "Non-existent file",
 			keyPath:     "non-existent.pem",
-			wantLoadErr: true,
+			wantLoadErr: errors.NotFound,
```
codingllama: SGTM.

crypto/keys/private.go
codingllama: SGTM.

server/admin/admin_server.go
codingllama: A way we could do this is to add key generation settings to the CreateTreeRequest. Actually, that might be even better than behaving differently depending on the supplied private key.

server/admin/admin_server_test.go
```diff
@@ -188,96 +188,125 @@ func TestAdminServer_CreateTree(t *testing.T) {
...
 				commitErr:     errors.New("commit error"),
 			},
 		},
 		{
-			desc: "non-existent key and SignerFactory does not implement keys.Generator",
-			req:  &trillian.CreateTreeRequest{Tree: testonly.LogTree},
-			opts: testOptions{
-				signerFactory: ktestonly.NewSignerFactoryWithErr(terrors.New(terrors.NotFound, "key not found")),
+			desc: "nilPrivateKeyAndNoKeyGenerator",
+			req:  &trillian.CreateTreeRequest{Tree: &nilPrivateKeyTree},
```
codingllama: Aren't we missing some sort of createErr or wantErr option here? Same for the following error tests.

RJPercival: wantErr is inferred from the setupOptions (see line 352). It's probably better to make it explicit in the test case though, so I can change that.

codingllama: Oh, I see, it's because of the setupOptions defaults.

A previous comment suggested for the defaults to be the success scenario. If we do that this comment may solve itself, as we'd have some sort of error flag to true here.

crypto/keys/keys_test.go
RJPercival: Raised issue #508 to track that.

server/admin/admin_server.go
RJPercival: Sounds reasonable, I'll do that.

RJPercival:
I've re-implemented this in PR #582, taking a slightly different, cleaner approach.

PR #481, opened 2017-03-28
Cache signers in SequenceManager

RJPercival:
This avoids having to reload the private key for each tree at the beginning of every sequencing pass.

### DIFF
--- a/server/sequencer_manager.go
+++ b/server/sequencer_manager.go
@@ -17,6 +17,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/google/trillian"
@@ -28,9 +29,10 @@ import (
 
 // SequencerManager provides sequencing operations for a collection of Logs.
 type SequencerManager struct {
-	guardWindow time.Duration
-	registry    extension.Registry
-	signers     map[int64]*crypto.Signer
+	guardWindow  time.Duration
+	registry     extension.Registry
+	signers      map[int64]*crypto.Signer
+	signersMutex sync.Mutex
 }
 
 // NewSequencerManager creates a new SequencerManager instance based on the provided KeyManager instance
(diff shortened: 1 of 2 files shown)


server/sequencer_manager.go
```diff
@@ -66,6 +68,17 @@ func (s SequencerManager) ExecutePass(logIDs []int64, logctx LogOperationManager
 	toSeq := make(chan int64, len(logIDs))
 
 	for _, logID := range logIDs {
+		// Get a signer for this log, if there isn't one in s.signers already.
+		if _, ok := s.signers[logID]; !ok {
+			ctx := util.NewLogContext(logctx.ctx, logID)
+			signer, err := newSigner(ctx, s.registry, logID)
+			if err != nil {
+				glog.Errorf("Could not get signer for log %d: %v", logID, err)
+				continue
```
codingllama: Is continue the right action to take here? Won't that cause a null pointer later on log.Sequencer?

Could please add a test for this scenario?

RJPercival: Yes, continue is the right action, as it will prevent the logID being added to the `toSeq` channel, which avoids the null pointer scenario I think you're talking about. I'll add a test to confirm.

codingllama: You're right. Thanks for adding the test, though.

RJPercival: Done.

RJPercival:
Rebased and ready for review.

server/sequencer_manager.go
```diff
@@ -37,6 +39,7 @@ func NewSequencerManager(registry extension.Registry, gw time.Duration) *Sequenc
 	return &SequencerManager{
 		guardWindow: gw,
 		registry:    registry,
+		signers:     make(map[int64]*crypto.Signer),
```
codingllama: Should we protect this with a mutex, or are Sequencers not meant to be thread safe? If the latter, please document.

I believe this is tied to the Travis failure as well.

RJPercival: Done.

PR #483, opened 2017-03-28
Use inclusion proof verifier in integration test

### LINKED ISSUE #333: Use Inclusion Proof Verifier in tests
Current tests verify inclusion proofs by rebuilding a parallel tree and checking that the inclusion proofs are the same.  We should migrate these tests to use a Log Verifier that computes the root hash from the neighbor nodes.

gdbelvin:
Strengthen integration test by doing a full inclusion and consistency proof verification. 

Simplify the test by removing checks that are done inside of the Merkle proof verifier.

Closes #333

### DIFF
--- a/integration/log.go
+++ b/integration/log.go
@@ -352,7 +352,11 @@ func checkLogRootHashMatches(tree *merkle.InMemoryMerkleTree, client trillian.Tr
 func checkInclusionProofLeafOutOfRange(logID int64, client trillian.TrillianLogClient, params TestParameters) error {
 	// Test is a leaf index bigger than the current tree size
 	ctx, cancel := getRPCDeadlineContext(params)
-	proof, err := client.GetInclusionProof(ctx, &trillian.GetInclusionProofRequest{LogId: logID, LeafIndex: params.leafCount + 1, TreeSize: int64(params.leafCount)})
+	proof, err := client.GetInclusionProof(ctx, &trillian.GetInclusionProofRequest{
+		LogId:     logID,
+		LeafIndex: params.leafCount + 1,
+		TreeSize:  int64(params.leafCount),
+	})
 	cancel()
 
 	if err == nil {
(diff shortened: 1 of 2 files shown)


integration/log.go
```diff
@@ -382,6 +390,21 @@ func checkInclusionProofTreeSizeOutOfRange(logID int64, client trillian.Trillian
 // proofs returned should match ones computed by the alternate Merkle Tree implementation, which differs
 // from what the log uses.
 func checkInclusionProofsAtIndex(index int64, logID int64, tree *merkle.InMemoryMerkleTree, client trillian.TrillianLogClient, params TestParameters) error {
+	// Get MerkleLeafHash to use in VerifyInclusionProof.
```
Martin2112: I think this should use a local copy of the leaf hash. If there are weird bugs in the log where it uses the wrong hash they could be hidden if that makes the proof still verify.

gdbelvin: I didn't see a local copy of the leaf hash. Please point it out if I'm missing it. 

In the meantime, I've adjusted the test to compute the MerkleLeafHash after fetching the leaf value.

Martin2112:
The in memory tree was loaded with the leaves so has the same hashes. I
think the method you want is not exported leafHash(index), which could be
changed.

Martin

gdbelvin:
Would it be better to use the leaf values from the tree directly?  If I understand correctly, the log doesn't provide any guarantees about where particular leaf values will be placed.

Martin2112:
The leaves are already read back from the tree and checked that all the
ones that were submitted were present. The in memory tree is then built in
the same order that the log sequenced them. I don't mind where the data
comes from but there didn't seem to be a need to read everything again.

Martin

gdbelvin:
Using an off by one offset allowed this test to use `InMemory.LeafHash` rather than making an extra get request to the server.

PR #497, opened 2017-03-29
Remove node internals from RPC proof responses to client

### LINKED ISSUE #350: API: Hide node_revision and node_id in message Node
Public clients only need an array of neighbor hashes to verify proofs. 
Including extra data invites API missuse.

Martin2112:
Closes #350.


trillian_log_api.proto
```diff
@@ -53,10 +53,9 @@ message LogLeaf {
 }
 
 message Node {
```
gdbelvin: Now that `Node` only only contains `node_hash`, could we convert 
`repeated Node proof_node` into `repeated bytes proof_node`? 
This would make verifying inclusion proofs easier. 

The current proto requires something like:
```go
proof := make([]byte, 0, len(Proof.ProofNode))
for _, n := range Proof.ProofNode {
   proof = append(proof, n.NodeHash)
}
VerifyInclusionProof(..., proof, ...) 
```

If `Proof.ProofNode` was `repeated bytes` This could be simplified to 
```go
VerifyInclusionProof(..., Proof.ProofNode, ...) 

```

Martin2112:
If we do that it should be a separate PR as it's a more intrusive change than this.

gdbelvin:
Sounds good.

Martin2112:
If you've got time to write the PR I'll review it. It could be a while before I get to it and it probably affects the client most.

PR #500, opened 2017-03-29
Client.VerifyLeaf

gdbelvin:
Add functions for clients to verify inclusion of known leaves at any
location and at a particular index.

This supports Key Transparency's need to verify inclusion of
SignedMapHeads at particular locations in Trillian.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -217,7 +217,7 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
 	return nil
 }
 
-// VerifyInclusion ensures that the given leaf data has been included in the log.
+// VerifyInclusion updates the log root and ensures that the given leaf data has been included in the log.
 func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
 	leaf := c.buildLeaf(data)
 	if err := c.UpdateRoot(ctx); err != nil {
@@ -226,7 +226,7 @@ func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
 	return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.root.TreeSize)
 }
 
-// VerifyInclusionAtIndex ensures that the given leaf data has been included in the log at a particular index.
+// VerifyInclusionAtIndex updates the log root and ensures that the given leaf data has been included in the log at a particular index.
 func (c *LogClient) VerifyInclusionAtIndex(ctx context.Context, data []byte, index int64) error {
 	leaf := c.buildLeaf(data)
 	if err := c.UpdateRoot(ctx); err != nil {
@@ -240,9 +240,6 @@ func (c *LogClient) VerifyInclusionAtIndex(ctx context.Context, data []byte, ind
 	resp, err := c.client.GetInclusionProof(ctx, req)
 	if err != nil {
 		return err
-	}
-	if len(resp.GetProof().ProofNode) < 1 {
-		return errors.New("no inclusion proof supplied")
 	}
 
 	v := merkle.NewLogVerifier(c.hasher)
--- a/client/client.go
+++ b/client/client.go
@@ -217,6 +217,40 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
 	return nil
 }
 
+// VerifyInclusion ensures that the given leaf data has been included in the log.
+func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -217,6 +217,40 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
...
+// VerifyInclusion ensures that the given leaf data has been included in the log.
+func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
+	leaf := c.buildLeaf(data)
+	if err := c.UpdateRoot(ctx); err != nil {
+		return fmt.Errorf("UpdateRoot(): %v", err)
+	}
+	return c.getInclusionProof(ctx, leaf.MerkleLeafHash, c.root.TreeSize)
+}
+
+// VerifyInclusionAtIndex ensures that the given leaf data has been included in the log at a particular index.
```
codingllama: Please document that this causes a root update (as per UpdateRoot)

Btw, is that something we *need* to do? It's easier to reason about client state if it doesn't have implied calls.

client/client.go
```diff
@@ -217,6 +217,40 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
...
+	req := &trillian.GetInclusionProofRequest{
+		LogId:     c.LogID,
+		LeafIndex: index,
+		TreeSize:  c.root.TreeSize,
+	}
+	resp, err := c.client.GetInclusionProof(ctx, req)
+	if err != nil {
+		return err
+	}
+	if len(resp.GetProof().ProofNode) < 1 {
```
codingllama: nit: I wouldn't expect the server to return 0-length proofs - that sounds like a bug to me. If you want to keep it as a defensive check that's ok, but maybe add a comment saying so?

client/client_test.go
```diff
@@ -121,6 +121,66 @@ func TestListByIndex(t *testing.T) {
...
+		[]byte("A"),
+		[]byte("B"),
+	}
+
+	if err := addSequencedLeaves(env, client, leafData); err != nil {
+		t.Errorf("Failed to add leaves: %v", err)
+	}
+
+	for _, l := range leafData {
+		if err := client.VerifyInclusion(context.Background(), l); err != nil {
```
codingllama: nit: Use a single context for all tests?

https://github.com/golang/go/wiki/CodeReviewComments#contexts - "Contexts are immutable, so it's fine to pass the same ctx to multiple calls that share the same deadline, cancellation signal, credentials, parent trace, etc."

client/client_test.go
```diff
@@ -121,6 +121,66 @@ func TestListByIndex(t *testing.T) {
...
+		[]byte("A"),
+		[]byte("B"),
+	}
+
+	if err := addSequencedLeaves(env, client, leafData); err != nil {
+		t.Errorf("Failed to add leaves: %v", err)
+	}
+
+	for i, l := range leafData {
+		if err := client.VerifyInclusionAtIndex(context.Background(), l, int64(i)); err != nil {
```
codingllama: nit: Single context too?

client/interface.go
```diff
@@ -26,9 +26,13 @@ type VerifyingLogClient interface {
 	// is available. If no proof is available within the ctx deadline, DeadlineExceeded
 	// is returned.
 	AddLeaf(ctx context.Context, data []byte) error
-	// GetByIndex returns a single leaf.
+	// VerifyInclusion ensures that data has been included in the log.
```
codingllama: Should we mention the verification mechanisms (ie, that it checks via inclusion proof), or are we assuming the client's audience understands that?

client/client.go
gdbelvin: Added documentation.

I personally find it easier to reason about the state of the client if the client *always* first updates the root, otherwise "not found" failures are ambiguous. Did we request an inclusion proof to too small of a tree size, or is the requested item not there? 

Is there a better, simpler algorithm for interacting with the log?

client/client.go
gdbelvin: This check is not needed.

client/client_test.go
gdbelvin: I didn't know they were immutable. Thanks for pointing that out!

client/client_test.go

client/interface.go
gdbelvin: added comments.

gdbelvin:
Updated according to reviewer comments. PTAL

gdbelvin:
Looks like it's passing now.

PR #506, opened 2017-03-31
ListTrees RPC implementation (admin.Server)

codingllama:


server/admin/admin_server_test.go
```diff
@@ -111,7 +112,86 @@ func TestAdminServer_BeginError(t *testing.T) {
...
+		want := deepCopy(storedTrees, func(t *trillian.Tree) { t.PrivateKey = nil })
+		if diff := pretty.Compare(resp.Tree, want); diff != "" {
+			t.Errorf("%v: post-ListTrees diff:\n%v", test.desc, diff)
+		}
+	}
+}
+
+// deepCopy makes a deep copy of a slice, allowing for an optional redact function to run on every
+// element.
+func deepCopy(s []*trillian.Tree, f func(*trillian.Tree)) []*trillian.Tree {
```
RJPercival: `copyAndUpdate` might be a better name, given the `f` parameter.

server/admin/admin_server_test.go
```diff
@@ -111,7 +112,86 @@ func TestAdminServer_BeginError(t *testing.T) {
...
+		if hasErr, wantErr := err != nil, test.listErr || test.commitErr; hasErr != wantErr {
+			t.Errorf("%v: ListTrees() = (_, %q), wantErr = %v", test.desc, err, wantErr)
+			continue
+		} else if hasErr {
+			continue
+		}
+
+		want := deepCopy(storedTrees, func(t *trillian.Tree) { t.PrivateKey = nil })
+		if diff := pretty.Compare(resp.Tree, want); diff != "" {
+			t.Errorf("%v: post-ListTrees diff:\n%v", test.desc, diff)
```
RJPercival: The convention for an test failure message with a diff is to have something like "-want +got", to make it easy to identify what is extra and what is missing.

integration/admin/admin_integration_test.go
```diff
@@ -160,6 +154,73 @@ func TestAdminServer_GetTree(t *testing.T) {
...
+		{desc: "oneTree", numTrees: 1},
+		{desc: "threeTrees", numTrees: 3},
+	}
+
+	ctx := context.Background()
+	createdTrees := []*trillian.Tree{}
+	for _, test := range tests {
+		if l := len(createdTrees); l > test.numTrees {
+			t.Fatalf("%v: numTrees = %v, but we already have %v stored trees", test.desc, test.numTrees, l)
+		} else if l < test.numTrees {
```
codingllama: Done.

integration/admin/admin_integration_test.go
```diff
@@ -160,6 +154,73 @@ func TestAdminServer_GetTree(t *testing.T) {
...
+			for i := l; i < test.numTrees; i++ {
+				var tree *trillian.Tree
+				if i%2 == 0 {
+					tree = testonly.LogTree
+				} else {
+					tree = testonly.MapTree
+				}
+				resp, err := client.CreateTree(ctx, &trillian.CreateTreeRequest{Tree: tree})
+				if err != nil {
+					t.Fatalf("%v: CreateTree() = (_, %q), want = (_, nil)", test.desc, err)
```
codingllama: Done.

server/admin/admin_server_test.go
```diff
@@ -111,7 +112,86 @@ func TestAdminServer_BeginError(t *testing.T) {
...
+		{desc: "threeTrees", numTrees: 3},
+	}
+
+	ctx := context.Background()
+	nextTreeID := int64(17)
+	storedTrees := []*trillian.Tree{}
+	for _, test := range tests {
+		if l := len(storedTrees); l > test.numTrees {
+			t.Fatalf("%v: numTrees = %v, but we already have %v stored trees", test.desc, test.numTrees, l)
+		} else if l < test.numTrees {
```
codingllama: Done.

server/admin/admin_server_test.go
codingllama: I find that to be a bit redundant as all diffs work like that. I don't think we've been doing this consistently either.

server/admin/admin_server_test.go
codingllama: Done.

integration/admin/admin_integration_test.go
codingllama: Actually, I've kept the else so the scope of *l* can remain as the if block.

server/admin/admin_server_test.go
codingllama: As above: kept the else so *l* is scoped to the if.

codingllama:
Issue #310

codingllama:
Thanks, Rob. PTAL?

server/admin/admin_server_test.go
RJPercival: [You've done it elsewhere in this file](https://github.com/google/trillian/blob/81e5d9cb63be1cd56d8026988be709058a820900/server/admin/admin_server_test.go#L164), so it'd at least be consistent with this file, if not the whole codebase.

codingllama: Fair enough.

codingllama:
PTAL?

Issue #495, opened 2017-03-28
Create scripts / configs to deploy a clustered CT log


Martin2112:
Current plan is to use Kubernetes. Assigning to M8 but would be better if it could be done earlier.

Martin2112:
I think this may be done now?

PR #507, opened 2017-04-03
Add a Kubernetes config and Docker image for running a Galera MySQL cluster

RJPercival:
This can be used to easily start up a new Galera MySQL cluster running on a Google Container Cluster.

The MySQL servers will be backed by 500Gi SSDs.
They will be running Percona Server 5.7 and use the XtraDB storage engine.
The root user will have the password "zaphod".

Once the cluster is running, Trillian's `scripts/resetdb.sh` script can be used to create a suitable test database and user.


Martin2112:
You've started on this early - well done!

For the moment could we do something in the package name maybe to mark it as experimental. I've held off putting my configs and stuff in until they were more clear. And at some point we'll want to write configs to run a complete log cluster so we might reorganize things.

Martin2112:
Towards #495.

RJPercival:
I've added an "EXPERIMENTAL" header at the top of README.md and any images generated by the `push.sh` script will be labelled "experimental" by default. Is that sufficient?

Martin2112:
Yes I think that will do for now but I might want to add some experimental scripts to help people run things so might add an experimental dir anyway.

PR #513, opened 2017-04-04
Backoff and retry if a QueueLeaves request fails during integration testing

RJPercival:
This could occur due to transient issues, e.g. database server is momentarily unavailable.

TODO: Add retry logic to other requests.

### DIFF
--- a/client/backoff/backoff.go
+++ b/client/backoff/backoff.go
@@ -15,6 +15,7 @@
 package backoff
 
 import (
+	"context"
 	"math"
 	"math/rand"
 	"time"
@@ -51,4 +52,19 @@ func (b *Backoff) Duration() time.Duration {
 // Reset sets the internal iteration count back to 0.
 func (b *Backoff) Reset() {
 	b.x = 0
+}
+
+// Retry calls a function until it succeeds or the context is done.
+// It will backoff if the function returns an error.
+// Once the context is done, retries will end and the most recent error will be returned.
+// Backoff is not reset by this function.
+func (b *Backoff) Retry(ctx context.Context, f func() error) error {
+	var err error
+	for ctx.Err() == nil {
+		if err = f(); err == nil {
+			break
+		}
+		time.Sleep(b.Duration())
+	}
+	return err
 }
--- a/client/backoff/backoff.go
+++ b/client/backoff/backoff.go
@@ -59,12 +59,23 @@ func (b *Backoff) Reset() {
 // Once the context is done, retries will end and the most recent error will be returned.
 // Backoff is not reset by this function.
 func (b *Backoff) Retry(ctx context.Context, f func() error) error {
-	var err error
-	for ctx.Err() == nil {
-		if err = f(); err == nil {
-			break
-		}
-		time.Sleep(b.Duration())
+	// If the context is already done, don't make any attempts to call f.
+	if ctx.Err() != nil {
+		return ctx.Err()
 	}
-	return err
+
+	// Try calling f until it doesn't return an error or ctx is done.
+	for {
+		err := f()
+		if err == nil {
+			return err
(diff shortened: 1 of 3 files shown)


integration/log.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
```
Martin2112: We've got a client/backoff package. Not sure the server should depend on client code though.

RJPercival: I noticed that package, but this one seems significantly better. Is there a particularly good reason that we implemented backoff ourselves?

codingllama: Drive-by comment.

@gdbelvin initially used a 3rd party backoff package, but I asked for it to be removed due to licensing concerns. I don't think it was the same package.

`client/backoff` is general purpose, IIRC. It's just a matter of moving it somewhere else.

RJPercival: This package has an MIT license, so that's not an issue. I could use `client/backoff`, and I don't think its location is problematic since the integration test is acting as a client. I'd have to write more code though, since our backoff package is extremely minimal compared to this one. This lets me do things like "retry this function with exponential backoff until the context deadline is reached", which is rather nice. Really, I'd prefer to delete our package in favour of this one.

gdbelvin:
Here's another backoff library that might help:
https://github.com/GoogleCloudPlatform/google-cloud-go/blob/e62b316e22eb3b7c2ffe26d3fc428ea01810f709/spanner/backoff.go

RJPercival:
That's almost the same as ours. Both lack the nice context integration that this library provides.

Martin2112:
I can see several options but they all seem to have undesirable dependencies. The integration test is a client as Rob mentioned so we could use client / backoff unless that's too much work?

RJPercival:
I've modified this to use the `trillian/client/backoff` package.

client/backoff/backoff.go
```diff
@@ -52,3 +53,18 @@ func (b *Backoff) Duration() time.Duration {
...
+// It will backoff if the function returns an error.
+// Once the context is done, retries will end and the most recent error will be returned.
+// Backoff is not reset by this function.
+func (b *Backoff) Retry(ctx context.Context, f func() error) error {
+	var err error
+	for ctx.Err() == nil {
+		if err = f(); err == nil {
+			break
+		}
+		time.Sleep(b.Duration())
```
AlCutter: Just a thought - duration could grow to be quite large, and if someone cancels the `ctx` they could end up still waiting quite a while for this to give up and return.
Could do something along the lines of:
```go
select {
case _ = <-time.After(Duration()):
	continue
case _ = <-ctx.Done():
	return lastErr
}
```
Quick demo: https://play.golang.org/p/kw3UMmr7YE

RJPercival: Done.

client/backoff/backoff.go
```diff
@@ -52,3 +53,29 @@ func (b *Backoff) Duration() time.Duration {
...
+func (b *Backoff) Retry(ctx context.Context, f func() error) error {
+	// If the context is already done, don't make any attempts to call f.
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	// Try calling f until it doesn't return an error or ctx is done.
+	for {
+		err := f()
+		if err == nil {
```
AlCutter: ooo, sorry should've thought of this before, how about:
```go
for {
  err := f()
  if err != nil {
    select {
      …
    }
  }
  return nil
}
```
to stick with the indent-error-flow idiom?

(https://play.golang.org/p/r6Zc-WHqFS)

RJPercival: Done.

PR #517, opened 2017-04-07
Fix DataStructures property table in Readme

RaghavaDhanya:
DataStructures property table in readme was broken due to changes in github md syntax.


RaghavaDhanya:
I signed it!

AlCutter:
Many thanks for spotting fixing that up!

PR #524, opened 2017-04-11
Code cleanups: merkle.Factory, crypto.NewSHA256Signer, util/context.go

codingllama:


codingllama:
A few remarks:

* `merkle.Factory` and `util/context.go` are just unnecessary.
* `crypto.NewSHA256Signer` is actually useful because of the package name conflict between `"crypto"` and `"trillian/crypto"`. To create a Trillian signer you need both, so `crypto.NewSHA256Signer` saves an aliased import.
* I've changed a few format strings to use `%v`, as it's more flexible in face of refactors (such as the ones here).

PR #532, opened 2017-04-14
Don't require a PEM password in cmd/createtree

rolandshoemaker:
PEM file may not be password protected so don't require one.

Also fixes the `cmd/createtree` tests, previously they would `return` from the loop at the first `test.wantErr` or `hasErr` instead of using `continue` which prevented the rest of the tests from running.

### DIFF
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -146,13 +146,9 @@ func newPK(opts *createOpts) (*any.Any, error) {
 		if _, err := os.Stat(path); err != nil {
 			return nil, fmt.Errorf("error reading PEM key file at %v: %v", path, err)
 		}
-		pass := opts.pemKeyPass
-		if pass == "" {
-			return nil, errors.New("empty PEM key password")
-		}
 		pemKey := &trillian.PEMKeyFile{
 			Path:     path,
-			Password: pass,
+			Password: opts.pemKeyPass,
 		}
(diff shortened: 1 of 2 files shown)


Martin2112:
I don't think we want to encourage the use of private keys without passwords, which is why all our test keys have them. If it was confined to test code code it wouldn't matter. I should probably give the people working on this part of the code a chance to comment though.

If you feel like splitting out the test fix we can probably merge that now.

rolandshoemaker:
> I don't think we want to encourage the use of private keys without passwords, which is why all our test keys have them. If it was confined to test code code it wouldn't matter. I should probably give the people working on this part of the code a chance to comment though.

This makes sense, I think the main problem though is that it isn't enforced consistently. You can run the CT frontend without using a password for your key file, the only restriction exists in `cmd/createtree`.

> If you feel like splitting out the test fix we can probably merge that now.

Sounds good, next time I'm in front of a real computer I'll split this out.

Martin2112:
Thanks. It's possibly become inconsistent as a result of recent changes in key management and admin API. We should fix that if so.

codingllama:
@Martin2112 has said it all. Happy to take the test fix, but the consistency change should be to have the CT frontend not accept non-encrypted PEMs.

rolandshoemaker:
Sounds good.

PR #533, opened 2017-04-18
Split the mysql storage related code from sql statements

Martin2112:
And introduce a DBWrapper that encapsulates the statement cache and creating Stmts for the common code. Hide MySQL specifics in the wrapper. Repackage everything into sql/coresql and sql/mysql and rename entites that are no longer tightly couple to mysql. Clean up some tests that relied on forcing revision numbers to avoid the need to expose hacks for testing now they cross package boundaries.

This is groundwork for adding a postgres storage implementation.


storage/sql/coresql/wrapper/wrapper.go
```diff
@@ -0,0 +1,108 @@
...
+// then impossible to avoid import cycles.
+package wrapper
+
+import (
+	"context"
+	"database/sql"
+)
+
+// TreeStatementProvider provides SQL statement objects for raw tree storage.
+type TreeStatementProvider interface {
```
AlCutter: Currently there's a somewhat undocumented contract between the wrapper and coresql packages in that some of the `Stmt` types returned here require positional parameters in a particular ordering, I'm wondering if we can refactor that a bit so that wrapperDBs execute the statements themselves after being passed in the params, but in the meantime adding some coments on these interface methods detailing what the expected params & ordering are for the returned `Stmt` would help make the contract clearer.
WDYT?

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// EnsureAllLeavesDistinct checks that the supplied leaves do not contain any duplicate
+// LeafIdentityHash Values.	All the leaf hashes should be distinct because the leaves were
+// created with distinct leaf data. If only we had maps with slices as keys or sets or pretty
```
Martin2112: Ah, this is obviously some strange usage of the word 'fine' that I wasn't previously aware of.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/storage"
+)
+
+// CheckLeafContents compares two log leaves to see if they match. It fails a test with
+// a descriptive message if not.
+func CheckLeafContents(leaf *trillian.LogLeaf, seq int64, rawHash, hash, data, extraData []byte, t *testing.T) {
```
Martin2112: Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+		if bytes.Equal(bl.LeafIdentityHash, leaf.LeafIdentityHash) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// EnsureAllLeavesDistinct checks that the supplied leaves do not contain any duplicate
+// LeafIdentityHash Values.	All the leaf hashes should be distinct because the leaves were
```
Martin2112: Looked OK in editor. Possibly a tab char? If so then it's fixed.

storage/sql/coresql/testonly/checks.go
Martin2112: Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+// created with distinct leaf data. If only we had maps with slices as keys or sets or pretty
+// much any kind of usable data structures we could do this properly.
+func EnsureAllLeavesDistinct(leaves []*trillian.LogLeaf, t *testing.T) {
```
Martin2112: Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+func NodesAreEqual(lhs []storage.Node, rhs []storage.Node) error {
+	if ls, rs := len(lhs), len(rhs); ls != rs {
+		return fmt.Errorf("different number of nodes, %d vs %d", ls, rs)
+	}
+	for i := range lhs {
+		if l, r := lhs[i].NodeID.String(), rhs[i].NodeID.String(); l != r {
+			return fmt.Errorf("NodeIDs are not the same,\nlhs = %v,\nrhs = %v", l, r)
+		}
+		if l, r := lhs[i].Hash, rhs[i].Hash; !bytes.Equal(l, r) {
+			return fmt.Errorf("Hashes are not the same for %s,\nlhs = %v,\nrhs = %v", lhs[i].NodeID.CoordString(), l, r)
```
Martin2112: Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// NodesAreEqual tests if two slices of storage nodes contain the same set of nodes.
+func NodesAreEqual(lhs []storage.Node, rhs []storage.Node) error {
+	if ls, rs := len(lhs), len(rhs); ls != rs {
+		return fmt.Errorf("different number of nodes, %d vs %d", ls, rs)
+	}
+	for i := range lhs {
+		if l, r := lhs[i].NodeID.String(), rhs[i].NodeID.String(); l != r {
+			return fmt.Errorf("NodeIDs are not the same,\nlhs = %v,\nrhs = %v", l, r)
```
Martin2112: Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		if l, r := lhs[i].Hash, rhs[i].Hash; !bytes.Equal(l, r) {
+			return fmt.Errorf("Hashes are not the same for %s,\nlhs = %v,\nrhs = %v", lhs[i].NodeID.CoordString(), l, r)
+		}
+	}
+	return nil
+}
+
+// DiffNodes compares two slices of storage nodes and returns slices containing extra and
```
Martin2112: I was wondering who'd be the first to spot that. Done.

storage/sql/coresql/testonly/checks.go
```diff
@@ -0,0 +1,114 @@
...
+// missing nodes (if any). If the slices contain the same data both returned slices will
+// be empty.
+func DiffNodes(got, want []storage.Node) ([]storage.Node, []storage.Node) {
```
Martin2112: Yes they are expected to be distinct. I've made it fail if either slice contains duplicates.

storage/sql/coresql/testonly/trees.go
```diff
@@ -0,0 +1,81 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testonly
+
+import (
+	"fmt"
+
+	"context"
```
Martin2112: Done.

storage/sql/coresql/wrapper/wrapper.go
Martin2112: Yes this is true. I didn't like the idea of the provider doing work on the database, that was where I drew up the split. This isn't meant to be fully generic, if we hit a case where we have to have a different number of params for some operation it will break but this can be addressed then.

I'll comment them for now but the ordering isn't fixed because the variable params need to be shuffled around to cope with positional parameters so not sure if the doc could be confusing at the interface level.

Martin2112: Just done an example of a half-way sort of option where the wrapper does the args marshalling as well. Bit more flexible and makes the API dependency clearer but not sure I like it. I'm going to try moving the execution into the wrapper and see what that looks like.

Martin2112: OK this is how it would look if the wrapper ran the query. Error handling is a bit messy as defer stmt.Close() in the wrapper would break as it closes the result set before it's read. This means the statement has to be returned from the wrapper. I could pass in a func that processes the Rows but this is starting to look a bit weird.

storage/sql/coresql/tree_storage.go
```diff
@@ -86,18 +86,13 @@ func (t *treeTX) getSubtrees(treeRevision int64, nodeIDs []storage.NodeID) ([]*s
...
 	}
 
-	stmt, args, err := t.ts.wrap.GetSubtreeStmt(t.tx, t.treeID, treeRevision, nodeIDs)
-	if err != nil {
-		return nil, err
-	}
-	defer stmt.Close()
-
-	rows, err := stmt.Query(args...)
+	stmt, rows, err := t.ts.wrap.GetSubtrees(t.tx, t.treeID, treeRevision, nodeIDs)
```
AlCutter: An option here might be to pass in a `func(...)` here which gets called once for each row, then you could `Close()` the stmt and rows inside the call itself.

storage/sql/coresql/testonly/checks.go
```diff
@@ -17,34 +17,35 @@ package testonly
...
+// otherwise a descriptive error.
+func CheckLeafContents(leaf *trillian.LogLeaf, seq int64, rawHash, hash, data, extraData []byte) error {
 	if got, want := leaf.MerkleLeafHash, hash; !bytes.Equal(got, want) {
-		t.Fatalf("Wrong leaf hash in returned leaf got\n%v\nwant:\n%v", got, want)
+		return fmt.Errorf("wrong leaf hash in returned leaf got\n%v\nwant:\n%v", got, want)
 	}
 
 	if got, want := leaf.LeafIdentityHash, rawHash; !bytes.Equal(got, want) {
-		t.Fatalf("Wrong raw leaf hash in returned leaf got\n%v\nwant:\n%v", got, want)
+		return fmt.Errorf("Wrong raw leaf hash in returned leaf got\n%v\nwant:\n%v", got, want)
```
Martin2112: Go frees you from tedious debates about how code should be formatted, so you can spend the time on tedious debates about how error messages should be formatted. I'll have another look at them.

storage/sql/coresql/tree_storage.go
Martin2112: I've now tried out about 5 options. None of them are a perfect separation of responsibilities. What do you think?

storage/sql/coresql/tree_storage.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// Otherwise we have to look at the result of the operation
+	rowsAffected, rowsError := res.RowsAffected()
+
+	if rowsError != nil {
+		return rowsError
+	}
+
+	if rowsAffected != count {
+		return fmt.Errorf("Expected %d row(s) to be affected but saw: %d", count,
```
AlCutter: /Expected/expected/

Martin2112: Done.

Martin2112:
Yes there doesn't seem to be a perfect option. This has almost no impact on external APIs and I think I can refactor again to get rid of the one thing that does - the dedup related fetch by id hash.

There is quite a lot of work first to make the API match this style throughout and then do it again for the postgres version in the other branch.

storage/sql/coresql/testonly/checks.go
Martin2112: All fixed.

storage/sql/coresql/wrapper/wrapper.go
```diff
@@ -0,0 +1,110 @@
...
+// database and transaction flow.
+type LifecycleHooks interface {
+	OnOpenDB() error
+}
+
+// DBWrapper encapsulates a database and provides customized SQL statement objects for all types
+// of tree storage as well as abstracting some operations that differ between databases.
+// Statements returned by any of public functions belong to the caller transaction and must be
+// closed on completion of the work.
+type DBWrapper interface {
```
codingllama: For discussion:

A DBWrapper is a mix of "statement provider" (ie, a thing-that-has-queries) and a DAO (particularly for trees). Mixing different concepts usually makes things harder in the long run, as we're experiencing now.

What we're trying to do, as I understand, is allow support for distinct databases. The usual way to do that is via DAOs (or whatever other fancy name people use nowadays). The problem *we* have is that our storage layer mixes several different concepts and responsibilities into the storage and transaction objects.

A few of the issues are:
1. We represent collection of entities as "storages", instead of having per-entity DAOs
1. We confuse transactions with DAOs
1. We have too much tree-manipulation logic on the storage layer

The combination of those mean our "Storage" objects do too much work, do work in unexpected places (like TX.Begin() methods) and make it borderline impossible to have cross-"storage" transactions. Re-writing the storage layer to support different databases is cumbersome, complex, and error prone. We also don't provide interface-based test suites, so each implementor is on its own (and we can't assess quality uniformly between implementations).

A DAO-like approach, particularly with "dumb" DAOs, fixes all of those and allows for database-specific forks. We could even provide SQL-but-non-database-specific DAOs, into which one just plugs slightly different query strings, but with a dumb-DAO layer and a few utility methods that shouldn't be necessary (nor is much desirable, IMO). Transactions should be moved to method parameters, instead of the object used to manipulate entities.

I think we should strive to understand where we are, where we'd like to be (ideally), and what venues we can take to get as close as possible to that ideal. Since we felt the need to rework this layer, we could take the time to put together a proposal that moves us to a permanently better place.

Martin2112: Yes I'm well aware of all that. Let's try to address some of these points.

This PR didn't start out as a mix of statement provider and DAO and won't be once I finish the requested rework so that it all looks like the tree interface.

The design goals were for performance and abstraction at the storage level, which I think have been achieved. "dumb DAOs" may be OK for leaf entities but did not perform well enough for subtrees in early prototyping so we were pushed towards a more complex solution.

The mixture of layers is a fair point but I don't agree that supporting multiple databases using this approach is complex. It needs a different wrapper but it shares most of the code that drives it.  I may be able to fully encapsulate the database access within the wrapper, which will help with testing the common code. But you can't test implementations that exercise specific databases with mocks with any real confidence.

If you look at the early history of this PR I did start off with something that just plugged queries into entirely code but @AlCutter asked me to move away from that. Several other options were unworkable or had undesirable side effects.

We have a fairly immediate requirement to support Postgres, which is what I'm trying to address at the moment without making a copy of all the code. We can always revisit the overall design again later.

storage/sql/coresql/db/storage.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/storage/sql/coresql/wrapper"
+	"github.com/google/trillian/storage/sql/mysql"
+)
+
+// OpenDB opens, and creates a wrapper for, the SQL databases that we know about. The wrapper
+// encapsulates things that differ between database implementations and abstracts away
+// from the SQL queries.
+func OpenDB(driver, dbURL string) (wrapper.DBWrapper, error) {
```
Martin2112: I'll have another look at it but I previously couldn't avoid import cycles unless it was broken out.

storage/log_storage.go
```diff
@@ -119,6 +119,10 @@ type LeafReader interface {
 	// same hash but different sequence numbers. If orderBySequence is true then the returned data
 	// will be in ascending sequence number order.
 	GetLeavesByHash(leafHashes [][]byte, orderBySequence bool) ([]*trillian.LogLeaf, error)
+	// GetLeavesByLeafIdentityHash returns leaves by their identity hash. Note: This is for
+	// internal use only and is not part of the public API. The returned leaf objects may not
+	// be fully populated
```
Martin2112: I haven't confirmed but I think I can probably move this to the wrapper. It's only for use in handling duplicate submissions. It can't fully populate the leaf struct because they might not have been sequenced yet.

Martin2112:
Closing this until we need it.

Issue #477, opened 2017-03-27
Allow servers to bind at user-specified addresses, not just ports


codingllama:

Martin2112:
Does this need more than #452?

Martin2112:
Ah that patch was just for CT so it probably does.

codingllama:
Yep, #452 was just the example server. Not that I expect this to be any different, we just have to fix a few integration tests as we go.

PR #538, opened 2017-04-21
Allow servers to bind at user-specified addresses

codingllama:
Issue #477


integration/ct_prep_test.sh
```diff
@@ -36,7 +36,7 @@ for ((i=0; i < RPC_SERVER_COUNT; i++)); do
   RPC_SERVERS="${RPC_SERVERS},localhost:${port}"
 
   echo "Starting Log RPC server on port ${port}"
```
codingllama: Discussed in person.

I'll make sure the messages match the flag format (ie, we should log *host:port*, not just *port*). We didn't reach a clear decision on the endpoint vs uri debate, so I'll go ahead with endpoint for now. I do prefer it slightly over uri, but I'm happy to change if people think otherwise.

Martin2112:
LGTM.

PR #539, opened 2017-04-21
Fix a case where docs + code have diverged

Martin2112:
And try to explain the hash recomputation stuff a bit more.

### DIFF
--- a/docs/storage/storage.md
+++ b/docs/storage/storage.md
@@ -199,18 +199,32 @@ leaf index followed by formatting and marshaling the data to be returned to the 
 API requests for proofs involve more work but both inclusion and consistency proofs follow the 
 same pattern.
 
-The node revision number for the tree size involved is obtained from storage and will be used in
-the subsequent node fetch.
+Path calculations in the tree need to be based on a particular tree revision. It is possible to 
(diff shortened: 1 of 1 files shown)


docs/storage/storage.md
```diff
@@ -199,18 +199,32 @@ leaf index followed by formatting and marshaling the data to be returned to the
...
+be fetched from storage and a set of flags indicating required hash recomputations. After a 
+successful read the hashes are extracted from the nodes, rehashed if necessary and returned to
+the client.
+
+Recomputation is needed because we don't necessarily store snapshots on disk for every tree size. 
+To serve proofs at a version intermediate between two stored versions it can be necessary to
+recompute hashes on the rightmost path. This requires extra nodes to be fetched but is bounded
+by the depth of the tree so this never becomes unmanageable.
 
 Nodes may have been overwritten in the second snapshot when a right hand sibling was added
```
codingllama: This sentence seems a bit off. Maybe start directly on "consider the state of the tree as ...", as you've already explained a good deal of it above.

Martin2112: Done. Also removed "on the right" later because if you have a narrow window it ends up below.

Issue #508, opened 2017-04-03
Document intended usage of the errors package


RJPercival:
The `errors` package could use some extra documentation to indicate how it should be used. In particular, it should state that the error codes should describe the problem from the perspective of an RPC client, rather than from the perspective of the function itself.

For example, in [this code review](https://github.com/google/trillian/pull/480#discussion_r108390398), a function (`NewFromPrivatePEMFile`) could not find the specified file and so returned an error with code `errors.NotFound`. However, review feedback indicated that it should return `errors.FailedPrecondition` instead, because the caller of the RPC that led to that function call (`CreateTree`) would consider it a failed precondition.

codingllama:
Fair enough. I'll have a go at it soon-ish. :)

PR #540, opened 2017-04-21
Document user-visible characteristic of TrillianErrors

codingllama:
Issue #508


codingllama:
If you can't pick the correct error at a certain function, then it's likely you're trying to pick errors at the wrong layer. In any case, the important part is that there's a mechanism to do so (and we have somewhat better documentation now). How we apply it may evolve as we go along.

PR #542, opened 2017-04-24
Fix cmd/createtree tests

rolandshoemaker:
Split off from #532.

Uses `t.Fatalf` instead of `t.Errorf` in the first case to maintain previous fail-fast behavior on test failures, otherwise just uses `continue` instead of `return` so all tests are actually ran.

### DIFF
--- a/cmd/createtree/main_test.go
+++ b/cmd/createtree/main_test.go
@@ -157,10 +157,9 @@ func TestRun(t *testing.T) {
 		tree, err := createTree(ctx, test.opts)
 		switch hasErr := err != nil; {
 		case hasErr != test.wantErr:
-			t.Errorf("%v: createTree() returned err = '%v', wantErr = %v", test.desc, err, test.wantErr)
-			return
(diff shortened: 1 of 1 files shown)


cmd/createtree/main_test.go
```diff
@@ -157,10 +157,9 @@ func TestRun(t *testing.T) {
 		tree, err := createTree(ctx, test.opts)
 		switch hasErr := err != nil; {
 		case hasErr != test.wantErr:
-			t.Errorf("%v: createTree() returned err = '%v', wantErr = %v", test.desc, err, test.wantErr)
-			return
+			t.Fatalf("%v: createTree() returned err = '%v', wantErr = %v", test.desc, err, test.wantErr)
```
codingllama: Could we do `t.Errorf(...) / continue`? I think that was the original intent.

I probably had the test wrapped on a `func(){ ... }()` due to defers, but forgot to change the `returns` to `continues` after removing it.

PR #544, opened 2017-04-24
Require a PEM priv key password in CT frontend

### LINKED ISSUE #543: Standardize PEM password requirements
Per #532 `cmd/createtree` requires a password for the PEM priv key file or will fail out but `examples/ct/ct_server` does not.

Either require a password for both or neither.

rolandshoemaker:
Fixes #543.

### DIFF
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -139,20 +139,15 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
 func newPK(opts *createOpts) (*any.Any, error) {
 	switch opts.privateKeyType {
 	case "PEMKeyFile":
-		path := opts.pemKeyPath
-		if path == "" {
+		if opts.pemKeyPath == "" {
 			return nil, errors.New("empty PEM path")
 		}
-		if _, err := os.Stat(path); err != nil {
-			return nil, fmt.Errorf("error reading PEM key file at %v: %v", path, err)
-		}
-		pass := opts.pemKeyPass
-		if pass == "" {
-			return nil, errors.New("empty PEM key password")
+		if opts.pemKeyPass == "" {
+			return nil, fmt.Errorf("empty password for PEM key file %q", opts.pemKeyPath)
 		}
 		pemKey := &trillian.PEMKeyFile{
-			Path:     path,
-			Password: pass,
+			Path:     opts.pemKeyPath,
+			Password: opts.pemKeyPass,
 		}
 		return ptypes.MarshalAny(pemKey)
 	default:
--- a/crypto/keys/private.go
+++ b/crypto/keys/private.go
@@ -38,8 +38,11 @@ type SignerFactory interface {
 }
 
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
-// The key may be protected by a password.
+// The key must be protected by a password.
 func NewFromPrivatePEMFile(keyFile, keyPassword string) (crypto.Signer, error) {
+	if keyPassword == "" {
+		return nil, fmt.Errorf("empty password for PEM key file %q", keyFile)
+	}
 	pemData, err := ioutil.ReadFile(keyFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read private key from file %q: %v", keyFile, err)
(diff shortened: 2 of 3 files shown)


rolandshoemaker:
I guess it might also make sense to just to do this from `crypto/keys.NewFromPrivatePEMFile` so it doesn't accidentally drift out of sync again.

codingllama:
Requiring a password at `keys.NewFromPrivatePEMFile` sounds good to me. It doesn't seem to break any tests either, so should be an easy change.

@RJPercival @Martin2112, any thoughts? For those without previous context in this discussion, the idea is to disallow unencrypted PEMs to be used with Trillian, with the intent to foster good practices.

Martin2112:
If we do it this way we'll have to enforce it in each app. So I suspect it's better to do it in NewFromPrivatePEMFile. Rob?

RJPercival:
Agreed, doing this in NewFromPrivatePEMFile seems best. All of our test keys are encrypted so that shouldn't cause any issues.

codingllama:
Thanks, folks. Roland, you mind doing the fix here, instead of at *examples/ct/instance.go*? I'm happy to review it.

rolandshoemaker:
Yup, will do.

crypto/keys/private.go
```diff
@@ -38,8 +38,11 @@ type SignerFactory interface {
 }
 
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
-// The key may be protected by a password.
+// The key must be protected by a password.
 func NewFromPrivatePEMFile(keyFile, keyPassword string) (crypto.Signer, error) {
+	if keyPassword == "" {
+		return nil, errors.New("empty PEM key password")
```
RJPercival: Could you include the `keyFile` value in the error message please, to make it obvious which file needs a password?

cmd/createtree/main.go
```diff
@@ -147,9 +147,6 @@ func newPK(opts *createOpts) (*any.Any, error) {
 			return nil, fmt.Errorf("error reading PEM key file at %v: %v", path, err)
 		}
```
RJPercival: The code above that checks for existence of the PEM file needs to be deleted as well. This only works if the createtree command is run on the same machine as the Trillian server, since the file path is resolved by the server, not the client.

cmd/createtree/main.go
```diff
@@ -139,20 +139,12 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
...
 	case "PEMKeyFile":
-		path := opts.pemKeyPath
-		if path == "" {
+		if opts.pemKeyPath == "" {
 			return nil, errors.New("empty PEM path")
 		}
-		if _, err := os.Stat(path); err != nil {
-			return nil, fmt.Errorf("error reading PEM key file at %v: %v", path, err)
-		}
-		pass := opts.pemKeyPass
```
codingllama: nit: Since we've kept the `path != ""` validation, would you mind keeping the `pass != ""` check too, for consistency?

Both checks seem reasonable to keep, as they inform users of missing flags.

rolandshoemaker: Seems reasonable.

rolandshoemaker:
Interesting: so it turns out that after #542 merged a few tests broke, mainly there is one that `os.Stat`'d a invalid path. Looking at `Server.CreateTree` doesn't actually need the private key in order to create a tree (since it doesn't actually sign the head).

If I completely don't initialized the `trillian.Tree` `PrivateKey` field everything seems to work as expected and tests pass. Any reason to not just carve all that extra code for reading the PEM file/pass out from `cmd/createtree`?

codingllama:
Hey Roland, sorry for the slow turnover.

I thought we just did away with the `os.Stat` check (which I completely agree is for the best). Feel free to remove the `invalidPEMPath` test case from `cmd/createtree/main_test.go`. I'd like to keep the not empty checks for path/pass though, as they relate directly to mandatory flags that haven't been provided.

Just to clarify, as I may have misunderstood the comment, I fully expect the *real* CreateTree RPC to fail if a PEM is not provided; even though we don't do signing (yet), the server does attempt to create a Signer instead, for which the key needs to be read.

rolandshoemaker:
Whoops, yup, somehow I missed the call to `trees.Signer` in `Server.CreateTree`, my bad!

rolandshoemaker:
Done.

PR #556, opened 2017-04-26
Make ct_hammer submit multiple certs in parallel

AlCutter:


integration/hammer.go
```diff
@@ -203,6 +221,31 @@ func newHammerState(cfg *HammerConfig) (*hammerState, error) {
...
 }
 
+// addMultiple calls the passed in function a random number
+// (1 <= n < MaxParallelAdds) of times.
+// The first of any errors returned by calls to addOne will be returned by this function.
+func (s *hammerState) addMultiple(ctx context.Context, addOne func(context.Context) error) error {
+	var wg sync.WaitGroup
+	numAdds := rand.Intn(s.cfg.MaxParallelAdds) + 1
+	errs := make(chan error, numAdds)
+	for i := 0; i < rand.Intn(s.cfg.MaxParallelAdds)+1; i++ {
```
codingllama: I think you meant `for i := 0; i < numAdds; i++`.

integration/hammer.go
```diff
@@ -110,23 +114,34 @@ type submittedCert struct {
 // most recent, but new entries are only appended when enough time has
 // passed since the last append, so the SCTs that get checked are spread
 // out across the MMD period.
-type pendingCerts [sctCount]*submittedCert
+type pendingCerts struct {
+	mu    sync.Mutex
+	certs [sctCount]*submittedCert
+}
+
+func (pc *pendingCerts) tryAppendCert(now time.Time, mmd time.Duration, submitted *submittedCert) {
```
codingllama: We should probably document the locking behavior. Same for the other funcs.

integration/ct_hammer/main.go
```diff
@@ -44,6 +44,7 @@ var (
 var (
 	addChainBias          = flag.Int("add_chain", 20, "Bias for add-chain operations")
 	addPreChainBias       = flag.Int("add_pre_chain", 20, "Bias for add-pre-chain operations")
+	maxChainsToAdd        = flag.Int("max_chains_to_add", 100, "Maximum number of chains to add in parallel (will always add at least 1 chain)")
```
AlCutter: Done.

integration/hammer.go
AlCutter: Done.

integration/hammer.go
```diff
@@ -135,30 +150,33 @@ func (pc *pendingCerts) appendCert(submitted *submittedCert) {
 	// Require caller to have checked canAppend().
 	which := 0
 	for ; which < sctCount; which++ {
-		if pc[which] == nil {
+		if pc.certs[which] == nil {
```
AlCutter: I was trying to be the least invasive possible, but you're right this is an unexpected sharp edge. Done.

integration/hammer.go
AlCutter: I did indeed, done!

PR #557, opened 2017-04-26
Allow for storage-specific settings to be supplied via Admin API

codingllama:
Allow storage specific settings to be supplied via an `Any` proto, similarly to `PrivateKey`.

In the general storage layers any valid proto is allowed to go through. The MySQL implementation, for now, disallows any settings to exist. We may change it later so that `TreeControl` table settings are exposed via the APIs.

Different storage implementations are meant to define and validate their protos on their own, possibly outside of the Trillian codebase.


codingllama:
Issue #310

storage/mysql/admin_storage.go
```diff
@@ -395,3 +401,10 @@ func (t *adminTX) UpdateTree(ctx context.Context, treeID int64, updateFunc func(
 func toMillisSinceEpoch(t time.Time) int64 {
 	return t.UnixNano() / 1000000
 }
+
+func validateStorageSettings(tree *trillian.Tree) error {
+	if tree.StorageSettings != nil {
+		return fmt.Errorf("storage_settings not supported, but found %v", tree.StorageSettings)
```
Martin2112: "but found" -> "got:" ?

codingllama: Done.

PR #563, opened 2017-04-27
UpdateTree RPC implementation

codingllama:


codingllama:
#310

codingllama:
I've snuck in a small change to the GetTree RPC impl as well, so that it takes advantage of `trees.GetTree`.

PR #566, opened 2017-04-27
Separate offline from online tasks in client

gdbelvin:
Personalitites will proxy responses from trillian logs back to their
callers. These callers will need a client that can verify those
responses even though they can't reach out and talk to the trillian log
itself.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -193,7 +193,7 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
 		}
 	}
 
-	// Verify Consistency proof.
+	// Verify root update.
 	if err := c.LogVerifier.UpdateRoot(resp, consistency); err != nil {
 		return err
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -36,15 +36,19 @@ import (
 type LogClient struct {
 	LogID  int64
 	client trillian.TrillianLogClient
-	*LogVerifier
+	*logVerifier
 }
 
 // New returns a new LogClient.
 func New(logID int64, client trillian.TrillianLogClient, hasher merkle.TreeHasher, pubKey crypto.PublicKey) *LogClient {
 	return &LogClient{
-		LogID:       logID,
-		client:      client,
-		LogVerifier: NewLogVerifier(hasher, pubKey),
+		LogID:  logID,
+		client: client,
+		logVerifier: &logVerifier{
+			hasher: hasher,
+			pubKey: pubKey,
+			v:      merkle.NewLogVerifier(hasher),
+		},
 	}
 }
 
@@ -56,7 +60,7 @@ func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
 		return err
 	}
 
-	leaf := c.LogVerifier.buildLeaf(data)
+	leaf := c.logVerifier.buildLeaf(data)
 	err := c.queueLeaf(ctx, leaf)
 	switch s, ok := status.FromError(err); {
 	case ok && s.Code() == codes.AlreadyExists:
@@ -194,7 +198,7 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
 	}
 
 	// Verify root update.
-	if err := c.LogVerifier.UpdateRoot(resp, consistency); err != nil {
+	if err := c.logVerifier.UpdateRoot(resp, consistency); err != nil {
 		return err
 	}
 	return nil
@@ -202,7 +206,7 @@ func (c *LogClient) UpdateRoot(ctx context.Context) error {
 
 // VerifyInclusion updates the log root and ensures that the given leaf data has been included in the log.
 func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
-	leaf := c.LogVerifier.buildLeaf(data)
+	leaf := c.logVerifier.buildLeaf(data)
 	if err := c.UpdateRoot(ctx); err != nil {
 		return fmt.Errorf("UpdateRoot(): %v", err)
 	}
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -177,56 +166,43 @@ func (c *LogClient) waitForRootUpdate(ctx context.Context) error {
...
-	hash := crypto.HashLogRoot(*str)
-	if err := crypto.Verify(c.pubKey, hash, str.Signature); err != nil {
-		return err
-	}
-
-	// Verify Consistency proof.
-	if str.TreeSize == c.root.TreeSize &&
-		bytes.Equal(str.RootHash, c.root.RootHash) {
+	if resp.SignedLogRoot.TreeSize == c.root.TreeSize &&
+		bytes.Equal(resp.SignedLogRoot.RootHash, c.root.RootHash) {
```
codingllama: As it's written, the intent behind the if is a bit hard to grasp (I'd expect hashes of equally-sized roots to match). IIUC, we could write it like this:

```go
if c.root.TreeSize > 0 && c.root.TreeSize == resp.SignedLogRoot.TreeSize {
    return nil
}`.
```

...which makes it clear that `TreeSize = 0` is the special case we care about.

client/offline.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Root returns the last valid root seen by UpdateRoot.
+// Returns an empty SignedLogRoot if UpdateRoot has not been called.
+func (c *LogVerifier) Root() trillian.SignedLogRoot {
+	return c.root
+}
+
+// UpdateRoot applies a GetLatestSignedLogRootResponse to Root(), if valid.
+func (c *LogVerifier) UpdateRoot(resp *trillian.GetLatestSignedLogRootResponse,
```
codingllama: Please document that consistency may be nil if the tree size is zero.

client/offline.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+	c.root = *str
+	return nil
+}
+
+// VerifyInclusionAtIndex verifies that the inclusion proof for data at index matches the currently trusted root.
+func (c *LogVerifier) VerifyInclusionAtIndex(data []byte, leafIndex int64, resp *trillian.GetInclusionProofResponse) error {
+	leaf := c.buildLeaf(data)
+	// XXX: c.root.TreeSize used to be req.TreeSize.
```
codingllama: What do you mean by `// XXX:`? Leftover comment, maybe?

client/offline.go
```diff
@@ -0,0 +1,111 @@
...
+	return &trillian.LogLeaf{
+		LeafValue:        data,
+		MerkleLeafHash:   c.hasher.HashLeaf(data),
+		LeafIdentityHash: hash[:],
+	}
+}
+
+// convertProof returns a slice of neighbor nodes from a trillian Proof.
+// TODO(martin): adjust the public API to do this in the server before returning a proof.
+func convertProof(proof *trillian.Proof) [][]byte {
```
codingllama: nit: Rename to `proofNodeHashes` (or something in that spirit?)

client/offline.go
```diff
@@ -0,0 +1,111 @@
...
+	hash := sha256.Sum256(data)
+	return &trillian.LogLeaf{
+		LeafValue:        data,
+		MerkleLeafHash:   c.hasher.HashLeaf(data),
+		LeafIdentityHash: hash[:],
+	}
+}
+
+// convertProof returns a slice of neighbor nodes from a trillian Proof.
+// TODO(martin): adjust the public API to do this in the server before returning a proof.
```
codingllama: nit: s/martin/Martin2112/

Out of curiosity, what's the API change we expect to happen here? I don't see any change in ordering, so do we expect Proof to have a `repeated bytes` ìnstead of a `repeated Node`?

client/offline.go
```diff
@@ -0,0 +1,111 @@
...
+	"crypto"
+	"crypto/sha256"
+
+	"github.com/google/trillian"
+	tcrypto "github.com/google/trillian/crypto"
+	"github.com/google/trillian/merkle"
+)
+
+// LogVerifier contains state needed to verify output from Trillian Logs.
+type LogVerifier struct {
```
codingllama: As it stands, I don't see much benefit in having LogVerifier. It's basically duplicating the merkle.LogVerifier API, with little added benefits (other than a few conversions).

If you do intent to keep it, please add tests for all public methods.

client/client.go
gdbelvin: This if statement is just a short circuit when nothing has changed. 
eg. Running UpdateRoot, which returns the same treesize and root hash. 

Hashes of equally sized roots should match, but if they don't, we want to be sure we return an error. "Trust but verify" :-)

client/offline.go
gdbelvin: done

client/offline.go
gdbelvin: Removed and added proper comment.

client/offline.go
gdbelvin: Yes, the request is for a `repeated bytes` now that `Node` doesn't have any other additional data.

client/offline.go
gdbelvin: done

gdbelvin:
Review comments addressed

client/client.go
codingllama: We don't return an error on hash mismatch here, though, do we? Could you add a test for this scenario too, please?

client/offline.go
codingllama: Comment not addressed?

client/client.go
gdbelvin: If the hashes mismatch, we skip the short cut and the consistency proof will fail.

gdbelvin: added test.

codingllama: Thanks.

client/offline.go
gdbelvin: missed this comment.  
The LogVerify interface does have some added value:  tracking client state for consistency proofs, and (later) gossiping.

PR #568, opened 2017-04-28
Use context-aware database/sql methods

codingllama:
Replace Begin/Exec/Prepare/Query/QueryRow/Stmt with their context-aware counterparts.

LogStorage and MapStorage now have a ctx method to all their methods. Places that still need a proper ctx to be propagated are using context.TODO() for easy future discovery.


codingllama:
I've touched most places that needed a ctx, except for TreeStorage methods (which we may address in future PRs). I've kept separate commits, so I can move them around to different PRs if necessary. Best not review commit-by-commit, though, unless you're interested in breaking the PR, as there's churn between the 1st and other commits.

codingllama:
#419

PR #570, opened 2017-04-28
Add public_key field to trillian.Tree

RJPercival:
Allows callers of the GetTree admin RPC to discover the public key of a tree. This will be particularly useful once Trillian is able to generate private keys, as this will be the only way to obtain the public key.


server/admin/admin_server_test.go
```diff
@@ -65,6 +66,23 @@ func TestServer_BeginError(t *testing.T) {
...
+
+	publicKeyDER, err := x509.MarshalPKIXPublicKey(privateKey.Public())
+	if err != nil {
+		t.Fatalf("Error marshaling public key: %v", err)
+	}
+
+	// Need to change the public key to correspond with the private key generated above.
+	validTree := *testonly.LogTree
+	validTree.PublicKey = &trillian.PublicKey{
+		Der: publicKeyDER,
```
codingllama: nit: Don't set the public key here? It's not relevant for this test.

server/admin/admin_server_test.go
```diff
@@ -260,24 +277,35 @@ func TestServer_CreateTree(t *testing.T) {
...
+	}
+
```
codingllama: Add a test for a tree that doesn't have the PubKey set, and check that the returned key is correct?

server/admin/admin_server_test.go
RJPercival: In fact it is. This func tests that `AdminServer` properly handles `AdminStorage.Begin()` returning an error but, if the public key is not set, CreateTree won't get as far as calling `Begin()`, because the tree will fail validation.

codingllama: I expected the server to derive the public key from the private before running storage validation, but maybe I'm missing some subtlety on the tests?

RJPercival: It will do if `PublicKey` is nil, but it's not nil in `testonly.LogTree`. This test could just do `validTree.PublicKey = nil` though, for simplicity. Would you like me to make that change?

server/admin/admin_server_test.go
RJPercival: Done, along with a test for if the public key does not match up with the private key.

Issue #571, opened 2017-04-28
CT: hardcoded low get-entries limit, with 400 otherwise


FiloSottile:
https://github.com/google/trillian/blob/0147e742d0dcfc42dbb4a73ef577e9b49692f8b1/examples/ct/handlers.go#L50

crt.sh requires 1024 get-entries calls, so it should probably be configurable.

Also, if the client exceeds it, a truncated response should probably be returned instead of a 400.

AGWA:
Not should, but SHALL according to RFC6962 section 4.6:

>    Logs MAY restrict the number of entries that can be retrieved per "get-entries" request.  If a client requests more than the permitted number of entries, the log SHALL return the maximum number of entries permissible.  These entries SHALL be sequential beginning with the entry specified by "start".

PR #572, opened 2017-04-29
CT: cap get-entries instead of erroring out - updates #571

FiloSottile:


rolandshoemaker:
[`TestGetEntriesRange`](https://github.com/FiloSottile/trillian/blob/13faf97d881b61613fc2407f75e676e7d2407cfd/examples/ct/handlers_test.go#L783) needs to be updated for the case where the range is truncated.

FiloSottile:
Yeah, my bad. I saw it after I lost wifi.

I'll patch it and add a new test when I figure out gomock.

daviddrysdale:
Subsumed into #577 and merged -- thanks!

PR #580, opened 2017-05-02
Use [][]byte for inclusion proofs

### LINKED ISSUE #499: Collapse Nodes in proof responses to bytes
There is no real point having the intermediate Node proto now it only contains one item after #497. The API can be simplified to repeated bytes.

gdbelvin:
There is no real point having the intermediate Node proto now it only
contains one item after #497. The API can be simplified to repeated
bytes.

Closes #499


Martin2112:
Can you fix the build before I review this? Thanks.

gdbelvin:
The build is now fixed and ready for your review :-)

PR #582, opened 2017-05-03
Generate private key for tree if none provided

RJPercival:
If a `CreateTreeRequest` is received by the admin server, but the provided `trillian.Tree` has no `PrivateKey` set, then the admin server will check whether the `KeyGen` field of the request is set. If it is, this will be passed to the `SignerFactory`, which will generate a key, store it and return an updated `trillian.Tree` that uses that key.

The default `SignerFactory`, `PEMSignerFactory`, will store the new key within the `trillian.Tree` proto. With the MySQL storage backend, this means the private keys will be held in the database. This is less secure than encrypted PEM files, because an attacker need only gain access to the database to obtain the key (whereas an attacker would only get the password in the case of a PEM file, and would then need to separately gain access to the file itself). However, it should suffice for the time being.


trillian_admin_api.proto
```diff
@@ -21,6 +21,7 @@ option java_outer_classname = "TrillianApiProto";
 package trillian;
 
 import "trillian.proto";
```
codingllama: Would you mind changing this line to `github.com/google/trillian/...` too?

trillian_admin_api.proto
```diff
@@ -47,6 +48,9 @@ message GetTreeRequest {
 message CreateTreeRequest {
   // Tree to be created. See Tree and CreateTree for more details.
   Tree tree = 1;
+  // Describes how the tree's keys should be generated.
```
codingllama: nit: Add a blank line between fields.

trillian_admin_api.proto
```diff
@@ -47,6 +48,9 @@ message GetTreeRequest {
...
+  // Only needs to be set if tree.private_key is not set.
+  keyspb.Generator key_gen = 2;
```
codingllama: nit: s/key_gen/generator/ ? (Or possibly something else if you take the comment on the generator rename.)

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package keyspb;
+
+// Generator defines how a new pair of asymmetric keys should be generated.
+message Generator {
+  message ECDSA {
```
codingllama: nit: Please document all messages / fields. This is part of our public API, so it should have as much documentation as possible.

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+    enum Curve {
+      P256 = 0;
```
codingllama: Suggestion: (Assuming you want P256 to be the default)

```proto
enum Curve {
  // Default curve size. Currently equivalent to P256.
  DEFAULT_CURVE = 0
  P224 = 1;
  P256 = 2;
  P384 = 3;
  P521 = 4;
}
```

Whereas, DEFAULT_CURVE is replaced by P256 server-side (for now). This allows us to increase the default without breaking the interface. (Note that responses should explicitly say P256 when DEFAULT_ comes in.)

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+      P224 = 1;
+      P384 = 2;
+      P521 = 3;
+    }
+    Curve curve = 1;
+  }
+
+  message RSA {
+    // Size of the keys in bits.
```
codingllama: Please document constraints: power of two? Min / max size? Is zero = default or is zero invalid? So on.

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package keyspb;
+
+// Generator defines how a new pair of asymmetric keys should be generated.
```
codingllama: I think it a bit strange that we keep the Godoc convention in protos. It makes for unusual comments when compared to proto libraries as used by other languages. (Very minor, feel free to push back.)

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+    ECDSA ecdsa_params = 1;
+    RSA rsa_params = 2;
+  }
+}
+
+// PEMKeyFile identifies a private key stored in a PEM-encoded file.
+message PEMKeyFile {
+  // File path of the private key.
+  string path = 1;
+  // Password for decrypting the private key.
```
codingllama: nit: blank line between fields.

Ditto for others.

crypto/keyspb/keyspb.proto
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package keyspb;
+
+// Generator defines how a new pair of asymmetric keys should be generated.
+message Generator {
```
codingllama: nit: "Generator" implies action, as in the "thing" that actually creates the key, whereas the message is more of an spec of how the key should be generated. Maybe rename?

crypto/keys/keys_test.go
```diff
@@ -220,3 +221,68 @@ func TestSignatureAlgorithm(t *testing.T) {
...
+		},
+		{
+			name:    "No params",
+			keygen:  &keyspb.Generator{},
+			wantErr: true,
+		},
+	} {
+		key, err := NewFromGenerator(test.keygen)
+		if gotErr := err != nil; gotErr != test.wantErr {
+			t.Errorf("%s: NewFromGenerator(_) = (_, %v), want err? %t", test.name, err, test.wantErr)
```
codingllama: nit: Use `%v` instead of `%s` / `%t`. It makes life a lot easier when refactoring code.

nit2: If you don't care about the params to NewFromGenerator, consider omitting them in the message (`"NewFromGenerator() = ..."`).

Ditto for others.

crypto/keys/keys_test.go
```diff
@@ -220,3 +221,68 @@ func TestSignatureAlgorithm(t *testing.T) {
...
+					EcdsaParams: &keyspb.Generator_ECDSA{
+						Curve: keyspb.Generator_ECDSA_P521,
+					},
+				},
+			},
+		},
+		{
+			name: "RSA with default params",
+			keygen: &keyspb.Generator{
+				Params: &keyspb.Generator_RsaParams{},
```
codingllama: What's the default key size? Should we check it?

crypto/keys/keys_test.go
```diff
@@ -220,3 +221,68 @@ func TestSignatureAlgorithm(t *testing.T) {
...
+func TestGenerateKey(t *testing.T) {
+	for _, test := range []struct {
+		name    string
+		keygen  *keyspb.Generator
+		wantErr bool
+	}{
+		{
+			name: "ECDSA with default params",
+			keygen: &keyspb.Generator{
+				Params: &keyspb.Generator_EcdsaParams{},
```
codingllama: Should we check that the resulting curve is P256?

crypto/keys/keys_test.go
```diff
@@ -220,3 +221,68 @@ func TestSignatureAlgorithm(t *testing.T) {
...
+			keygen: &keyspb.Generator{
+				Params: &keyspb.Generator_RsaParams{},
+			},
+		},
+		{
+			name: "RSA with explicit params",
+			keygen: &keyspb.Generator{
+				Params: &keyspb.Generator_RsaParams{
+					RsaParams: &keyspb.Generator_RSA{
+						Bits: 4096,
```
codingllama: Please add tests for bad number of bits (<0, non power of two)

crypto/keys/pem_signer_factory_test.go
```diff
@@ -107,3 +113,101 @@ func TestPEMSignerFactoryNewSigner(t *testing.T) {
...
+				Params: &keyspb.Generator_EcdsaParams{
+					EcdsaParams: &keyspb.Generator_ECDSA{},
+				},
+			},
+			tree: &trillian.Tree{
+				SignatureAlgorithm: sigpb.DigitallySigned_ECDSA,
+			},
+		},
+		{
+			name: "Replace existing key",
```
codingllama: Should we be conservative and make this error? Is there a situation where we'd want this to happen?

crypto/keys/private.go
```diff
@@ -26,15 +28,24 @@ import (
...
 // A signers may be created by loading a private key, interfacing with a HSM,
 // or sending network requests to a remote key management service, to give a few
 // examples.
 type SignerFactory interface {
 	// NewSigner returns a signer for the given tree.
+	// It consults Tree.PrivateKey to determine how to retrieve the key.
 	NewSigner(context.Context, *trillian.Tree) (crypto.Signer, error)
+
+	// Generate creates a new asymmetric key pair for a tree according to some parameters.
+	// It returns the tree with updated PrivateKey and PublicKey fields.
```
codingllama: nit: "It returns a new tree with updated ..."

Any reason you chose this route instead of modifying the input tree?

Conceptually, I would prefer it to do neither and instead return `(PrivateKey, error)`; That would mean moving all the key-derivation and Any-handling logic to the Admin server, which seems better as well (it already does key derivation).

crypto/keys/private.go
```diff
@@ -26,15 +28,24 @@ import (
...
 )
 
+// The size of an RSA key generated by this library, in bits, if not overridden.
+const defaultRsaKeySizeInBits = 2048
+
 // SignerFactory creates signers for Trillian trees.
 // A signers may be created by loading a private key, interfacing with a HSM,
 // or sending network requests to a remote key management service, to give a few
 // examples.
 type SignerFactory interface {
```
codingllama: (pre-existing nit): SignerFactory is publicly used, but declared in private.go. That doesn't seem right.

server/admin/admin_server.go
```diff
@@ -82,24 +83,39 @@ func (s *Server) CreateTree(ctx context.Context, request *trillian.CreateTreeReq
...
-		return nil, status.Errorf(codes.InvalidArgument, "failed to marshal public key: %v", err.Error())
-	}
-	// If a public key was provided, check that it matches the one we derived. If it doesn't, this indicates a mistake by the caller.
-	if tree.PublicKey != nil && !bytes.Equal(tree.PublicKey.Der, publicKeyDER) {
-		return nil, status.Error(codes.InvalidArgument, "the public and private keys are not a pair")
-	}
-	// If no public key was provided, use the DER that we just marshaled.
-	if tree.PublicKey == nil {
-		tree.PublicKey = &trillian.PublicKey{Der: publicKeyDER}
+	// If a private key was provided, validate it. If a key was not provided, generate one.
```
codingllama: If a request has both `tree.PrivateKey` and `KeyGen` it should fail - the client has to clearly pick one. (Please add a test too.)

crypto/keys/pem_signer_factory_test.go
```diff
@@ -107,3 +113,101 @@ func TestPEMSignerFactoryNewSigner(t *testing.T) {
...
+			tree: &trillian.Tree{
+				SignatureAlgorithm: sigpb.DigitallySigned_RSA,
+			},
+			wantErr: true,
+		},
+	} {
+		ctx := context.Background()
+		var sf PEMSignerFactory
+
+		newTree, err := sf.Generate(ctx, test.tree, test.keygen)
```
codingllama: As it stands, `Generate()` could be modifying unrelated tree fields and this test would be none the wiser. I think the better approach here is to change Generate's return, as stated in a following comment.

server/admin/admin_server.go
```diff
@@ -82,24 +83,39 @@ func (s *Server) CreateTree(ctx context.Context, request *trillian.CreateTreeReq
...
+		}
+
+		// If no public key was provided, use the DER that we just marshaled.
+		if tree.PublicKey == nil {
+			tree.PublicKey = &keyspb.PublicKey{Der: publicKeyDER}
+		}
+	case request.KeyGen != nil:
+		newTree, err := s.registry.SignerFactory.Generate(ctx, tree, request.KeyGen)
+		if err != nil {
+			return nil, status.Errorf(codes.FailedPrecondition, "failed to generate private key: %v", err.Error())
```
codingllama: `InvalidArgument` instead of `FailedPrecondition`? There isn't much system state that could interfere in key generation (at least not in the current impl, it seems).

server/admin/admin_server.go
```diff
@@ -82,24 +83,39 @@ func (s *Server) CreateTree(ctx context.Context, request *trillian.CreateTreeReq
...
+		// If a public key was provided, check that it matches the one we derived. If it doesn't, this indicates a mistake by the caller.
+		if tree.PublicKey != nil && !bytes.Equal(tree.PublicKey.Der, publicKeyDER) {
+			return nil, status.Error(codes.InvalidArgument, "the public and private keys are not a pair")
+		}
+
+		// If no public key was provided, use the DER that we just marshaled.
+		if tree.PublicKey == nil {
+			tree.PublicKey = &keyspb.PublicKey{Der: publicKeyDER}
+		}
+	case request.KeyGen != nil:
```
codingllama: Please add an admin-layer test for key generation. No need to test all key types, one happy path and one fail test (bad params) should be fine.

crypto/keyspb/keyspb.proto
RJPercival: Done.

crypto/keyspb/keyspb.proto
RJPercival: Go's crypto/rsa package doesn't document any constraints but, looking at its implementation, the only constraint appears to be that `bits` must be large enough to generate 2 primes. I'll document that 0 is treated as "use a default key size" too.

crypto/keyspb/keyspb.proto
RJPercival: Done.

trillian_admin_api.proto
codingllama: Update: this may break things. If it does, just ignore the comment.

crypto/keyspb/keyspb.proto
codingllama: Maybe something like `KeySpecification`? (Don't feel like you have to take it.)

crypto/keyspb/keyspb.proto
codingllama: crypto/rsa doesn't seem to be opinionated about key sizes, but I think we should. Maybe disallow <2048?

The same reasoning applies to curve length. Maybe we should remove P224 as an option?

crypto/keys/keys_test.go
codingllama: Other than being unusual, not sure. :)

crypto/keys/private.go
codingllama: Oh, I missed that. My bad. :/

Yep, a comment or a slight rename might same me from making the same comment again in a few days.

crypto/keys/private.go
codingllama: SG. WDYT of `Generate(...) (Any, crypto.PublicKey, error)`?

trillian_admin_api.proto
codingllama: Update2: aaaaand this broke Travis. It's not as simple as I thought. Sorry about that, please revert.

crypto/keyspb/keyspb.proto
codingllama: It's better, thanks. This is a pet peeve of mine, but I like APIs to be fully documented.

trillian_admin_api.proto
RJPercival: Renamed to be `key_spec`.

trillian_admin_api.proto
RJPercival: Done.

trillian_admin_api.proto
RJPercival: Reverted.

RJPercival:
Ready for another pass - I think I've addressed all of your comments.

crypto/keyspb/keyspb.proto
codingllama: There's also a distinction between the keys we allow and the keys we generate. If someone has a pre-existing P224 key, they are free to go ahead and use it, but if we're generating a new key I think it's OK to be more picky.

crypto/keys/pem_signer_factory.go
```diff
@@ -41,9 +47,49 @@ func (f PEMSignerFactory) NewSigner(ctx context.Context, tree *trillian.Tree) (c
...
+
+	switch key := key.(type) {
+	case *ecdsa.PrivateKey:
+		privateKeyDER, err = x509.MarshalECPrivateKey(key)
+	case *rsa.PrivateKey:
+		privateKeyDER = x509.MarshalPKCS1PrivateKey(key)
+	default:
+		return nil, fmt.Errorf("unsupported key type: %T", key)
+	}
+	if err != nil {
```
codingllama: nit: Move `var err error`  and the if below line 84? That's the only branch with errors.

server/admin/admin_server_test.go
```diff
@@ -328,6 +335,41 @@ func TestServer_CreateTree(t *testing.T) {
...
+			req: &trillian.CreateTreeRequest{
+				Tree: &omittedKeys,
+				KeySpec: &keyspb.Specification{
+					Params: &keyspb.Specification_EcdsaParams{},
+				},
+			},
+			wantCommit: true,
+		},
+		{
+			desc: "privateKeySpecandPrivateKeyProvided",
```
codingllama: nit: privateKeySpec**And**PrivateKeyProvided (capitalized "And").

server/admin/admin_server_test.go
```diff
@@ -369,7 +411,25 @@ func TestServer_CreateTree(t *testing.T) {
 
 	ctx := context.Background()
 	for _, test := range tests {
-		setup := setupAdminServer(ctrl, privateKey, false /* snapshot */, test.wantCommit, test.commitErr)
+		sf := keys.NewMockSignerFactory(ctrl)
+		if test.req.GetKeySpec() != nil {
+			if test.req.Tree.GetPrivateKey() != nil || test.req.Tree.GetPublicKey() != nil {
```
codingllama: nil PrivateKey but non-nil PublicKey doesn't cause an error on PEMSignerFactory. Would you mind adding a test case for this scenario on pem_signer_factory_test.go too?

trillian_admin_api.proto
```diff
@@ -21,6 +21,7 @@ option java_outer_classname = "TrillianApiProto";
 package trillian;
 
 import "trillian.proto";
+import "github.com/google/trillian/crypto/keyspb/keyspb.proto";
```
codingllama: import "crypto/keyspb/keyspb.proto", so it's consistent with other internal imports?

RJPercival: As best as I can tell, you can't import protos from a different directory with a relative path like that. It causes the compilation error `cannot find package "crypto/keyspb"`.

RJPercival: It's also not consistent with other internal imports. The `crypto/sigpb/sigpb.proto` is imported as `github.com/google/trillian/crypto/sigpb/sigpb.proto`, for example. I think it's only protos that are in the same directory that are included using relative paths.

server/admin/admin_server_test.go

server/admin/admin_server_test.go

crypto/keys/pem_signer_factory.go
RJPercival: Done.

PR #583, opened 2017-05-03
Set session affinity of MySQL load balancer to "ClientIP"

RJPercival:
This avoids a single sequencer connecting to multiple database replicas, which results in a large number of conflicts (it appears the parallel goroutines can start multiple transactions, with identical sequence numbers, only one of which can actually succeed).


Martin2112:
I think this is worth changing and should benefit performance. 

There should only be one sequencer goroutine per tree. So I'm not sure we fully understand what's happening here. Possibly they're being started at fixed intervals regardless of prior completion. Something like that could cause it but I don't see that happening in the logs for my test runs.

Issue #578, opened 2017-05-02
Move towards file based configuration instead of flag based


rolandshoemaker:
Ideally I'd like to manage `trillian_log_server`, `trillian_log_signer`, and `ct_server` as services but given that a large portion of their configuration is provided via flags this is somewhat of a pain to do. The majority of these flags could be moved into JSON/YAML configuration files to make this easier to accomplish.

Martin2112:
What do you mean by 'service' in this context, what's your environment like?

rolandshoemaker:
`systemd`, it would be entirely possible to create a management script for the current design but that means needing to customize that script for each different deployment to change things like MySQL/backend URIs etc instead of being able to manage those at the configuration file level with something like Salt.

Martin2112:
OK. I don't write systemd configs or use Salt so I'm not clear whether we'd need a different main that reads a fixed config path or we could support both by adding a --config <file> flag to the existing one.

Can you give an example of what would work best for your planned setup using e.g. log_server, which only has 4 flags?

rolandshoemaker:
> OK. I don't write systemd configs or use Salt so I'm not clear whether we'd need a different main that reads a fixed config path or we could support both by adding a --config flag to the existing one.

Adding a `--config` flag seems like the easiest option and work for our purposes, general design would be you would set a default path of something like `/etc/trillian/log_config.json` and then allow an override with the `--config` flag but I think it would also work fine to just set an empty default and only override the existing flags if a path was provided.

> Can you give an example of what would work best for your planned setup using e.g. log_server, which only has 4 flags?

JSON would probably be the simplest configuration format since you wouldn't need any extra third-party dependencies, then you could have something basic like

```
{
  "http_endpoint": "localhost:8091",
  "rpc_endpoint": "localhost:8090",
  "mysql_uri": "test:zaphod@tcp(127.0.0.1:3306)/test",
  "log_dir": "/var/log/trillian",
  ...
}
```

For our purposes we would then generate this file from a template stored on our central configuration server whenever we brought up a new host/cluster and store it at some standard path we set in the `--config` flag.

Martin2112:
Thanks, that's helpful.

I'm not overly keen on a default path for cross-platform compatibility reasons and the maybe minor risk of confusion by accidentally picking up an existing file and having your flags ignored.

Otherwise, I think it's worth doing. Would be great if you had some time to work on PRs to add it!

rolandshoemaker:
[...]

PR #584, opened 2017-05-04
Use relative import paths in protos

gdbelvin:
The current mix of relative and absolute import paths in protos prevents other projects from referencing them. 

This PR standardizes on relative import paths in the trillian protos to resolve this issue. 

Related issue: https://github.com/coreos/etcd/pull/5194


trillian_admin_api.proto
```diff
@@ -20,7 +20,7 @@ option java_outer_classname = "TrillianApiProto";
 
 package trillian;
 
-import "trillian.proto";
+import "github.com/google/trillian/trillian.proto";
```
codingllama: I've been looking at proto files in other Google APIs as a reference, but none seem to include the repo path in their imports.

See this [search on github.com/googleapis/googleapis](https://github.com/googleapis/googleapis/search?utf8=%E2%9C%93&q=import&type=), for example.

I wonder if we're being a bit unorthodox here. Maybe instead of the full repo path, we could go with just `trillian/` (eg, `trillian/trillian.proto`)?

trillian_log_api.proto
```diff
@@ -20,8 +20,8 @@ option java_outer_classname = "TrillianApiProto";
 
 package trillian;
 
-import "trillian.proto";
-import "google/rpc/status.proto";
+import "github.com/google/trillian/trillian.proto";
+import "github.com/googleapis/googleapis/google/rpc/status.proto";
```
codingllama: I don't think this import should change.

trillian_admin_api.proto
Martin2112: Can we give that a try and see if it works?

gdbelvin: The problem is that protoc can't distinguish 
`trillian.proto` with `-I=.` from
`github.com/google/trillian/trillian.proto` with `-I=$GOPATH/src`

The current scheme of using local paths eg `trillian.proto` works great as long as these proto files don't get referenced from other proto files outside this directory. But as soon as we have  a personality that wants to proxy some of these types (eg. #578), we'll have a problem:
- The personality imports `github.com/google/trillian/trillian.proto`
- The personality imports `github.com/google/trillian/trillian_log_api.proto`
- which imports `trillian.proto`, which is interpreted as a third file rather than the same as the first, causing issues.

gdbelvin:
This PR is just one example of how the proto issue could be resolved. Please feel free to close this and use the ideas as inspiration if you prefer.

trillian_admin_api.proto
gdbelvin: Another strategy could be enforcing a rule of only one proto file per package. 
- We could collapse all the protos into a single file
- We could split the various proto files into separate packages.

codingllama: I don't like so much the idea of enforcing one proto file per package. I think I'd rather have the "github.com/" import in this case.

I can see how the "trillian/" import wouldn't work now (at least not in the current arrangement). In a 2nd attempt of mirroring googleapis, WDYT if we moved the proto files under a `proto` folder, like this:

```
github.com/google/trillian
  proto/
    trillian/
      crypto/
        sigpb/
          sigpb.proto
      trillian.proto
      trillian_admin_api.proto
      trillian_log_api.proto
      trillian_map_api.proto
```
It seems to me that, in this case, we could use `-I=$GOPATH/github.com/google/trillian/proto` and `import "trillian/trillian.proto"` as well. Mockgen commands would need to change a bit, but I think we could still output the .pb.go files to the root, as we do now.

@gdbelvin, thanks for the patience with my suggestions!

gdbelvin: - I like the idea of putting all the proto files under `trillian/proto`
- Typically pb.go files are placed in the same directory as their proto files. This makes it obvious which gen.go needs to be invoked to update it. Is there a different pattern that you're seeing? 
- What import path do you suggest that external projects use? Do we want to require relying projects to add `-I=$GOPATH/github.com/google/trillian/proto` to their protoc invocations?

codingllama: 1. :)
1. I'm not sure if that will work due to Go package names. If it does, I actually prefer it as well.
1. Yes, I'd be OK with that.

gdbelvin:
Ok, post discussion I think the easiest way for external projects to use trillian protos is to have them add the trillian project path to one of their protoc imports `"-I=$GOPATH/src/github.com/google/trillian"`. 

This PR now just makes sure that all proto reference paths are local.

gdbelvin:
Proto best practices: https://johanbrandhorst.github.io/blog/post/go-protobuf-tips/

RJPercival:
Is this abandoned or do you still intend to make this change at some point?

gdbelvin:
This is a change we should still make. However it doesn't have a lot of
concrete value until grpc/grpc#10304
<https://github.com/grpc/grpc/issues/10304> has been fixed.

gdbelvin:
Reviving this PR so others can generate protos correctly.  PTAL

anandwana001:
If I use complete path of proto, generateProto of protoc Gradle plugin fails and If I use just the name of the proto file (as both of my proto files are in the same package) lint check is getting failed. 

`protoc -I . --buf-check-lint_out=. $(find . -name '*.proto')`
`buf check lint`

gdbelvin:
@anandwana001 Please open a new issue for this so we can prioritize it correctly.

PR #585, opened 2017-05-05
Make clients responsible for managing the trusted root

gdbelvin:
Previously, I had pulled out the logic around updating the trusted root into this sub-module. 
When I tried using it in key transparency, I realized I had introduced a module with difficult-to-test state as a deep dependency, making things difficult to reason about and test.

This PR keeps the authentication submodules stateless. 

While this reduces the responsibilities of this submodule to a thin wrapper, it still has value as an abstractable interface for testing and mocking, as well as clarify of function invocation.

### DIFF
--- a/client/offline.go
+++ b/client/offline.go
@@ -39,8 +39,8 @@ func NewLogVerifier(hasher merkle.TreeHasher, pubKey crypto.PublicKey) LogVerifi
 	}
 }
 
-// VerifyRoot verifies that a resp is a valid append-only operation from trusted.
-// If trusted.TreeSize is zero, an append-only proof is not needed.
+// VerifyRoot verifies that newRoot is a valid append-only operation from trusted.
+// If trusted.TreeSize is zero, a consistency proof is not needed.
 func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
 	consistency [][]byte) error {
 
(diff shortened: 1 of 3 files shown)


client/offline.go
```diff
@@ -40,51 +39,42 @@ func NewLogVerifier(hasher merkle.TreeHasher, pubKey crypto.PublicKey) LogVerifi
...
-func (c *logVerifier) Root() trillian.SignedLogRoot {
-	return c.root
-}
-
-// UpdateRoot applies a GetLatestSignedLogRootResponse to Root(), if valid.
-// consistency may be nil if Root().TreeSize is zero.
-func (c *logVerifier) UpdateRoot(resp *trillian.GetLatestSignedLogRootResponse,
-	consistency *trillian.GetConsistencyProofResponse) error {
-	str := resp.SignedLogRoot
+// VerifyRoot verifies that a resp is a valid append-only operation from trusted.
```
codingllama: nit: s/a resp/newRoot/

client/offline.go
```diff
@@ -40,51 +39,42 @@ func NewLogVerifier(hasher merkle.TreeHasher, pubKey crypto.PublicKey) LogVerifi
...
+// If trusted.TreeSize is zero, an append-only proof is not needed.
```
codingllama: nit: s/an append-only/a consistency/

client/offline.go
gdbelvin: done

client/offline.go
gdbelvin: done

PR #587, opened 2017-05-05
Reformat handlers_test.go

codingllama:


codingllama:
Aka: scripts/presubmit.sh --fix

Issue #490, opened 2017-03-28
Rate limiting for logs


Martin2112:
The log server must implement rate limiting and prevent a large queue backlog building up. Otherwise, sequencing could fail to catch up with entries before the MMD expires.

Edit by @codingllama: s/admission control/rate limiting/

Note that this is global-per-log rate limiting, per-tenant rate-limiting is separate (see XXX)

codingllama:
FYI, access control is tracked by #311. This is on my backlog.

Martin2112:
I was reading access control to be permissions, rather than rate limiting. I think they're different things with different priorities.

codingllama:
You said "The log server must implement admission control" in the 1st comment, but now I'm guessing you meant rate limiting instead? I think we're on the same page here.

Martin2112:
#311 didn't have a description and I may have added to the confusion. I think that this one means rate limiting and #311 means limiting access to particular trees, possibly a no-op in the initial version.

codingllama:
For details on the rate limiting approach see #588 / [quota/doc.go](https://github.com/google/trillian/blob/master/quota/doc.go).

codingllama:
After a period of design and experimentation, I'm back with #726, which kicks off an etcd-backed quota mechanism. This mechanism is flexible enough to support all (up until now) theoretic quota features:

* Global, per Tree and per User quotas, split into read and write 
* Sequencing-based replenishment
* Time-based replenishment

Global and tree quotas work out of the box. User quotas require defining a "quota user". While there isn't a definite answer (yet) in this area, a few options present themselves, eg, defining users as a Trillian fork or allowing the personality to specify the quota user on RPCs, the latter which would be my choice ATM.

Default trees and users quotas is a planned feature, but I won't add it at the first iteration. Unfortunately, this means all quotas must be pre-configured, which means extra setup for every new tree. For most Trillian installations that won't be much of problem, but that's not practical for installations with a great number of trees (or a great number of defined quotas users). That's when default quotas offer the highest gain.

Etcd is already required for the signer, so I'm hoping it won't set the bar any higher for adoption of the quota system.

Quota setup and occasional maintenance will be performed via REST endpoints. I'll add some docs when I've got it all working and plugged together.

PR #588, opened 2017-05-05
	Interface and noop implementation for Trillian quotas

codingllama:
Quotas are meant as a protection against traffic peaks / service DoS. Each
request requires certain tokens, which are replenished over time. If there
aren't enough tokens, requests are denied.

An interface for acquiring / replenishing tokens is defined, along with a Noop
implementation that acts as if it had infinite tokens.


codingllama:
#490

PR #589, opened 2017-05-05
Add ctx to NodeReader / NodeWriter

codingllama:
Similar to #568, but for tree storage.

### DIFF
--- a/merkle/sparse_merkle_tree_test.go
+++ b/merkle/sparse_merkle_tree_test.go
@@ -147,7 +147,7 @@ func TestRootAtRevision(t *testing.T) {
 	r, tx := getSparseMerkleTreeReaderWithMockTX(mockCtrl, 100)
 	node := getRandomRootNode(t, 14)
 	tx.EXPECT().Commit().AnyTimes().Return(nil)
-	tx.EXPECT().GetMerkleNodes(gomock.Any(), int64(23), rootNodeMatcher{}).Return([]storage.Node{node}, nil)
+	tx.EXPECT().GetMerkleNodes(ctx, int64(23), rootNodeMatcher{}).Return([]storage.Node{node}, nil)
 	root, err := r.RootAtRevision(ctx, 23)
 	if err != nil {
 		t.Fatalf("Failed when calling RootAtRevision(23): %v", err)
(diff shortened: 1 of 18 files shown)


codingllama:
The Yak is finally naked.

#419

merkle/sparse_merkle_tree_test.go
```diff
@@ -138,14 +139,16 @@ func getRandomNonRootNode(t *testing.T, rev int64) storage.Node {
...
+
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
 	r, tx := getSparseMerkleTreeReaderWithMockTX(mockCtrl, 100)
 	node := getRandomRootNode(t, 14)
 	tx.EXPECT().Commit().AnyTimes().Return(nil)
-	tx.EXPECT().GetMerkleNodes(int64(23), rootNodeMatcher{}).Return([]storage.Node{node}, nil)
-	root, err := r.RootAtRevision(23)
+	tx.EXPECT().GetMerkleNodes(gomock.Any(), int64(23), rootNodeMatcher{}).Return([]storage.Node{node}, nil)
```
Martin2112: In these tests can we check that the supplied context was propagated rather than use Any? Or does that not work.

server/vmap/trillian_map_server.go
```diff
@@ -120,9 +120,12 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	defer tx.Close()
 
 	glog.Infof("%v: Writing at revision %v", mapID, tx.WriteRevision())
```
Martin2112: While you're here can you demote this logging to debug level? Thanks.

merkle/sparse_merkle_tree_test.go
codingllama: Done (used `glog.V(2).Infof(...)`).

server/vmap/trillian_map_server.go
codingllama: Done (used `glog.V(2).Infof(...)`).

merkle/sparse_merkle_tree_test.go
codingllama: Done.

PR #592, opened 2017-05-06
Update java_outer_classname on new log/map APIs so they are distinct.…

shemnon:
… Otherwise protoc will fail with "Tried to write the same file twice."

### DIFF
--- a/trillian.proto
+++ b/trillian.proto
@@ -14,11 +14,14 @@
 
 syntax = "proto3";
 
+option java_multiple_files = true;
+option java_package = "com.google.trillian.proto";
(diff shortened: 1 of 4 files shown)


trillian.proto
```diff
@@ -18,6 +18,7 @@ package trillian;
 
 import "github.com/google/trillian/crypto/sigpb/sigpb.proto";
 import "google/protobuf/any.proto";
+option java_outer_classname = "TrillianProto";
```
codingllama: Would you mind setting `option java_package = "com.google.trillian.proto";` here too?

PR #602, opened 2017-05-10
Update blob types to fix 64K leaf size limit

### LINKED ISSUE #596: CT: leaf_input truncated if > 65,535 bytes
The MySQL schema stores the `LeafValue` and `ExtraData` in a `BLOB`, so they're silently truncated if they're over 65,535 bytes.

I think this is causing problems with Let's Encrypt's test CT log, which I believe runs Trillian. The `leaf_input` of entry 6984671 (http://clicky.ct.letsencrypt.org/ct/v1/get-entries?start=6984671&end=6984671) has been truncated to exactly 65,535 bytes despite the certificate being much longer - the 10k bad SANs cert from https://10000-sans.badssl.com/.

This also causes Trillian to produce invalid Merkle tree hashes as the `leaf_input` hash is calculated before MySQL ends up truncating the `leaf_input`, whereas monitors will compute the hash of the truncated entries. (I think this is ultimately causing mine to mark every Let's Encrypt Clicky STH as invalid - see https://ct.grahamedgecombe.com/logs/38.)

The `ASN1.Cert` structure in RFC 6962 has a 3 byte length so it can be up to 16 megabytes. However, a `MEDIUMBLOB` still isn't enough in the worst case as a few extra bytes are consumed by the timestamp/entry type, so the schema probably needs to use `LONGBLOB` instead (or reject submissions which are too long, though I don't know whether this would be considered acceptable under the current Chrome policy).

Martin2112:
Leaf / client supplied values -> LONGBLOB, keys and internal stuff -> MEDIUMBLOB.

Closes #596.

### DIFF
--- a/storage/mysql/storage.sql
+++ b/storage/mysql/storage.sql
@@ -22,8 +22,8 @@ CREATE TABLE IF NOT EXISTS Trees(
   Description           VARCHAR(200),
   CreateTimeMillis      BIGINT NOT NULL,
   UpdateTimeMillis      BIGINT NOT NULL,
-  PrivateKey            BLOB NOT NULL,
-  PublicKey             BLOB NOT NULL,
+  PrivateKey            MEDIUMBLOB NOT NULL,
+  PublicKey             MEDIUMBLOB NOT NULL,
   PRIMARY KEY(TreeId)
 );
 
@@ -80,10 +80,10 @@ CREATE TABLE IF NOT EXISTS LeafData(
   LeafIdentityHash     VARBINARY(255) NOT NULL,
   -- This is the data stored in the leaf for example in CT it contains a DER encoded
   -- X.509 certificate but is application dependent
-  LeafValue            BLOB NOT NULL,
+  LeafValue            LONGBLOB NOT NULL,
   -- This is extra data that the application can associate with the leaf should it wish to.
   -- This data is not included in signing and hashing.
-  ExtraData            BLOB,
+  ExtraData            LONGBLOB,
   PRIMARY KEY(TreeId, LeafIdentityHash),
   FOREIGN KEY(TreeId) REFERENCES Trees(TreeId) ON DELETE CASCADE
 );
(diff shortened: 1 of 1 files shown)


storage/mysql/storage.sql
```diff
@@ -22,8 +22,8 @@ CREATE TABLE IF NOT EXISTS Trees(
   Description           VARCHAR(200),
   CreateTimeMillis      BIGINT NOT NULL,
   UpdateTimeMillis      BIGINT NOT NULL,
-  PrivateKey            BLOB NOT NULL,
-  PublicKey             BLOB NOT NULL,
+  PrivateKey            MEDIUMBLOB NOT NULL,
```
codingllama: I *think* BLOB is good enough for both Private and Public keys. Private are serialized protos (currently containing a file path), publics are DER-encoded keys (but not certs).

storage/mysql/storage.sql
```diff
@@ -80,10 +80,10 @@ CREATE TABLE IF NOT EXISTS LeafData(
   LeafIdentityHash     VARBINARY(255) NOT NULL,
   -- This is the data stored in the leaf for example in CT it contains a DER encoded
   -- X.509 certificate but is application dependent
-  LeafValue            BLOB NOT NULL,
+  LeafValue            LONGBLOB NOT NULL,
   -- This is extra data that the application can associate with the leaf should it wish to.
   -- This data is not included in signing and hashing.
-  ExtraData            BLOB,
+  ExtraData            LONGBLOB,
```
codingllama: Do we really want to allow extra data up to 4GB? Maybe cap at MEDIUMBLOB?

storage/mysql/storage.sql
Martin2112: I'm not certain that 64KB is enough. It might be but the cost of avoiding that limit is low.

storage/mysql/storage.sql
Martin2112: I think we should probably allow for it as we don't know what applications will really need. Clients won't be able to submit huge blobs unless the database server is configured for it.

Martin2112:
If we hit the packet size limit in testing we can look at increasing it. I think it will produce a reasonable error message if it happens.

From the department of coincidences I think the default server packet size setting is the same as the max allowed size defined in RFC 6962.

storage/mysql/storage.sql
codingllama: Fair enough.

PR #605, opened 2017-05-11
Allow Requesting SignedMapRoot by Revision

cesarghali:
Requesting `SignedMapRoot` by revision is a helpful feature for some Trillian personalities like Key Transparency. In fact, Key Transparency's monitor API (`GetMutations`) requires fetching `SignedMapRoot` given a specific revision number. This allows monitors to verify map revisions older than the latest one.


storage/mysql/map_storage.go
```diff
@@ -33,6 +33,9 @@ const (
 	selectLatestSignedMapRootSQL = `SELECT MapHeadTimestamp, RootHash, MapRevision, RootSignature, MapperData
 		 FROM MapHead WHERE TreeId=?
 		 ORDER BY MapHeadTimestamp DESC LIMIT 1`
+	selectGetSignedMapRootSQL = `SELECT MapHeadTimestamp, RootHash, MapRevision, RootSignature, MapperData
+		 FROM MapHead WHERE TreeId=? AND MapRevision=?
+		 ORDER BY MapHeadTimestamp DESC LIMIT 1`
```
Martin2112: Don't think you need the order by clause as the above query should already restrict to one row. There a unique constraint on (TreeId, MapRevision).

cesarghali: Good catch, thanks Martin. The code is fixed.

PR #607, opened 2017-05-11
Add Docker build files

gdbelvin:
Support building trillian clusters by defining docker build files.
They should be run like so:
`docker build -f server/trillian_log_server/Dockerfile .` from `$GOPATH/src/github.com/google/trillian`

Please correct me if these are the wrong binaries to be defining definitions for. 
I see that there's also a `server/main.go` which has some duplicate functionality.

### DIFF
--- a/Dockerfile.db
+++ b/Dockerfile.db
@@ -1,0 +1,8 @@
+FROM mysql
+
+ENV MYSQL_PASSWORD=zaphod \
+    MYSQL_USER=test \
+    MYSQL_DATABASE=test \
+    MYSQL_RANDOM_ROOT_PASSWORD=yes
+
+ADD storage/mysql/storage.sql /docker-entrypoint-initdb.d/storage.sql
--- a/server/trillian_log_server/Dockerfile
+++ b/server/trillian_log_server/Dockerfile
@@ -1,0 +1,29 @@
+FROM golang
+
+ENV DB_USER=test \
+    DB_PASSWORD=zaphod \
+    DB_DATABASE=test \
+    DB_HOST=127.0.0.0:3306 
+
+ENV HOST=0.0.0.0 \
+    RPC_PORT=8090 \
+    HTTP_PORT=8091 
+
+ENV DUMP_METRICS 0s
+
+ADD . /go/src/github.com/google/trillian
+WORKDIR /go/src/github.com/google/trillian 
+
+RUN go get -v ./server/trillian_log_server
+
+ENTRYPOINT /go/bin/trillian_log_server \
+	--mysql_uri="${DB_USER}:${DB_PASSWORD}@tcp(${DB_HOST})/${DB_DATABASE}" \
+	--rpc_endpoint="$HOST:$RPC_PORT" \
+	--http_endpoint="$HOST:$HTTP_PORT" \
+	--dump_metrics_interval="$DUMP_METRICS"
+
+EXPOSE $RPC_PORT
+EXPOSE $HTTP_PORT
+
+HEALTHCHECK --interval=5m --timeout=3s \
+  CMD curl -f http://localhost:$HTTP_PORT/debug/vars || exit 1
--- a/server/trillian_log_server/Dockerfile
+++ b/server/trillian_log_server/Dockerfile
@@ -11,7 +11,7 @@ ENV HOST=0.0.0.0 \
 
 ENV DUMP_METRICS 0s
 
-ADD . /go/src/github.com/google/trillian
+ADD . /go/src/github.com/google/trillian 
 WORKDIR /go/src/github.com/google/trillian 
 
 RUN go get -v ./server/trillian_log_server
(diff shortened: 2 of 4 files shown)


server/trillian_log_server/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ENV RPC_PORT 8090
+ENV HTTP_PORT 8091
+
+ENV DUMP_METRICS 0s
+
+ADD . /go/src/github.com/google/trillian
+WORKDIR /go/src/github.com/google/trillian 
+
+RUN go get -d -v ./server/trillian_log_server
+RUN go install ./server/trillian_log_server
```
RJPercival: Why do you tell `go get` not to install `trillian_log_server` (using the -d flag), then immediately install it using `go install`?

server/trillian_log_server/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ENV DB_USER test
+ENV DB_PASSWORD zaphod
+ENV DB_DATABASE test
+ENV DB_HOST 127.0.0.0:3306
+
+ENV HOST 0.0.0.0
+ENV RPC_PORT 8090
+ENV HTTP_PORT 8091
+
+ENV DUMP_METRICS 0s
```
RJPercival: [It'd be better for caching to set all of these using a single `ENV` command.](https://docs.docker.com/engine/reference/builder/#env)

gdbelvin: Good call

server/trillian_log_server/Dockerfile
gdbelvin: Here I'm following Dockerfile [examples](https://github.com/docker-library/golang/blob/132cd70768e3bc269902e4c7b579203f66dc9f64/1.8/onbuild/Dockerfile)

Looks like doing it in one line is ok too.

AlCutter:
How we doing with this one?
There seems to be a bunch of unrelated proto changes in here too, did they sneak in by accident?

gdbelvin:
This is updated and ready for review. 
@AlCutter the proto changes slipped in, they're now removed. 
@RJPercival How does this PR intersect with some of your work to use Kubernetes?

server/vmap/trillian_map_server/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ENV HOST=0.0.0.0 \
+    RPC_PORT=8090 \
+    HTTP_PORT=8091 
+
+ADD . /go/src/github.com/google/trillian
+WORKDIR /go/src/github.com/google/trillian 
+
+RUN go get ./server/vmap/trillian_map_server
+
+# Run the outyet command by default when the container starts.
```
RJPercival: "outyet command"?

server/trillian_log_signer/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN go get -v ./server/trillian_log_signer
+
+# Run the outyet command by default when the container starts.
+ENTRYPOINT /go/bin/trillian_log_signer \
+	--mysql_uri="${DB_USER}:${DB_PASSWORD}@tcp(${DB_HOST})/${DB_DATABASE}" \
+	--http_endpoint="$HOST:$HTTP_PORT" \
+	--dump_metrics_interval="$DUMP_METRICS" \
+	--sequencer_guard_window="$SEQUENCER_GUARD_WINDOW" \
+	--force_master="$FORCE_MASTER"
```
RJPercival: This is missing the --alsologtostderr flag that the other Dockerfiles have.

server/trillian_log_server/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM golang
+
+ENV DB_USER=test \
```
RJPercival: It might be better to use [ARG](https://docs.docker.com/engine/reference/builder/#arg) rather than ENV, so that these values can be overridden when building the image. I'm ok with ENV though.

Dockerfile.db
```diff
@@ -0,0 +1,8 @@
+FROM mysql
+
+ENV MYSQL_PASSWORD=zaphod \
+    MYSQL_USER=test \
+    MYSQL_DATABASE=test \
+    MYSQL_RANDOM_ROOT_PASSWORD=yes
+
+ADD storage/mysql/storage.sql /docker-entrypoint-initdb.d/storage.sql
```
RJPercival: What is this Dockerfile for?

server/vmap/trillian_map_server/Dockerfile
gdbelvin: Removed. This was left over from an example file.

server/trillian_log_signer/Dockerfile
gdbelvin: added.

server/trillian_log_server/Dockerfile
gdbelvin: ENV allows the arguments to be overridden at runtime which might be even more flexible than fixing them at build time.

Dockerfile.db
gdbelvin: This is the Docker equivalent of `./script/resetdb` since this particular mysql image runs all the sql scripts in `/docker-entrypoint-initdb.d` when it's being initialized.

gdbelvin:
PTAL

server/trillian_log_server/Dockerfile
RJPercival: The documentation suggests using both, by declaring an ARG first and then using its value to initialize the ENV. I'm fine with leaving it as it is though, just thought I'd highlight this option.

RJPercival:
To answer your question about Kubernetes, I'll use these Dockerfiles to create images that can be run by Kubernetes. Then I need to write some Kubernetes config files that tell it how to configure and run the Trillian server and signer, and where to find the MySQL server.

PR #610, opened 2017-05-12
MySQL-based quota.Manager implementation

codingllama:
Add a MySQL-based `quota.Manager` implementation that kicks in once `Unsequenced` goes beyond a certain number of rows.

### DIFF
--- a/integration/quota/quota_test.go
+++ b/integration/quota/quota_test.go
@@ -1,0 +1,165 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package quota
+
+import (
+	"context"
+	"fmt"
+	"hash"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/quota/mysql"
+	"github.com/google/trillian/server"
+	"github.com/google/trillian/server/admin"
+	"github.com/google/trillian/server/interceptor"
+	"github.com/google/trillian/storage/testonly"
+	"github.com/google/trillian/testonly/integration"
+	"github.com/google/trillian/trees"
+	"github.com/google/trillian/util"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestRateLimiting(t *testing.T) {
+	maxUnsequenced := 20
+
+	adminClient, logClient, closeFn, err := setupLogServer(maxUnsequenced)
+	if err != nil {
+		t.Fatalf("setupLogServer() returned err = %v", err)
+	}
+	defer closeFn()
+
+	ctx := context.Background()
+	tree, err := adminClient.CreateTree(ctx, &trillian.CreateTreeRequest{Tree: testonly.LogTree})
+	if err != nil {
+		t.Fatalf("CreateTree() returned err = %v", err)
+	}
+	hasherFn, err := trees.Hash(tree)
+	if err != nil {
+		t.Fatalf("Hash() returned err = %v", err)
+	}
+	hasher := hasherFn.New()
+	lw := &leafWriter{client: logClient, hash: hasher, treeID: tree.TreeId}
+
+	// Requests where leaves < maxUnsequenced should work
+	for i := 0; i < maxUnsequenced; i++ {
+		if err := lw.QueueLeaf(ctx); err != nil {
+			t.Errorf("QueueLeaf() returned err = %v", err)
+		}
+	}
+
+	// Some point after now requests should start to fail
+	stop := false
+	timeout := time.After(1 * time.Second)
+	for !stop {
+		select {
+		case <-timeout:
+			t.Error("Timed out before rate limiting kicked in")
+			stop = true
+		default:
+			err := lw.QueueLeaf(ctx)
+			if err == nil {
+				continue // Rate liming hasn't kicked in yet
+			}
+			if s, ok := status.FromError(err); !ok || s.Code() != codes.ResourceExhausted {
+				t.Fatalf("QueueLeaf() returned err = %v", err)
+			}
+			stop = true
+		}
+	}
+}
+
+func setupLogServer(maxUnsequenced int) (trillian.TrillianAdminClient, trillian.TrillianLogClient, func(), error) {
+	var lis net.Listener
+	var s *grpc.Server
+	var conn *grpc.ClientConn
+	closeFn := func() {
+		if conn != nil {
+			conn.Close()
+		}
+		if s != nil {
+			s.GracefulStop()
+		}
+		if lis != nil {
+			lis.Close()
+		}
+	}
+
+	registry, err := integration.NewRegistryForTests("RateLimitingTest")
+	if err != nil {
+		closeFn()
+		return nil, nil, nil, err
+	}
+
+	qm, ok := registry.QuotaManager.(*mysql.QuotaManager)
+	if !ok {
+		closeFn()
+		return nil, nil, nil, fmt.Errorf("unexpected QuotaManager type: %T", registry.QuotaManager)
+	}
+	qm.MaxUnsequencedRows = maxUnsequenced
+
+	intercept := &interceptor.TrillianInterceptor{
+		Admin:        registry.AdminStorage,
+		QuotaManager: registry.QuotaManager,
+	}
+	s = grpc.NewServer(grpc.UnaryInterceptor(interceptor.WrapErrors(intercept.UnaryInterceptor)))
+	trillian.RegisterTrillianAdminServer(s, admin.New(registry))
+	trillian.RegisterTrillianLogServer(s, server.NewTrillianLogRPCServer(registry, util.SystemTimeSource{}))
+
+	lis, err = net.Listen("tcp", ":0")
+	if err != nil {
+		closeFn()
+		return nil, nil, nil, err
+	}
+	go s.Serve(lis)
+
+	conn, err = grpc.Dial(lis.Addr().String(), grpc.WithInsecure())
+	if err != nil {
+		closeFn()
+		return nil, nil, nil, err
+	}
+	return trillian.NewTrillianAdminClient(conn), trillian.NewTrillianLogClient(conn), closeFn, nil
+}
+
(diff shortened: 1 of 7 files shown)


codingllama:
I have plans for a more sophisticated rate limiting mechanism, which could allow for other kinds of quotas (per user, per tree, readonly, so on), but that will take some time. The MySQL count-based strategy is easy enough to write and reason about, so here it is.

All "unit" tests pass if executed in "information_schema" mode, but it seemed prudent to code them a bit more defensively in case it's works-in-my-machine syndrome. The same goes for the integration test, but there's an additional intent in making it a bit more flexible, which is to make my life easier once I have the other rate limiting mechanism mentioned above (it's likely to behave a bit less predictably than a simple count).

#490

quota/mysql/mysql_quota.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return countFromInformationSchema(ctx, m.DB)
+}
+
+func countFromInformationSchema(ctx context.Context, db *sql.DB) (int, error) {
+	// information_schema.tables doesn't have an explicit PK, so let's play it safe and ensure
+	// the cursor returns a single row.
+	rows, err := db.QueryContext(
+		ctx,
+		`SELECT table_rows
```
Martin2112: Please pull out all SQL statements as constants.

quota/mysql/mysql_quota.go
```diff
@@ -0,0 +1,149 @@
...
+	// Note that this is a Global/Write quota suggestion, so it applies across trees.
+	DefaultMaxUnsequenced = 500000 // About 2h of non-stop signing at 70QPS.
+)
+
+// QuotaManager is a MySQL-based quota.Manager implementation.
+//
+// It has two working modes: one queries the information schema for the number of Unsequenced rows,
+// the other does a select count(*) on the Unsequenced table. Information schema queries are
+// default, even though they are approximate, as they're constant time (select count(*) on InnoDB
+// based MySQL needs to traverse the index and may take quite a while to complete).
```
Martin2112: I'm quite worried about this option given the length of time I've seen this type of query take. I wonder if we should have a table of counts by tree and update this using DB triggers.

codingllama: Yep, I've considered this option too, but I would prefer to avoid a single point of contact if possible.

The `count(*)` mode is present mostly so we have fully deterministic tests. As I stated in a comment above, all tests *seem* to run fine on the information_schema mode, but I'm afraid they might get flaky in other environments (other people's machines, Travis, so on). I could completely remove the `count(*)` mode, if you prefer, in which case we would deal with flakiness if/when it happens.

quota/mysql/mysql_quota.go
codingllama: Any particular reason for that? I usually find that moving queries to constants hinders readability, specially if the queries aren't being reused. It just forces us to scroll back and forth.

quota/mysql/mysql_quota.go
Martin2112: I think we can leave it for now but I'm slightly worried that we have two mechanisms that could go wrong in different ways outside our control.

quota/mysql/mysql_quota.go
Martin2112: Mostly the work on supporting multiple databases. The code for pgsql is about 95% the same but not the SQL.

codingllama: Done. Let me know if you'd like to me to refactor it to be more cross-DB friendly (in a follow up PR).

quota/mysql/mysql_quota.go
codingllama: Ack.

codingllama:
PTAL?

codingllama:
Friendly ping?

quota/mysql/mysql_quota.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, spec := range specs {
+		if spec.Group != quota.Global || spec.Kind != quota.Write {
+			continue
+		}
+		// Only allow global writes if Unsequenced is under the expected limit
+		count, err := m.countUnsequenced(ctx)
+		if err != nil {
+			return err
+		}
+		if count+numTokens > m.MaxUnsequencedRows {
```
Martin2112: I think this can go wrong as closely made N requests for M using the count won't be treated as involving N*M tokens. Not absolutely sure if this is a real problem though. It could be avoided by maintaining a count though that's got it's own problems.

Also we're counting the unsequenced table every time? I think that's going to cause a problem unless we don't expect people to use that scheme in production.

quota/mysql/mysql_quota.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return countFromInformationSchema(ctx, m.DB)
+}
+
+func countFromInformationSchema(ctx context.Context, db *sql.DB) (int, error) {
+	// information_schema.tables doesn't have an explicit PK, so let's play it safe and ensure
+	// the cursor returns a single row.
+	rows, err := db.QueryContext(ctx, countFromInformationSchemaQuery, "Unsequenced", "BASE TABLE")
+	if err != nil {
+		return -1, err
```
Martin2112: These should maybe return 0, err unless -1 has some clear benefit.

quota/mysql/mysql_quota.go
```diff
@@ -0,0 +1,149 @@
...
+	if rows.Next() {
+		return -1, errors.New("too many rows returned from information_schema query")
+	}
+	return count, nil
+}
+
+func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
+	var count int
+	if err := db.QueryRowContext(ctx, countFromUnsequencedQuery).Scan(&count); err != nil {
+		return -1, err
```
Martin2112: and here.

quota/mysql/mysql_quota_test.go
```diff
@@ -0,0 +1,364 @@
...
+		tree, err := createTree(ctx, db)
+		if err != nil {
+			t.Errorf("%v: createTree() returned err = %v", desc, err)
+			continue
+		}
+
+		qm := &mysqlq.QuotaManager{DB: db, MaxUnsequencedRows: maxUnsequenced, UseSelectCount: test.useSelectCount}
+
+		// All GetTokens() calls where leaves < maxUnsequenced should succeed:
+		// information_schema may be outdated, but it should still point to a valid point in
```
Martin2112: "point to a valid point in the past" reads a bit unclear, please reword. We also don't know how out of date this could be in practice.

quota/mysql/mysql_quota_test.go
```diff
@@ -0,0 +1,364 @@
...
+		// Allow some time for information_schema to "catch up".
+		stop := false
+		timeout := time.After(1 * time.Second)
+		for !stop {
+			select {
+			case <-timeout:
+				t.Errorf("%v: Timed out", desc)
+				stop = true
+			default:
+				// An error means that GetTokens is working correctly
```
Martin2112: Is it safe to test for any error type?

server/trillian_log_server/main.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/extension"
 	"github.com/google/trillian/monitoring"
-	"github.com/google/trillian/quota"
+	mysqlq "github.com/google/trillian/quota/mysql"
```
Martin2112: This may be a bad idea and might not help but could we move the mysql query of the quota manager out to storage/mysql? or storage/mysql/quota?

integration/quota/quota_test.go
```diff
@@ -0,0 +1,165 @@
...
+	// Requests where leaves < maxUnsequenced should work
+	for i := 0; i < maxUnsequenced; i++ {
+		if err := lw.QueueLeaf(ctx); err != nil {
+			t.Errorf("QueueLeaf() returned err = %v", err)
+		}
+	}
+
+	// Some point after now requests should start to fail
+	stop := false
+	timeout := time.After(1 * time.Second)
```
codingllama: I ran it a few hundred times and didn't have it flake. The quota system should begin to block requests *soon-ish*, but I didn't want to assume a specific number, in particular because information_schema is an estimate (although it seems to be pretty precise for small numbers). I'm open to suggestions if you have any.

quota/mysql/mysql_quota.go
codingllama: It isn't meant to be an exact quota, so we may go a few requests above the limit if they are close in time and close to the blocking threshold. I think that's fine, as there has to be some safety margin in the chosen limit anyway, plus further requests are likely to be blocked.

I don't expect the count(*) strategy to be applied in prod. It's not the default either. It's here so it's easier to reason about QuotaManager's behavior in some tests (as counts are deterministic).

quota/mysql/mysql_quota.go
codingllama: Done (throughout the PR).

There's no particular meaning to the returned count if `err != nil`.

quota/mysql/mysql_quota.go
codingllama: Done.

quota/mysql/mysql_quota_test.go
codingllama: Rephrased.

I spent a happy afternoon playing with this, my verdict is that it is "close enough". The assumption here is that if sequencing stops, or falls short in face of the received write QPS, the actual count will keep increasing, which will be reflected in information_schema. So, once we get in the ballpark of the defined threshold, we start blocking. It's not precise, but it ought to be reasonable enough that the table won't keep growing wildly past the limit.

quota/mysql/mysql_quota_test.go
codingllama: Good call, changed to test for the appropriate error.

server/trillian_log_server/main.go
codingllama: I'd rather stay clear of storage/mysql - there's too much code in that package already. That, along with Go's default package visibility, means things are getting a bit intertwined.

We could move to storage/mysql/quota, but I'm not sure exactly how you want it to be done (or why). What do you have a mind, a file with query constants? Also, if we're not sure about this, could we leave it to a separate PR? If we move the queries out it might make sense to rename some things (`sql.QuotaManager` instead of `mysql.QuotaManager`, for example).

codingllama:
PTAL?

integration/quota/quota_test.go
Martin2112: Can we use a fake time source for the test?

quota/mysql/mysql_quota.go
Martin2112: If it's not for production use it's not as important. It might not be a few requests over the limit e.g. if 25 people ask for 1000 at the same time. Might be worth bounding how many tokens can be taken at once.

server/trillian_log_server/main.go
Martin2112: There could be advantages in having all the mysql code together. As we currently only have one implementation we can probably worry about this later.

codingllama: OK. I'm happy to do any changes required for multi-dialect support, just let me know what/when you need it.

codingllama:
Thanks, Martin.

PR #612, opened 2017-05-12
Send PEM files to server when provisioning a log. 

gdbelvin:
The admin client currently only supports provisioning logs and maps with keys that are locally on disk.  Since the files being supplied by the admin client may not already be on the server's local disk, this presents a bit of a tricky puzzle. 

This PR sends the contents of the private key and the server stores the key directly in the admin db. 

This PR also removes the requirement for PEM files to have passwords since the server does not have this requirement.

### DIFF
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -43,6 +43,7 @@ import (
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/google/trillian"
 	"github.com/google/trillian/cmd"
+	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/crypto/sigpb"
 	"google.golang.org/grpc"
@@ -59,7 +60,7 @@ var (
 	displayName        = flag.String("display_name", "", "Display name of the new tree")
 	description        = flag.String("description", "", "Description of the new tree")
 
-	privateKeyFormat = flag.String("private_key_format", "PEMKeyFile", "Type of private key to be used")
+	privateKeyFormat = flag.String("private_key_format", "PrivateKey", "Type of private key to be used")
 	pemKeyPath       = flag.String("pem_key_path", "", "Path to the private key PEM file")
 	pemKeyPassword   = flag.String("pem_key_password", "", "Password of the private key PEM file")
 
@@ -145,7 +146,7 @@ func newPK(opts *createOpts) (*any.Any, error) {
 	switch opts.privateKeyType {
 	case "PEMKeyFile":
 		if opts.pemKeyPath == "" {
-			return nil, errors.New("empty PEM path")
+			return nil, errors.New("empty pem_key_path")
 		}
 		if opts.pemKeyPass == "" {
 			return nil, fmt.Errorf("empty password for PEM key file %q", opts.pemKeyPath)
@@ -155,6 +156,20 @@ func newPK(opts *createOpts) (*any.Any, error) {
 			Password: opts.pemKeyPass,
 		}
 		return ptypes.MarshalAny(pemKey)
+	case "PrivateKey":
+		if opts.pemKeyPath == "" {
+			return nil, errors.New("empty pem_key_path")
+		}
+		pemSigner, err := keys.NewFromPrivatePEMFile(
+			opts.pemKeyPath, opts.pemKeyPass)
+		if err != nil {
+			return nil, err
+		}
+		der, err := keys.MarshalPrivateKey(pemSigner)
+		if err != nil {
+			return nil, err
+		}
+		return ptypes.MarshalAny(&keyspb.PrivateKey{Der: der})
 	default:
 		return nil, fmt.Errorf("unknown private key type: %v", opts.privateKeyType)
 	}
--- a/cmd/createtree/main_test.go
+++ b/cmd/createtree/main_test.go
@@ -19,9 +19,12 @@ import (
 	"net"
 	"testing"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/crypto/sigpb"
 	"github.com/kylelemons/godebug/pretty"
@@ -29,14 +32,23 @@ import (
 	"google.golang.org/grpc"
 )
 
-func TestRun(t *testing.T) {
-	pemKey := &keyspb.PEMKeyFile{
-		Path:     "../../testdata/log-rpc-server.privkey.pem",
-		Password: "towel",
-	}
-	anyKey, err := ptypes.MarshalAny(pemKey)
+func marshalAny(p proto.Message) *any.Any {
+	anyKey, err := ptypes.MarshalAny(p)
 	if err != nil {
-		t.Fatalf("Can't marshall pemKey: %v", err)
+		panic(err)
+	}
+	return anyKey
+}
+
+func TestRun(t *testing.T) {
+	pemPath, pemPassword := "../../testdata/log-rpc-server.privkey.pem", "towel"
+	pemSigner, err := keys.NewFromPrivatePEMFile(pemPath, pemPassword)
+	if err != nil {
+		t.Fatalf("NewFromPrivatPEM(): %v", err)
+	}
+	pemDer, err := keys.MarshalPrivateKey(pemSigner)
+	if err != nil {
+		t.Fatalf("MashalPrivateKey(): %v", err)
 	}
 
 	// defaultTree reflects all flag defaults with the addition of a valid pk
@@ -46,7 +58,7 @@ func TestRun(t *testing.T) {
 		HashStrategy:       trillian.HashStrategy_RFC_6962,
 		HashAlgorithm:      sigpb.DigitallySigned_SHA256,
 		SignatureAlgorithm: sigpb.DigitallySigned_RSA,
-		PrivateKey:         anyKey,
+		PrivateKey:         marshalAny(&keyspb.PrivateKey{Der: pemDer}),
 	}
 
 	server, lis, stopFn, err := startFakeServer()
@@ -57,8 +69,8 @@ func TestRun(t *testing.T) {
 
 	validOpts := newOptsFromFlags()
 	validOpts.addr = lis.Addr().String()
-	validOpts.pemKeyPath = pemKey.Path
-	validOpts.pemKeyPass = pemKey.Password
+	validOpts.pemKeyPath = pemPath
+	validOpts.pemKeyPass = pemPassword
 
 	nonDefaultTree := *defaultTree
 	nonDefaultTree.TreeType = trillian.TreeType_MAP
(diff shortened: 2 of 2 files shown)


cmd/createtree/main.go
```diff
@@ -141,16 +142,28 @@ func newPK(opts *createOpts) (*any.Any, error) {
 	switch opts.privateKeyType {
 	case "PEMKeyFile":
 		if opts.pemKeyPath == "" {
-			return nil, errors.New("empty PEM path")
```
codingllama: Please keep the checks - they are useful for tool users (and slightly more readable than the resulting server-side error message they would get).

cmd/createtree/main.go
```diff
@@ -141,16 +142,28 @@ func newPK(opts *createOpts) (*any.Any, error) {
...
 			Path:     opts.pemKeyPath,
 			Password: opts.pemKeyPass,
 		}
 		return ptypes.MarshalAny(pemKey)
+	case "PrivateKey":
+		if opts.pemKeyPath == "" {
+			return nil, errors.New("empty pem_key_path")
+		}
+		pemSigner, err := keys.NewFromPrivatePEMFile(
+			opts.pemKeyPath, opts.pemKeyPass)
```
codingllama: It seems conceptually strange to me to force a file to exist in this case. We could allow for a literal key to be passed via a flag. That possibly means the key gets logged on shell history, unless the user is careful, so there's that as well. I'm not too sure TBH.

WDYT?

crypto/keys/pem_signer_factory.go
```diff
@@ -72,10 +71,11 @@ func (f PEMSignerFactory) Generate(ctx context.Context, tree *trillian.Tree, spe
 		return nil, fmt.Errorf("Specification produces %T, but tree.SignatureAlgorithm = %v", key, tree.GetSignatureAlgorithm())
 	}
 
-	return marshalPrivateKey(key)
+	return MarshalPrivateKey(key)
 }
 
-func marshalPrivateKey(key crypto.Signer) (*any.Any, error) {
+// MarshalPrivateKey returns an any.Any with a keyspb.PrivateKey inside.
```
codingllama: I think this would be better placed at `private.go` now, considering that we've just made it public. I would also prefer for it to have its own set of tests.

cmd/createtree/main.go
```diff
@@ -141,16 +142,28 @@ func newPK(opts *createOpts) (*any.Any, error) {
...
 		pemKey := &keyspb.PEMKeyFile{
 			Path:     opts.pemKeyPath,
 			Password: opts.pemKeyPass,
 		}
 		return ptypes.MarshalAny(pemKey)
+	case "PrivateKey":
+		if opts.pemKeyPath == "" {
+			return nil, errors.New("empty pem_key_path")
+		}
+		pemSigner, err := keys.NewFromPrivatePEMFile(
```
codingllama: Please add a `pemKeyPass == ""` check here too, as above. (Reasoning in another comment.)

cmd/createtree/main_test.go
```diff
@@ -87,6 +102,9 @@ func TestRun(t *testing.T) {
 	emptyPEMPass := *validOpts
 	emptyPEMPass.pemKeyPass = ""
 
+	pemkey := *validOpts
```
codingllama: nit: pemKey

crypto/keys/private.go
```diff
@@ -56,9 +56,6 @@ type SignerFactory interface {
 // NewFromPrivatePEMFile reads a PEM-encoded private key from a file.
 // The key must be protected by a password.
 func NewFromPrivatePEMFile(keyFile, keyPassword string) (crypto.Signer, error) {
-	if keyPassword == "" {
```
codingllama: Please don't remove this check, we've added it on purpose. The intention is to foster good practices along with Trillian usage. People really shouldn't be storing unencrypted PEM files in their servers.

There's some context at #544. If you feel strongly about this, please move the discussion to a separate PR.

cmd/createtree/main.go
```diff
@@ -141,16 +142,28 @@ func newPK(opts *createOpts) (*any.Any, error) {
...
-		if opts.pemKeyPass == "" {
-			return nil, fmt.Errorf("empty password for PEM key file %q", opts.pemKeyPath)
+			return nil, errors.New("empty pem_key_path")
 		}
 		pemKey := &keyspb.PEMKeyFile{
 			Path:     opts.pemKeyPath,
 			Password: opts.pemKeyPass,
 		}
 		return ptypes.MarshalAny(pemKey)
+	case "PrivateKey":
```
codingllama: Would mind making this the default (line 61)? @RJPercival and I discussed this a bit, and we both seem to prefer it slightly over PEMKeyFile. (I'm guessing you do too.)

cmd/createtree/main.go
gdbelvin: adding checks back in

cmd/createtree/main.go
gdbelvin: done

cmd/createtree/main.go
gdbelvin: done

cmd/createtree/main.go
gdbelvin: I think it's probably better to focus on generating the key server side.

cmd/createtree/main_test.go
gdbelvin: done

crypto/keys/pem_signer_factory.go
gdbelvin: moved

crypto/keys/private.go
gdbelvin: added back

gdbelvin:
@codingllama or @RJPercival Would you mind reviewing?

crypto/keys/keys_test.go
```diff
@@ -324,3 +324,25 @@ func TestGenerateKey(t *testing.T) {
 		}
 	}
 }
+
+func TestMarshalPrivateKey(t *testing.T) {
+	for _, tc := range []struct {
+		pem     string
+		typeURL string
+	}{
+		{ecdsaPrivateKey, "type.googleapis.com/keyspb.PrivateKey"},
```
codingllama: nit: Please use field keys, it makes future changes less painful.

`{pem: ecdsaPrivateKey, typeURL: "..."},`

crypto/keys/keys_test.go
```diff
@@ -324,3 +324,25 @@ func TestGenerateKey(t *testing.T) {
...
+		{rsaPrivateKey, "type.googleapis.com/keyspb.PrivateKey"},
+	} {
+		k, err := NewFromPrivatePEM(tc.pem, "")
+		if err != nil {
+			t.Errorf("NewFromPrivatePEM(%v): %v", tc.pem, err)
```
codingllama: Add `continue` after t.Errorf, otherwise it'll probably panic on failed tests. Ditto for others.

crypto/keys/keys_test.go
```diff
@@ -324,3 +324,25 @@ func TestGenerateKey(t *testing.T) {
...
+		}
+		a, err := MarshalPrivateKey(k)
+		if err != nil {
+			t.Errorf("MarshalPrivateKey(%v): %v", k, err)
+		}
+		if got, want := a.TypeUrl, tc.typeURL; got != want {
```
codingllama: We should check if the DER on the PrivateKey proto is correct, otherwise the test isn't adding much.

crypto/keys/private_keys.go
```diff
@@ -159,3 +160,23 @@ func curveFromParams(params *keyspb.Specification_ECDSA) elliptic.Curve {
...
+
+	switch key := key.(type) {
+	case *ecdsa.PrivateKey:
+		privateKeyDER, err = x509.MarshalECPrivateKey(key)
+	case *rsa.PrivateKey:
+		privateKeyDER = x509.MarshalPKCS1PrivateKey(key)
+	default:
+		return nil, fmt.Errorf("unsupported key type: %T", key)
+	}
+	if err != nil {
```
codingllama: nit: Move `err`'s declaration and the if inside the ecdsa case? Neither need to be at this scope.

cmd/createtree/main_test.go
```diff
@@ -134,6 +149,32 @@ func TestRun(t *testing.T) {
 			desc:    "emptyPEMPass",
 			opts:    &emptyPEMPass,
 			wantErr: true,
+			wantTree: &trillian.Tree{
```
gdbelvin: done

crypto/keys/keys_test.go
gdbelvin: done

crypto/keys/keys_test.go
gdbelvin: done

crypto/keys/keys_test.go
gdbelvin: I added a check to make sure the type contained in the DER was correct. I hope that's close enough?

crypto/keys/private_keys.go
gdbelvin: done

crypto/keys/keys_test.go
gdbelvin: Actually, a rebase removed the need for this method.

gdbelvin:
A rebase simplified this PR. PTAL

cmd/createtree/main.go
RJPercival: I think it's best to limit it to files since, as you point out, there are risks to allowing a private key to appear on the command-line. Besides, if someone doesn't want it to touch disk, they can do something like:
```shell
openssl ecparam -name secp256k1 -genkey -noout | createtree --pem_key_path /dev/stdin --pem_key_password "foo"
```

cmd/createtree/main_test.go
```diff
@@ -46,7 +58,7 @@ func TestRun(t *testing.T) {
 		HashStrategy:       trillian.HashStrategy_RFC_6962,
 		HashAlgorithm:      sigpb.DigitallySigned_SHA256,
 		SignatureAlgorithm: sigpb.DigitallySigned_RSA,
-		PrivateKey:         anyKey,
+		PrivateKey:         marshalAny(&keyspb.PrivateKey{Der: pemDer}),
```
RJPercival: nit: It might be nicer to still do the marshaling at the start of the test, so you can report an error using `t.Fatalf()`.

cmd/createtree/main_test.go
```diff
@@ -136,6 +151,22 @@ func TestRun(t *testing.T) {
 			wantErr: true,
 		},
 		{
+			desc:    "PEMKeyFile",
+			opts:    &pemKeyOpts,
+			wantErr: false,
+			wantTree: &trillian.Tree{
```
RJPercival: nit: Making this tree at the start of the test might be nicer, in the same way that nonDefaultTree was created. Then you could get rid of `marshalAny()` as well.

cmd/createtree/main.go
RJPercival: Ah actually, that wouldn't work, because we require that the private key has a password. There's bound to be an openssl command to put in the middle that will (somewhat pointlessly) encrypt the key though.

cmd/createtree/main_test.go
gdbelvin: `marshalAny` is used twice.  I could remove it and do all of the marshaling before the test to take advantage of `t.Fatalf`, but my thought was that the values we are marshaling are test configuration values which won't change as a result of code, and that it was nice to be able to use the output of `marshalAny` directly in building the table driven tests.

cmd/createtree/main_test.go
gdbelvin: Good idea. done.

PR #613, opened 2017-05-12
Return empty list when no leaves are requested.

gdbelvin:
The map server currently panics if no map leaves are requested.
This PR adds a test for this and returns an empty list.

### DIFF
--- a/integration/map.go
+++ b/integration/map.go
@@ -116,8 +116,37 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
 			}
 			leafHash := h.HashLeaf(leaf.LeafValue)
 			if err := merkle.VerifyMapInclusionProof(leaf.Index, leafHash, latestRoot.RootHash, incl.Inclusion, h); err != nil {
-				return fmt.Errorf("inclusion proof failed to verify for key %s: %v", incl.Leaf.Index, err)
+				return fmt.Errorf("verifyMapInclusionProof(%x): %v", leaf.Index, err)
 			}
+		}
+	}
+	if err := testForNonExistentLeaf(ctx, mapID, client, latestRoot); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Ensure that a query for a leaf that does not exist results in a valid inclusion proof.
+func testForNonExistentLeaf(ctx context.Context, mapID int64,
+	client trillian.TrillianMapClient, latestRoot trillian.SignedMapRoot) error {
+	h := merkle.NewMapHasher(testonly.Hasher)
+	index1 := []byte("doesnotexist....................")
+	r, err := client.GetLeaves(ctx, &trillian.GetMapLeavesRequest{
+		MapId:    mapID,
+		Revision: latestRoot.MapRevision,
+		Index:    [][]byte{index1},
+	})
+	if err != nil {
+		return fmt.Errorf("GetLeaves(%s): %v", index1, err)
+	}
+	for _, incl := range r.MapLeafInclusion {
+		leaf := incl.Leaf
+		if got, want := len(leaf.LeafValue), 0; got != want {
+			return fmt.Errorf("len(GetLeaves(%s).LeafValue): %v, want, %v", index1, got, want)
+		}
+		leafHash := h.HashLeaf(leaf.LeafValue)
+		if err := merkle.VerifyMapInclusionProof(leaf.Index, leafHash, latestRoot.RootHash, incl.Inclusion, h); err != nil {
+			return fmt.Errorf("VerifyMapInclusionProof(%x): %v", leaf.Index, err)
 		}
 	}
 	return nil
--- a/server/vmap/trillian_map_server.go
+++ b/server/vmap/trillian_map_server.go
@@ -17,17 +17,26 @@ package vmap
 import (
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/google/trillian"
 	spb "github.com/google/trillian/crypto/sigpb"
 	"github.com/google/trillian/extension"
 	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/trees"
+
+	"github.com/golang/glog"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
-// TODO: There is no access control in the server yet and clients could easily modify
+const (
+	// indexSize is the number of bytes all indexes should be.
+	// TODO(gdbelvin): specify the index length in the tree specification.
+	indexSize = 32
+)
+
+// TODO(codingllama): There is no access control in the server yet and clients could easily modify
 // any tree.
 
 // TrillianMapServer implements the RPC API defined in the proto
@@ -45,7 +54,9 @@ func (t *TrillianMapServer) IsHealthy() error {
 	return t.registry.MapStorage.CheckDatabaseAccessible(context.Background())
 }
 
-// GetLeaves implements the GetLeaves RPC method.
+// GetLeaves implements the GetLeaves RPC method.  Each requested index will
+// return an inclusion proof to either the leaf, or nil if the leaf does not
+// exist.
 func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapLeavesRequest) (*trillian.GetMapLeavesResponse, error) {
 	mapID := req.MapId
 
@@ -74,33 +85,51 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
 
 	smtReader := merkle.NewSparseMerkleTreeReader(req.Revision, hasher, tx)
 
-	leaves, err := tx.Get(ctx, req.Revision, req.Index)
-	if err != nil {
-		return nil, err
-	}
-	glog.Infof("%v: wanted %v leaves, found %v", mapID, len(req.Index), len(leaves))
-
-	resp := &trillian.GetMapLeavesResponse{
-		MapLeafInclusion: make([]*trillian.MapLeafInclusion, len(leaves)),
-	}
-	for i, leaf := range leaves {
-		proof, err := smtReader.InclusionProof(ctx, req.Revision, leaf.Index)
+	inclusions := make([]*trillian.MapLeafInclusion, 0, len(req.Index))
+	found := 0
+	for _, index := range req.Index {
+		// TODO(gdbelvin): specify the index length in the tree specification.
+		if got, want := len(index), indexSize; got != want {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"index len(%x): %v, want %v", index, got, want)
+		}
+		// Fetch the leaf if it exists.
+		leaves, err := tx.Get(ctx, req.Revision, [][]byte{index})
 		if err != nil {
 			return nil, err
 		}
-		// Copy the leaf from the iterator, which gets overwritten
-		value := leaf
-		resp.MapLeafInclusion[i] = &trillian.MapLeafInclusion{
-			Leaf:      &value,
-			Inclusion: proof,
+		var leaf *trillian.MapLeaf
+		if len(leaves) == 1 {
+			leaf = &leaves[0]
+			found++
+		} else {
+			// Empty leaf for proof of non-existence.
+			leaf = &trillian.MapLeaf{
+				Index:     index,
+				LeafValue: nil,
+			}
 		}
+
+		// Fetch the proof regardless of whether the leaf exists.
+		proof, err := smtReader.InclusionProof(ctx, req.Revision, index)
+		if err != nil {
+			return nil, err
+		}
+
+		inclusions = append(inclusions, &trillian.MapLeafInclusion{
+			Leaf:      leaf,
+			Inclusion: proof,
+		})
 	}
+	glog.Infof("%v: wanted %v leaves, found %v", mapID, len(req.Index), found)
 
 	if err := tx.Commit(); err != nil {
 		return nil, err
 	}
 
-	return resp, nil
+	return &trillian.GetMapLeavesResponse{
+		MapLeafInclusion: inclusions,
+	}, nil
 }
 
 // SetLeaves implements the SetLeaves RPC method.
@@ -131,7 +160,10 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	}
 
 	for _, l := range req.Leaves {
-		// TODO(gbelvin) Verify that Index is of the proper length.
+		if got, want := len(l.Index), indexSize; got != want {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"len(%x): %v, want %v", l.Index, got, want)
+		}
 		// TODO(gbelvin) use LeafHash rather than computing here.
 		l.LeafHash = hasher.HashLeaf(l.LeafValue)
 
--- a/storage/mysql/map_storage.go
+++ b/storage/mysql/map_storage.go
@@ -179,9 +179,14 @@ func (m *mapTreeTX) Set(ctx context.Context, keyHash []byte, value trillian.MapL
 	return err
 }
 
-// MapLeaf indexes are overwritten rather than returning the MapLeaf proto provided in Set.
-// TODO: return a map[_something_]Mapleaf or []IndexValue to separate the index from the value.
+// Get returns a list of map leaves indicated by indexes.
+// If an index is not found, no corresponding entry is returned.
+// Each MapLeaf.Index is overwritten with the index the leaf was found at.
 func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) ([]trillian.MapLeaf, error) {
+	// If no indexes are requested, return an empty set.
+	if len(indexes) == 0 {
+		return []trillian.MapLeaf{}, nil
+	}
 	stmt, err := m.ms.getStmt(ctx, selectMapLeafSQL, len(indexes), "?", "?")
 	if err != nil {
 		return nil, err
@@ -195,8 +200,6 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
 	}
 	args = append(args, m.treeID)
 	args = append(args, revision)
-
-	glog.Infof("args size %d", len(args))
 
 	rows, err := stx.QueryContext(ctx, args...)
 	// It's possible there are no values for any of these keys yet
@@ -230,7 +233,6 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
 		ret = append(ret, mapLeaf)
 		nr++
 	}
-	glog.Infof("%d rows, %d empty", nr, er)
 	return ret, nil
 }
 
--- a/storage/mysql/map_storage_test.go
+++ b/storage/mysql/map_storage_test.go
@@ -253,22 +253,30 @@ func TestMapSetSameKeyInSameRevisionFails(t *testing.T) {
 	}
 }
 
-func TestMapGetUnknownKey(t *testing.T) {
+func TestMapGet0Results(t *testing.T) {
 	cleanTestDB(DB)
 	mapID := createMapForTests(DB)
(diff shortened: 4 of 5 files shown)


storage/mysql/map_storage_test.go
```diff
@@ -190,6 +190,26 @@ var mapLeaf = trillian.MapLeaf{
...
+	s := NewMapStorage(DB)
+
+	readRev := int64(1)
+	ctx := context.Background()
+	tx := beginMapTx(ctx, s, mapID, t)
+	defer tx.Close()
+	leaves := [][]byte{}
+	readValues, err := tx.Get(ctx, readRev, leaves)
+	if err != nil {
+		t.Fatalf("Failed to get %v:  %v", leaves)
```
codingllama: Missing `, err)` at the end?

storage/mysql/map_storage_test.go
```diff
@@ -190,6 +190,26 @@ var mapLeaf = trillian.MapLeaf{
 	ExtraData: []byte("Some Extra Data"),
 }
 
+func TestMapEmptyGet(t *testing.T) {
```
codingllama: Would you mind combining this and `TestMapGetUnknownKey` in single a table-driven test method? I've had a quick look and they seem close enough that it shouldn't cause you too much trouble.

storage/mysql/map_storage_test.go
gdbelvin: done, merged with test below

storage/mysql/map_storage_test.go
gdbelvin: done

gdbelvin:
Just thinking about this further, requesting 0 leaves and requesting a leaf that doesn't exist should actually be different. 
- 0 leaves should return 0 results. 
- Requesting a leaf that doesn't exist should return a proof down to a nil node, so the client can verify that the server isn't hiding a result that does exist.

storage/mysql/map_storage_test.go
```diff
@@ -253,22 +253,29 @@ func TestMapSetSameKeyInSameRevisionFails(t *testing.T) {
...
-	readValues, err := tx.Get(ctx, 1, [][]byte{[]byte("This doesn't exist.")})
-	if err != nil {
-		t.Fatalf("Read returned error %v", err)
-	}
-	if got, want := len(readValues), 0; got != want {
-		t.Fatalf("Unexpectedly read %d values, expected %d", got, want)
+	for _, tc := range []struct {
+		index [][]byte
+	}{
+		{index: [][]byte{}}, //empty list.
```
codingllama: Suggestion: Use `nil` instead of `[][]byte{}`?

server/vmap/trillian_map_server.go
```diff
@@ -74,24 +74,33 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
 
 	smtReader := merkle.NewSparseMerkleTreeReader(req.Revision, hasher, tx)
 
-	leaves, err := tx.Get(ctx, req.Revision, req.Index)
-	if err != nil {
-		return nil, err
-	}
-	glog.Infof("%v: wanted %v leaves, found %v", mapID, len(req.Index), len(leaves))
+	// TODO: we need to return one response per request.
```
codingllama: Please use the `TODO(user): ` format.

server/vmap/trillian_map_server.go
```diff
@@ -74,24 +74,33 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
-		// Copy the leaf from the iterator, which gets overwritten
-		value := leaf
+
+		// Create placeholder leaf for non-existant leaves.
+		leaf := &trillian.MerkleLeaf{
+			Index: index,
+		}
+
+		// Fetch the leaf if it exists.
+		leafs, err := tx.Get(ctx, req.Revision, [][]byte{index})
```
codingllama: nit: s/leafs/leaves/

server/vmap/trillian_map_server.go
```diff
@@ -74,24 +74,33 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
+		if err != nil {
+			return nil, err
+		}
+		for l := range leafs {
```
codingllama: nit: I would prefer an if to a for here, so it's clearer we expect/want at most a size 1 slice.

```go
if len(leaves) == 1 {
  leaf = leaves[0]
}
```

server/vmap/trillian_map_server.go
```diff
@@ -74,24 +74,33 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
+			leaf = l
+			break
+		}
+
```
codingllama: I've given this a bit more thought and I think the corner cases we're handling here actually merit error responses, as they're likely to stem from bugs / mistaken requests.

E.g.:
* `len(req.Index) == 0` -> `InvalidArgument` (no sense in asking for zero leaves)
* `len(index) == 0` -> `InvalidArgument` (no sense in asking for a leaf with an empty index)
* leaf not in storage -> `NotFound` (unambiguous response, unlike a empty-but-valid proof)

(See [status.Errorf](https://godoc.org/google.golang.org/grpc/status#Errorf).)

server/vmap/trillian_map_server.go
gdbelvin: Removed the comment and added better documentation on the method function.

server/vmap/trillian_map_server.go
gdbelvin: I keep doing that! - fixed.

server/vmap/trillian_map_server.go
gdbelvin: done

server/vmap/trillian_map_server.go
gdbelvin: This might deserve a GVC discussion, but because this is a crypto API for an untrusted server, the behavior might need to diverge slightly from what a typical API might do. 

- Leaf not in storage needs to return a proof of non-inclusion, unlike typical APIs. Suppose the server has updated a leaf and wants to hide it from a particular user. The server could return `NotFound` instead of the real value. To prevent this, the server must return proofs, even for `NotFound`. 
- `len(req.Index) == 0` is a syntactically valid request. Treating len == 0 as a special case means clients need to special case their logic as well, and special cases are ugly. 
- `len(index) == 0 != tree.IndexSize` is certainly an error case.

storage/mysql/map_storage_test.go
gdbelvin: done

server/vmap/trillian_map_server.go
gdbelvin: @AlCutter could you weigh in while @codingllama is OOO?

gdbelvin:
@AlCutter This PR is ready to review.

server/vmap/trillian_map_server.go
```diff
@@ -74,33 +78,47 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
-	leaves, err := tx.Get(ctx, req.Revision, req.Index)
-	if err != nil {
-		return nil, err
-	}
-	glog.Infof("%v: wanted %v leaves, found %v", mapID, len(req.Index), len(leaves))
+	inclusions := make([]*trillian.MapLeafInclusion, 0, len(req.Index))
+	found := 0
+	for _, index := range req.Index {
+		// TODO: specify the index length in the tree specification.
+		if got, want := len(index), 32; got != want {
```
daviddrysdale: (personal preference) as this isn't a test, personally I'd just do `if len(index) != indexSize`

server/vmap/trillian_map_server.go
```diff
@@ -74,33 +78,47 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
 			return nil, err
 		}
-		// Copy the leaf from the iterator, which gets overwritten
-		value := leaf
-		resp.MapLeafInclusion[i] = &trillian.MapLeafInclusion{
-			Leaf:      &value,
-			Inclusion: proof,
+		if len(leaves) == 1 {
+			leaf = &leaves[0]
+			found++
```
daviddrysdale: I don't really know the map stuff, so I found this a bit hard to follow (both before and after!).

Should (exists? leaf) == (exists? proof) ? If so, it might be nice to check for internal consistency (which I guess is `(len(leaves) == 0)` iff `(proof == nil)`)

server/vmap/trillian_map_server.go
```diff
@@ -74,33 +78,47 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
 
-	resp := &trillian.GetMapLeavesResponse{
-		MapLeafInclusion: make([]*trillian.MapLeafInclusion, len(leaves)),
-	}
-	for i, leaf := range leaves {
-		proof, err := smtReader.InclusionProof(ctx, req.Revision, leaf.Index)
+		// Create placeholder leaf for non-existent leaves.
+		leaf := &trillian.MapLeaf{
+			Index: index,
+		}
```
daviddrysdale: maybe shuffle this down to after the `tx.Get` block

integration/map.go
```diff
@@ -116,9 +116,34 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
 			}
 			leafHash := h.HashLeaf(leaf.LeafValue)
 			if err := merkle.VerifyMapInclusionProof(leaf.Index, leafHash, latestRoot.RootHash, incl.Inclusion, h); err != nil {
-				return fmt.Errorf("inclusion proof failed to verify for key %s: %v", incl.Leaf.Index, err)
+				return fmt.Errorf("VerifyMapInclusionProof(%x): %v", leaf.Index, err)
```
gdbelvin: Errors should be lower case in most cases, but this is a proper noun and I'm not sure it makes sense in this case.


The purpose of lower case error messages is to make messages of the form `log.Printf("Reading %s: %v", filename, err)` look sane.  In the case of proper nouns (the function name in this case), the lower case requirement doesn't make sense, and there's an exception for it in the [style guide](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

integration/map.go
```diff
@@ -116,9 +116,34 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
...
 		}
 	}
+	if err := testForNonExistantLeaf(ctx, mapID, client, latestRoot); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Ensure that a query for a leaf that does not exist results in a valid inclusion proof.
+func testForNonExistantLeaf(ctx context.Context, mapID int64,
```
gdbelvin: done

integration/map.go
```diff
@@ -116,9 +116,34 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
...
+	client trillian.TrillianMapClient, latestRoot trillian.SignedMapRoot) error {
+	h := merkle.NewMapHasher(testonly.Hasher)
+	r, err := client.GetLeaves(ctx, &trillian.GetMapLeavesRequest{
+		MapId:    mapID,
+		Revision: latestRoot.MapRevision,
+		Index:    [][]byte{[]byte("doesnotexist....................")},
+	})
+	if err != nil {
+		return err
```
gdbelvin: done

integration/map.go
```diff
@@ -116,9 +116,34 @@ func RunMapIntegration(ctx context.Context, mapID int64, client trillian.Trillia
...
+	}
+	for _, incl := range r.MapLeafInclusion {
+		leaf := incl.Leaf
+		leafHash := h.HashLeaf(leaf.LeafValue)
+		if err := merkle.VerifyMapInclusionProof(leaf.Index, leafHash, latestRoot.RootHash, incl.Inclusion, h); err != nil {
```
gdbelvin: I do.  `LeafValue` will be `nil` which makes the inclusion proof a "proof of empty". 
Just added a test for expecting `LeafValue` to be `nil`.

storage/mysql/map_storage_test.go
```diff
@@ -253,22 +253,29 @@ func TestMapSetSameKeyInSameRevisionFails(t *testing.T) {
...
+		index [][]byte
+	}{
+		{index: nil}, //empty list.
+		{index: [][]byte{[]byte("This doesn't exist.")}},
+	} {
+		tx := beginMapTx(ctx, s, mapID, t)
+		defer tx.Close()
+		readValues, err := tx.Get(ctx, 1, tc.index)
+		if err != nil {
+			t.Fatalf("Read returned error %v", err)
```
gdbelvin: done

storage/mysql/map_storage_test.go
```diff
@@ -253,22 +253,29 @@ func TestMapSetSameKeyInSameRevisionFails(t *testing.T) {
...
+		}
+		if got, want := len(readValues), 0; got != want {
+			t.Fatalf("Unexpectedly read %d values, expected %d", got, want)
```
gdbelvin: done

storage/mysql/map_storage.go
```diff
@@ -179,9 +179,14 @@ func (m *mapTreeTX) Set(ctx context.Context, keyHash []byte, value trillian.MapL
 	return err
 }
 
-// MapLeaf indexes are overwritten rather than returning the MapLeaf proto provided in Set.
-// TODO: return a map[_something_]Mapleaf or []IndexValue to separate the index from the value.
+// Get returns a list of map leafs indicated by indexes.
```
gdbelvin: done

server/vmap/trillian_map_server.go
```diff
@@ -46,6 +49,7 @@ func (t *TrillianMapServer) IsHealthy() error {
 }
 
 // GetLeaves implements the GetLeaves RPC method.
+// Each requested index will return an inclusion proof to either the leaf, or nil if the leaf does not exit.
```
gdbelvin: done. also added to the proto.

server/vmap/trillian_map_server.go
```diff
@@ -74,33 +78,47 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
 
-	leaves, err := tx.Get(ctx, req.Revision, req.Index)
-	if err != nil {
-		return nil, err
-	}
-	glog.Infof("%v: wanted %v leaves, found %v", mapID, len(req.Index), len(leaves))
+	inclusions := make([]*trillian.MapLeafInclusion, 0, len(req.Index))
+	found := 0
+	for _, index := range req.Index {
+		// TODO: specify the index length in the tree specification.
```
gdbelvin: done

server/vmap/trillian_map_server.go
gdbelvin: good point.  Constants are better. Thx

server/vmap/trillian_map_server.go
gdbelvin: done

server/vmap/trillian_map_server.go
gdbelvin: Added a few comments and adjusted the code flow for better readability. 

We return the leaf if it exists. If it doesn't exist we still return the index, but use nil for `LeafValue`.  Regardless of whether a leaf exists, we always return it's inclusion proof. If the leaf doesn't exist, this proof simply uses nil for the `LeafValue`.  

We can think of it like this: maps are a giant merkle tree, most of who's leaves are nil.  As time goes along, we replace these nil values with a proper leaf.

server/vmap/trillian_map_server.go
```diff
@@ -131,7 +149,11 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	}
 
 	for _, l := range req.Leaves {
-		// TODO(gbelvin) Verify that Index is of the proper length.
+		// TODO: specify the index length in the tree specification.
```
gdbelvin: done

server/vmap/trillian_map_server.go
```diff
@@ -131,7 +149,11 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		if got, want := len(l.Index), 32; got != want {
```
gdbelvin: done

server/vmap/trillian_map_server.go
```diff
@@ -131,7 +149,11 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+			return nil, status.Errorf(codes.InvalidArgument,
+				"len(%x): %v, want %v", l.Index, got, want)
```
gdbelvin: done

gdbelvin:
PTAL

daviddrysdale:
I only look at pull requests with a passing Travis build :-)

gdbelvin:
Fair enough :-)

gdbelvin:
Passing.

crypto/keyspb/keyspb.pb.go
```diff
@@ -303,7 +303,7 @@ func init() { proto.RegisterFile("keyspb.proto", fileDescriptor0) }
 
 var fileDescriptor0 = []byte{
 	// 306 bytes of a gzipped FileDescriptorProto
-	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x09, 0x6e, 0x88, 0x02, 0xff, 0x7c, 0x91, 0x4d, 0x4b, 0xf3, 0x40,
+	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0xff, 0x7c, 0x91, 0x4d, 0x4b, 0xf3, 0x40,
```
daviddrysdale: Why does this PR now include a change to keyspb.pb.go?

trillian_map_api.proto
```diff
@@ -77,6 +77,8 @@ message GetSignedMapRootResponse {
 // TrillianMap defines a service which provides access to a Verifiable Map as
 // defined in the Verifiable Data Structures paper.
 service TrillianMap {
+  // GetLeaves returns an inclusion proof for each index requested.
+  // for indexes that do not exist, the inclusion proof will use nil for the empty leaf value.
```
daviddrysdale: s/for/For/

server/vmap/trillian_map_server.go
daviddrysdale: I also meant that the whole `got, want` dance seems odd in non-test code.

server/vmap/trillian_map_server.go
```diff
@@ -74,33 +85,51 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
...
+				LeafValue: nil,
+			}
+		}
 
-	resp := &trillian.GetMapLeavesResponse{
-		MapLeafInclusion: make([]*trillian.MapLeafInclusion, len(leaves)),
-	}
-	for i, leaf := range leaves {
-		proof, err := smtReader.InclusionProof(ctx, req.Revision, leaf.Index)
+		// Fetch the proof regardless of whether the leaf exists.
```
daviddrysdale: maybe add " (if it doesn't, the proof will be nil)"

server/vmap/trillian_map_server.go
daviddrysdale: Thanks for the re-jigging -- it's much easier to follow now. 

I see that a non-existent leaf uses nil for the `LeafValue`, but what value does the `Inclusion` proof take?

crypto/keyspb/keyspb.pb.go
gdbelvin: I'll remove it

trillian_map_api.proto
gdbelvin: done

server/vmap/trillian_map_server.go
gdbelvin: If the leaf does not exist, the proof will still be a a full merkle tree inclusion path. It's just that the leaf value will be nil. Hashing that with all the neighbor values up to the root will produce the correct root hash.

daviddrysdale: Ah, OK, thanks for explaining.  I should really find out more about how the map works at some point...

PR #614, opened 2017-05-15
Remove liveness probe for Galera instances

RJPercival:
If an SST was underway to replicate the database state, the recipient Galera instance would appear to be dead, according to this probe, until the SST completed. This would result in Kubernetes killing the recipient.

TODO: Devise a better liveness probe, perhaps checking whether the file "/data/mysql/sst_in_progress" exists.

### DIFF
--- a/storage/mysql/kubernetes/galera.yaml
+++ b/storage/mysql/kubernetes/galera.yaml
@@ -86,8 +86,6 @@ spec:
           readinessProbe:
             exec:
               command: ["mysql", "-h", "127.0.0.1", "-u", "dummy", "-e", "SELECT 1"]
-            initialDelaySeconds: 5
-            timeoutSeconds: 1
           image: us.gcr.io/trillian-test/galera:experimental
           imagePullPolicy: Always
           name: mysql
--- a/storage/mysql/kubernetes/galera.yaml
+++ b/storage/mysql/kubernetes/galera.yaml
@@ -83,11 +83,6 @@ spec:
         - resources:
             limits:
               cpu: 2
-          livenessProbe:
-            exec:
-              command: ["mysqladmin", "-u", "dummy", "ping"]
-            initialDelaySeconds: 30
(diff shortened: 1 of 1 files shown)


RJPercival:
From reading the documentation, it looks like you need to be able to connect to the server to get the wsrep status. In this scenario, the server isn't accepting connections (that's why the liveness probe was failing).

Martin2112:
I thought the donor was up but in a desynced state. Aggregating the state
from the available nodes would maybe work. Suspiciously their monitoring
cluster status doc doesn't really cover SST monitoring.

http://galeracluster.com/documentation-webpages/monitoringthecluster.html

RJPercival:
The donor will be up but the liveness probe I've been referring to runs on the joiner, which is down.

Martin2112:
OK. If we were doing real monitoring we'd want to be collecting data from
all nodes to detect partitions etc. This should be OK for now as it won't
recover in the time this probe previously allowed.

RJPercival:
Something like [mysqld_exporter for Prometheus](https://github.com/prometheus/mysqld_exporter) might be a good way to monitor the cluster in the future. That allows Prometheus to collect the wsrep_* vars, amongst others, and then presumably alerts could be set up for when a partition forms.

Martin2112:
Yes, that sounds like a good plan.

gdbelvin:
Slightly unrelated issue, but I tried to use the containers in the file and didn't have permissions. 
Should they be public, or is there a special group that consumers of these images need to be part of?

RJPercival:
By default, the container image is uploaded to trillian-test's GKE container registry on GCP - you'll need to be made a member of trillian-test to access it. However, you could just upload the image elsewhere (the push.sh script has a "-t" flag) and tweak the galera.yaml file to use your image. You'd need to do this to run it on GCE under any other project anyway.

RJPercival:
@Martin2112, ok to merge?

PR #628, opened 2017-05-19
Add a resetdb.sh script that works with Kubernetes

RJPercival:
Adds a script that can reset a Trillian MySQL database running on Kubernetes, by:

1. Setting up port forwarding to the MySQL port of a Galera node.
2. Passing connection details to scripts/resetdb.sh.
3. Shutting down port forwarding.

### DIFF
--- a/storage/mysql/kubernetes/resetdb.sh
+++ b/storage/mysql/kubernetes/resetdb.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/bin/bash -e
+
+readonly TRILLIAN_PATH=$(go list -f '{{.Dir}}' github.com/google/trillian)
 
 # Pick a random port between 2000 - 34767. Will fail if unlucky and the port is in use.
 readonly port=$((${RANDOM} + 2000))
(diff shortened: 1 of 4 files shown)


RJPercival:
*ping*

storage/mysql/kubernetes/resetdb.sh
```diff
@@ -0,0 +1,25 @@
...
+chmod 0600 "${options_file}"
+echo > ${options_file} <<EOF
+[client]
+host=localhost
+port=${port}
+user=root
+password="$(kubectl get secrets mysql-credentials --template '{{index .data "root-password"}}' | base64 -d)"
+EOF
+
+"${GOPATH}/src/github.com/google/trillian/scripts/resetdb.sh" \
```
Martin2112: All the stuff that relies on gopath will presumably break if there's more than one entry in it. not sure this matters.

RJPercival: I've fixed this here. The examples in README.md still use $GOPATH, but I think that's clearer than `go list -f '{{.Dir}}' github.com/google/trillian`.

RJPercival:
Does that actually work? $GOPATH is supposed to be a colon-separated list, not a Bash array. There's no guarantee that Trillian is in the first of the directories listed in $GOPATH either.

Martin2112:
I think it could be made to work via suitable hacks but as you say it could be out of order so I don't currently think it's really worth doing.

PR #636, opened 2017-05-23
Fix flakiness of test by eliminating race condition in Backoff.Retry()

RJPercival:
On rare occasions, the select statement would receive from `time.After(b.Duration())` sooner than it received from `ctx.Done()`, even when `ctx` had been canceled beforehand.

### DIFF
--- a/client/backoff/backoff.go
+++ b/client/backoff/backoff.go
@@ -68,6 +68,16 @@ func (b *Backoff) Retry(ctx context.Context, f func() error) error {
 	for {
 		err := f()
 		if err != nil {
+			// If the context has errored, return immediately.
+			// Can't just wait for Done() in this case, because
+			// this can race with the backoff, resulting in more
+			// calls to f() than expected.
+			if ctx.Err() != nil {
+				return err
+			}
+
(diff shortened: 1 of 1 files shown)


client/backoff/backoff.go
```diff
@@ -68,6 +68,16 @@ func (b *Backoff) Retry(ctx context.Context, f func() error) error {
 	for {
 		err := f()
 		if err != nil {
+			// If the context has errored, return immediately.
+			// Can't just wait for Done() in this case, because
+			// this can race with the backoff, resulting in more
+			// calls to f() than expected.
+			if ctx.Err() != nil {
```
gdbelvin: If the context errors, does it invoke ctx.Done()? If it does, wouldn't the select return it first?

RJPercival: I think this was because the initial backoff duration is ridiculously low (1ns). That meant `time.After()` fired practically immediately, and so the select could choose it instead of `ctx.Done()` in some rare cases.

RJPercival:
I've created #637 which is a better solution than this.

PR #643, opened 2017-05-24
mv server/vmap/trillian_map_server -> server/trillian_map_server

gdbelvin:
Improve consistency by keeping all the binaries at the same level and in
the same place.


AlCutter:
Can you fix the map integration test?

gdbelvin:
Fixed

PR #649, opened 2017-05-31
fix resetdb script for undefined GOPATH

schachmat:
If GOPATH was not defined in the shell environment, this script failed. With the change it reads the value from the go environment.


codingllama:
Hi @schachmat, thanks for the PR. Would you mind signing the CLA before we proceed?

codingllama:
FYI, there are other places where GOPATH is assumed to be set: integration test scripts, code generation, etc. Those are needed for devs/contributors, but not necessarily for people trying out Trillian. If you want to contribute, you'd probably be better off just setting the GOPATH instead of trying to update all scripts.

schachmat:
Sure, I have set GOPATH on my dev machines. I noticed this issue when testing Trillian using a fresh VM and thought it would be nice for people wanting to try Trillian to just run without having to set GOPATH first. The integration tests apparently worked without setting GOPATH, not sure if they have changed since then.

PR #652, opened 2017-05-31
Map Storage random optimizations

gdbelvin:
- Add documentaion links to storage README.
- Remove if statement from bit length calculation.
- Use anonymous type for readonlyMapTX.
  - Saves us from defining passthrough functions.

### DIFF
--- a/storage/mysql/map_storage.go
+++ b/storage/mysql/map_storage.go
@@ -73,7 +73,7 @@ func (m *mySQLMapStorage) CheckDatabaseAccessible(ctx context.Context) error {
 }
 
 type readOnlyMapTX struct {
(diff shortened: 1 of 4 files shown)


storage/mysql/map_storage.go
```diff
@@ -73,7 +73,7 @@ func (m *mySQLMapStorage) CheckDatabaseAccessible(ctx context.Context) error {
 }
 
 type readOnlyMapTX struct {
-	tx *sql.Tx
+	*sql.Tx
```
Martin2112: I think this OK because the code is MySQL specific and doesn't need to do anything special on commit. That might not be true on other databases but we don't need to worry about it here.

PR #655, opened 2017-06-03
Remove trillian.Tree parameter from SignerFactory interface

RJPercival:
Allows SignerFactory to be used for things other than Trillian trees, e.g. personalities.

### DIFF
--- a/crypto/keys/pem_signer_factory.go
+++ b/crypto/keys/pem_signer_factory.go
@@ -17,33 +17,25 @@ package keys
 import (
 	"context"
 	"crypto"
-	"crypto/ecdsa"
-	"crypto/rsa"
-	"crypto/x509"
 	"fmt"
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
-	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keyspb"
 )
 
 // PEMSignerFactory handles PEM-encoded private keys.
-// It supports trees whose PrivateKey field is a:
-// - keyspb.PEMKeyFile
-// - keyspb.PrivateKey
 // It implements keys.SignerFactory.
 type PEMSignerFactory struct{}
 
-// NewSigner returns a crypto.Signer for the given tree.
-func (f PEMSignerFactory) NewSigner(ctx context.Context, tree *trillian.Tree) (crypto.Signer, error) {
-	if tree.GetPrivateKey() == nil {
-		return nil, fmt.Errorf("tree %d has no PrivateKey", tree.GetTreeId())
-	}
-
+// NewSigner uses the information in pb to return a crypto.Signer.
+// pb must be one of the following types:
+// - keyspb.PEMKeyFile
+// - keyspb.PrivateKey
+func (f PEMSignerFactory) NewSigner(ctx context.Context, pb *any.Any) (crypto.Signer, error) {
 	var privateKey ptypes.DynamicAny
-	if err := ptypes.UnmarshalAny(tree.GetPrivateKey(), &privateKey); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal private key for tree %d: %v", tree.GetTreeId(), err)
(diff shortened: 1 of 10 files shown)


RJPercival:
I've rebased this on master now that PR #639 has been merged.

crypto/keys/pem_signer_factory.go
```diff
@@ -17,33 +17,25 @@ package keys
...
-func (f PEMSignerFactory) NewSigner(ctx context.Context, tree *trillian.Tree) (crypto.Signer, error) {
-	if tree.GetPrivateKey() == nil {
-		return nil, fmt.Errorf("tree %d has no PrivateKey", tree.GetTreeId())
-	}
-
+// NewSigner uses the information in pb to return a crypto.Signer.
+// pb must be one of the following types:
+// - keyspb.PEMKeyFile
+// - keyspb.PrivateKey
+func (f PEMSignerFactory) NewSigner(ctx context.Context, pb *any.Any) (crypto.Signer, error) {
```
codingllama: Suggestion: Have you considered using proto.Message instead of any.Any in the signature? If it turns out to be an Any we can unmarshal, but otherwise we don't force callers to wrap "regular" protos into Any, if they have regular ones available.

Similarly, we could have Generate() return keyspb.PrivateKey and push the marshaling out of signer_factory boundaries.

crypto/keys/signer_factory_tester.go
```diff
@@ -65,22 +63,20 @@ func (tester *SignerFactoryTester) RunAllTests(t *testing.T) {
...
 	for _, test := range append(tester.NewSignerTests, []NewSignerTest{
 		{
-			Name: "Unsupported PrivateKey type",
-			Tree: &trillian.Tree{
-				PrivateKey: mustMarshalAny(&empty.Empty{}),
-			},
-			WantErr: true,
-		},
-		{
-			Name:    "No PrivateKey",
```
codingllama: Would it be worth it to keep this test and pass a nil KeyProto?

server/admin/admin_server_test.go
```diff
@@ -271,21 +272,21 @@ func TestServer_CreateTree(t *testing.T) {
 	defer ctrl.Finish()
 
 	// PEM on the testonly trees is ECDSA, so let's use an ECDSA key for tests.
-	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	var defaultPrivateKey crypto.Signer
+	var err error
+	defaultPrivateKey, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
-		t.Fatalf("Error generating test key: %v", err)
+		t.Fatalf("failed to generate test private key: %v", err)
```
codingllama: nit: s/failed/Failed. (Start with a capitalized word on t.Fatalf.)

server/admin/admin_server_test.go
```diff
@@ -271,21 +272,21 @@ func TestServer_CreateTree(t *testing.T) {
...
 	}
 
-	publicKeyDER, err := x509.MarshalPKIXPublicKey(privateKey.Public())
+	defaultPublicKeyDER, err := x509.MarshalPKIXPublicKey(defaultPrivateKey.Public())
 	if err != nil {
-		t.Fatalf("Error marshaling public key: %v", err)
+		t.Fatalf("failed to marshal test public key as DER: %v", err)
```
codingllama: Ditto.

crypto/keys/signer_factory_tester.go
RJPercival: We don't typically check for nil pointers being passed into funcs. I'm happy to add this if it's a convention we want to start following. I've heard people argue in favour of crashing on nil pointers, but returning an error seems better to me.

server/admin/admin_server_test.go

server/admin/admin_server_test.go

crypto/keys/pem_signer_factory.go
RJPercival: We could certainly push the marshaling out to the caller. I'll do this in a follow-up PR.

PR #662, opened 2017-06-07
Add IsHealthy to AdminServer implementation + wire to storage

Martin2112:
Similar to what was done for the log server.

### DIFF
--- a/server/admin/admin_server.go
+++ b/server/admin/admin_server.go
@@ -44,6 +44,11 @@ func New(registry extension.Registry) *Server {
 	return &Server{registry}
 }
 
+// IsHealthy returns nil if the server is healthy, error otherwise.
+func (s *Server) IsHealthy() error {
+	return s.registry.AdminStorage.CheckDatabaseAccessible(context.Background())
+}
+
 // ListTrees implements trillian.TrillianAdminServer.ListTrees.
 func (s *Server) ListTrees(ctx context.Context, req *trillian.ListTreesRequest) (*trillian.ListTreesResponse, error) {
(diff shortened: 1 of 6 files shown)


storage/mysql/admin_storage.go
```diff
@@ -72,6 +72,10 @@ func (s *mysqlAdminStorage) Begin(ctx context.Context) (storage.AdminTX, error)
 	return &adminTX{tx: tx}, nil
 }
 
+func (s *mysqlAdminStorage) CheckDatabaseAccessible(ctx context.Context) error {
```
codingllama: nit: Add a simple test for CheckDatabaseAccessible?

server/admin/admin_server.go
```diff
@@ -42,6 +42,11 @@ func New(registry extension.Registry) *Server {
 	return &Server{registry}
 }
 
+// IsHealthy returns nil if the server is healthy, error otherwise.
+func (s *Server) IsHealthy() error {
+	return s.registry.AdminStorage.CheckDatabaseAccessible(context.Background())
```
codingllama: I realize the other IsHealthy methods don't take a ctx either, but maybe we should break the consistency here and take a ctx as parameter. WDYT?

Martin2112:
You could have an admin implementation that doesn't share the same storage as the log data. Not sure how likely this is.

server/admin/admin_server.go
Martin2112: I'll TODO this so they can be all made consistent.

Martin2112:
PTAL.

Martin2112:
I think it got hit by the flake in #617.

PR #663, opened 2017-06-07
Replace SignerFactory use of *any.Any with proto.Message

RJPercival:
Avoids having to marshal a protobuf message as a protobuf `Any` before passing it or returning it from a `SignerFactory`. If the caller then wants to store it in a `trillian.Tree`, they will need to do this marshaling.

### DIFF
--- a/crypto/keys/pem_signer_factory.go
+++ b/crypto/keys/pem_signer_factory.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"fmt"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/google/trillian/crypto/keyspb"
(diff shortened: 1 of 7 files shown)


crypto/keys/pem_signer_factory.go
```diff
@@ -32,25 +33,28 @@ type PEMSignerFactory struct{}
...
 	}
 
-	return nil, fmt.Errorf("unsupported PrivateKey type: %T", privateKey.Message)
+	return nil, fmt.Errorf("unsupported private key protobuf type: %T", pb)
 }
 
 // Generate creates a new private key based on a key specification.
 // It returns a proto that can be passed to NewSigner() to get a crypto.Signer.
-func (f PEMSignerFactory) Generate(ctx context.Context, spec *keyspb.Specification) (*any.Any, error) {
+func (f PEMSignerFactory) Generate(ctx context.Context, spec *keyspb.Specification) (proto.Message, error) {
```
codingllama: Return keyspb.PrivateKey? It won't hurt to be more specific here, I think.

codingllama: Nevermind, the proto may vary depending on the SignerFactory implementation.

RJPercival: Not all SignerFactory implementation may want to return an actual DER-encoded private key here. A PKCS11 SignerFactory, for instance, might return a protobuf containing the key's name.

PR #631, opened 2017-05-22
Add Flexible Flexible ResetDB and Docker Support 

witoff:
**Problem** 
Trillian has a number of hardcoded assumptions that DB and Trillian are running on the same host.  This is problematic running most cloud environments and in most configurations of docker. 

**Updates**

This PR adds Docker support, cherrypicked from @gdbelvin's work in [607](https://github.com/google/trillian/pull/607). I'd prefer that his PR is merged first and I rebase on his work.  Even though that PR is recent (10 days), it's out of date enough that this PR was easier to cherrypick in directly. 

ResetDB is generalized to run on any host
```shell 
./resetdb.sh --verbose --force -h localhost
```

**Usage**
```shell
# Build
docker build -t trillian-server -f server/trillian_log_server/Dockerfile .
docker build -t trillian-signer -f server/trillian_log_signer/Dockerfile .
docker build -t trillian-db-client -f server/trillian_db_client/Dockerfile .

# Run
docker run \
  --name mysql \
  -p 3306:3306 \
  -e MYSQL_ROOT_PASSWORD=password123 \
  mysql

# Seed the DB
docker run --rm -it \
  --link mysql \
  -e DB_USER=root \
  -e DB_PASSWORD=password123 \
  trillian-db-client ./scripts/resetdb.sh --verbose --force -h mysql

# Startup the Server
docker run --rm -it \
  --link mysql \
  -p 8091:8091 \
  -p 8090:8090 \
  -e DB_USER=root \
  -e DB_PASSWORD=password123 \
  -e DB_HOST=mysql:3306 \
  trillian-server
```

PR #667, opened 2017-06-08
Automate Local and AWS Deployments

witoff:
**Summary**

Per #666 (yikes!), we'd like to better document deployment configurations for trillian.   This builds on pull #631 to provide simple, few-line local and cloud deployments using docker-compose and terraform.

**Notes**

* This removes the MySQL strict clause  in `storage.sql`'s which is incompatible with aurora.
* This merges Dockerfiles into `/examples/deployments/` directory.   Shipping these opinionated Dockerfiles next to source code felt like a stretch.
* This pulls [wait-for-it](https://github.com/vishnubob/wait-for-it) in at docker build time.  Is this an appropriate way to use a dependency with our license?


examples/deployment/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+To use AWS's managed databases, start here:
+
+```shell
+# Create Resources
+terraform plan aws/
+terraform apply aws/
+
+# Set Variables
+HOST=...
```
witoff: merged this with automated launch userdata within terraform (more automation!) and changed to `DB_HOST`.

examples/deployment/aws/terraform.tf
```diff
@@ -0,0 +1,59 @@
...
+  identifier          = "trillian-${count.index}"
+  cluster_identifier  = "${aws_rds_cluster.trillian.id}"
+  instance_class      = "db.r3.large"
+  publicly_accessible = true
+  apply_immediately   = true
+}
+
+resource "aws_security_group" "trillian" {
+  name        = "trillian"
+  description = "Allow MySQL from Single IP"
```
witoff: updated ;)

examples/deployment/scripts/download-wait-for-it.sh
```diff
@@ -0,0 +1,14 @@
...
+
+download() {
+  COMMIT=8f52a814ef0cc70820b87fbf888273f3aa7f5a9b
+  URL=https://raw.githubusercontent.com/vishnubob/wait-for-it/${COMMIT}/wait-for-it.sh
+  curl -sO $URL
+  chmod a+x wait-for-it.sh
+}
+
+download
+echo "ffe253ce564df22adbcf9c799e251ca0  wait-for-it.sh" > file.md5
```
witoff: doh, thanks.  fixed!

AlCutter:
BTW - we generally rebase topic branches onto latest `master` rather than merge `master` into the topic branch

witoff:
rebased :)

scripts/resetdb.sh
```diff
@@ -1,16 +1,59 @@
...
+  [ -z ${DB_PASSWORD+x} ] || FLAGS="${FLAGS} -p$DB_PASSWORD"
+
+  # handle flags
+  FORCE=false
+  VERBOSE=false
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      --force) FORCE=true ;;
+      --verbose) VERBOSE=true ;;
+      *) FLAGS="${FLAGS} $1"
```
RJPercival: I think this will remove quotation marks around arguments.

scripts/resetdb.sh
```diff
@@ -1,16 +1,59 @@
...
+    echo "-- using DB_USER: ${DB_USER}"
+  fi
+  echo "Warning: about to destroy and reset database '${DB_NAME}'"
+
+  [[ ${FORCE} = true ]] || read -p "Are you sure? " -n 1 -r
+
+  if [ -z ${REPLY+x} ] || [[ $REPLY =~ ^[Yy]$ ]]
+  then
+      echo "Resetting DB..."
+      mysql -u $DB_USER $FLAGS -e "DROP DATABASE IF EXISTS ${DB_NAME};"
```
RJPercival: Some flags, e.g. [--defaults-extra-file](https://dev.mysql.com/doc/refman/5.7/en/option-file-options.html#option_general_defaults-extra-file), will only work if they are the first flag passed to mysql. This flag ordering makes passing that flag impossible.

scripts/resetdb.sh
```diff
@@ -1,16 +1,59 @@
...
+      mysql -u $DB_USER $FLAGS -e "CREATE DATABASE ${DB_NAME};"
+      mysql -u $DB_USER $FLAGS -e "GRANT ALL ON ${DB_NAME}.* TO '${DB_NAME}'@'localhost' IDENTIFIED BY 'zaphod';"
```
RJPercival: I'm still not sure how this is supposed to work. Granting access for the user to connect only from localhost shouldn't work if the Trillian server/signer is running on a different machine, as will likely be the case on cloud infrastructure. It *may* work if $DB_NAME == "test", because some MySQL installations come with default privileges that grant unrestricted access to that database for all users.

scripts/resetdb.sh
RJPercival: Consider making $FLAGS an array.

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+Want to deploy/use the Trillian General Transparency project in the cloud?  Here are some common ways of getting off the ground with Docker.
+
+## Setup
+
+**Clone Source**
+
+Both build and example deployment files are stored within this repo.  For any of the below deployment methods, start here:
+
+```shell
+git clone git@github.com:google/trillian.git
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+cd trillian
+```
+
+## Local Deployments
+
+**Run With Docker Compose**
+
+For simple deployments running in a container is an easy way to get up and running with a local database.  To use Docker to run and interact with Trillian, start here:
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Set a random password and bring up the services defined in the provided compose file.  This includes a local mysql database database, a one-shot container to create the schema and the trillian server:
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```shell
+# Set a random password
+export DB_PASSWORD="$(openssl rand -hex 16)"
+
+# Bring up services defined in this compose file.  This includes:
+# - local MySQL databse
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+# - container to seed the database
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+# - the trillian server
+docker-compose -f examples/deployment/docker-compose.yml up
+```
+
+Verify that your local installation by checking the metrics endpoint:
```
witoff: fixed

examples/deployment/README.md
```diff
@@ -0,0 +1,64 @@
...
+TODO
+
+### Run in AWS
+
+With a pair of AWS keys [accessible to Terraform](https://www.terraform.io/docs/providers/aws/), this template deploys a simple Trillian setup in AWS using EC2 and RDS MySQL.
+
+```shell
+# Set a random password
+export TF_VAR_DB_PASSWORD="$(openssl rand -hex 16)"
+export TF_VAR_ingress_cidr="0.0.0.0/0"
```
witoff: fixed

examples/deployment/aws/terraform.tf
```diff
@@ -0,0 +1,159 @@
...
+curl -o /tmp/go.tar.gz https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz
+tar -C /usr/local -xzf /tmp/go.tar.gz
+export PATH=$PATH:/usr/local/go/bin
+
+mkdir -p /go
+export GOPATH=/go
+
+# Install Trillian
+mkdir -p /go/src/github.com/google/trillian
+git clone https://github.com/google/trillian.git /go/src/github.com/google/trillian
```
witoff: good catch!  compressed down to one line with `go get`.

examples/deployment/aws/terraform.tf
```diff
@@ -0,0 +1,159 @@
...
+cd /go/src/github.com/google/trillian
+go get ./server/trillian_map_server
```
witoff: replaced with `trillian_log_server`.  this came from an earlier example.

examples/deployment/docker/db_client/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.8
+
+RUN apt-get update && \
+    apt-get install -y mysql-client
+
+ADD . /go/src/github.com/google/trillian
+WORKDIR /go/src/github.com/google/trillian
+
+RUN go get -v ./server/trillian_log_server
```
witoff: removed this line but going to keep the source code included so `/scripts/*.sh` can be run

examples/deployment/docker/db_client/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ENV DB_USER=test \
+    DB_PASSWORD=zaphod \
+    DB_DATABASE=test
+
+RUN ./examples/deployment/scripts/download-wait-for-it.sh
```
witoff: in the `[docker-compose.yaml](https://github.com/google/trillian/pull/667/files#diff-971a49e212add79c1dc8922a40d5e753R14)` file, this is used to wait for MySQL to become available before attempting to run `resetdb.sh`

scripts/resetdb.sh
```diff
@@ -1,16 +1,59 @@
...
+  echo " - DB_PASSWORD"
+}
+
+collect_vars() {
+  # set unset environment variables to defaults
+  [ -z ${DB_USER+x} ] && DB_USER="root"
+  [ -z ${DB_NAME+x} ] && DB_NAME="test"
+  # format reused supplied environment variables
+  FLAGS=""
+  [ -z ${DB_PASSWORD+x} ] || FLAGS="${FLAGS} -p$DB_PASSWORD"
```
witoff: fixed with: `-p'${DB_PASSWORD}'`

scripts/resetdb.sh
witoff: reordered to: `mysql -u $FLAGS $DB_USER` and preserved user supplied order in `collect_vars` fn.

scripts/resetdb.sh
witoff: i'm also not entirely clear on this.  this change mirrors what's currently in the master branch and just makes DB_NAME a variable.  after this is merged i'd like work on better integration testing here which may prove this unnecessary.

scripts/resetdb.sh
witoff: added a verbose logging statement that will print args to clarify what's being run.  even as an array i'd still have to reconstruct with proper flags to pass to mysql.  if you think this is a blocker we can revert back to passing all flags via `$@` to mysql and only accepting other params via envars.

examples/deployment/docker/db_client/Dockerfile
RJPercival: Could you add a comment here saying that please.

witoff: yep, added:
```
# This is used to wait for new MySQL deployments to become ready e.g.
#  ./wait-for-it.sh localhost:3306 -- mysql
```

RJPercival:
I think you need to rebase - looks like master was somehow broken when you last rebased.

daviddrysdale:
Yeah, Prometheus changed underneath us upstream -- pull #724 should have fixed it though.

witoff:
k, rebased and tests passing now.

examples/deployment/docker/db_client/Dockerfile
RJPercival: Could you mention in the comment that it's used in docker-compose.yaml (or otherwise indicate that it's expected to be run by some monitoring system). This makes it clear that nothing in the container itself uses this.

scripts/resetdb.sh
```diff
@@ -1,16 +1,64 @@
...
+
+  [[ ${FORCE} = true ]] || read -p "Are you sure? " -n 1 -r
+
+  if [ -z ${REPLY+x} ] || [[ $REPLY =~ ^[Yy]$ ]]
+  then
+      echo "Resetting DB..."
+      mysql -u $FLAGS $DB_USER -e "DROP DATABASE IF EXISTS ${DB_NAME};"
+      mysql -u $FLAGS $DB_USER -e "CREATE DATABASE ${DB_NAME};"
+      mysql -u $FLAGS $DB_USER -e "GRANT ALL ON ${DB_NAME}.* TO '${DB_NAME}'@'localhost' IDENTIFIED BY 'zaphod';"
+      mysql -u $FLAGS $DB_USER -D ${DB_NAME} < ${TRILLIAN_PATH}/storage/mysql/storage.sql
```
RJPercival: I don't think you meant to separate "-u" and "$DB_USER". FYI, I've fixed this in the branch I referenced above, as a side-effect of adding "-u $DB_USER" to $FLAGS in collect_vars().

scripts/resetdb.sh
RJPercival: Suggested fix: https://github.com/RJPercival/trillian/commits/witoff/aws

This will properly handle quoted arguments. I made another couple of small improvements while I was poking around in that file, which you can find just after that commit.

scripts/resetdb.sh
RJPercival: With the database name hardcoded as "test", it's going to get the special treatment that MySQL gives that table by default. By replacing this with a variable, that's no longer true unless $DB_NAME == "test". Therefore, I think replacing 'localhost' with '%' on this line is warranted. Access to the database server should be restricted to certain IP ranges, so it won't allow brute force attacks on this user by just anyone.With the database name hardcoded as "test", it's going to get the special treatment that MySQL gives that table by default. By replacing this with a variable, that's no longer true unless $DB_NAME == "test". Therefore, I think replacing 'localhost' with '%' on this line is warranted. Access to the database server should be restricted to certain IP ranges, so it won't allow brute force attacks on this user by just anyone.With the database name hardcoded as "test", it's going to get the special treatment that MySQL gives that table by default. By replacing this with a variable, that's no longer true unless $DB_NAME == "test". Therefore, I think replacing 'localhost' with '%' on this line is warranted. Access to the database server should be restricted to certain IP ranges, so it won't allow brute force attacks on this user by just anyone.

RJPercival:
These should be my final comments; it's looking good to merge bar those.

witoff:
edits are merged (thank you!) and waiting on #756 to fix broken golang/mocks updates before we can merge.

gdbelvin:
Hi @witoff, thanks for this PR.  Rather than moving the opinionated Dockerfiles to the example directory, would you mind keeping them in-place and making them less opinionated? Having some kind of starting Dockerfile would help other projects adopt Trillian faster.  

If you feel that the opinionated Dockerfiles have value, please use the `from` Docker command to start with the less opinionated Dockerfiles, and add the opinionated bits. 

Thanks!
-Gary

witoff:
@gdbelvin updated PR with a proposal on how these could work.  my view is that `main.go` exposes a set of flags, default `Dockerfiles` define what's needed to build/run the binary and it's up to the user to configure from there, likely via docker-compose, docker run etc.  

Specifying flags/healthcheck/envars in the Dockerfile makes assumptions about my infrastructure that could reasonably be incorrect which would force me to write a separate Dockerfile anyways. 

I've also kept the `/examples/...` Dockerfiles separate, since inheriting from the `/server/...` Dockerfiles would force a 2-phase build (or more scripts) which feels like it would work against the goal here of making it easy to reason about and run one example deployment. 

wdyt? if you're okay with this approach does this break any other uses of these dockerfiles that i can help update?

gdbelvin:
This is great.  I now remember why I had the flags in the original docker files, and that is because go flags doesn't pick up environment variables, so I basically duplicated the flags with an additional (confusing) set of defaults.  These non-opinionated Docker files would be great if we had a way for go flags to pick up passed-in environment variables. 

 @RJPercival @Martin2112 Do we have a plan for passing flags via environment variables?

RJPercival:
No, we had no plans to implement that. It's easy to do, probably something like this in cmd/flags.go:
```go
func ParseFlagsFromEnv() {
  flag.VisitAll(func (f *Flag) {
    // Might want to modify the flag name before using it as the envar name, e.g. uppercase and/or prefix with "TRILLIAN_".
    if env, ok := os.LookupEnv(f.Name); ok {
      f.Value.Set(env)
    }
  })
}
```
The only question is how much value there is in supporting this. Presumably this is to support people following the [12 Factor App](https://12factor.net/config) guidelines, which lays out the advantages. The main cost is that it adds a non-obvious, third way of controlling the behaviour of Trillian's binaries.

Martin2112:
We've deliberately kept flags isolated and e.g. the logserver main.go is
just over 100 lines of code. Anybody who needs this should easily be able
create an alternate main() that gets the configuration from somewhere else.

Martin

gdbelvin:
I think we can use the `command` option in `docker-compose` to add commandline flags now that we're using `ENTRYPOINT` 

LGTM.

PR #669, opened 2017-06-08
Replace gogo/protobuf/proto imports with golang/protobuf/proto

RJPercival:
This was presumably done by accident.


RJPercival:
Indeed. I certainly don't think we'd want a situation like this, where we're using it for parts of Trillian and not others. I think this happened because some tools that automatically add imports will add gogo/protobuf rather than golang/protobuf.

PR #671, opened 2017-06-12
Document how to restrict the IP ranges that can access a Galera cluster

RJPercival:
By default, a Galera cluster created using Kubernetes is publicly accessible.

### DIFF
--- a/storage/mysql/kubernetes/README.md
+++ b/storage/mysql/kubernetes/README.md
@@ -1,13 +1,15 @@
 :exclamation: **EXPERIMENTAL** :exclamation:
 
-# Usage #
+# MySQL on Kubernetes
+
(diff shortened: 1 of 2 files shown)


storage/mysql/kubernetes/README.md
```diff
@@ -19,6 +19,16 @@ kubectl proxy
 This dashboard will also show the external IP of the cluster on the
 "Services" page, on the row for the "mysql" service.
 
+#### :exclamation: WARNING :exclamation:
```
Martin2112: Should we make this the other way round? So they have to explicitly expose it.

RJPercival: Done.

PR #676, opened 2017-06-13
Replenish quota tokens based on sequenced entries

codingllama:
Sequencing-based quotas need a signal to be able to replenish tokens based on
the sequencer progress. Different quota implementations may simply choose to
ignore the signal.


codingllama:
Thanks, Martin!

PR #678, opened 2017-06-13
admin_storage_tester.go no longer needs to load keys from disk

RJPercival:
This is helpful because some test infrastructure moves the binaries around, which results in the tests not being able to find the key files.

### DIFF
--- a/storage/testonly/admin_storage_tester.go
+++ b/storage/testonly/admin_storage_tester.go
@@ -41,15 +41,31 @@ func mustMarshalAny(pb proto.Message) *any.Any {
 	return value
 }
 
-// loadPublicPEMFileAsDER reads a public key from a PEM file and returns the key in DER encoding.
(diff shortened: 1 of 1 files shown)


storage/testonly/admin_storage_tester.go
```diff
@@ -56,6 +56,22 @@ func loadPublicPEMFileAsDER(path string) []byte {
 	return der
 }
 
+// privatePEMToDER reads a PEM-encoded private key and returns it in DER encoding.
+// The PEM must be encrypted, and the password must be provided in order to decrypt it.
+// If an error occurs, it panics.
+func privatePEMToDER(pem string, password string) []byte {
```
codingllama: nit: `pem, password string`

RJPercival: Done.

PR #680, opened 2017-06-14
Make integration tests use demo keys from testonly package

RJPercival:
Rather than loading keys from testdata/log-rpc-server-{pub,priv}key.pem files. 

Finishes what PR #678 started.

### DIFF
--- a/testonly/integration/clientserver.go
+++ b/testonly/integration/clientserver.go
@@ -46,8 +46,13 @@ var (
 	// SequencerInterval is the time between runs of the sequencer.
 	SequencerInterval = 100 * time.Millisecond
 	timeSource        = util.SystemTimeSource{}
-	publicKeyPath     = testonly.RelativeToPackage("../../testdata/log-rpc-server.pubkey.pem")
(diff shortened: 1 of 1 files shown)


testonly/integration/clientserver.go
```diff
@@ -46,13 +46,25 @@ var (
...
-	privateKeyInfo    = &keyspb.PEMKeyFile{
-		Path:     testonly.RelativeToPackage("../../testdata/log-rpc-server.privkey.pem"),
-		Password: "towel",
+	publicKey         = testonly.DemoPublicKey
+	privateKeyInfo    = &keyspb.PrivateKey{
+		Der: privatePEMToDER(testonly.DemoPrivateKey, testonly.DemoPrivateKeyPass),
 	}
 )
 
+func privatePEMToDER(pem, password string) []byte {
```
codingllama: Is this the same logic as the one on admin_storage_tester.go? Could we reuse it? (We can do it in a separate PR if you prefer.)

RJPercival: It is the same logic, yes. I'll see if I can find a common place for it to live in a separate PR.

PR #682, opened 2017-06-14
Improve error messages for not found trees

codingllama:


codingllama:
As discussed on https://github.com/google/keytransparency/issues/604.

I've decided against making the change in the interceptor, as chaining the errors there removes the error code from the translation layer. I'd have to unpack the error, get the code, chain the msgs and replace the code, which seemed a bit weird to do.

Instead I went to the error source and improved the message that flows up.

PR #684, opened 2017-06-15
Move code for marshaling PEM to DER into crypto/keys/testonly package

RJPercival:
PrivatePEMToDER() is used in a couple of different tests, so this eliminates some duplication.

As requested in PR #680.

### DIFF
--- a/crypto/keys/testonly/keys.go
+++ b/crypto/keys/testonly/keys.go
@@ -1,0 +1,51 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


crypto/keys/testonly/keys.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return der
+}
+
+// MustMarshalPrivatePEMToDER decrypts a PEM-encoded private key and returns it in DER encoding.
+// If an error occurs, it panics.
+func MustMarshalPrivatePEMToDER(pem, password string) []byte {
+	key, err := keys.NewFromPrivatePEM(ttestonly.DemoPrivateKey, ttestonly.DemoPrivateKeyPass)
```
codingllama: Use pem / password as the params? I think you've left DemoPrivateKey / DemoPrivateKeyPass by accident there.

RJPercival: Oops, fixed!

PR #691, opened 2017-06-19
CONIKS hasher

### LINKED ISSUE #670: Map: CONIKS map hasher
To provide l bits of security in the map we need the following:
- A collision and pre-image resistant hash function H of 2l bits. 
- A tree specific nonce to prevent multi-tree attacks [1][2]
- An index specific value for both leafs and empty branches to prevent attacking multiple locations within the same tree [1][2]

In particular, this means that the HStar2 algorithm does not provide the full l bits of security in a multi-tree setting, and we need to adjust the tree hasher interface to support supplying location and tree specific values during the computation of leaves and empty branches.

[1]: https://eprint.iacr.org/2016/683.pdf Section 5.1
[2]: https://eprint.iacr.org/2014/1004.pdf Section 3.1

Proposal for MapHasher:
```go
type MapHasher interface {
	HashEmpty(treeNonce int64, index []byte, height int) []byte
	HashLeaf(treeNonce int64, index []byte, depth int, leaf []byte) []byte
	HashChildren(l, r []byte) []byte
	BitLen() int
}
```

gdbelvin:
Add treeID, index, and depth to HashLeaf and HashEmtpy.
The fulfills the security requirements and closes #670.

Adds the CONIKS hasher to the set of map hashers.
The next PR will make it available through the command line.


merkle/coniks/coniks.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
Martin2112: 2017.

merkle/coniks/coniks.go
```diff
@@ -0,0 +1,89 @@
...
+// Domain separation prefixes
+var (
+	leafIdentifier  = []byte("L")
+	emptyIdentifier = []byte("E")
+)
+
+// Default is the standard CONIKS hasher.
+var Default = New(crypto.SHA512_256)
+
+// hasher implements the sparse merkel tree hashing algorithm specified in the CONIKS paper.
```
Martin2112: Same typo as before. Merkle. Please fix where you cut and pasted it from too.

merkle/coniks/coniks.go
```diff
@@ -0,0 +1,89 @@
...
+// HashEmpty returns the hash of an empty branch at a given depth.
+// A depth of 0 indicates the hash of an empty leaf.
+// Empty branches within the tree are plain interior nodes e1 = H(e0, e0) etc.
+func (m *hasher) HashEmpty(treeID int64, index []byte, height int) []byte {
+	depth := m.BitLen() - height
+
+	h := m.New()
+	h.Write(emptyIdentifier)
+	binary.Write(h, binary.BigEndian, uint64(treeID))
+	h.Write(index) // TODO block out the bits that are not part of the index.
```
Martin2112: What are the implications of this TODO. Does it work properly as is?

merkle/coniks/coniks.go
```diff
@@ -0,0 +1,89 @@
...
+
+// EmptyRoot returns the root of an empty tree.
+func (m *hasher) EmptyRoot() []byte {
+	panic("EmptyRoot() not defined for coniks.Hasher")
+}
+
+// HashEmpty returns the hash of an empty branch at a given depth.
+// A depth of 0 indicates the hash of an empty leaf.
+// Empty branches within the tree are plain interior nodes e1 = H(e0, e0) etc.
+func (m *hasher) HashEmpty(treeID int64, index []byte, height int) []byte {
```
Martin2112: The comment says 'depth' but the parameter is 'height'. This will cause confusion. If it really is height can you fix it in the interface definition and the other implementation. We've tried to use 'depth' everywhere because of previous confusion.

merkle/coniks/coniks.go
```diff
@@ -0,0 +1,89 @@
...
+	h := m.New()
+	h.Write(leafIdentifier)
+	binary.Write(h, binary.BigEndian, uint64(treeID))
+	h.Write(index) // TODO block out the bits that are not part of the index.
+	binary.Write(h, binary.BigEndian, uint32(depth))
+	h.Write(leaf)
+	return h.Sum(nil)
+}
+
+// HashChildren returns the inner Merkle tree node hash of the the two child nodes l and r.
```
Martin2112: replace 'inner' with 'internal'.

merkle/coniks/coniks_test.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
Martin2112: 2017.

merkle/hstar2_test.go
```diff
@@ -25,6 +24,8 @@ import (
 	"github.com/google/trillian/testonly"
 )
 
+const treeID = int64(0)
```
Martin2112: I'd suggest something other than zero for the ID. I'd also test that different treeID values result in different hashes.

merkle/map_verifier.go
```diff
@@ -40,6 +40,11 @@ func VerifyMapInclusionProof(index, leafHash, expectedRoot []byte, proof [][]byt
 	if got, want := len(leafHash)*8, hBits; got != want {
 		return fmt.Errorf("invalid leafHash length %d, want %d", got, want)
 	}
+	for i, element := range proof {
+		if got, wanta, wantb := len(element), 0, h.Size(); got != wanta && got != wantb {
+			return fmt.Errorf("invalid proof: len(proof[%v]) %d, want %d or %d", i, got, wanta, wantb)
```
Martin2112: Should use "got:, want:" formatting in Errorf.

merkle/objhasher/objhasher_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/google/trillian/merkle/rfc6962"
 )
 
+const treeID = int64(0)
```
Martin2112: Again, zero is probably not the best choice.

storage/cache/subtree_cache_test.go
```diff
@@ -56,8 +56,10 @@ var splitTestVector = []struct {
 var defaultLogStrata = []int{8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8}
 var defaultMapStrata = []int{8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 176}
 
+const treeID = int64(0)
```
Martin2112: Another use of zero.

merkle/coniks/coniks.go
gdbelvin: done

merkle/coniks/coniks.go
gdbelvin: done

merkle/coniks/coniks.go
gdbelvin: After merging this PR, I wouldn't be opposed to changing the tree computation algorithms to all use depth.  At the moment though, the algorithms themselves use height and it's a little simpler to hide the height to depth conversion inside this function.

merkle/coniks/coniks.go
gdbelvin: done

merkle/coniks/coniks_test.go
gdbelvin: done

merkle/hstar2_test.go
gdbelvin: I'm not exactly sure what to do here. The maphasher - which matches the Python / C++ merkle tree code - does not use TreeID and result in different hashes.  

We could 
- Change the maphasher to include TreeID and generate new test vectors. 
- Keep the existing ones and add a new test with CONIKS test vectors. But do we really want to make the merkle package dependent on all the sub hashing algorithms?

Martin2112:
OK. We can leave maphasher as is. The coniks one should test that the
treeID is being correctly handled.

merkle/map_verifier.go
gdbelvin: better?

merkle/objhasher/objhasher_test.go
gdbelvin: The subhashers that the test uses don't use treeID

storage/cache/subtree_cache_test.go
gdbelvin: same

merkle/coniks/coniks.go
gdbelvin: The result is that the hashes would be incorrect.  Thanks for poking at this. 
I've implemented and tested the proper masking function now.

gdbelvin:
I've implemented `maskIndex` PTAL.

gdbelvin:
Rebased now that #694 has been merged. PTAL

c633:
Do you want to hash the `treeID`, as proposed at https://github.com/coniks-sys/coniks-go/issues/177#issuecomment-310151341?

merkle/coniks/coniks.go
```diff
@@ -0,0 +1,124 @@
...
+
+// leftmask contains bitmasks indexed such that the left x bits are set.
+// 0 is special cased to 0xFF since 8 mod 8 is 0.
+var leftmask = map[int]byte{
+	0: 0xFF, 1: 0x80, 2: 0xC0, 3: 0xE0,
+	4: 0xF0, 5: 0xF8, 6: 0xFC, 7: 0xFE,
+}
+
+// maskIndex returns index with only the left depth bits set.
+// index must be of size m.Size() and 0 <= depth <= m.BitLen().
```
Martin2112: This could probably be an array. I think including an example in the comment would help here. Maybe also mention that it's only used for the last byte to explain the 0xff in position 0.

merkle/coniks/coniks_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}{
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 0, want: h2b("0000000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 1, want: h2b("8000000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 2, want: h2b("C000000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 3, want: h2b("E000000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 4, want: h2b("F000000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 5, want: h2b("F800000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 6, want: h2b("FC00000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 7, want: h2b("FE00000000000000000000000000000000000000")},
+		{index: h2b("FF00000000000000000000000000000000000000"), depth: 8, want: h2b("FF00000000000000000000000000000000000000")},
```
Martin2112: Can you add a couple of test cases that are less uniform and show the mask being applied across byte boundaries etc.

merkle/hstar2.go
```diff
@@ -176,3 +177,13 @@ func (s *valueSorter) Less(i, j int) bool {
 func indexLess(a, b *HStar2LeafHash) bool {
 	return a.Index.Cmp(b.Index) < 0
 }
+
+// PaddedBytes takes a big.Int and returns it's value, left padded with zeros.
+// e.g. 1 -> 0000000000000000000000000000000000000001
+func PaddedBytes(i *big.Int, size int) []byte {
+	b := i.Bytes()
+	ret := make([]byte, size)
```
Martin2112: Does this work if there is no padding needed?

merkle/hstar2_test.go
```diff
@@ -204,3 +207,25 @@ func TestHStar2NegativeTreeLevelOffset(t *testing.T) {
...
 	}
 }
+
+func TestPaddedBytes(t *testing.T) {
+	size := 160 / 8
+	for _, tc := range []struct {
+		i    int64
+		want []byte
+	}{
+		{i: 1, want: h2b("0000000000000000000000000000000000000001")},
```
Martin2112: This would be a good place to add the test for zero padding + possibly others.

merkle/objhasher/objhasher_test.go
Martin2112: OK then as long as hashers that do use it have their own tests.

server/trillian_log_server/main.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/extension"
 	_ "github.com/google/trillian/merkle/objhasher" // Load hashers
```
Martin2112: I don't think you need the comments on these imports + in other files. The '_' implies it's being imported for side effects.

merkle/coniks/coniks.go
gdbelvin: I would agree, but different hashers can have different sizes for their indexes, which makes using an array here difficult.

Martin2112: Not sure what you mean. It's only used by maskIndex and it's indexed by byte position from 0-7.

Martin2112: Yes, leftmask, which should probably also be leftMask.

merkle/coniks/coniks_test.go
gdbelvin: Better?

merkle/hstar2.go
gdbelvin: Added at test for this.

merkle/hstar2_test.go
gdbelvin: eyp

server/trillian_log_server/main.go
gdbelvin: go vet sometimes complains if there isn't a comment on underscore imports.

merkle/coniks/coniks_test.go
Martin2112: Not sure if github is showing me all the latest changes but I'd like to see a depth above 8.

gdbelvin:
Added more tests. PTAL

PR #693, opened 2017-06-20
Simplify MaxRootDuration validation

codingllama:
Removed validation from admin_server, as it's already performed by
tree_validation.


codingllama:
Thanks!

PR #694, opened 2017-06-20
Create Hasher Registry

gdbelvin:
Implements a hasher registry in the style of go's crypto package

Replaces `tree.Hasher` with `hashers.NewLogHasher` 

Moves tree_hasher.go to merkle/hashers to break some import cycles.

### DIFF
--- a/merkle/hashers/tree_hasher.go
+++ b/merkle/hashers/tree_hasher.go
@@ -50,25 +50,17 @@ type MapHasher interface {
 }
 
 var (
-	maxHashers = maxKey(trillian.HashStrategy_name) + 1
-	logHashers = make([]LogHasher, maxHashers)
-	mapHashers = make([]MapHasher, maxHashers)
+	logHashers = make(map[trillian.HashStrategy]LogHasher)
+	mapHashers = make(map[trillian.HashStrategy]MapHasher)
 )
-
-func maxKey(m map[int32]string) int32 {
-	var max int32
-	for k := range m {
-		if k > max {
-			max = k
-		}
-	}
-	return max
-}
 
 // RegisterLogHasher registers a hasher for use.
 func RegisterLogHasher(h trillian.HashStrategy, f LogHasher) {
 	if h == trillian.HashStrategy_UNKNOWN_HASH_STRATEGY {
 		panic(fmt.Sprintf("RegisterLogHasher(%v) of unknown hasher", h))
+	}
+	if logHashers[h] != nil {
+		panic(fmt.Sprintf("%v already registred as a LogHasher", h))
 	}
 	logHashers[h] = f
 }
@@ -78,10 +70,13 @@ func RegisterMapHasher(h trillian.HashStrategy, f MapHasher) {
 	if h == trillian.HashStrategy_UNKNOWN_HASH_STRATEGY {
 		panic(fmt.Sprintf("RegisterMapHasher(%v) of unknown hasher", h))
 	}
+	if mapHashers[h] != nil {
+		panic(fmt.Sprintf("%v already registred as a MapHasher", h))
+	}
 	mapHashers[h] = f
 }
 
-// GetLogHasher returns a LogHasher
+// GetLogHasher returns a LogHasher.
 func GetLogHasher(h trillian.HashStrategy) (LogHasher, error) {
 	f := logHashers[h]
 	if f != nil {
@@ -90,7 +85,7 @@ func GetLogHasher(h trillian.HashStrategy) (LogHasher, error) {
 	return nil, fmt.Errorf("LogHasher(%v) is unknown hasher", h)
 }
 
-// GetMapHasher returns a MapHasher
+// GetMapHasher returns a MapHasher.
 func GetMapHasher(h trillian.HashStrategy) (MapHasher, error) {
 	f := mapHashers[h]
 	if f != nil {
--- a/merkle/hashers/tree_hasher.go
+++ b/merkle/hashers/tree_hasher.go
@@ -81,8 +81,8 @@ func RegisterMapHasher(h trillian.HashStrategy, f MapHasher) {
 	mapHashers[h] = f
 }
 
-// LogHasher returns a LogHasher
-func LogHasher(h trillian.HashStrategy) (LogHasher, error) {
+// GetLogHasher returns a LogHasher
+func GetLogHasher(h trillian.HashStrategy) (LogHasher, error) {
 	f := logHashers[h]
 	if f != nil {
 		return f, nil
@@ -90,8 +90,8 @@ func LogHasher(h trillian.HashStrategy) (LogHasher, error) {
 	return nil, fmt.Errorf("LogHasher(%v) is unknown hasher", h)
 }
 
-// MapHasher returns a MapHasher
-func MapHasher(h trillian.HashStrategy) (MapHasher, error) {
+// GetMapHasher returns a MapHasher
+func GetMapHasher(h trillian.HashStrategy) (MapHasher, error) {
 	f := mapHashers[h]
 	if f != nil {
 		return f, nil
--- a/merkle/hashers/tree_hasher.go
+++ b/merkle/hashers/tree_hasher.go
@@ -50,7 +50,7 @@ type MapHasher interface {
 }
 
 var (
-	maxHashers = trillian.HashStrategy(maxKey(trillian.HashStrategy_name))
+	maxHashers = maxKey(trillian.HashStrategy_name) + 1
 	logHashers = make([]LogHasher, maxHashers)
 	mapHashers = make([]MapHasher, maxHashers)
 )
--- a/merkle/hashers/tree_hasher.go
+++ b/merkle/hashers/tree_hasher.go
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package merkle
+package hashers
+
+import (
+	"fmt"
+
+	"github.com/google/trillian"
+)
 
 // LogHasher provides the hash functions needed to compute dense merkele trees.
 type LogHasher interface {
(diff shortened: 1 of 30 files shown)


merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,47 @@ type MapHasher interface {
 	// TODO(gbelvin): Replace Size() with BitLength().
 	Size() int
 }
+
+var (
+	maxHashers = trillian.HashStrategy(len(trillian.HashStrategy_value))
```
codingllama: There's a dangerous assumption here that `len(trillian.HashStrategy_value)` is greater than the highest proto tag used. This works for now, but will break if we retire a value from the enum or if we simply skip tag numbers.

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,47 @@ type MapHasher interface {
...
+	logHashers = make([]LogHasher, maxHashers)
+	mapHashers = make([]MapHasher, maxHashers)
+)
+
+// RegisterLogHasher registers a hasher for use.
+func RegisterLogHasher(h trillian.HashStrategy, f LogHasher) {
+	if h < 0 || h >= maxHashers {
+		panic(fmt.Sprintf("RegisterLogHasher(%v) of unknown hasher", h))
```
codingllama: nit: If it's on the enum, then the value has to be known. On a related note, we probably shouldn't let people register hashers for `UNKNOWN_HASH_STRATEGY`.

I would probably prefer an error return to a panic as well.

Ditto for others.

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,47 @@ type MapHasher interface {
...
+// RegisterMapHasher registers a hasher for use.
+func RegisterMapHasher(h trillian.HashStrategy, f MapHasher) {
+	if h < 0 || h >= maxHashers {
+		panic(fmt.Sprintf("RegisterMapHasher(%v) of unknown hasher", h))
+	}
+	mapHashers[h] = f
+}
+
+// NewLogHasher returns a LogHasher
+func NewLogHasher(h trillian.HashStrategy) (LogHasher, error) {
```
codingllama: nit: Technically, it's not a new hasher. GetLogHasher, maybe?

Ditto for NewMapHasher.

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,47 @@ type MapHasher interface {
...
 }
+
+var (
+	maxHashers = trillian.HashStrategy(len(trillian.HashStrategy_value))
+	logHashers = make([]LogHasher, maxHashers)
+	mapHashers = make([]MapHasher, maxHashers)
+)
+
+// RegisterLogHasher registers a hasher for use.
+func RegisterLogHasher(h trillian.HashStrategy, f LogHasher) {
```
codingllama: This solution assumes a hasher is a function of HashStrategy, but in fact it's a mix of HashAlgorithm and HashStrategy. We had this right on the (now deleted) trees functions.

I don't mind about the interface changes, but I'd strongly prefer if you kept `trees.Hash` and `trees.XHasher` as they were. I don't think we need a registry for hashers either; we've only got a few (hand-coded) of them anyway.

merkle/hashers/tree_hasher.go
gdbelvin: Good point. This should be fixed now.

merkle/hashers/tree_hasher.go
gdbelvin: I wanted to ask you about this.  I saw hash algorithm being incorporated into the call to `rfc6962.New`.  

HashAlgorithm is used to determine the signature algorithm -- that's why it's a subfield of `DigitallySigned`. Could we consider not overloading the HashAlgorith field? In my experience, its better to include algorithms into the suite identifier, eg. RFC69626_SHA256. Would you be open to that change?  Happy to create a separate PR for it.

merkle/hashers/tree_hasher.go
gdbelvin: Good point. should I remove the check? 

I'm following the style of golang's crypto package here. Since this function should only be called in an init method, I assumed that a panic was more ok (ish). Ie. it should only be triggered right on launch.

merkle/hashers/tree_hasher.go
gdbelvin: Get* isn't very Go style friendly...  How's `LogHasher()`?

gdbelvin:
Let's chat about this tomorrow. @Martin2112  Would love your thoughts as well.

Martin2112:
Still doesn't build.

merkle/hashers/tree_hasher.go
codingllama: I'd suggest a map[HashStrategy]XHasher instead. No need to worry about proto tags that way.

merkle/hashers/tree_hasher.go
codingllama: SGTM.

merkle/hashers/tree_hasher.go
codingllama: Discussed offline, we'll go ahead with #698.

merkle/hashers/tree_hasher.go
codingllama: I think we should check for:

* hash == UNKNOWN_HASH
* hasher == nil (possibly?)
* hash previously registered (ie, warn about overrides)

Re panic or non-panic, would all callers do:

```go
if err := RegisterLogHash(h, hasher); err != nil {
  panic(err)
}
```

... or something to that effect? If yes, then I guess the panic inside Register() is appropriate.

codingllama:
I'm not totally sold on the need for a hasher registry, specially considering that we're unlikely to have many variations. I'd probably stick to a simpler approach, eg, a switch-case hardcoded for each type.

I won't hold you back if you prefer this, but could you explain why a simpler approach won't work?

merkle/hashers/tree_hasher.go
gdbelvin: done

gdbelvin: Turns out `LogHasher` conflicts with the interface name. 
Even though this isn't technically create new objects (because there's no need to), it seems like it fits in the New* category because it's like a factory method.  What do you think?

merkle/hashers/tree_hasher.go
gdbelvin: This is what's being done in the lines below.

gdbelvin: Good idea. Done.

merkle/hashers/tree_hasher.go
gdbelvin: - We're checking for `UNKNOWN_HASH`.  
- `hasher == nil` is ok, and will be caught when requested. 
- I'll add a check for overrides.

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,48 @@ type MapHasher interface {
...
+
+var (
+	logHashers = make(map[trillian.HashStrategy]LogHasher)
+	mapHashers = make(map[trillian.HashStrategy]MapHasher)
+)
+
+// RegisterLogHasher registers a hasher for use.
+func RegisterLogHasher(h trillian.HashStrategy, f LogHasher) {
+	if h == trillian.HashStrategy_UNKNOWN_HASH_STRATEGY {
+		panic(fmt.Sprintf("RegisterLogHasher(%v) of unknown hasher", h))
```
codingllama: nit: Either use `%s` or `h.String()`, so the message gets the enum name?

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,48 @@ type MapHasher interface {
...
+	}
+	if logHashers[h] != nil {
+		panic(fmt.Sprintf("%v already registred as a LogHasher", h))
```
codingllama: Ditto.

Same for others.

codingllama: Typo on "registered".

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,48 @@ type MapHasher interface {
...
+	logHashers[h] = f
+}
+
+// RegisterMapHasher registers a hasher for use.
+func RegisterMapHasher(h trillian.HashStrategy, f MapHasher) {
+	if h == trillian.HashStrategy_UNKNOWN_HASH_STRATEGY {
+		panic(fmt.Sprintf("RegisterMapHasher(%v) of unknown hasher", h))
+	}
+	if mapHashers[h] != nil {
+		panic(fmt.Sprintf("%v already registred as a MapHasher", h))
```
codingllama: Typo on "registered".

merkle/hashers/tree_hasher.go
```diff
@@ -42,3 +48,48 @@ type MapHasher interface {
...
+	mapHashers[h] = f
+}
+
+// GetLogHasher returns a LogHasher.
+func GetLogHasher(h trillian.HashStrategy) (LogHasher, error) {
+	f := logHashers[h]
+	if f != nil {
+		return f, nil
+	}
+	return nil, fmt.Errorf("LogHasher(%v) is unknown hasher", h)
```
codingllama: nit: "is unknown hasher" -> "is an unknown hasher".

Same below.

merkle/objhasher/objhasher.go
```diff
@@ -17,26 +17,26 @@ package objhasher
...
 
 type objloghasher struct {
-	merkle.LogHasher
+	hashers.LogHasher
 }
 
-// NewMapHasher returns a new ObjectHasher based on the passed in MapHasher
-func NewMapHasher(baseHasher merkle.MapHasher) merkle.MapHasher {
+// GetMapHasher returns a new ObjectHasher based on the passed in MapHasher
+func GetMapHasher(baseHasher hashers.MapHasher) hashers.MapHasher {
```
codingllama: nit: I think this one was better named as NewMapHasher.

Same below (NewLogHasher).

storage/testonly/admin_storage_tester.go
```diff
@@ -28,6 +28,7 @@ import (
 	ktestonly "github.com/google/trillian/crypto/keys/testonly"
 	"github.com/google/trillian/crypto/keyspb"
 	spb "github.com/google/trillian/crypto/sigpb"
+	_ "github.com/google/trillian/merkle/maphasher" // TESET_MAP_HASHER
```
codingllama: nit: Typo on TEST_MAP_HASHER

merkle/hashers/tree_hasher.go
gdbelvin: done

merkle/hashers/tree_hasher.go
gdbelvin: done

gdbelvin: done

merkle/hashers/tree_hasher.go
gdbelvin: done

merkle/hashers/tree_hasher.go
gdbelvin: yep. done

storage/testonly/admin_storage_tester.go
gdbelvin: done

gdbelvin:
Updated. PTAL.

PR #695, opened 2017-06-20
Remove interceptor's reliance on grpc.UnaryServerInfo

codingllama:
All data required for interception is now extracted from the request message.
This makes tests simpler (no need to specify the RPC's FullMethod) and makes it
easier to adapt interceptors to other environments.

### DIFF
--- a/server/interceptor/interceptor.go
+++ b/server/interceptor/interceptor.go
@@ -16,6 +16,7 @@
 package interceptor
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/google/trillian"
@@ -29,6 +30,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// requestTypePrefix is the expected prefix for all request protos.
+	requestTypePrefix = "*trillian."
+)
+
 // TrillianInterceptor checks that:
 // * Requests addressing a tree have the correct tree type and tree state;
 // * TODO(codingllama): Requests are properly authenticated / authorized ; and
@@ -39,10 +45,12 @@ type TrillianInterceptor struct {
 }
 
 // UnaryInterceptor executes the TrillianInterceptor logic for unary RPCs.
-func (i *TrillianInterceptor) UnaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-	quotaUser := i.QuotaManager.GetUser(ctx, req)
+func (i *TrillianInterceptor) UnaryInterceptor(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	// IMPORTANT: Do not rely on grpc.UnaryServerInfo in this filter. It makes life a lot harder
+	// when adapting the code to our internal branch.
 
-	rpcInfo, err := getRPCInfo(req, info.FullMethod, quotaUser)
+	quotaUser := i.QuotaManager.GetUser(ctx, req)
+	rpcInfo, err := getRPCInfo(req, quotaUser)
 	if err != nil {
 		return nil, err
 	}
@@ -78,17 +86,11 @@ type rpcInfo struct {
 }
 
 // getRPCInfo returns the rpcInfo for the given request, or an error if the request is not mapped.
-// Full method is the full RPC method name, as acquired from grpc.UnaryServerInfo (e.g.,
-// /trillian.TrillianLog/GetInclusionProof).
 // RPCs are mapped using the following logic:
-// Tree IDs are acquired via one of the "Request" interfaces defined below (treeIDRequest,
+// treeID is acquired via one of the "Request" interfaces defined below (treeIDRequest,
 // logIDRequest, mapIDRequest, etc). Requests must implement to one of those.
-// Tree type is determined by the RPC service name: TrillianAdmin is unrestricted,
-// TrillianLog = LOG, TrillianMap = MAP.
-// Readonly status is determined by the RPC method name: if it starts with Get or List it's
-// considered readonly.
-// Finally, a few RPCs are hand-mapped as doesNotHaveTree, such as Create and ListTree.
-func getRPCInfo(req interface{}, fullMethod, quotaUser string) (*rpcInfo, error) {
+// TreeType and Readonly are determined based on the request type.
+func getRPCInfo(req interface{}, quotaUser string) (*rpcInfo, error) {
 	var treeID int64
 	switch req := req.(type) {
 	case *trillian.CreateTreeRequest:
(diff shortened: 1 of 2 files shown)


server/interceptor/interceptor.go
```diff
@@ -39,10 +45,12 @@ type TrillianInterceptor struct {
 }
 
 // UnaryInterceptor executes the TrillianInterceptor logic for unary RPCs.
-func (i *TrillianInterceptor) UnaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-	quotaUser := i.QuotaManager.GetUser(ctx, req)
+func (i *TrillianInterceptor) UnaryInterceptor(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	// IMPORTANT: Do not rely on grpc.UnaryServerInfo in this filter. It makes life a lot harder
+	// when adapting the code to our internal branch.
```
daviddrysdale: Maybe "when adapting the code to other environments."  (ditto for commit message), to avoid an us/them distinction.

server/interceptor/interceptor.go
```diff
@@ -107,24 +109,24 @@ func getRPCInfo(req interface{}, fullMethod, quotaUser string) (*rpcInfo, error)
...
-	case "trillian.TrillianMap":
+	case isMapRPC(req):
 		treeType = trillian.TreeType_MAP
 	default:
-		return nil, status.Errorf(codes.Internal, "cannot determine treeType for request: %T", req)
+		return nil, status.Errorf(codes.Internal, "unmapped request type: %T", req)
 	}
 
-	readonly := strings.HasPrefix(methodName, "Get") || strings.HasPrefix(methodName, "List")
+	reqType := fmt.Sprintf("%T", req)
```
daviddrysdale: I'm not keen on this approach -- the magic reflection stuff makes it hard for tools (and humans) to spot the connection between the RPC types and their policing.

Given there is already a set of type switches below, why not extend them to return (e.g.) `(isAdmin bool, readonly bool)`?

server/interceptor/interceptor.go

server/interceptor/interceptor.go
codingllama: I tried a bunch of variations of "isAdmin" and others before deciding on this one. The advantage here is that it doesn't need a lot of human intervention: if you add a new RPC, you add it to the appropriate list and that's it. There isn't a huge mapping of request to {treeID, treeType, readonly} that has to be written (and reviewed) by humans. We have a simple list of requests and a consistent set of rules, applied across all requests.

codingllama:
PTAL?

server/interceptor/interceptor.go
daviddrysdale: I don't see that this approach has significantly less human intervention than what I suggest -- in either case you need to add an entry to an existing list, whether it's:
```go
func isAdminRPC(req interface{}) bool {
	switch req.(type) {
	// Please keep in alphabetical order
	case *trillian.CreateTreeRequest:
	case *trillian.DeleteTreeRequest:
	case *trillian.GetTreeRequest:
	case *trillian.ListTreesRequest:
	case *trillian.UpdateTreeRequest:
	default:
		return false
	}
	return true
}
```
or:
```go
func isAdminRPC(req interface{}) (bool, bool) {
	isAdmin := false
	readonly := false
	switch req.(type) {
	case *trillian.CreateTreeRequest, *trillian.DeleteTreeRequest, *trillian.UpdateTreeRequest:
		isAdmin = true
		readonly = false
	case *trillian.GetTreeRequest, *trillian.ListTreesRequest:
		isAdmin = true
		readonly = true
	}
	return isAdmin, readonly
}
```
is hardly a big deal (and note I'm not suggesting a `treeID` mapping, so that's a red herring).  

What is more significant for maintainability is that the convention that `List..` / `Get...` methods are readonly is more explicit (and greppable), and less likely to go wrong if someone foolishly chooses the wrong name.

Fundamentally, I think that doing reflection and string matching on the names of types should be a last resort, reserved for when there's no other way to get the same effect.

gdbelvin:
Could we consider a proto annotation to better document what rpcs are
privileged?

server/interceptor/interceptor.go
codingllama: OK, fair enough about no reflection if not necessary.

About "someone foolishly choos(ing) the wrong name" for an RPC: I'd be really surprised to have this happen, and even more so if we let the error stand. A non-readonly Get or List method is a significant offence to API style. But, I've taken the comment, so let's leave it at that.

codingllama:
@gdbelvin, I'm not sure what you mean by privileged in this context, but I think proto annotations are not a good medium for the data we're dealing with in this PR.

server/interceptor/interceptor.go
```diff
@@ -29,6 +28,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// requestTypePrefix is the expected prefix for all request protos.
+	requestTypePrefix = "*trillian."
```
daviddrysdale: Is this still used?

server/interceptor/interceptor.go
```diff
@@ -151,17 +137,69 @@ func getRPCInfo(req interface{}, fullMethod, quotaUser string) (*rpcInfo, error)
...
+	return trillian.TreeType_UNKNOWN_TREE_TYPE, false, fmt.Errorf("unmapped request type: %T", req)
+}
+
+func getAdminRequestInfo(req interface{}) (bool, bool) {
+	isAdmin := true
+	readonly := false
+	switch req.(type) {
+	case *trillian.CreateTreeRequest,
+		*trillian.DeleteTreeRequest,
+		*trillian.UpdateTreeRequest:
```
daviddrysdale: Personal preference, but I think these would be easier to read if the `isXxx` flag is set in the case arms, rather than being set by default then cleared in a default arm.  You could even get to use the rarely-spotted `fallthrough` keyword!:
```go
func getAdminRequestInfo(req interface{}) (bool, bool) {
	isAdmin := false
	readonly := false
	switch req.(type) {
	case *trillian.GetTreeRequest,
		*trillian.ListTreesRequest:
		readonly = true
		fallthrough
	case *trillian.CreateTreeRequest,
		*trillian.DeleteTreeRequest,
		*trillian.UpdateTreeRequest:
		isAdmin = true
	}
	return isAdmin, readonly
}
```

server/interceptor/interceptor.go
```diff
@@ -151,17 +137,69 @@ func getRPCInfo(req interface{}, fullMethod, quotaUser string) (*rpcInfo, error)
...
+	case *trillian.GetTreeRequest,
+		*trillian.ListTreesRequest:
+		readonly = true
```
daviddrysdale: (nit) the other 2 function have the `readonly` cases before the mutating cases.

server/interceptor/interceptor.go
codingllama: Well spotted, removed.

server/interceptor/interceptor.go

server/interceptor/interceptor.go
codingllama: fallthrough is not allowed in type switch. I've kept it as is.

codingllama:
Thanks, David!

server/interceptor/interceptor.go
daviddrysdale: Ah, I didn't know that.

PR #696, opened 2017-06-21
Add a caching quota.Manager decorator

codingllama:
The cached quota.Manager keeps an in-memory count of tokens per quota.Spec. It
may be used together with any quota.Manager implementation that is friendly to
token caching.

### DIFF
--- a/quota/cache/cache.go
+++ b/quota/cache/cache.go
@@ -29,8 +29,8 @@ import (
 var now = time.Now
 
 type manager struct {
-	qm                    quota.Manager
-	batchSize, maxEntries int
+	qm                       quota.Manager
+	minBatchSize, maxEntries int
 
 	// mu guards cache
 	mu    sync.Mutex
@@ -47,18 +47,18 @@ type bucket struct {
 
 // NewCachedManager wraps a quota.Manager with an implementation that caches tokens locally.
 //
-// batchSize determines the minimum number of tokens requested from qm for each GetTokens() request.
-// More tokens may be requested in order to fulfill the present request.
+// minBatchSize determines the minimum number of tokens requested from qm for each GetTokens()
+// request.
 //
 // maxEntries determines the maximum number of cache entries, apart from global quotas. The oldest
 // entries are evicted as necessary, their tokens replenished via PutTokens() to avoid excessive
 // leakage.
-func NewCachedManager(qm quota.Manager, batchSize, maxEntries int) quota.Manager {
+func NewCachedManager(qm quota.Manager, minBatchSize, maxEntries int) quota.Manager {
 	return &manager{
-		qm:         qm,
-		batchSize:  batchSize,
-		maxEntries: maxEntries,
-		cache:      make(map[quota.Spec]*bucket),
+		qm:           qm,
+		minBatchSize: minBatchSize,
+		maxEntries:   maxEntries,
+		cache:        make(map[quota.Spec]*bucket),
 	}
 }
 
@@ -102,9 +102,10 @@ func (m *manager) GetTokens(ctx context.Context, numTokens int, specs []quota.Sp
 			}()
 		}()
 
-		// A more accurate count would be numTokens+m.batchSize-bucket.tokens, but that might force
-		// us to make a GetTokens call for each spec. A single call is likely to be more efficient.
-		tokens := numTokens + m.batchSize
+		// A more accurate count would be numTokens+m.minBatchSize-bucket.tokens, but that might
+		// force us to make a GetTokens call for each spec. A single call is likely to be more
+		// efficient.
+		tokens := numTokens + m.minBatchSize
 		if err := m.qm.GetTokens(ctx, tokens, specsToRefill); err != nil {
 			return err
 		}
--- a/quota/cache/cache.go
+++ b/quota/cache/cache.go
@@ -1,0 +1,200 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cache contains a caching quota.Manager implementation.
+package cache
+
+import (
+	"context"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/quota"
+)
+
+// now is used in place of time.Now to allow tests to take control of time.
+var now = time.Now
+
+type manager struct {
+	qm                    quota.Manager
+	batchSize, maxEntries int
+
+	// mu guards cache
+	mu    sync.Mutex
+	cache map[quota.Spec]*bucket
+
+	// evictWg tracks evict() goroutines.
+	evictWg sync.WaitGroup
+}
+
+type bucket struct {
+	tokens       int
+	lastModified time.Time
+}
+
+// NewCachedManager wraps a quota.Manager with an implementation that caches tokens locally.
+//
+// batchSize determines the minimum number of tokens requested from qm for each GetTokens() request.
+// More tokens than batchSize may be requested in order to fulfill the present request.
+//
+// maxEntries determines the maximum number of cache entries, apart from global quotas. The oldest
+// entries are evicted as necessary, their tokens replenished via PutTokens() to avoid excessive
+// leakage.
+func NewCachedManager(qm quota.Manager, batchSize, maxEntries int) quota.Manager {
+	return &manager{
+		qm:         qm,
+		batchSize:  batchSize,
+		maxEntries: maxEntries,
+		cache:      make(map[quota.Spec]*bucket),
+	}
+}
+
+func (m *manager) GetUser(ctx context.Context, req interface{}) string {
+	return m.qm.GetUser(ctx, req)
+}
+
+func (m *manager) PeekTokens(ctx context.Context, specs []quota.Spec) (map[quota.Spec]int, error) {
+	return m.qm.PeekTokens(ctx, specs)
+}
+
+func (m *manager) PutTokens(ctx context.Context, numTokens int, specs []quota.Spec) error {
+	return m.qm.PutTokens(ctx, numTokens, specs)
+}
+
+func (m *manager) ResetQuota(ctx context.Context, specs []quota.Spec) error {
+	return m.qm.ResetQuota(ctx, specs)
+}
+
+func (m *manager) GetTokens(ctx context.Context, numTokens int, specs []quota.Spec) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// Verify which buckets need more tokens, if any
+	specsToRefill := []quota.Spec{}
+	for _, spec := range specs {
+		bucket, ok := m.cache[spec]
+		if !ok || bucket.tokens < numTokens {
+			specsToRefill = append(specsToRefill, spec)
+		}
+	}
+
+	// Request the required number of tokens and add them to buckets
+	if len(specsToRefill) != 0 {
+		defer func() {
+			// Do not hold GetTokens on eviction, it won't change the result.
+			m.evictWg.Add(1)
+			go func() {
+				m.evict(ctx)
+				m.evictWg.Done()
+			}()
+		}()
+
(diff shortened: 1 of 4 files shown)


codingllama:
Issue #490.

codingllama:
Ping?

codingllama:
Friendly ping?

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+			m.evictWg.Add(1)
+			go func() {
+				m.evict(ctx)
+				m.evictWg.Done()
+			}()
+		}()
+
+		// A more accurate count would be numTokens+m.batchSize-bucket.tokens, but that might force
+		// us to make a GetTokens call for each spec. A single call is likely to be more efficient.
+		tokens := numTokens + m.batchSize
```
RJPercival: I'm not sure this is the behaviour someone would expect from reading the documentation of `NewCachedManager()`. I'd expect something more like `math.Max(numTokens, m.batchSize)`, or perhaps `(numTokens / m.batchSize + 1) * m.batchSize`.

RJPercival: Having said that, I'm ok with doing this, but maybe the documentation should be reworded slightly to be more explicit about the actual behaviour.

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+		if err := m.qm.GetTokens(ctx, tokens, specsToRefill); err != nil {
```
RJPercival: Happy to keep `m.mu` locked for the time it takes to get a response to an RPC? That seems like a long time to keep a mutex locked, and blocks additional `GetTokens()` calls in the meantime, but I can see how you could end up making lots of duplicate GetTokens requests otherwise.

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+
+	// Subtract tokens from cache
+	lastModified := now()
+	for _, spec := range specs {
+		bucket, ok := m.cache[spec]
+		// Sanity check
+		if !ok || bucket.tokens < 0 || bucket.tokens < numTokens {
+			glog.Warningf("Bucket invariants failed for spec %+v: ok = %v, bucket = %+v", spec, ok, bucket)
```
RJPercival: I'd make that `glog.Errorf()`, since for this to happen, something is very wrong.

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+	// m.mu is explicitly unlocked, so we don't have to hold it while we wait for goroutines to
+	// complete.
+
+	if len(m.cache) <= m.maxEntries {
+		m.mu.Unlock()
+		return
+	}
+
+	// Find and evict the oldest entries. To avoid excessive token leakage, let's try and
+	// replenish the tokens held for the evicted entries.
```
RJPercival: s/replenish/return/

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+		if spec.Group != quota.Global {
+			buckets = append(buckets, specBucket{bucket: b, spec: spec})
+		}
+	}
+	sort.Sort(buckets)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < len(m.cache)-m.maxEntries; i++ {
+		b := buckets[i]
+		glog.Infof("Evicting bucket for spec %+v (%v tokens held)", b.spec, b.tokens)
```
RJPercival: Given the frequency this could occur at, it's probably best as verbose logging, e.g. `glog.V(1).Infof()`.

RJPercival: This log message could probably be worded in a friendlier way for people not familiar with the quota code, e.g. 
```go
glog.V(1).Infof("Too many tokens cached - returning least recently used (%v tokens for %+v)", b.tokens, b.spec)
```

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+		delete(m.cache, b.spec)
+
+		// goroutines must not access the cache, the lock is released before they complete.
+		wg.Add(1)
+		go func() {
+			if err := m.qm.PutTokens(ctx, b.tokens, []quota.Spec{b.spec}); err != nil {
+				glog.Warningf("Error replenishing tokens from evicted bucket (spec = %+v, bucket = %+v): %v", b.spec, b.bucket, err)
```
RJPercival: I think you'd get almost the same message if you replace "spec = %+v, bucket = %+v" with "%+v" and `b.spec, b.bucket` with `b`.

quota/cache/cache.go
RJPercival: Actually, replenish is fine. I see that it's the accepted terminology for this when dealing with quotas and tokens.

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+			wg.Done()
+		}()
+	}
+
+	m.mu.Unlock()
+	wg.Wait()
```
RJPercival: Why does `evict()` have to wait for these goroutines?

quota/cache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+			buckets = append(buckets, specBucket{bucket: b, spec: spec})
+		}
+	}
+	sort.Sort(buckets)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < len(m.cache)-m.maxEntries; i++ {
+		b := buckets[i]
+		glog.Infof("Evicting bucket for spec %+v (%v tokens held)", b.spec, b.tokens)
+		delete(m.cache, b.spec)
```
RJPercival: Doing a delete here means the loop will only run over approximately half as many entries as you'd expect. This is because `len(m.cache)` will count down while `i` counts upwards. The tests don't expose this because there is an eviction goroutine per bucket (you never call GetTokens with more than one spec), and this loop will always run at least one iteration.

RJPercival: Here's a patch that'll exhibit the problem:
```diff
--- a/quota/cache/cache_test.go
+++ b/quota/cache/cache_test.go
@@ -171,17 +171,18 @@ func TestCachedManager_GetTokens_EvictsCache(t *testing.T) {
                }
                tree++
        }
 
        // All entries added now must cause eviction of the oldest entries
-       evicts := 5
-       for i := 0; i < evicts; i++ {
+       evicts := 20
+       for i := 0; i < evicts; i += 2 {
                mock.EXPECT().PutTokens(ctx, batchSize, specForTree(firstTree+int64(i))).Return(nil)
-               if err := qm.GetTokens(ctx, tokens, specForTree(tree)); err != nil {
+               mock.EXPECT().PutTokens(ctx, batchSize, specForTree(firstTree+int64(i+1))).Return(nil)
+               if err := qm.GetTokens(ctx, tokens, append(specForTree(tree), specForTree(tree+1)...)); err != nil {
                        t.Fatalf("GetTokens() returned err = %v (i = %v)", err, i)
                }
-               tree++
+               tree += 2
        }
 
        qm.(*manager).wait()
 }
```
Run the test with `go-test ./quota/cache --test.run CachedManager_GetTokens_EvictsCache -test.count 10` (i.e. multiple times), as the error is non-deterministic (it depends on the eviction goroutines not always running immediately between GetTokens calls).

quota/cache/cache_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	ctx := context.Background()
+	tokens := 5
+	for _, want := range []error{nil, errors.New("llama ate all tokens")} {
+		mock := quota.NewMockManager(ctrl)
+		qm := NewCachedManager(mock, batchSize, maxEntries)
+
+		mock.EXPECT().PutTokens(ctx, tokens, specs).Return(want)
+		if err := qm.PutTokens(ctx, tokens, specs); err != want {
+			t.Errorf("PutTokens() returned err = %v, want = %v", err, want)
```
RJPercival: %#v might be better here, as it will display `nil` as "<nil>", rather than "".

quota/cache/cache_test.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func TestCachedManager_GetTokens_CachesTokens(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ctx := context.Background()
+	tokens := 3
+	mock := quota.NewMockManager(ctrl)
+	mock.EXPECT().GetTokens(ctx, tokens+batchSize, specs).Times(2).Return(nil)
```
RJPercival: To make the test less brittle, this should just expect that GetTokens is called with at least `tokens` or `batchSize`, whichever is greater (that's all that the documentation guarantees).

quota/cache/cache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			t.Fatalf("GetTokens() returned err = %v (i = %v)", err, i)
+		}
+		tree++
+	}
+
+	// All entries added now must cause eviction of the oldest entries
+	evicts := 5
+	for i := 0; i < evicts; i++ {
+		mock.EXPECT().PutTokens(ctx, batchSize, specForTree(firstTree+int64(i))).Return(nil)
+		if err := qm.GetTokens(ctx, tokens, specForTree(tree)); err != nil {
```
RJPercival: Should test this when more than 1 spec is passed to GetTokens.

quota/cache/cache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			t.Fatalf("GetTokens() returned err = %v (i = %v)", err, i)
+		}
+		tree++
+	}
+
+	qm.(*manager).wait()
```
RJPercival: The test will deadlock if an unexpected function on the mock is called by an eviction goroutine (or any of the goroutines that they start), because the goroutine will panic and therefore not call `wg.Done()` (it isn't deferred) . That's not *necessarily* worth defending against, but something to be aware of.

quota/cache/cache.go
codingllama: I tried to acknowledge this with the following sentence in NewCachedManager: "More tokens than batchSize may be requested in order to fulfill the present request". I didn't want to exactly document the algorithm there, though, so we have some leeway in how to implement it.

quota/cache/cache.go
codingllama: Yes, I decided to go for the simpler approach because:

1. It's easier to maintain the contract without the possibility of concurrent RPCs (otherwise we may get multiple token batches at once)
1. It's likely that the leftover tokens of the current RPC are enough to address the other blocked calls. As they would block on an RPC anyway, blocking on mu shouldn't be slower.

quota/cache/cache.go
codingllama: Done.

quota/cache/cache.go
codingllama: Ack.

quota/cache/cache.go
codingllama: I've taken the message, but didn't do the verbose logging. I don't think eviction should be too frequent, so I'd rather see it on the logs if it is.

The cache grows based on the number of trees and number of users. The number of trees is likely to be small for most Trillian installations (a handful, maybe?) and quota users are not implemented. If we go the route of using root certs / intermediates as users, the number gets a bit higher, but it may still be realistically cacheable.

quota/cache/cache.go
codingllama: b.bucket is a pointer, so ("%+v", b) prints the pointer address :/

quota/cache/cache.go
codingllama: Currently, it's mostly so tests can tell when eviction ends (via `m.wait()`) . Having evict() block on the goroutines it spawns allows us to correctly account for in-flight evict calls at line 101 (`m.evictWg.Done()`).

quota/cache/cache_test.go
codingllama: Done for all messages where errs might be nil.

quota/cache/cache_test.go
codingllama: Done with `AtLeast(batchSize)`, which is what the documentation states.

quota/cache/cache_test.go
codingllama: Done, moved the wait to a select with a timeout.

quota/cache/cache.go
RJPercival: I think it's the name "batchSize" that is misleading. It suggests (to me) that tokens are fetched in batches of that size, so to not round up to it, or a multiple of it, is surprising. Maybe it's just me though. Perhaps just calling it "minBatchSize" or "minTokens" would be better.

quota/cache/cache.go
RJPercival: :disappointed:

quota/cache/cache.go
codingllama: Done. Good call.

quota/cache/cache_test.go
codingllama: Done.

codingllama:
PTAL?

Issue #698, opened 2017-06-22
Specify Algorithm and Hash function together in enums


gdbelvin:
Kill the hash algorithm in the tree spec.

Martin2112:
This is fixed I think?

gdbelvin:
This was partially implemented, but not completed. 
Future work should focus on a potential migration to google/tink

Martin2112:
OK I think we can close this one then and `tink` can be a new thing.

PR #699, opened 2017-06-22
Specify algorithms in cipher suites.

gdbelvin:
This PR specifies allowed combinations of algorithms for signing and tree hashing. 

First step towards #698


crypto/sigpb/sigpb.proto
```diff
@@ -44,10 +44,24 @@ message DigitallySigned {
     ECDSA = 3;
   }
 
+  // SignatureCipherSuite defines the set of algorithms used for signing.
+  enum SignatureCipherSuite {
```
codingllama: Have you considered moving this message to the top level? It's used in other contexts, outside of DigitallySigned, so it could stand on its own.

(Same would apply for HashAlgo and SigAlgo, assuming we wouldn't delete them.)

trees/trees.go
```diff
@@ -119,7 +119,7 @@ func LogHasher(tree *trillian.Tree) (merkle.LogHasher, error) {
 	}
 
 	switch tree.HashStrategy {
-	case trillian.HashStrategy_RFC_6962:
+	case trillian.HashStrategy_RFC6962_SHA256:
 		return rfc6962.New(hash), nil
```
codingllama: This should be `return rfc6962.New(crypto.SHA256), nil`. No need to call `Hash(tree)` above either.

crypto/sigpb/sigpb.proto
```diff
@@ -44,10 +44,24 @@ message DigitallySigned {
...
+  }
+
   // hash_algorithm contains the hash algorithm used.
+  // TODO(gdbelvin): deprecate
   HashAlgorithm hash_algorithm = 1;
   // sig_algorithm contains the signing algorithm used.
+  // TODO(gdbelvin): deprecate 
   SignatureAlgorithm signature_algorithm = 2;
+  // cipher_suite contains the combination of ciphers used for this signature.
+  SignatureCipherSuite cipher_suite = 4;
```
codingllama: Rename to `signature_cipher_suite`, like trillian.Tree?

crypto/sigpb/sigpb.proto
```diff
@@ -44,10 +44,24 @@ message DigitallySigned {
...
+    // Signature and hash algorithm are unknown / not set. 
+    UNKNOWN_CIPHER_SUITE = 0;
+    // RSA with SHA256.
+    RSA_SHA256 = 1;
+    // ECDSA with SHA256.
+    ECDSA_SHA256 = 2;
+  }
+
   // hash_algorithm contains the hash algorithm used.
+  // TODO(gdbelvin): deprecate
```
codingllama: nit: Maybe "Deprecate in favor of signature_cipher_suite"?

crypto/sigpb/sigpb.proto
```diff
@@ -44,10 +44,24 @@ message DigitallySigned {
...
   HashAlgorithm hash_algorithm = 1;
   // sig_algorithm contains the signing algorithm used.
+  // TODO(gdbelvin): deprecate
```
codingllama: Ditto ("deprecate in favor of signature_cipher_suite")

trillian.proto
```diff
@@ -109,12 +109,17 @@ message Tree {
 
   // Hash algorithm to be used by the tree.
   // Readonly.
+  // TODO(gdbelvin): Deprecate.
```
codingllama: nit: "Deprecate in favor of hash_strategy"?

trillian.proto
```diff
@@ -109,12 +109,17 @@ message Tree {
...
   sigpb.DigitallySigned.HashAlgorithm hash_algorithm = 5;
 
   // Signature algorithm to be used by the tree.
   // Readonly.
+  // TODO(gdbelvin): Deprecate.
```
codingllama: nit: "Deprecate in favor of signature_cipher_suite"?

crypto/sigpb/sigpb.proto
gdbelvin: It used to be at the top level, but I thought it should be specific to signature generation so I moved it here.  Perhaps other contexts should define their own enums?

crypto/sigpb/sigpb.proto
gdbelvin: done

crypto/sigpb/sigpb.proto
gdbelvin: done

crypto/sigpb/sigpb.proto
gdbelvin: This is a member of `DigitallySigned` I would think the context implies signature, or is there another reason to use a longer name?

gdbelvin: done

trees/trees.go
gdbelvin: done

trillian.proto
gdbelvin: done

trillian.proto
gdbelvin: done

trillian.proto
```diff
@@ -109,12 +109,12 @@ message Tree {
 
   // Hash algorithm to be used by the tree.
   // Readonly.
-  // TODO(gdbelvin): Deprecate.
+  // TODO(gdbelvin): Deprecate in favor of signature_cipher_suite.
```
codingllama: nit: In favor of hash_strategy, in this case.

crypto/sigpb/sigpb.proto
codingllama: If the enums could deviate, declaring copies would be OK, but it doesn't seem we want that. I do think a top level enum would be more appropriate, considering it's used outside DigitallySigned, but I won't push any further.

trillian.proto
gdbelvin: Well, this field is overloaded, so it's both :-)  
I'll add a note.

PR #702, opened 2017-06-22
Add some flags/ENV vars to log-signer's Dockerfile

### LINKED ISSUE #701: Add some useful flags to the Dockerfile
As we currently deploy keytransparency to gke using kubernetes, it would be great to experiment with some of the available flags. In particular: 
```
--sequencerIntervalFlag
--numSeqFlag
--batchSizeFlag
```

Liamsi:
It would be helpful if one could change these flags for running kubernetes pods. This pull adds a few environment variables to the Dockerfile to make this possible.

Resolves #701

### DIFF
--- a/server/trillian_log_signer/Dockerfile
+++ b/server/trillian_log_signer/Dockerfile
@@ -8,11 +8,11 @@ ENV DB_USER=test \
 ENV HOST=0.0.0.0 \
     HTTP_PORT=8091
 
-ENV SEQUENCER_GUARD_WINDOW=0s
-ENV FORCE_MASTER=true
-ENV SEQUENCER_INTERVAL=10s
-ENV NUM_SEQ_FLAG=10
-ENV BATCH_SIZE=50
+ENV SEQUENCER_GUARD_WINDOW=0s \
+    FORCE_MASTER=true \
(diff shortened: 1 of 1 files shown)


server/trillian_log_signer/Dockerfile
```diff
@@ -8,8 +8,11 @@ ENV DB_USER=test \
 ENV HOST=0.0.0.0 \
     HTTP_PORT=8091
 
-ENV SEQUENCER_GUARD_WINDOW=0s \
-    FORCE_MASTER=true
+ENV SEQUENCER_GUARD_WINDOW=0s
```
gdbelvin: Docker creates a new "layer" for each command in the docker file. 
If you group the commands (via `\`) you can save a bunch of time when building.

Liamsi: Thanks. Done. 
Interestingly, travis now complains that the pb generated files differ from what is checked in: https://travis-ci.org/google/trillian/jobs/246141769#L380 
seems unrrelated to this change, though.

codingllama: FYI: https://github.com/google/trillian/pull/704

PR #706, opened 2017-06-26
Map: remove storage dependency in verifier

gdbelvin:
This PR cleans up map_verifier so that it does not depend on the storage layer.

Fixes:
```
// TODO(al): Remove this dep on storage, since clients will want to use this code.
```


Martin2112:
This looks like a near dup of the other PR? It has the same issue of turning bits into strings.

gdbelvin:
Yeah, I'll just work on the other PR and close this one

PR #708, opened 2017-06-26
Choose ECDSA for default signature algorithm

gdbelvin:
Choose a sane signature algorithm default in createtree.


gdbelvin:
We'll use `curl` as recommended.  Closing

PR #709, opened 2017-06-27
Add http handlers for simple log and map operations.

gdbelvin:
Use the grpc-gateway to make common functions available in the browser.
Greatly simplifies debugging and opens up the possibility for better
http health monitoring too.

The proposed http paths are mere initial ideas. Comments welcome on how to improve, especially for compliance with the style guide.


server/main.go
```diff
@@ -72,6 +72,12 @@ func (m *Main) Run(ctx context.Context) error {
 		if err := trillian.RegisterTrillianAdminHandlerFromEndpoint(ctx, mux, m.RPCEndpoint, opts); err != nil {
 			return err
 		}
+		if err := trillian.RegisterTrillianLogHandlerFromEndpoint(ctx, mux, m.RPCEndpoint, opts); err != nil {
```
codingllama: We shouldn't register Log / Map gateway handlers here, as we don't know here which are really available. (We register Admin because it's present in all servers.)

Instead, do this by taking advantage of RegisterHandlerFunc. See links for [logserver](https://github.com/google/trillian/blob/master/server/trillian_log_server/main.go#L111) and [mapserver](https://github.com/google/trillian/blob/master/server/trillian_map_server/main.go#L90).

trillian_log_api.proto
```diff
@@ -182,23 +183,42 @@ service TrillianLog {
 
     // No direct equivalent at the storage level
     rpc GetInclusionProof (GetInclusionProofRequest) returns (GetInclusionProofResponse) {
+      option (google.api.http) = {
+        get: "/v1beta1/log/{log_id}/size/{tree_size}/inclusion/index/{leaf_index}"
```
codingllama: Regarding the URL scheme:

* Use plural for collections, ie, **/logs** instead of **/log** (https://cloud.google.com/apis/design/resource_names#collection_id)

* Size doesn't feel like a collection. There isn't a "Size" resource. I think **/logs/\_/leaves/\_** fits better, with size being a regular parameter when appropriate.

I would go for the following collections, using custom methods for most of the non-standard reads:

**/v1beta1/logs/{log_id}**
**/v1beta1/logs/{log_id}/leaves/{leaf_index}**
**/v1beta1/logs/{log_id}/roots**

That gives us the following:

GetInclusionProof:
GET **/v1beta1/logs/{log_id}/leaves/{leaf_index}:inclusion_proof**

GetInclusionProofByHash:
GET **/v1beta1/logs/{log_id}/leaves/{leaf_index}:inclusion_proof_by_hash**

GetConsistencyProof:
GET **/v1beta1/logs/{log_id}:consistency_proof**

GetLatestSignedLogRoot:
GET **/v1beta1/logs/{log_id}/roots:latest**

GetSequencedLeafCount:
GET **/v1beta1/logs/{log_id}/leaves:sequenced_count**

GetEntryAndProofRequest:
GET **/v1beta1/logs/{log_id}/leaves/{leaf_index}**

(I'm ignoring the methods you chose not to map as well.)

TrillianLog and TrillianMap's design is not resource oriented, so not all APIs map as nicely as I'd hope for. Ideally I'd like to give this more thought, but I guess we can go forward with the mapping as long as there's an understanding that it's not stable and is likely to change in the future. To be truly compliant to Google's API style we'd have to change most of our request and response messages as well.

Note that the paragraph above doesn't apply to the Admin (and soon to be) Quota APIs, which are resource oriented from the start and fit nicely in this model.

trillian_log_api.proto
```diff
@@ -182,23 +183,42 @@ service TrillianLog {
...
+      };
     }
     rpc GetInclusionProofByHash (GetInclusionProofByHashRequest) returns (GetInclusionProofByHashResponse) {
+      // Enable when grpc-ecosystem/grpc-gateway#423 is resolved. Bytes is not currently supported as a URL parameter.
```
codingllama: See the comment above: we don't need the hash to be part of the URL.

trillian_map_api.proto
```diff
@@ -81,6 +82,14 @@ service TrillianMap {
   // For indexes that do not exist, the inclusion proof will use nil for the empty leaf value.
   rpc GetLeaves(GetMapLeavesRequest) returns(GetMapLeavesResponse) {}
   rpc SetLeaves(SetMapLeavesRequest) returns(SetMapLeavesResponse) {}
-  rpc GetSignedMapRoot(GetSignedMapRootRequest) returns(GetSignedMapRootResponse) {}
-  rpc GetSignedMapRootByRevision(GetSignedMapRootByRevisionRequest) returns(GetSignedMapRootResponse) {}
+  rpc GetSignedMapRoot(GetSignedMapRootRequest) returns(GetSignedMapRootResponse) {
```
codingllama: Similarly as the comment above, I'd go for:

GetSignedMapRoot:
GET **/v1beta1/maps/{map_id}/roots:latest**

GetSignedMapRootByRevision:
GET **/v1beta1/maps/{map_id}/roots:revision**

(I'm working under the assumption that {revision} is not the logical ID for map roots.)

server/main.go
gdbelvin: done

trillian_log_api.proto
gdbelvin: Thanks for your review! 

Agreed that the API wasn't designed to be resource oriented. Hopefully someday we can add a set of resource oriented methods and deprecate the ones that aren't.  I created this PR mostly so I could debug the log easily without resorting to `grpc_cli`.  The methods here are most certainly alpha mode and can change.

gdbelvin:
Updated the resource paths. Would you mind taking a look?

PR #710, opened 2017-06-27
Verify index length from hasher

gdbelvin:
Fixes `// TODO(gdbelvin): specify the index length in the tree specification.`

### DIFF
--- a/merkle/common.go
+++ b/merkle/common.go
@@ -26,6 +26,7 @@ func isRightChild(leafIndex int64) bool {
 }
 
 // bit returns the i'th bit of index from the right.
+// eg. bit(0x80000000, 31) -> 1
 func bit(index []byte, i int) uint {
 	IndexBits := len(index) * 8
 	bIndex := (IndexBits - i - 1) / 8
--- a/merkle/common_test.go
+++ b/merkle/common_test.go
@@ -32,6 +32,9 @@ func TestBit(t *testing.T) {
 		{index: h2b("000b"), i: 3, want: 1},
 		{index: h2b("0001"), i: 0, want: 1},
 		{index: h2b("8000"), i: 15, want: 1},
+		{index: h2b("0000000000000001"), i: 0, want: 1},
+		{index: h2b("0000000000010000"), i: 16, want: 1},
+		{index: h2b("8000000000000000"), i: 63, want: 1},
 	} {
 		if got, want := bit(tc.index, tc.i), tc.want; got != want {
 			t.Errorf("bit(%x, %d): %v, want %v", tc.index, tc.i, got, want)
--- a/merkle/common_test.go
+++ b/merkle/common_test.go
@@ -12,22 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package merkle provides Merkle tree manipulation functions.
 package merkle
 
-// parent returns the index of the parent node in the parent level of the tree.
-func parent(leafIndex int64) int64 {
-	return leafIndex >> 1
-}
+import (
+	"testing"
+)
 
-// isRightChild returns true if the node is a right child.
-func isRightChild(leafIndex int64) bool {
-	return leafIndex&1 == 1
-}
-
-// bit returns the i'th bit of index from the right.
-func bit(index []byte, i int) uint {
-	IndexBits := len(index) * 8
-	bIndex := (IndexBits - i - 1) / 8
-	return uint((index[bIndex] >> uint(i%8)) & 0x01)
+func TestBit(t *testing.T) {
+	for _, tc := range []struct {
+		index []byte
+		i     int
+		want  uint
+	}{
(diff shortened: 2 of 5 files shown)


merkle/map_verifier.go
```diff
@@ -30,34 +30,29 @@ import (
...
-	nID := storage.NewNodeIDFromHash(index)
 
 	runningHash := make([]byte, len(leafHash))
 	copy(runningHash, leafHash)
 
-	for bit := 0; bit < hBits; bit++ {
-		proofIsRightHandElement := nID.Bit(bit) == 0
-		pElement := proof[bit]
+	fmtString := fmt.Sprintf("%%0%db", h.BitLen())
+	bits := fmt.Sprintf(fmtString, new(big.Int).SetBytes(index))
```
Martin2112: Can we avoid using strings for binary data and bringing in big.Int for this. It should be more efficient as bit manipulation. Just write something that can do the equivalent of nID.Bit(x).

merkle/map_verifier.go
```diff
@@ -30,34 +30,29 @@ import (
...
+		pElement := proof[height]
 		if len(pElement) == 0 {
-			pElement = h.HashEmpty(treeID, index, bit)
-		}
-		if got, want := len(pElement)*8, hBits; got != want {
-			return fmt.Errorf("invalid proof: element has length %d, want %d", got, want)
+			pElement = h.HashEmpty(treeID, index, height)
 		}
-		if proofIsRightHandElement {
+		if bit == '0' {
```
Martin2112: This was more readable before.

gdbelvin: fixed

merkle/map_verifier.go
gdbelvin: done

gdbelvin:
Updated PTAL

merkle/common.go
```diff
@@ -24,3 +24,10 @@ func parent(leafIndex int64) int64 {
 func isRightChild(leafIndex int64) bool {
 	return leafIndex&1 == 1
 }
+
+// bit returns the i'th bit of index from the right.
```
Martin2112: I think index 0 being rightmost bit is OK as it matches std::bitset in C++. Can you add an example for increased clarity? Preferably a multi-byte one. For example if there's 2 bytes of input bit 7 is the MSB of index[1] and bit 8 is the LSB of index[0]?

merkle/common_test.go
```diff
@@ -0,0 +1,40 @@
...
+		index []byte
+		i     int
+		want  uint
+	}{
+		{index: h2b("00"), i: 0, want: 0},
+		{index: h2b("00"), i: 7, want: 0},
+		{index: h2b("000b"), i: 0, want: 1},
+		{index: h2b("000b"), i: 1, want: 1},
+		{index: h2b("000b"), i: 2, want: 0},
+		{index: h2b("000b"), i: 3, want: 1},
```
Martin2112: This doesn't fully test crossing byte boundaries. I'd add a couple of test cases.

merkle/map_verifier.go
```diff
@@ -30,32 +29,26 @@ import (
...
-	// TODO(al): Remove this dep on storage, since clients will want to use this code.
-	nID := storage.NewNodeIDFromHash(index)
-
 	runningHash := make([]byte, len(leafHash))
 	copy(runningHash, leafHash)
 
-	for bit := 0; bit < hBits; bit++ {
-		proofIsRightHandElement := nID.Bit(bit) == 0
-		pElement := proof[bit]
+	for height := 0; height < h.BitLen(); height++ {
```
Martin2112: Is that height or depth? These terms are somewhat confusing.

gdbelvin: We're reading from the bottom of the tree upwards height 0 to h.BitLen() == root, so it's height.

merkle/common.go
gdbelvin: added.

merkle/common_test.go
gdbelvin: added.

merkle/map_verifier.go
Martin2112: There are still some cases where we've used 'depth' for this elsewhere. I can't see anywhere else that uses 'height' for coordinates. We've tried to switch to 'level' e.g. in the path code because this is defined in our tree coordinate space where 0 is leaf level and it increases upwards.

gdbelvin:
Updated. PTAL

merkle/map_verifier.go
gdbelvin: hmm, level is fine with me. It's nice that it has a proper definition.  Should I switch to that?

Martin2112: We decided on level based on review feedback that depth and height were confusing.

Martin2112:
OK I think renaming the param to level is the last thing then it's ready.

PR #711, opened 2017-06-27
Rename PEMSignerFactory to DefaultSignerFactory

RJPercival:
There is nothing PEM-specific about this SignerFactory implementation now.
It supports PEM files, DER and PKCS#11.


RJPercival:
Travis will never pass, because this breaks the [certificate-transparency-go](https://github.com/google/certificate-transparency-go) repo. PR that fixes it is google/certificate-transparency-go#37, which I'll merge shortly after merging this.

PR #713, opened 2017-06-27
Test DefaultSignerFactory.NewSigner() with PrivateKey protos

RJPercival:


crypto/keys/pem_signer_factory_test.go
```diff
@@ -21,6 +21,16 @@ import (
 )
 
 func TestPEMSignerFactory(t *testing.T) {
+	key, err := NewFromPrivatePEMFile("../../testdata/log-rpc-server.privkey.pem", "towel")
```
codingllama: Could we use `testonly.DemoPrivateKey` instead of reading a file?

RJPercival: We could, but the rest of the tests read that same file so it doesn't buy us much. Happy to make this change if you prefer though.

codingllama: Sorry, I didn't look at the other tests. No need to change just this one.

PR #603, opened 2017-05-10
Delete from unsequenced using all fields in the index

Martin2112:
Also the queue table didn't get cleaned up properly when we changed the duplicate handling to be via LeafIdentityHash. The message id column has been removed as you can't insert dupes because of the pk of LeafData. There was also no index on queue timestamp, which wasn't helping.

Unfortunately this needs a schema change. There will probably be at least one more change to unsequenced based on the current testing.

Closes #457.

PR #717, opened 2017-06-28
Improve dequeuing performance

rolandshoemaker:
These changes significantly increase the performance of sequencing by batching the deletes from the `Unsequenced` table and inserts into the `SequencedLeafData` table. This undoes some of the work done in #603.

This is a breaking change as it requires adding a new field to the `Unsequenced` table. It would be possible to do a similar thing without making this change but then requires some slightly hacky uses of the existing indexes to avoid removing rows with duplicate `LeafIdentityHash`s. That said I think the changes are worth the short term pain as they significantly improve performance and reduce load on the database.

As I have yet to put together a god Prometheus dashboard I don't have any good charts to illustrate the speed up so here is some log output from a unpatched and patched `trillian_log_signer` using a batch size of `1000` while using `ct_hammer` with `-max_parallel_chains 50 -invalid_chance 0` and an empty tree. This shows (at least as far as I can tell) that this isn't re-introducing any of the previous locking problems on the `Unsequenced` and `SequencedLeafData` tables that were constraining sequencing performance. Note these numbers are from my laptop so the boost should actually be a bit better on real machines.

```
# master
5487756799089772688: processed 1000 items in 0.84 seconds (1190.85 qps)
5487756799089772688: processed 1000 items in 0.84 seconds (1192.62 qps)
5487756799089772688: processed 1000 items in 0.95 seconds (1052.98 qps)
5487756799089772688: processed 1000 items in 0.95 seconds (1047.63 qps)
5487756799089772688: processed 1000 items in 1.03 seconds (972.74 qps)
5487756799089772688: processed 1000 items in 1.00 seconds (1001.31 qps)
5487756799089772688: processed 1000 items in 1.05 seconds (949.57 qps)
5487756799089772688: processed 1000 items in 1.13 seconds (887.82 qps)
5487756799089772688: processed 1000 items in 1.22 seconds (817.28 qps)
5487756799089772688: processed 1000 items in 1.06 seconds (944.25 qps)

# patched
5487756799089772688: processed 1000 items in 0.14 seconds (7335.50 qps)
5487756799089772688: processed 1000 items in 0.16 seconds (6397.87 qps)
5487756799089772688: processed 1000 items in 0.14 seconds (6920.55 qps)
5487756799089772688: processed 1000 items in 0.17 seconds (5918.10 qps)
5487756799089772688: processed 1000 items in 0.21 seconds (4784.84 qps)
5487756799089772688: processed 1000 items in 0.11 seconds (9004.58 qps)
5487756799089772688: processed 1000 items in 0.13 seconds (7871.44 qps)
5487756799089772688: processed 1000 items in 0.17 seconds (5987.53 qps)
5487756799089772688: processed 1000 items in 0.12 seconds (8514.92 qps)
5487756799089772688: processed 1000 items in 0.10 seconds (10255.44 qps)
```

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -308,7 +308,7 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 		}
 
 		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
-			return nil, errors.New("Dequeued a leaf with incorrect hash size")
+			return nil, errors.New("dequeued a leaf with incorrect hash size")
 		}
 
 		leaves = append(leaves, leaf)
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -39,25 +39,13 @@ import (
 )
 
 const (
-	// If this statement ORDER BY clause is changed refer to the comment in removeSequencedLeaves
-	selectQueuedLeavesSQL = `SELECT LeafIdentityHash,MerkleLeafHash,QueueTimestampNanos
-			FROM Unsequenced
-			WHERE TreeID=?
-			AND Bucket=0
-			AND QueueTimestampNanos<=?
-			ORDER BY QueueTimestampNanos,LeafIdentityHash ASC LIMIT ?`
 	insertUnsequencedLeafSQL = `INSERT INTO LeafData(TreeId,LeafIdentityHash,LeafValue,ExtraData)
-			VALUES(?,?,?,?)`
-	insertUnsequencedEntrySQL = `INSERT INTO Unsequenced(TreeId,Bucket,LeafIdentityHash,MerkleLeafHash,QueueTimestampNanos)
-			VALUES(?,0,?,?,?)`
-	insertSequencedLeafSQL = `INSERT INTO SequencedLeafData(TreeId,LeafIdentityHash,MerkleLeafHash,SequenceNumber)
 			VALUES(?,?,?,?)`
 	selectSequencedLeafCountSQL   = "SELECT COUNT(*) FROM SequencedLeafData WHERE TreeId=?"
 	selectUnsequencedLeafCountSQL = "SELECT TreeId, COUNT(1) FROM Unsequenced GROUP BY TreeId"
 	selectLatestSignedLogRootSQL  = `SELECT TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature
 			FROM TreeHead WHERE TreeId=?
 			ORDER BY TreeHeadTimestamp DESC LIMIT 1`
-	deleteUnsequencedSQL = "DELETE FROM Unsequenced WHERE TreeId=? AND Bucket=0 AND QueueTimestampNanos=? AND LeafIdentityHash=?"
 
 	// These statements need to be expanded to provide the correct number of parameter placeholders.
 	selectLeavesByIndexSQL = `SELECT s.MerkleLeafHash,l.LeafIdentityHash,l.LeafValue,s.SequenceNumber,l.ExtraData
@@ -280,12 +268,6 @@ func (t *logTreeTX) WriteRevision() int64 {
 	return t.treeTX.writeRevision
 }
 
-// dequeuedLeaf is used internally and contains some data that is not part of the client API.
-type dequeuedLeaf struct {
-	queueTimestampNanos int64
-	leafIdentityHash    []byte
-}
-
 func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime time.Time) ([]*trillian.LogLeaf, error) {
 	start := time.Now()
 	stx, err := t.tx.PrepareContext(ctx, selectQueuedLeavesSQL)
@@ -296,7 +278,7 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	}
 
 	leaves := make([]*trillian.LogLeaf, 0, limit)
-	dql := make([]*dequeuedLeaf, 0, limit)
+	dq := make([]interface{}, 0, limit)
 	rows, err := stx.QueryContext(ctx, t.treeID, cutoffTime.UnixNano(), limit)
 
 	if err != nil {
@@ -307,30 +289,17 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	defer rows.Close()
 
 	for rows.Next() {
-		var leafIDHash []byte
-		var merkleHash []byte
-		var queueTimeNanos int64
-
-		err := rows.Scan(&leafIDHash, &merkleHash, &queueTimeNanos)
-
+		leaf, dqi, err := t.dequeueLeaf(rows)
 		if err != nil {
-			glog.Warningf("Error scanning work rows: %s", err)
 			return nil, err
 		}
 
-		if len(leafIDHash) != t.hashSizeBytes {
+		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
 			return nil, errors.New("Dequeued a leaf with incorrect hash size")
 		}
 
-		// Note: the LeafData and ExtraData being nil here is OK as this is only used by the
-		// sequencer. The sequencer only writes to the SequencedLeafData table and the client
-		// supplied data was already written to LeafData as part of queueing the leaf.
-		leaf := &trillian.LogLeaf{
-			LeafIdentityHash: leafIDHash,
-			MerkleLeafHash:   merkleHash,
-		}
 		leaves = append(leaves, leaf)
-		dql = append(dql, &dequeuedLeaf{queueTimestampNanos: queueTimeNanos, leafIdentityHash: leafIDHash})
+		dq = append(dq, dqi)
 	}
 
 	if rows.Err() != nil {
@@ -343,7 +312,7 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	// The convention is that if leaf processing succeeds (by committing this tx)
 	// then the unsequenced entries for them are removed
 	if len(leaves) > 0 {
-		err = t.removeSequencedLeaves(ctx, dql)
+		err = t.removeSequencedLeaves(ctx, dq)
 	}
 
 	if err != nil {
@@ -399,13 +368,17 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
 		}
 
 		// Create the work queue entry
-		_, err = t.tx.ExecContext(
-			ctx,
-			insertUnsequencedEntrySQL,
+		args := []interface{}{
 			t.treeID,
 			leaf.LeafIdentityHash,
 			leaf.MerkleLeafHash,
-			queueTimestamp.UnixNano())
+		}
+		args = append(args, queueArgs(t.treeID, leaf.LeafIdentityHash, queueTimestamp)...)
+		_, err = t.tx.ExecContext(
+			ctx,
+			insertUnsequencedEntrySQL,
+			args...,
+		)
 		if err != nil {
 			glog.Warningf("Error inserting into Unsequenced: %s", err)
 			return nil, fmt.Errorf("Unsequenced: %v", err)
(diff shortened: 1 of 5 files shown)


Martin2112:
I tested something similar and it didn't really make much difference on large trees, which is why it remains on one of my branches. This split the queue into logical buckets and tried to add and remove from different buckets using a time-based ring approach.

I think we really need to see numbers for a large tree, of at least several hundred million entries. This is where #603 came from. Also can we see the query plans being used for the deletes?

Rather than adding the new field you can possibly reuse the 'bucket' field. It's not currently being used and is there to avoid breakage if we did want to adopt the split queue approach.

rolandshoemaker:
Sorry this has taken me so long to get back to, I've been juggling a bunch of stuff!

I've done some tests with a much larger tree than my initial tests (~200 million entries) which show an admittedly smaller performance improvement (between 30% and 40%) while no new leaves are being added to `Unsequenced` and a small decrease (although this is probably statistically irrelevant given the variance) and added the `ANALYZE` output for the modified DELETE (both attached below).

Given that for certain applications (such as CT) trees are unlikely to actually grow to this size (except in certain cases such as the Pilot style archival/research style logs) I think this performance improvement is still beneficial (also note that the performance is still significantly improved even under insert load for smaller trees, let me know if you'd like to see stats for that as well) and makes sense.

Using the existing `Bucket` column is possible but we'd probably want to come up with a better way of generating the insert ID since truncating the existing hash down to 4 bytes (the default size of INTEGER) would drop the collision resistance to 16 bits which doesn't seem viable (I think? My math might be shaky here) given that it is quite possible tjat the size of `Unsequenced` may grow quite large depending on what parameters are used for the signer. Since the migration to add this new column would be _relatively_ simple (although it'd need to be done when the `Unsequenced` table is empty since there is no way to easily add a default value without some back-fill tool) I'd probably go with just adding a new column but ¯\_(ツ)_/¯.

---

Perf tests against a tree containing ~200 million entries (note these tests were done on a different machine to those above):

```
# current master

## no insert load
2764478521680768074: processed 1000 items in 0.80 seconds (1244.43 qps)
2764478521680768074: processed 1000 items in 0.70 seconds (1436.39 qps)
2764478521680768074: processed 1000 items in 0.67 seconds (1495.31 qps)
2764478521680768074: processed 1000 items in 0.68 seconds (1467.36 qps)
2764478521680768074: processed 1000 items in 0.62 seconds (1602.25 qps)
2764478521680768074: processed 1000 items in 0.81 seconds (1229.71 qps)
2764478521680768074: processed 1000 items in 0.92 seconds (1089.21 qps)
2764478521680768074: processed 1000 items in 0.85 seconds (1175.28 qps)
2764478521680768074: processed 1000 items in 0.86 seconds (1163.13 qps)
2764478521680768074: processed 650 items in 0.57 seconds (1142.57 qps)

min 1089 max 1602 avg 

## insert load
2764478521680768074: processed 1000 items in 1.12 seconds (890.91 qps)
2764478521680768074: processed 1000 items in 1.03 seconds (974.92 qps)
2764478521680768074: processed 1000 items in 1.02 seconds (981.63 qps)
2764478521680768074: processed 1000 items in 0.99 seconds (1011.97 qps)
2764478521680768074: processed 1000 items in 0.98 seconds (1020.74 qps)
2764478521680768074: processed 1000 items in 0.95 seconds (1057.50 qps)
2764478521680768074: processed 1000 items in 0.95 seconds (1053.20 qps)
2764478521680768074: processed 1000 items in 0.98 seconds (1015.64 qps)
2764478521680768074: processed 1000 items in 1.15 seconds (868.32 qps)
2764478521680768074: processed 1000 items in 0.96 seconds (1041.35 qps)

min 868 max 1057 avg 991 

# patched

## no insert load
2764478521680768074: processed 1000 items in 0.58 seconds (1713.23 qps)
2764478521680768074: processed 1000 items in 0.48 seconds (2083.28 qps)
2764478521680768074: processed 1000 items in 0.44 seconds (2297.43 qps)
2764478521680768074: processed 1000 items in 0.46 seconds (2157.61 qps)
2764478521680768074: processed 1000 items in 0.48 seconds (2085.26 qps)
2764478521680768074: processed 1000 items in 0.48 seconds (2065.70 qps)
2764478521680768074: processed 1000 items in 0.49 seconds (2048.88 qps)
2764478521680768074: processed 1000 items in 0.45 seconds (2224.29 qps)
2764478521680768074: processed 1000 items in 0.44 seconds (2296.30 qps)
2764478521680768074: processed 1000 items in 0.45 seconds (2222.24 qps)

min 1713 max 2297 avg 2119

## insert load
2764478521680768074: processed 1000 items in 0.77 seconds (1304.07 qps)
2764478521680768074: processed 1000 items in 0.88 seconds (1130.97 qps)
2764478521680768074: processed 1000 items in 0.96 seconds (1040.52 qps)
2764478521680768074: processed 1000 items in 0.94 seconds (1063.56 qps)
2764478521680768074: processed 1000 items in 0.98 seconds (1025.39 qps)
2764478521680768074: processed 1000 items in 1.00 seconds (1004.06 qps)
2764478521680768074: processed 1000 items in 1.10 seconds (912.86 qps)
2764478521680768074: processed 1000 items in 1.13 seconds (881.95 qps)
2764478521680768074: processed 1000 items in 1.23 seconds (812.91 qps)

min 812 max 1304 avg 917
```

ANALYZE DELETE:

```
MariaDB [second]> ANALYZE DELETE FROM Unsequenced WHERE QueueID IN (...);
+------+-------------+-------------+------+---------------+------+---------+------+------+---------+----------+------------+-------------+
| id   | select_type | table       | type | possible_keys | key  | key_len | ref  | rows | r_rows  | filtered | r_filtered | Extra       |
+------+-------------+-------------+------+---------------+------+---------+------+------+---------+----------+------------+-------------+
|    1 | SIMPLE      | Unsequenced | ALL  | QueueID       | NULL | NULL    | NULL | 8010 | 7999.00 |   100.00 |      12.50 | Using where |
+------+-------------+-------------+------+---------------+------+---------+------+------+---------+----------+------------+-------------+
```

Martin2112:
Thanks for doing the extra stats. Can you give the specs of the machines you tested with?

I'm now thinking about how we should proceed. I tested several similar schemes and several times convinced myself they were better, only to be disappointed later. However, if you think it works better for you then we should probably consider making it available.

I think it would be good if this could be made an option while we collect more data on performance, and it takes a long time to get trees up to the 800M entries or so we did before. Ideally, people would only need to make the schema change if they intend to enable the option to avoid disrupting existing log users.

Does that sound sensible?

rolandshoemaker:
> Thanks for doing the extra stats. Can you give the specs of the machines you tested with?

The initial tests were done on a c4.large EC2 instance with a 3 node m4.large EC2 Galera cluster. The follow up tests were done on a single machine with a AMD 1920X chip (12 core/3.5GHz) with 32GB of ram and local (non-NVMe) SSD storage. Neither the Galera cluster or local MariaDB instances had any special performance tuning applied.

> I'm now thinking about how we should proceed. I tested several similar schemes and several times convinced myself they were better, only to be disappointed later. However, if you think it works better for you then we should probably consider making it available.
>
> I think it would be good if this could be made an option while we collect more data on performance, and it takes a long time to get trees up to the 800M entries or so we did before. Ideally, people would only need to make the schema change if they intend to enable the option to avoid disrupting existing log users.
>
> Does that sound sensible?

Yup, I think hiding this behind a feature flag would make sense and be relatively easy with the current code. I'm planning on standing up a new testing log for LE that will have proper monitoring so once this is landed and we have run that log for a while I should be able to come back with better information on how the sequencing rate scales with real-life log size/usage.

I'll take a crack at updating this PR over the weekend to reflect this.

rolandshoemaker:
Updated to hide this new functionality behind a build tag, I started working on a version that hid it behind a CLI flag but that ended up being super messy and I think this is a bit simpler and allows the code to live side by side without too much hassle.

Let me know what you think!

Martin2112:
Thanks. Your machines have 2x the amount of RAM we tested with but comparing like with like that shouldn't really matter.

Some of the build combinations don't seem to be working. I think one of them was a flake but the others may be actual problems.

rolandshoemaker:
The required ones look good, it looks like the failures in the allowed failures are map integration test timeouts, which I can't replicate locally, and attempting to use `apt-get` on OS X.

storage/mysql/log_storage.go
```diff
@@ -296,7 +278,7 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	}
 
 	leaves := make([]*trillian.LogLeaf, 0, limit)
-	dql := make([]*dequeuedLeaf, 0, limit)
+	dq := make([]interface{}, 0, limit)
```
Martin2112: Not sure why this changed to interface{}, seems very generic. Possibly make dequeuedLeaf a storage level type to get round it? or is that no good?

storage/mysql/queue_batching.go
```diff
@@ -0,0 +1,125 @@
...
+			WHERE TreeID=?
+			AND Bucket=0
+			AND QueueTimestampNanos<=?
+			ORDER BY QueueTimestampNanos,LeafIdentityHash ASC LIMIT ?`
+	insertUnsequencedEntrySQL = `INSERT INTO Unsequenced(TreeId,Bucket,LeafIdentityHash,MerkleLeafHash,QueueTimestampNanos,QueueID) VALUES(?,0,?,?,?,?)`
+	insertSequencedLeafSQL    = `INSERT INTO SequencedLeafData(TreeId,LeafIdentityHash,MerkleLeafHash,SequenceNumber) VALUES`
+	deleteUnsequencedSQL      = "DELETE FROM Unsequenced WHERE QueueID IN (<placeholder>)"
+)
+
+func (t *logTreeTX) dequeueLeaf(rows *sql.Rows) (*trillian.LogLeaf, interface{}, error) {
```
Martin2112: Is this identical to the other impl? If so can you TODO for possible refactoring.

storage/mysql/queue.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return leaf, &dequeuedLeaf{queueTimestampNanos: queueTimeNanos, leafIdentityHash: leafIDHash}, nil
+}
+
+func queueArgs(treeID int64, identityHash []byte, queueTimestamp time.Time) []interface{} {
+	return []interface{}{queueTimestamp.UnixNano()}
+}
+
+func (t *logTreeTX) UpdateSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) error {
+	// TODO: In theory we can do this with CASE / WHEN in one SQL statement but it's more fiddly
```
Martin2112: Might as well remove that comment as it refers to something we probably won't implement.

Martin2112:
Sorry, we've been having trouble with some tests over the past few days. I've been waiting until that was fixed. I'm rerunning the failed ones now in case they were spurious.

storage/mysql/log_storage.go
```diff
@@ -309,28 +291,28 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
...
 		// Note: the LeafData and ExtraData being nil here is OK as this is only used by the
 		// sequencer. The sequencer only writes to the SequencedLeafData table and the client
 		// supplied data was already written to LeafData as part of queueing the leaf.
 		leaf := &trillian.LogLeaf{
 			LeafIdentityHash: leafIDHash,
 			MerkleLeafHash:   merkleHash,
 		}
+
+		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
+			return nil, errors.New("Dequeued a leaf with incorrect hash size")
```
Martin2112: should be lower case "dequeued", sorry, was my original error.

storage/mysql/queue.go
```diff
@@ -0,0 +1,101 @@
...
+
+func queueArgs(treeID int64, identityHash []byte, queueTimestamp time.Time) []interface{} {
+	return []interface{}{queueTimestamp.UnixNano()}
+}
+
+func (t *logTreeTX) UpdateSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) error {
+	for _, leaf := range leaves {
+		// This should fail on insert but catch it early
+		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
+			return errors.New("Sequenced leaf has incorrect hash size")
```
Martin2112: and here.

Martin2112:
I think it's pretty much ready if you could just tweak those error messages. Thanks.

PR #722, opened 2017-06-29
Add --mysql_max_conns flag to Trillian binaries

RJPercival:
Controls the maximum number of connections each binary can make to the MySQL database. Defaults to approximately one third of the default limit imposed by MySQL. This could be increased if using a MySQL cluster, or if the max_connections setting on the MySQL server has been increased.

This avoids a "[too many connections](https://dev.mysql.com/doc/refman/en/too-many-connections.html)" error under heavy load.

### DIFF
--- a/server/trillian_log_server/main.go
+++ b/server/trillian_log_server/main.go
@@ -51,7 +51,11 @@ import (
 )
 
 var (
-	mySQLURI           = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
+	mySQLURI = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
+	// MySQL supports 151 connections by default (https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_max_connections).
+	// Default to a maximum of ~1/2 of this, because other Trillian binaries will also need to make connections.
+	mySQLMaxConns = flag.Int("mysql_max_conns", 75, "Maximum number of connections that may be made to the MySQL database")
+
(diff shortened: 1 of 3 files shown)


Martin2112:
It's hard to pick a number of this as it's going to vary based on config and load. Maybe we can document some suggested settings later.

BTW not sure about the logserver setting as it's likely that multiple copies will be run.

gdbelvin:
Hmm, I wonder if a broader discussion about how to have high QPS served from a front end would be helpful? Does each RPC produce it's own connection to the database?

RJPercival:
Each transaction can produce its own connection to the database, so each RPC can produce one or more database connections, depending on how many transactions are required. The Go sql package creates a new connection if it doesn't have an idle one to reuse and, by default, there is no limit to how many connections it will open.

From a quick look at Travis, this change appears to be tanking performance. The integration test looks like it's timing out.

Martin2112:
We might be better off managing this with rate limiting rather than limiting connections.

codingllama:
@Martin2112 I don't think rate limiting removes the need for connection limits. For one, the limits may be high enough that we'd run into this issue even with rate limiting on.

Martin2112:
Typically you'd load test it to find the limits. At some point it won't scale up any more, which might be before it runs out of connections.

RJPercival:
I'm not convinced that reducing the maximum number of SQL connections to 50 was actually having a performance impact. Instead, I think it may have been stalling the integration test for some reason. If I increase that limit to 75 connections, it runs just as fast as with an unlimited number (in ~10 seconds). If we want to go ahead with this PR, I'd best do some more investigating to figure out what's going on.

gdbelvin:
Is this still something that we want?

Martin2112:
I don't think we can solve the problem this way.

Martin

gdbelvin:
Maybe we can close this PR then

PR #727, opened 2017-07-07
Map test cleanup

gdbelvin:
In order to facilitate easier debugging and prepare for the needed changes to support #691, I did some preliminary code cleanup: 

- Use canonical sort functions in HStar2
- Use `hasher.BitLen()` in place of `hasher.Size()*8`
- Use index ([]byte)/ value pairs in tests instead of  key (string)/value pairs.
   - This removed a dependency on `testonly.HashKey` 
   - `map[key]value` -> `[][]byte` of alternating index / value pairs.  
- Removed unused test vectors
- Combined `EmptyRoot*` tests into `simpleTestVector`

### DIFF
--- a/integration/map.go
+++ b/integration/map.go
@@ -32,6 +32,8 @@ import (
 
 const treeID = int64(0)
 
+var hasher = maphasher.Default
+
 // RunMapIntegration runs a map integration test using the given map ID and client.
 func RunMapIntegration(ctx context.Context, mapID int64, pubKey crypto.PublicKey, client trillian.TrillianMapClient) error {
 	{
@@ -90,7 +92,6 @@ func RunMapIntegration(ctx context.Context, mapID int64, pubKey crypto.PublicKey
 
 	// Check values
 	// Mix up the ordering of requests
-	h := maphasher.Default
 	randIndexes := make([][]byte, len(tests))
 	for i, r := range rand.Perm(len(tests)) {
 		randIndexes[i] = tests[r].Index
@@ -129,8 +130,8 @@ func RunMapIntegration(ctx context.Context, mapID int64, pubKey crypto.PublicKey
 			if got, want := leaf.LeafValue, ev.LeafValue; !bytes.Equal(got, want) {
 				return fmt.Errorf("got value %s, want %s", got, want)
 			}
-			leafHash := h.HashLeaf(treeID, leaf.Index, h.BitLen(), leaf.LeafValue)
-			if err := merkle.VerifyMapInclusionProof(treeID, leaf.Index, leafHash, r.GetMapRoot().GetRootHash(), incl.Inclusion, h); err != nil {
+			leafHash := hasher.HashLeaf(treeID, leaf.Index, hasher.BitLen(), leaf.LeafValue)
+			if err := merkle.VerifyMapInclusionProof(treeID, leaf.Index, leafHash, r.GetMapRoot().GetRootHash(), incl.Inclusion, hasher); err != nil {
 				return fmt.Errorf("verifyMapInclusionProof(%x): %v", leaf.Index, err)
 			}
 		}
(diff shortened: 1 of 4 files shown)


integration/map.go
```diff
@@ -32,6 +32,8 @@ import (
 
 const treeID = int64(0)
 
+var hasher = maphasher.Default
```
Martin2112: Why move this?

merkle/hstar2_test.go
```diff
@@ -28,59 +29,56 @@ import (
...
+	{testonly.HashKey("a"), []byte("2"), b64("2rAZz4HJAMJqJ5c8ClS4wEzTP71GTdjMZMe1rKWPA5o=")},
 }
 
-func TestHStar2EmptyRootKAT(t *testing.T) {
-	s := NewHStar2(treeID, maphasher.Default)
-	root, err := s.HStar2Root(s.hasher.Size()*8, []HStar2LeafHash{})
-	if err != nil {
-		t.Fatalf("Failed to calculate root: %v", err)
+// createHStar2Leaves returns a []HStar2LeafHash formed by the mapping of index, value ...
+// createHStar2Leaves panics if len(iv) is odd. Duplicate i/v pairs get over written.
```
Martin2112: Can you reference the doc that this came from and/or explain how we know the results are reliable?

merkle/hstar2_test.go
```diff
@@ -224,6 +201,17 @@ func TestPaddedBytes(t *testing.T) {
 		}
 	}
 }
+
+// b64 converts a base64 string into []byte.
```
Martin2112: The name b64 sounds like an encoder rather than a decoder.

merkle/map_verifier_test.go
```diff
@@ -98,182 +69,34 @@ var mapInclusionTestVector = []struct {
 			testonly.MustDecodeBase64("CtC2GCsc3/zFn1DNkoUThUnn7k+DMotaNXvmceKIL4Y="),
 		},
 		testonly.MustDecodeBase64("U6ANU1en3BSbbnWqhV2nTGtQ+scBlaZf9kRPEEDZsHM="),
-	}, {
-		"key-1-848",
```
Martin2112: Are we reducing coverage by removing these cases?

integration/map.go
gdbelvin: I moved the hasher variable to a package variable in order to make testing with a different hasher easier - one place to change it.

Martin2112: Wouldn't we need a test matrix for that though?

merkle/hstar2_test.go
gdbelvin: This algorithms is inspired by gRPC's `[metadata.Pairs](https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L65)` which also has a key / value pair format. 

I use a map internally to also guarantee that emitted values don't have duplicate keys.

merkle/map_verifier_test.go
gdbelvin: We're not losing any coverage.  The cases weren't being used, and I don't think they were exercising any additional code paths.  

In the next PR, I'll be adding a fuller suite of merkle tree tests that don't rely on test vectors, and can thus be used with any hasher.

merkle/hstar2_test.go
gdbelvin: would `fromB64` be better?  I was hoping to keep a short name because this function gets used a lot when forming table driven tests and a long name just pushes the test vectors off the screen.

integration/map.go
gdbelvin: To do it properly, yes. 

In the next PR, I'll be adding a `MapEnv` we can can do this with a table driven test.

merkle/map_verifier_test.go
Martin2112: OK.

merkle/hstar2_test.go
Martin2112: unB64 or deB64?

merkle/hstar2_test.go
Martin2112: I think we're at cross purposes. I meant that before we had hardcoded values that couldn't go wrong. If this leaf generation code code is basically a copy of the code we're testing for example then it's not ideal.

gdbelvin: oh, hmm.  

This isn't a new function - it replaces the previous `createHStar2Leaves` which was also in the test. It's just a test helper function. There's nothing like this function in the non-test codebase.

merkle/hstar2_test.go
gdbelvin: `deB64` sounds good.

Martin2112:
Can you fix Travis and then it's probably ready.

gdbelvin:
Looks like it was the `mysql` driver being flakey.

PR #732, opened 2017-07-11
NewMapEnv() for map integration test matrix

gdbelvin:
This PR creates a MapEnv, similar to LogEnv to support pure go
integration tests for the map.

This supports targeted testing of round-trip AddMapLeaf /
VerifyMapInclusionProof operations for each of the map hashers.

At this time, the TEST_MAP_HASHER is the only one that is known to work.

Also included:
- Added function names along error return paths in the map.


testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/google/trillian/quota"
+	"github.com/google/trillian/server"
+	"github.com/google/trillian/storage/mysql"
+	stestonly "github.com/google/trillian/storage/testonly"
+	"google.golang.org/grpc"
+)
+
+// MapEnv is a map server and connected client.
+type MapEnv struct {
+	registry     extension.Registry
```
codingllama: Suggestion: Make all fields public? I don't see any harm in exposing them to tests, plus it may be useful in some cases.

testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+	grpcServer := grpc.NewServer()
+	mapServer := server.NewTrillianMapServer(registry)
+	trillian.RegisterTrillianMapServer(grpcServer, mapServer)
+
+	// Listen and start server.
+	addr, lis, err := listen()
+	if err != nil {
+		return nil, err
+	}
+	go grpcServer.Serve(lis)
```
codingllama: We should probably save "lis" on a MapEnv field and ensure it's properly closed on Close().

testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+	registry := extension.Registry{
+		AdminStorage:  mysql.NewAdminStorage(db),
+		MapStorage:    mysql.NewMapStorage(db),
+		QuotaManager:  quota.Noop(),
+		SignerFactory: &keys.DefaultSignerFactory{},
+	}
+
+	ret, err := NewMapEnvWithRegistry(ctx, testID, registry)
+	if err != nil {
+		return nil, err
```
codingllama: Whenever NewMapEnv / NewMapEnvWithRegistry exit on error, they're letting resources (potentially) open: db, grpcServer, clientConn, etc.

We could rework MapEnv.Close() so it works even if env is not properly initialized, then it's just a matter of calling env.Close() before error returns.

testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+func (env *MapEnv) Close() {
+	env.ClientConn.Close()
+	env.grpcServer.GracefulStop()
+	if env.DB != nil {
+		env.DB.Close()
+	}
+}
+
+// CreateMap creates a map and signs the first empty tree head.
+func (env *MapEnv) CreateMap(hashStrategy trillian.HashStrategy) (*trillian.Tree, error) {
```
codingllama: Suggestion: You could pass in a "tree template" (ie, a tree with only the fields you care about set) and have the method set all other nil fields to reasonable defaults (possibly the ones as testonly.MapTree). It's a bit more work, but allows callers to evolve more naturally (without having to change the signature every time we want to add a new tree param).

testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+	ctx := context.Background()
```
codingllama: Please pass in the ctx.

testonly/integration/mapenv.go
```diff
@@ -0,0 +1,129 @@
...
+	tx, err := env.registry.AdminStorage.Begin(ctx)
```
codingllama: Please use the CreateTree RPC instead of writing directly to storage. There's logic you skip this way, which is especially important in the context of integration tests.

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
```
codingllama: Suggestion: Lately, I've been organizing imports like this:

```go
import (
	<golang libs>

	<other libs>

	<aliased imports>

	<_ imports>
)
```

Eg:

```go
import (
	"context"
	"encoding/hex"

	"github.com/google/trillian"

	stestonly "github.com/google/trillian/storage/testonly"

	_ "github.com/google/trillian/merkle/coniks"
	_ "github.com/google/trillian/merkle/maphasher"
)
```

I find it cleaner to look at, as blocks become more "uniform" (I saw this in an internal example somewhere). Totally optional.

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"encoding/hex"
+	"fmt"
+	"testing"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/merkle"
+	_ "github.com/google/trillian/merkle/coniks" // Register
```
codingllama: nit: Drop the "// Register". The idiom is common enough, I think.

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+// createHashKV returns a []*trillian.MapLeaf formed by the mapping of index, value ...
+// createHashKV panics if len(iv) is odd. Duplicate i/v pairs get over written.
+func createMapLeaves(iv ...[]byte) []*trillian.MapLeaf {
+	if len(iv)%2 != 0 {
+		panic(fmt.Sprintf("integration: createMapLeaves got odd number of iv pairs: %v", len(iv)))
+	}
+	r := []*trillian.MapLeaf{}
+	var index []byte
+	for i, b := range iv {
```
codingllama: Suggestion: I think writing a for with 2-step loops would make this clearer:

```go
	for i := 0; i < len(iv); i += 2 {
		r = append(r, &trillian.MapLeaf{
			Index:     iv[i],
			LeafValue: iv[i+1],
		})
	}
```

It took me a few moments to realize what you have is equivalent to the above.

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer env.Close()
+	for _, tc := range []struct {
+		trillian.HashStrategy
+		index []byte
+		value []byte
+	}{
+		{trillian.HashStrategy_TEST_MAP_HASHER, testonly.TransparentHash("A"), []byte("A")},
```
codingllama: Please use labels in the struct. It helps a lot with future maintenance.

Maybe test with >1 leaf as well?

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func TestInclusionWithEnv(t *testing.T) {
+	ctx := context.Background()
+	env, err := integration.NewMapEnv(ctx, "TestInclusionWithEnv")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer env.Close()
+	for _, tc := range []struct {
+		trillian.HashStrategy
```
codingllama: nit: Add a "desc string" attribute to use as test description and prefix it to all errors?

Ie, line 69 becomes `t.Errorf("%v: CreateMap() returned err = %v", tc.desc, err)`

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+		index []byte
```
codingllama: nit: `index, value []byte`.

Or not, not a big deal anyway. :)

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+			t.Errorf("CreateMap(): %v", err)
+			continue
+		}
+		mapID := tree.TreeId
+		hasher, err := hashers.NewMapHasher(tree.HashStrategy)
+		if err != nil {
+			t.Errorf("NewMapHasher(): %v", err)
+			continue
+		}
+		client := trillian.NewTrillianMapClient(env.ClientConn)
```
codingllama: Have you considered providing a MapClient within MapEnv? I expect all usages of MapEnv to need a MapClient.

integration/map_test.go
```diff
@@ -0,0 +1,123 @@
...
+			if err := merkle.VerifyMapInclusionProof(mapID, index,
+				leafHash, rootHash, proof, hasher); err != nil {
+				t.Errorf("VerifyMapInclusionProof(): %v", err)
+			}
+		}
+	}
+}
+
+// h2b converts a hex string into []byte.
+func h2b(h string) []byte {
```
codingllama: Unused function, remove?

server/map_rpc_server.go
```diff
@@ -57,13 +57,13 @@ func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapL
 
 	tree, hasher, err := t.getTreeAndHasher(ctx, mapID, true /* readonly */)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("getTreeAndHasher(): %v", err)
```
codingllama: Could you use textual descriptions instead of the function name? Eg, `"error creating tree hasher: %v"`. I expect those to be more human-friendly, especially for people not familiar with Trillian inner workings.

Ditto for all other changed messages.

testonly/integration/mapenv.go
gdbelvin: Maybe.. my thought is that by keeping them private, we keep a bias towards data encapsulation. We can make them public one by one as we need them on a case by case basis.  Maintenance of the Env struct is easier the fewer things there are poking around with it.

testonly/integration/mapenv.go
gdbelvin: I added the appropriate resource cleanup methods directly. 
I tend to prefer creating objects at the end of methods to avoid partially initialized objects.

testonly/integration/mapenv.go
gdbelvin: `grpcServer.Stop()` and `grpcServer.GracefulStop()` close `lis`

testonly/integration/mapenv.go
gdbelvin: I'll just call the admin RPC directly as you recommend.  I think this RPC was not present when I wrote LogEnv

testonly/integration/mapenv.go
gdbelvin: I removed `CreateMap` in lieu of `adminClient.CreateTree`

testonly/integration/mapenv.go
gdbelvin: done

integration/map_test.go
gdbelvin: looks cleaner 👍

integration/map_test.go
gdbelvin: I think govet requires underscore imports to have a comment.

integration/map_test.go
gdbelvin: How about I add a comment to that effect?  I think using range is more idiomatic

integration/map_test.go
gdbelvin: done

integration/map_test.go
gdbelvin: done

integration/map_test.go
gdbelvin: I'm currently working on making this work with the CONIKS hasher (which breaks)
After I have the rest of the codebase working, I'll certainly add more tests.  At the moment, fewer test cases mean fewer lines of output to sort through.

integration/map_test.go
gdbelvin: Hmm, not a bad idea, but this line is fairly easy.  Tests also need to construct an `AdminClient`. It's easier to provide one grpcConnection than pre-construct all the possible subtypes.

integration/map_test.go
gdbelvin: This will be used once I have other test vectors.

server/map_rpc_server.go
gdbelvin: Sounds good, especially since this will get returned to the user

gdbelvin:
Moved files and made most of the requested changes. PTAL

testonly/integration/mapenv.go
codingllama: Fair enough.

integration/map_test.go
codingllama: Are you sure? I think I saw of couple of "_" imports without comment around.

integration/map_test.go
codingllama: I don't see how that's better. More LoC, more operations performed, less obvious stepping, so on. I'd nitpick the standard library's PR if I had the chance too. :)

I'll leave it in your hands.

integration/map_test.go
codingllama: Let's add it when it gets used, but not before.

integration/map_test.go
codingllama: I think handing back both Admin and MapClient is worth it and will pay itself back rather quickly. That's what we're really after when getting a MapEnv, isn't it? The other fields don't matter nearly as much (not even ClientConn, assuming we hand back the specific clients needed).

integration/map_test.go
gdbelvin: we'll try without :-)

integration/map_test.go
gdbelvin: fair enough. I'll change it.

integration/map_test.go
gdbelvin: sounds good

integration/map_test.go
gdbelvin: sg

Issue #733, opened 2017-07-11
Integration tests timeout on Windows using local MySQL server


RJPercival:
```
$ go test github.com/google/trillian/integration
E0711 11:12:55.415149   12272 log_operation_manager.go:388] failed to execute operation on logs: failed to retrieve full list of log IDs: failed to get tx for retrieving logIDs: context canceled
*** Test killed: ran too long (10m0s).
```

RJPercival:
After seeing a similar timeout occur after reducing the max number of connections to the MySQL server (in PR #722), I tried increasing the max_connections MySQL server parameter from the default of 151 to 500. This had no effect.

RJPercival:
The cause of this was a bug in PR #734 that resulted in sequencing passes failing, but the integration test not reporting this as an error. Instead, it would just spin attempting pass after pass until the timeout was reached. The integration test should be made to abort if a terminal error occurs.

Martin2112:
Is this still an issue? I haven't seen this fail recently.

RJPercival:
Yes, [the LogOperationManager swallows sequencing errors](https://github.com/google/trillian/blob/master/server/log_operation_manager.go#L344) and so the integration tests can't detect that there's a problem - they just run until the timeout. There's a relevant TODO here: https://github.com/google/trillian/blob/master/server/log_operation_manager.go#L378

Martin2112:
The TODO seems to no longer be there. Not sure if this is fixed.

RJPercival:
[The LogOperationManager still swallows these errors](https://github.com/google/trillian/blob/master/server/log_operation_manager.go#L408).

Martin2112:
Fair enough.

RJPercival:
The TODO is now here: https://github.com/google/trillian/blob/master/server/log_operation_manager.go#L442

Martin2112:
I added error metrics in #957, doesn't help tests much but you can see signer errors on production logs.

paulmattei:
Next step: The integration test should use the new metric

pavelkalinnikov:
Blocked on #1640.

PR #734, opened 2017-07-12
Replace SignerFactory with global protobuf handlers for keys

RJPercival:
Previously, when adding support for a new way of accessing keys (e.g. PKCS#11), there were two options:
1. Modify DefaultSignerFactory.
    - Anyone using DefaultSignerFactory would be able to use this, but they'd also have any new dependencies imposed upon them.
2. Create a new SignerFactory implementation.
    - Requires running a fork of Trillian.
    - This new implementation would likely duplicate some code from DefaultSignerFactory, if it wanted to support some of the same ways of accessing keys.

An improved way of doing this is to instead have a global registry of functions that each accept a particular kind of protobuf message and return the corresponding private key. Functions can be added to this registry to provide support for new methods of accessing keys, or removed to disable particular methods. Disabled methods could be excluded from the build entirely using build tags, sparing users who do not use a particular method from having to install unnecessary dependencies.

TODO: In a follow-up PR, I intend to improve test coverage of the pre-existing crypto/keys/... packages.


RJPercival:
Judging by Travis, it looks like this PR might be responsible for the slow integration tests I reported in #733. I'm not sure where the performance issue is yet though - I'll investigate shortly.

RJPercival:
It turned out not to be a performance issue, but rather a sequencing pass failure that wasn't being (visibly) reported (see #738). I've now fixed that.

cmd/createtree/main.go
```diff
@@ -64,35 +59,24 @@ var (
...
 
-// createOpts contains all user-supplied options required to run the program.
-// It's meant to facilitate tests and focus flag reads to a single point.
-type createOpts struct {
-	addr                                                                                     string
-	treeState, treeType, hashStrategy, hashAlgorithm, sigAlgorithm, displayName, description string
-	maxRootDuration                                                                          time.Duration
-	privateKeyType, pemKeyPath, pemKeyPass, pkcs11ConfigPath                                 string
-}
+var keyHandlers = make(map[string]func() (proto.Message, error))
```
codingllama: FYI: I found it a bit odd that keyHandlers was declared here, but initialized elsewhere. There's more at a further comment.

cmd/createtree/main_test.go
```diff
@@ -43,147 +61,101 @@ func marshalAny(p proto.Message) *any.Any {
...
+			setFlags: func() {
+				*treeType = nonDefaultTree.TreeType.String()
+				*signatureAlgorithm = nonDefaultTree.SignatureAlgorithm.String()
+				*displayName = nonDefaultTree.DisplayName
+				*description = nonDefaultTree.Description
+			},
+			wantTree: &nonDefaultTree,
+		},
+		{
+			desc:     "defaultOptsOnly",
```
codingllama: Is this an error test because the private key won't be set? Maybe rename to "mandatoryOptsNotSet", or add a comment? At a glance it doesn't look different from "validOpts", it only became clearer after I realized runTest "fixes" a few flags.

cmd/createtree/main_test.go
```diff
@@ -43,147 +61,101 @@ func marshalAny(p proto.Message) *any.Any {
...
+			tree, err := createTree(ctx)
+			switch hasErr := err != nil; {
+			case hasErr != test.wantErr:
+				t.Errorf("%v: createTree() returned err = '%v', wantErr = %v", test.desc, err, test.wantErr)
+				return
+			case hasErr:
+				return
+			}
+
+			if diff := pretty.Compare(tree, test.wantTree); diff != "" {
```
codingllama: Please use `proto.Equal()` when comparing protos, `pretty.Compare` doesn't know about internal proto fields, so it may diff in cases where protos are meant to be equal.

The suggested idiom is:

```go
if !proto.Equal(pb1, pb2) {
    diff := pretty.Compare(pb1, pb2)
    t.Errorf("%v: post-Foo diff (-got +want):\n%v", test.desc, diff)
}
```

(Yep, it's a bit verbose, I know :/ )

cmd/createtree/pem.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/google/trillian/crypto/keyspb"
+)
+
+var (
+	pemKeyPath = flag.String("pem_key_path", "", "Path to the private key PEM file")
+	pemKeyPass = flag.String("pem_key_password", "", "Password of the private key PEM file")
+)
+
+func init() {
+	keyHandlers["PEMKeyFile"] = func() (proto.Message, error) {
```
codingllama: I would prefer keyHandlers to be declared and initialized in the same file. I would also prefer all flags to be defined together (at main.go), but since this is in the main package we're not breaking the rule too much.

Suggestion: for the sake of isolation / cleaner interface between createtree and key logic, we could do something like:

At createtree/keys/keys.go:

```go
package keys

type handler func() (proto.Message, error)

func SetHandler(name string, fn handler) { ... }

func Handler(name string) (handler, bool) { ... }

func NewPEMKeyHandler(path, pass) handler { ... }

```

At createtree/main.go:

```go
func main() {
    keys.RegisterHandler("PEMKeyFile", keys.NewPEMKeyHandler(*path, *pass))
    // ...
}
```

WDYT?

cmd/createtree/pem_test.go
```diff
@@ -0,0 +1,106 @@
...
+	keyProto, err := ptypes.MarshalAny(&keyspb.PEMKeyFile{
+		Path:     pemPath,
+		Password: pemPassword,
+	})
+	if err != nil {
+		t.Fatalf("MarshalAny(PEMKeyFile): %v", err)
+	}
+	wantTree.PrivateKey = keyProto
+
+	runTest(t, []*testCase{
```
codingllama: Could we have runTest and testCase moved to a separate, test-library like file? I'd be ok with a testonly package inside createtree, for example.

cmd/createtree/pem_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{
+			desc: "empty pemKeyPath",
```
codingllama: Since this is tested in an integration-like style, where we exercise it via createtree, I'd be tempted to use real handlers at main_test.go and move those test cases there. Or, possibly, just test the handlers directly here. If you take the comment on package separation, it'll force you down one of those roads regardless (which why I like it).

cmd/flagsaver/flagsaver.go
```diff
@@ -0,0 +1,36 @@
+// Package flagsaver provides a simple way to save and restore flag values.
+// TODO(RJPercival): Move this to its own GitHub project.
+//
+// Example:
+//   func TestFoo(t *testing.T) {
+//     defer flagsaver.Save().Restore()
+//     // Test code that changes flags
+//   } // flags are reset to their original values here.
+package flagsaver
```
codingllama: flagsaver isn't a command we run, so it shouldn't be placed inside cmd/. It definitely deserves its own package, though. Maybe util/flagsaver?

cmd/flagsaver/flagsaver.go
```diff
@@ -0,0 +1,36 @@
...
+
+import "flag"
+
+// Stash holds flag values so that they can be restored at the end of a test.
+type Stash struct {
```
codingllama: Maybe `type Stash map[string]string` ?

cmd/flagsaver/flagsaver_test.go
```diff
@@ -0,0 +1,67 @@
...
+	strFlag      = flag.String("str_flag", "foo", "test string flag")
+	durationFlag = flag.Duration("duration_flag", 5*time.Second, "test duration flag")
+)
+
+// TestRestore checks that flags are saved and restore correctly.
+// Checks are performed on flags with both their default values and with explicit values set.
+// Only a subset of the possible flag types are currently tested.
+func TestRestore(t *testing.T) {
+	tests := []struct {
+		// Test name
```
codingllama: nit: "// name is ..."

Ditto for the other comments, although I think they could all be removed.

nit2: "desc string"? That seems to be the recommended format.

cmd/flagsaver/flagsaver_test.go
```diff
@@ -0,0 +1,67 @@
...
+		name string
+		// Name of flag to save and restore.
+		flag string
+		// The value the flag should have when saved. If empty, this indicates the flag should have its default value.
+		oldValue string
+		// The value the flag should have just before being restored to oldValue.
+		newValue string
+	}{
+		{"RestoreDefaultIntValue", "int_flag", "", "666"},
```
codingllama: nit: Please use field labels. It's clearer, saves you from specifying defaults, and saves us from having to touch all structs if a new field is added.

crypto/keys/der_test.go
```diff
@@ -0,0 +1,73 @@
...
+	keyDER, err := base64.StdEncoding.DecodeString("MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgS81mfpvtTmaINn+gtrYXn4XpxxgE655GLSKsA3hhjHmhRANCAASwBWDdgHS04V/cN0LZgc8vZaK4I1HWLLCoaOO27Z0B1aS1aqBE7g1Oo8ldSCBJAvee866kcHhZkVniPdCG2ZZG")
+	if err != nil {
+		t.Fatalf("Could not decode test key: %v", err)
+	}
+
+	sf := NewSignerFactory()
+	sf.AddHandler(&keyspb.PrivateKey{}, NewFromPrivateKeyProto)
+
+	for _, test := range []struct {
+		name     string
```
codingllama: Suggestion: s/name/desc?  Same for others (if you take it, ofc).

crypto/keys/der_test.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+			wantErr: true,
+		},
+		{
+			name:     "PrivateKey with missing DER",
+			keyProto: &keyspb.PrivateKey{},
+			wantErr:  true,
+		},
+	} {
+		signer, err := sf.NewSigner(context.Background(), test.keyProto)
```
codingllama: Suggestion: Declare ctx outside the loop and reuse it?

crypto/keys/der_test.go
```diff
@@ -0,0 +1,73 @@
...
+func TestPrivateKeyProtoHandler(t *testing.T) {
+	// ECDSA private key in DER format.
+	keyDER, err := base64.StdEncoding.DecodeString("MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgS81mfpvtTmaINn+gtrYXn4XpxxgE655GLSKsA3hhjHmhRANCAASwBWDdgHS04V/cN0LZgc8vZaK4I1HWLLCoaOO27Z0B1aS1aqBE7g1Oo8ldSCBJAvee866kcHhZkVniPdCG2ZZG")
+	if err != nil {
+		t.Fatalf("Could not decode test key: %v", err)
+	}
+
+	sf := NewSignerFactory()
+	sf.AddHandler(&keyspb.PrivateKey{}, NewFromPrivateKeyProto)
+
```
codingllama: Should we test the handler for `*keyspb.Specification`? Ditto for the other public functions.

crypto/keys/keys_test.go
```diff
@@ -141,57 +142,14 @@ func TestLoadPrivateKeyAndSign(t *testing.T) {
...
-		}
-
-		if got := SignatureAlgorithm(key); got != test.want {
-			t.Errorf("%v: SignatureAlgorithm(%v) = %v, want %v", test.name, key, got, test.want)
-		}
 	}
 }
 
-func TestGenerateKey(t *testing.T) {
+func TestNewFromSpec(t *testing.T) {
```
codingllama: We shuffled some things around, but tests seem to have lagged behind. Could we make sure that the tests are in the matching _test.go file, compared to the one the function is defined?

cmd/createtree/pkcs11_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian/crypto/keyspb"
+)
+
+func TestRunPkcs11(t *testing.T) {
+	err := os.Chdir("../..")
```
codingllama: Wouldn't it be simpler to do "../../testdata/pkcs11-conf.json" below?

crypto/keys/pkcs11.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian/crypto/keyspb"
+	"github.com/letsencrypt/pkcs11key"
+)
+
+// Pkcs11ConfigProtoHandler returns a ProtoHandler configured to use the specified PKCS#11 modulePath.
+// This ProtoHandler will retrieve keys as specified by PKCS11Config proto messages.
+// It can be passed to SignerFactory.AddHandler().
+func Pkcs11ConfigProtoHandler(modulePath *string) ProtoHandler {
```
codingllama: nit: State the actual return type, instead of using ProtoHandler? None of the other handlers seem to do this.

crypto/keys/pkcs11_test.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keys
+
+import "testing"
+
+func TestPkcs11(t *testing.T) {
+	// PKCS11Config support is tested by integration/log_integration.sh (when $WITH_PKCS11 == "true").
+	t.Skip("Only integration testing is implemented for PKCS#11")
```
codingllama: Could we have a simple, happy path test? If we disable integration testing for some reason this will go without coverage.

crypto/keys/signer_factory.go
```diff
@@ -0,0 +1,86 @@
...
+// SignerFactory produces a crypto.Signer from a protobuf message describing a key.
+// If SignerFactory.Generate != nil, it can also generate new private keys.
+type SignerFactory struct {
+	// Generate creates a new private key based on a key specification.
+	// It returns a proto that can be passed to NewSigner() to get a crypto.Signer.
+	// If nil, key generation will not be possible.
+	Generate ProtoGenerator
+
+	// handlers convert a protobuf message into a crypto.Signer.
+	handlers map[string]ProtoHandler
```
codingllama: Have you considered making handlers global? Is there a situation where having factory-scoped handlers is useful?

A global handler registry would allow each handler-defining package to register its own handlers at init-time.

crypto/keys/signer_factory.go
```diff
@@ -0,0 +1,86 @@
...
+// AddHandler enables the SignerFactory to transform a protobuf message of the same
+// type as keyProto into a crypto.Signer by invoking handler.
+// The keyProto need only be an empty example of the type of protobuf message that
+// the handler can process - only its type is examined.
+// If a handler for this type of protobuf message has already been added, it will
+// be replaced.
+func (f SignerFactory) AddHandler(keyProto proto.Message, handler ProtoHandler) {
+	keyProtoType := proto.MessageName(keyProto)
+
+	if _, alreadyExists := f.handlers[keyProtoType]; alreadyExists {
```
codingllama: nit: s/alreadyExists/ok/, so we don't deviate from the idiom?

crypto/signatures.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	gocrypto "crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+
+	"github.com/google/trillian/crypto/sigpb"
+)
+
+// SignatureAlgorithm returns the algorithm used for this public key.
+// Only ECDSA and RSA keys are supported. Other key types will return sigpb.DigitallySigned_UNKNOWN.
```
codingllama: nit: s/sigpb.DigitallySigned_UNKNOWN/sigpb.DigitallySigned_ANONYMOUS/

In a somewhat related comment, is anonymous ever valid / allowed? Should we differentiate between anonymous and unknown? (This is possibly out of scope.)

server/admin/admin_server_test.go
```diff
@@ -318,10 +347,11 @@ func TestServer_CreateTree(t *testing.T) {
...
 	tests := []struct {
-		desc                           string
-		req                            *trillian.CreateTreeRequest
-		createErr                      error
-		commitErr, wantErr, wantCommit bool
+		desc                  string
+		req                   *trillian.CreateTreeRequest
+		createErr             error
+		commitErr, wantCommit bool
+		wantErr               string
```
codingllama: Asserting on error strings is likely to make tests brittle. Ideally, we'd use something more resilient to small changes, like error codes, but I that won't help if the codes are all the same. OTOH, it does make it easier to reason that the error is what you think it is, and not some problem.

WDYT?

cmd/createtree/main.go
RJPercival: It's because one of the handlers comes from a file that is conditionally built (if "pkcs11" build tag is used), so it can't be populated right here.

cmd/createtree/main_test.go
RJPercival: That's correct. I'll rename it.

cmd/createtree/main_test.go
RJPercival: That's not actually new code (the indentation has just changed), but I'll fix it up nonetheless.

cmd/flagsaver/flagsaver.go
RJPercival: Done.

cmd/flagsaver/flagsaver.go
RJPercival: That would preclude extending this type in the future and expose the implementation detail that it stores a map of flag names to values. The only reason I can see to do this is to allow the user to examine stashed flag values - do you see that being a use case?

cmd/flagsaver/flagsaver_test.go
RJPercival: Nits addressed, but left the comments, since I'm not sure `oldValue` and `newValue` are completely self-explanatory (perhaps I just need better names?).

cmd/flagsaver/flagsaver_test.go
RJPercival: Done.

RJPercival:
If you like, I think I could break this PR in two, with the first just doing some moving around of code?

crypto/keys/der.go
```diff
@@ -0,0 +1,102 @@
...
+		switch key2 := key2.(type) {
+		case *ecdsa.PrivateKey:
+			return key2, nil
+		case *rsa.PrivateKey:
+			return key2, nil
+		}
+		return nil, fmt.Errorf("der: unsupported private key type: %T", key2)
+	}
+
+	key3, err3 := x509.ParseECPrivateKey(keyDER)
```
RJPercival: No, these can be re-ordered.

crypto/keys/der_test.go
RJPercival: I'm planning on writing tests for all of the public functions in a follow-up PR - it that ok? I haven't added any new public functions, other than `NewPrivateKeyProtoFromSpec()`, that don't have tests (I'll add a test for that now).

crypto/keys/der_test.go
RJPercival: Done.

crypto/keys/der_test.go
codingllama: SGTM.

codingllama:
No need to break it now, I'll review the fixes commit-by-commit from now on. Just give me nudge when you finish replying to the comments.

crypto/keys/der_test.go
RJPercival: Done.

cmd/createtree/pkcs11_test.go
RJPercival: The testdata/pkcs11-conf.json config file contains a file path that is relative to the project root dir, so the test needs to change the working directory.

crypto/keys/keys_test.go
RJPercival: Done for TestNewFromSpec. I was intending to do this more thoroughly in the follow-up CL by replacing TestLoadPrivateKeyAndSign with individual tests for each of the public funcs (and put those tests in the appropriate _test.go file).

crypto/keys/pkcs11.go
RJPercival: That's because this func returns a ProtoHandler func, whereas the other funcs like this *are* the ProtoHandlers. This one is a bit different because the PKCS#11 ProtoHandler needs a bit of configuration (modulePath).

crypto/keys/signer_factory.go
RJPercival: I wasn't entirely convinced that "ok" was an appropriate name here, as it's semantically the opposite - if it already exists, that's not really ok. Originally I was returning an error here, which meant it was very much not ok. Perhaps that doesn't matter though.

crypto/keys/signer_factory.go
RJPercival: It's what I started with, but...
1. I've known people to be opposed to globals (e.g. @daviddrysdale lamented the scattering of flags through the codebase). 
1. I preferred the explicitness of adding a handler, rather than it being a side-effect of importing a package, since it's quite an important decision - do you really want unencrypted keys in your storage, for instance? It also means it's easier to control which handlers are available to tests.
1. Scoping it to a SignerFactory means you can start servers/signers in-process (like the integration tests do) and not be forced to share the same set of handlers.

Having said that, I do recognise that making the handlers global and having them register on package init is a more common pattern. I can put together a commit that makes that change and see how it compares if you like?

crypto/signatures.go
RJPercival: I think this is intended to mirror [the DigitallySigned struct in RFC5246](https://tools.ietf.org/html/rfc5246#section-7.4.1.4.1), which is why there isn't an "unknown" value. There are cases where anonymous is valid (see that RFC), but never in Trillian. Arguably, this should return an error rather than sigpb.DigitallySigned_ANONYMOUS.

RJPercival: Fixed typo.

server/admin/admin_server_test.go
RJPercival: I agree, and never usually check the error string but, in this case, there are a lot of different errors that could occur. I really want to know that the test is failing for the reason I think it is, rather than because the test is misconfigured, for example. Error codes would be a better solution except, as you suggest, I don't think they'd actually be different for most of these test cases (they're almost all cases of invalid arguments). Maybe that would be good enough for distinguishing the intended type of error from unintended errors though?

crypto/keys/pkcs11_test.go
RJPercival: This would be difficult because the PKCS#11 code calls out to https://github.com/letsencrypt/pkcs11key which attempts to load a dynamic library and communicate with a PKCS#11 interface. I could expose a package variable that allows the `pkcs11key.New()` call to be overridden during the tests, but that's about the only way I can think of making it unit testable.

At least by now having a skipped test for this, there is some (minimal) visibility of the fact that it lacks unit tests.

cmd/flagsaver/flagsaver.go
codingllama: It works fine as a struct, but since you don't have other fields an alias could have worked too. It's fine as it is.

cmd/flagsaver/flagsaver_test.go
codingllama: Looks good.

crypto/keys/signer_factory.go
codingllama: We didn't seem to shy away from global state in other places, like the [hashers registry](https://github.com/google/trillian/blob/master/merkle/hashers/tree_hasher.go#L61),  and an init-based registry would reduce the number of `+build pkcs11` files reaching for signerFactories, so it seemed worth to suggest it.

That said, your counter-arguments and the wariness about global state are certainly strong points (I found number "2" above particularly convincing). I'm happy going forward the way it is. Thanks for the detailed response, Rob.

crypto/keys/signer_factory.go
codingllama: OK  ;)

crypto/keys/pkcs11.go
gdbelvin: If we return the interface type everywhere, we benefit from some compile time checking in this package that the returned type matches the interface. It's also a nice way of documenting what interface we're trying to satisfy

cmd/createtree/pem.go
```diff
@@ -0,0 +1,68 @@
...
+			return nil, fmt.Errorf("empty password for PEM key file %q", *pemKeyPath)
+		}
+
+		return &keyspb.PEMKeyFile{
+			Path:     *pemKeyPath,
+			Password: *pemKeyPass,
+		}, nil
+	}
+
+	keyHandlers["PrivateKey"] = func() (proto.Message, error) {
```
gdbelvin: Consider making each of these anonymous functions real functions for ease on the eye and ease of testing.

crypto/keys/der_test.go
RJPercival: Added test for `NewPrivateKeyProtoFromSpec()`.

crypto/keys/der.go
RJPercival: Re-ordered to try EC first.

crypto/keys/pkcs11.go
RJPercival: I've changed the way handlers are obtained so that it is consistent with the above pattern (`<ProtoName>ProtoHandler(...) ProtoHandler`).

RJPercival:
I've just completed some changes @gdbelvin requested offline, which swaps over to a global registry of key handlers with auto-registration on import and ultimately deletes `SignerFactory`. I'll take on improvement to cmd/createtree next.

cmd/createtree/pem_test.go
RJPercival: It'll then lose access to the flags and won't be able to call `createTree()`. I can put them in their own file, but not their own package.

cmd/createtree/pem.go
RJPercival: `keyHandlers` can't be initialized in the same file, because the PKCS#11 handler needs to be behind a build tag. It needs to use the global registry / auto-registration pattern. I can move this registry out of main.go and into its own package though.

cmd/createtree/main.go
```diff
@@ -64,35 +59,24 @@ var (
...
 
-func createTree(ctx context.Context, opts *createOpts) (*trillian.Tree, error) {
```
gdbelvin: This is a great change of replacing `opts` with the relevant flags, but for the clarity of this already large PR, could I suggest that it be a separate PR?

cmd/createtree/main.go
gdbelvin: Were we going to move keyHandlers to their own file or package?

cmd/createtree/main.go
```diff
@@ -142,84 +126,32 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
...
-			Password: opts.pemKeyPass,
-		}
-		return ptypes.MarshalAny(pemKey)
-	case "PrivateKey":
-		if opts.pemKeyPath == "" {
-			return nil, errors.New("empty pem_key_path")
-		}
-		pemSigner, err := keys.NewFromPrivatePEMFile(
-			opts.pemKeyPath, opts.pemKeyPass)
+func newPK(format string) (*any.Any, error) {
```
gdbelvin: This function should probably live in it's own file or package along with the `keyHandlers` 

Also consider a clearer function name, say `NewPublicKeyProto`

cmd/createtree/main.go
```diff
@@ -142,84 +126,32 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
...
-		hashAlgorithm:    *hashAlgorithm,
-		sigAlgorithm:     *signatureAlgorithm,
-		displayName:      *displayName,
-		description:      *description,
-		maxRootDuration:  *maxRootDuration,
-		privateKeyType:   *privateKeyFormat,
-		pemKeyPath:       *pemKeyPath,
-		pemKeyPass:       *pemKeyPassword,
-		pkcs11ConfigPath: *pkcs11ConfigPath,
+func keyFormats() []string {
```
gdbelvin: A function comment would be great

cmd/createtree/main_test.go
```diff
@@ -43,147 +61,113 @@ func marshalAny(p proto.Message) *any.Any {
...
-	emptyPEMPass := *validOpts
-	emptyPEMPass.pemKeyPass = ""
-
-	pemKeyOpts := *validOpts
-	pemKeyOpts.privateKeyType = "PEMKeyFile"
-	pemKeyTree := *defaultTree
-	pemKeyTree.PrivateKey, err = ptypes.MarshalAny(&keyspb.PEMKeyFile{
-		Path:     pemPath,
-		Password: pemPassword,
+	runTest(t, []*testCase{
```
gdbelvin: If you'd like to run the test in a sub function, consider using a proper SubTest
https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

```go
for _, tc := range []struct{}{}{
  t.Run("A=1", func(t *testing.T) { ... })
}

File: cmd/createtree/main_test.go [hunk 28]
```diff
@@ -43,147 +61,113 @@ func marshalAny(p proto.Message) *any.Any {
...
-		t.Fatalf("MarshalAny(PKCS11Config): %v", err)
+// runTest executes the createtree command against a fake Trillian server for
+// each of the provided test cases, and checks that the tree in the request is
+// as expected, or an expected error occurs.
+// It changes a couple of flags from their default values before performing the
+// test. It sets the adminServerAddr flag to point to the fake server and sets
+// the privateKeyFormat flag to "test". It also registers a "test" key handler.
+// Finally, it calls the test's setFlags func (if provided) to allow it to
+// change flags specific to the test, then performs the test.
+func runTest(t *testing.T, tests []*testCase) {
```
gdbelvin: a slightly more descriptive name would be great

gdbelvin: a slightly more descriptive name would be great - and the test should probably only run one test case at a time

cmd/createtree/pem.go
gdbelvin: I think we figured out a way for everything except the `RegisterHander` function to be in it's own package?

cmd/createtree/pem.go
RJPercival: Done.

cmd/createtree/main.go
RJPercival: Done (though it's now called `RegisteredTypes()` in the createtree/keys package).

cmd/createtree/main.go
RJPercival: Done.

cmd/createtree/main.go
RJPercival: Sure, it'll have to merge first though, since having the PKCS#11 flag behind a build tag is dependent on createOpts behind removed.

cmd/createtree/main_test.go
RJPercival: I'm not sure this is so much a subtest as just delegating the actual test logic to a helper. However, I wonder if we should be using subtests for all of our tabular tests? It'd save having to prepend the test case name to every failure message, and defer would work nicely (running at the end of each test case, rather than at the end of the entire test).

cmd/createtree/main_test.go
RJPercival: They are only run one at a time.

RJPercival: Ah you mean that it should take a single testCase, rather than an array.

cmd/createtree/main_test.go
codingllama: The convention seems to prefer "regular" table-driven, but it may be because `t.Run` is relatively recent addition, iirc. Maybe we should take this question to stylists? Meanwhile, I'd vouch for the simplest approach (no `t.Run` unless you really need it).

crypto/keys/der/der_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"testing"
+
+	. "github.com/google/trillian/crypto/keys/der"
+	"github.com/google/trillian/crypto/keys/testonly"
+	"github.com/google/trillian/crypto/keyspb"
+)
+
+func TestFromProto(t *testing.T) {
+	// ECDSA private key in DER format.
+	keyDER, err := base64.StdEncoding.DecodeString("MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgS81mfpvtTmaINn+gtrYXn4XpxxgE655GLSKsA3hhjHmhRANCAASwBWDdgHS04V/cN0LZgc8vZaK4I1HWLLCoaOO27Z0B1aS1aqBE7g1Oo8ldSCBJAvee866kcHhZkVniPdCG2ZZG")
```
codingllama: Suggestion: Move the DER string to a const?

crypto/keys/der/der_test.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		// Get the key out of the proto, check that it matches the spec and test that it works.
+		key, err := FromProto(pb)
+		if err != nil {
+			t.Errorf("%v: FromProto(%#v) = (_, %q), want (_, nil)", test.desc, pb, err)
+		}
+
+		if err := testonly.CheckKeyMatchesSpec(key, test.keySpec); err != nil {
+			t.Errorf("%v: NewProtoFromSpec() => %v", test.desc, err)
```
codingllama: nit: *s/NewProtoFromSpec/CheckKeyMatchesSpec/*

crypto/keys/der/proto/register_test.go
```diff
@@ -12,62 +12,63 @@
...
 	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian/crypto/keys"
+	"github.com/google/trillian/crypto/keys/testonly"
 	"github.com/google/trillian/crypto/keyspb"
 )
 
-func TestPrivateKeyProtoHandler(t *testing.T) {
+func TestProtoHandler(t *testing.T) {
 	// ECDSA private key in DER format.
 	keyDER, err := base64.StdEncoding.DecodeString("MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgS81mfpvtTmaINn+gtrYXn4XpxxgE655GLSKsA3hhjHmhRANCAASwBWDdgHS04V/cN0LZgc8vZaK4I1HWLLCoaOO27Z0B1aS1aqBE7g1Oo8ldSCBJAvee866kcHhZkVniPdCG2ZZG")
```
codingllama: Ditto (move to const suggestion)

crypto/keys/generate_test.go
```diff
@@ -0,0 +1,107 @@
...
+		key, err := NewFromSpec(test.keySpec)
+		if gotErr := err != nil; gotErr != test.wantErr {
+			t.Errorf("%v: NewFromSpec() = (_, %v), want err? %v", test.desc, err, test.wantErr)
+			continue
+		} else if gotErr {
+			continue
+		}
+
+		if err := testonly.CheckKeyMatchesSpec(key, test.keySpec); err != nil {
+			t.Errorf("%v: NewFromSpec(): %v", test.desc, err)
```
codingllama: nit: *s/NewFromSpec/CheckKeyMatchesSpec/*

server/trillian_log_server/main.go
```diff
@@ -17,15 +17,23 @@
...
 
 import (
+	"context"
 	"flag"
 	_ "net/http/pprof"
 
 	_ "github.com/go-sql-driver/mysql" // Load MySQL driver
 
+	// Register key ProtoHandlers
+	_ "github.com/google/trillian/crypto/keys/der/proto"
```
codingllama: Suggestion: I've been splitting imports in the following blocks: golang, external libs, aliases and _ imports. I've found it more readable than mixing aliases and _ imports on the "external" block.

Eg:
```go
import (
    "context"
    "flag"

    "github.com/golang/glog"
    "github.com/google/trillian"
    "github.com/google/trillian/cmd"

    mysqlq "github.com/google/trillian/quota/mysql"

    _ "github.com/google/trillian/merkle/objhasher" // Load hashers
    _ "github.com/google/trillian/merkle/rfc6962"   // Load hashers
)
```

RJPercival:
Yes, this breaks ct-go so Travis will never be happy. PR google/certificate-transparency-go#50 will fix ct-go once this merges.

server/trillian_log_server/main.go
RJPercival: Done.

crypto/keys/generate_test.go
RJPercival: Done.

crypto/keys/der/der_test.go
RJPercival: Done.

AMarcedone:
What's the ETA for merging this PR?

cmd/createtree/main_test.go
RJPercival: I consulted a stylist, and they too leaned away from using subtests, but mostly because they are quite new and rarely used so far. I've decided to use them, but inside of `runTest()`, where it was already using an anonymous function to isolate each test case.

cmd/createtree/main.go
RJPercival: Done in #788.

cmd/createtree/main_test.go
RJPercival: See #788 for the use of subtests.

RJPercival:
Apologies for the delay getting this merged, I was on holiday for a week and have had competing priorities. I've now broken this PR up as requested by @gdbelvin (see PRs #791, #792, #793, #794). Once those PRs are merged, I'll rebase this and hopefully it'll be of a more easily reviewable size.

RJPercival:
This has now been rebased. The number of modified files has reduced from 63 to 46, and the number of changes in those files has reduced.

.travis.yml
```diff
@@ -9,12 +9,12 @@ go:
   - 1.8
 
 env:
-  - GOFLAGS=
-  - GOFLAGS=-race
-  - GOFLAGS=      WITH_ETCD=true
-  - GOFLAGS=-race WITH_ETCD=true
-  - GOFLAGS=      WITH_PKCS11=true
-  - GOFLAGS=-race WITH_PKCS11=true
+  - GOFLAGS=""
```
codingllama: nit: Use single quotes

.travis.yml
```diff
@@ -9,12 +9,12 @@ go:
...
+  - GOFLAGS="-race"
+  - GOFLAGS=""                     WITH_ETCD=true
+  - GOFLAGS="-race"                WITH_ETCD=true
+  - GOFLAGS="--tags pkcs11"        WITH_PKCS11=true
+  - GOFLAGS=""-race --tags pkcs11" WITH_PKCS11=true
```
codingllama: There seems to be an extra double quote here.

server/sequencer_manager_test.go
```diff
@@ -326,3 +350,12 @@ func createTestInfo(registry extension.Registry) *LogOperationInfo {
 		TimeSource:  fakeTimeSource,
 	}
 }
+
+func fakeKeyProtoHandler(pb proto.Message, signer crypto.Signer, err error) (proto.Message, keys.ProtoHandler) {
```
codingllama: Maybe rename pb to wantPB (or something to that effect)?

.travis.yml
RJPercival: Done.

server/sequencer_manager_test.go
RJPercival: Done.

PR #735, opened 2017-07-12
Travis should let tests timeout

RJPercival:
Travis will kill a build if there is no output for 10 minutes. 10 minutes is also the duration that `go test` will wait before a test times out. By reducing the test timeout, `go test` will get the chance to timeout a test and continue, rather than the build being killing right there. This allows it to print stack traces and detect any other test issues. Reducing the timeout also guarantees that no slow tests will be added in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ install:
 script:
   - set -e
   - cd $HOME/gopath/src/github.com/google/trillian
-  - travis_wait ./scripts/presubmit.sh
+  - ./scripts/presubmit.sh
   - |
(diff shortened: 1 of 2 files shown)


codingllama:
FYI, offending PR is #726.

RJPercival:
I'd be happy to change this to reduce the test timeout duration instead. What do we think a reasonable duration is? 5 minutes? 1 minute? I quite like the idea of enforcing fast tests, but not sure how pragmatic that is.

RJPercival:
I've changed this to reduce the `go test` timeout instead.

PR #737, opened 2017-07-12
Quota increase factor and other minor adjustments

codingllama:
Added a quota_increase_factor flag that multiplies the number of replenished
tokens. A value slightly higher than 1 adds resilience to leakages, without too
much impact on legitimate "token outages".

A few minor adjustments have been performed, like not calling Get/Put if tokens
= 0, correctly charging quota of batch write requests and not charging Admin
RPCs.

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -51,7 +51,25 @@ var (
 	seqStoreRootLatency    monitoring.Histogram
 	seqCommitLatency       monitoring.Histogram
 	seqCounter             monitoring.Counter
+
+	// QuotaIncreaseFactor is the multiplier used for the number of tokens added back to
+	// sequencing-based quotas. The resulting PutTokens call is equivalent to
+	// "PutTokens(_, numLeaves * QuotaIncreaseFactor, _)".
+	// A factor >1 adds resilience to token leakage, on the risk of a system that's overly
+	// optimistic in face of true token shortages. The higher the factor, the higher the quota
+	// "optimism" is. A factor that's too high (say, >1.5) is likely a sign that the quota
+	// configuration should be changed instead.
+	// A factor <1 WILL lead to token shortages, therefore it'll be normalized to 1.
+	QuotaIncreaseFactor = 1.1
 )
+
+func quotaIncreaseFactor() float64 {
+	if QuotaIncreaseFactor < 1 {
+		QuotaIncreaseFactor = 1
+		return 1
+	}
+	return QuotaIncreaseFactor
+}
 
 func createMetrics(mf monitoring.MetricFactory) {
 	if mf == nil {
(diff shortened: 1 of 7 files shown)


codingllama:
@AlCutter, I've added a new flag (quota_dry_run) and rebased on master. PTAL.

codingllama:
Issue #490

log/sequencer.go
```diff
@@ -360,18 +379,22 @@ func (s Sequencer) SequenceBatch(ctx context.Context, logID int64, limit int, gu
...
-		glog.Warningf("Failed to replenish tokens for tree %v: %v", logID, err)
+	if numLeaves > 0 {
+		tokens := int(float64(numLeaves) * quotaIncreaseFactor())
+		glog.V(2).Infof("Replenishing %v tokens for tree %v (numLeaves = %v)", tokens, logID, leaves)
+		if err := s.qm.PutTokens(ctx, tokens, []quota.Spec{
+			{Group: quota.Tree, Kind: quota.Read, TreeID: logID},
+			{Group: quota.Tree, Kind: quota.Write, TreeID: logID},
+			{Group: quota.Global, Kind: quota.Read},
+			{Group: quota.Global, Kind: quota.Write},
+		}); err != nil {
```
AlCutter: Would it be an entirely terrible idea to keep (local) count of the token "debt" here and add it on to future attempts to return tokens, zero-ing the debt if when you manage a successful token put (i.e. a best-effort attempt to avoid the loss of tokens)?

server/interceptor/interceptor.go
```diff
@@ -59,8 +64,13 @@ func (i *TrillianInterceptor) UnaryInterceptor(ctx context.Context, req interfac
...
 	}
 
-	if err := i.QuotaManager.GetTokens(ctx, 1 /* numTokens */, rpcInfo.specs); err != nil {
-		return nil, status.Errorf(codes.ResourceExhausted, "quota exhausted: %v", err)
+	if len(rpcInfo.specs) > 0 && rpcInfo.tokens > 0 {
+		if err := i.QuotaManager.GetTokens(ctx, rpcInfo.tokens, rpcInfo.specs); err != nil {
+			if !i.QuotaDryRun {
+				return nil, status.Errorf(codes.ResourceExhausted, "quota exhausted: %v", err)
+			}
+			glog.Warningf("(QuotaDryRun) Request %+v not denied due to dry run mode: %v", req, err)
```
AlCutter: I'd really love a monitoring metric here :)

log/sequencer.go
codingllama: I think you just described #741.

server/interceptor/interceptor.go
codingllama: I was planning on adding the metrics in a separate PR.

codingllama:
PTAL?

log/sequencer.go
AlCutter: Ok, excellent :)

server/interceptor/interceptor.go
AlCutter: Oke doke.

log/sequencer.go
codingllama: Oops, sorry, I misunderstood you. That's a different approach from #741.

I'll go ahead and merge this, as I think we want it anyway. Let's take the discussion offline.

PR #741, opened 2017-07-13
Replenish tokens for bad requests / duplicate leaves

codingllama:
Bad requests and, more importantly, duplicate leaves, may be a major source of token leakage for sequencing-based quotas (as tokens are taken when the request goes in, but never replenished via sequencing). Interceptors are changed to now return tokens in those scenarios.

This change introduces the first block of genuine after-handler logic. To allow for easy ports to non-gRPC environments, the concepts of a RequestProcessor and Before/After have been added.

### DIFF
--- a/server/interceptor/interceptor.go
+++ b/server/interceptor/interceptor.go
@@ -80,6 +80,7 @@ func (i *TrillianInterceptor) NewProcessor() RequestProcessor {
 
 type trillianProcessor struct {
 	parent *TrillianInterceptor
+	info   *rpcInfo
 }
 
 func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (context.Context, error) {
@@ -88,6 +89,7 @@ func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (conte
 	if err != nil {
 		return ctx, err
 	}
+	tp.info = info
 
 	if info.treeID != 0 {
 		tree, err := trees.GetTree(ctx, tp.parent.Admin, info.treeID, info.opts)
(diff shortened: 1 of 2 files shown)


codingllama:
Issue #490.

codingllama:
@AlCutter, this is branched off of #737, as they would invariably conflict. The first new commit is dc6072e / "Define a RequestProcessor ...". I suggest you only review from there forward.

Once #737 is in I'll rebase appropriately.

server/interceptor/interceptor.go
```diff
@@ -110,7 +112,42 @@ func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (conte
 	return ctx, nil
 }
 
-func (tp *trillianProcessor) After(ctx context.Context, resp interface{}, handlerErr error) {}
+func (tp *trillianProcessor) After(ctx context.Context, resp interface{}, handlerErr error) {
+	// Decide if we have to replenish tokens. There are a few situations that require tokens to be
+	// replenished:
+	// * Invalid requests (a bad request shouldn't spend sequencing-based tokens, as it won't cause
+	//   a corresponding sequencing to happen)
+	// * Requests that filter out duplicates (eg, QueueLeaf and QueueLeaves, for the same reason
```
AlCutter: "e.g."

AlCutter: "same reason as above"

codingllama:
PTAL?

AlCutter:
I can't see the bit from the comment #737 about storing up a count of tokens we couldn't return and trying to return them in future PutToken calls, am I being blind?

codingllama:
PTAL?

PR #742, opened 2017-07-13
Table driven tests for storage/types.go

gdbelvin:
Before changing anything, I thougt it would be a good idea to flesh out
the current unit tests to get a good understanding of what the code is
doing.

I've preserved all the existing test vectors, and added some additional
ones.

There are a few scenarios (marked with TODO's) where the current output
seems incorrect, but I could be wrong.  Would love to get clarity on the
correct behavior so I can use this set of unit tests as a bench mark on
whether my changes are consistent.

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -86,7 +86,13 @@ func NewEmptyNodeID(maxLenBits int) NodeID {
 }
 
 // NewNodeIDWithPrefix creates a new NodeID of nodeIDLen bits with the prefixLen MSBs set to prefix.
+// NewNodeIDWithPrefix places the lower prefixLenBits of prefix in the most significant bits of path.
+// Path will have enough bytes to hold maxLenBits
+//
 func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits int) NodeID {
+	if got, want := nodeIDLenBits%8, 0; got != want {
+		panic(fmt.Sprintf("nodeIDLenBits mod 8: %v, want %v", got, want))
+	}
 	maxLenBytes := bytesForBits(maxLenBits)
 	p := NodeID{
 		Path:          make([]byte, maxLenBytes),
@@ -156,6 +162,9 @@ func (n *NodeID) SetBit(i int, b uint) {
 
 // Bit returns 1 if the ith bit is true, and false otherwise.
 func (n *NodeID) Bit(i int) uint {
+	if got, want := i, n.PathLenBits-1; got > want {
+		panic(fmt.Sprintf("storage: Bit(%v) > (PathLenBits -1): %v", got, want))
+	}
 	bIndex := (n.PathLenBits - i - 1) / 8
 	return uint((n.Path[bIndex] >> uint(i%8)) & 0x01)
 }
--- a/storage/types_test.go
+++ b/storage/types_test.go
@@ -16,202 +16,295 @@ package storage
 
 import (
 	"bytes"
+	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"testing"
 )
 
-func TestZerosNewNodeIDWithPrefix(t *testing.T) {
-	n := NewNodeIDWithPrefix(0, 0, 0, 64)
-	if got, want := n.Path, []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}; !bytes.Equal(got, want) {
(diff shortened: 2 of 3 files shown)


Martin2112:
Before I look at this did you read the storage documentation and if so did this help with the TODO questions?

Martin2112:
Also I'm working on #743 which is a tool that will dump the subtree protos for a specified tree size.

gdbelvin:
I look forward to using your tool, it might be helpful in visualizing what's going on with the tree storage errors I'm observing in my feature branch. 

I did take a look at the documentation.  I've highlighted what might be deviations from the documentation in the TODOs.

gdbelvin:
Post chat, things are clearer.  All TODOs have been resolved.  PTAL

storage/types.go
```diff
@@ -58,7 +58,12 @@ type NodeID struct {
 	// e.g. if Path contains two bytes, and PrefixLenBits is 9, then the 8 bits
 	// in Path[0] are included, along with the lowest bit of Path[1]
 	PrefixLenBits int
-	PathLenBits   int
+}
+
+// PathLenBits returns the number of bits available to store the path.
```
Martin2112: I thought this PR wasn't making any functional changes.

storage/types.go
```diff
@@ -70,28 +75,33 @@ func bytesForBits(numBits int) int {
...
-		PathLenBits:   len(h) * 8,
 		PrefixLenBits: len(h) * 8,
 	}
 }
 
 // NewEmptyNodeID creates a new zero-length NodeID with sufficient underlying
 // capacity to store a maximum of maxLenBits.
 func NewEmptyNodeID(maxLenBits int) NodeID {
+	if got, want := maxLenBits%8, 0; got != want {
+		panic(fmt.Sprintf("storeage: NewEmptyNodeID() maxLenBits mod 8: %v, want %v", got, want))
```
Martin2112: storeage is a typo. It's arguable whether this is an error.

storage/types.go
```diff
@@ -70,28 +75,33 @@ func bytesForBits(numBits int) int {
...
 	}
 }
 
 // NewNodeIDWithPrefix creates a new NodeID of nodeIDLen bits with the prefixLen MSBs set to prefix.
+// NewNodeIDWithPrefix places the lower prefixLenBits of prefix in the most significant bits of path.
+// Path will have enough bytes to hold maxLenBits
+//
 func NewNodeIDWithPrefix(prefix uint64, prefixLenBits, nodeIDLenBits, maxLenBits int) NodeID {
-	maxLenBytes := bytesForBits(maxLenBits)
+	if got, want := nodeIDLenBits%8, 0; got != want {
```
Martin2112: Again I don't think this is an error, it's just something we don't currently do.

storage/types.go
```diff
@@ -156,16 +168,19 @@ func (n *NodeID) SetBit(i int, b uint) {
 
 // Bit returns 1 if the ith bit is true, and false otherwise.
 func (n *NodeID) Bit(i int) uint {
-	bIndex := (n.PathLenBits - i - 1) / 8
+	if got, want := i, n.PathLenBits()-1; got > want {
+		panic(fmt.Sprintf("storage: Bit(%v) > (PathLenBits -1): %v", got, want))
```
Martin2112: I'm not overjoyed about all these panics in low level code. It's duplicating the bounds check on the array with a chance to get it wrong.

storage/types.go
gdbelvin: Removed the functional changes

storage/types.go
gdbelvin: removed

storage/types.go
gdbelvin: removed

storage/types_test.go
```diff
@@ -230,3 +323,29 @@ func TestCoordString(t *testing.T) {
 		}
 	}
 }
+
+// h26 converts a hex string into an uint64.
```
gdbelvin: done

storage/types.go
gdbelvin: I found the panics slightly easier to read, and a nice documentation of the exact preconditions the function expects. 

There are a number of cases where the bounds check would succeed and produce garbage data, whereas this check would catch the logic bug. For example, imagine a case where `PathLenBits` is not a multiple of 8, say 7.  One could request `Bit(8)` without causing an out-of-bounds error, but the result may still be undefined.  

What do you advise?

gdbelvin:
Removed functional changes, added SuffixKeyTests.

PR #744, opened 2017-07-14
Removes GetActiveLogIDs() from LogTreeTX.

AlCutter:
This is a meta-data operation, and shouldn't be part of a per-tree TX.


AlCutter:
Nothing uses this particular call via LogTreeTX; the only place which does this type of operation is the sequencer operation manager, which already uses the non-tree version...

AlCutter:
(mocks regenerated btw - thanks!)

PR #745, opened 2017-07-14
Adds a GetUnsequencedCount() to LogStorage.

AlCutter:


storage/memory/log_storage.go
```diff
@@ -342,6 +342,22 @@ func (t *logTreeTX) GetActiveLogIDs(ctx context.Context) ([]int64, error) {
...
+	t.ms.mu.RLock()
+	defer t.ms.mu.RUnlock()
+
+	ret := make(map[int64]int64)
+	for id, tree := range t.ms.trees {
+		tree.mu.RLock()
+		defer tree.mu.RUnlock()
+
+		k := unseqKey(id)
+		q := tree.store.Get(k).(*kv).v.(*list.List)
```
codingllama: nit: what does "q" stand for? 

nit2: `tree.store` would benefit from a comment stating what it contains (ie, the tree items are `kv`s and the `kv`'s value is a `*list.List`). Bonus points if it explains why it's using "unusual" containers, instead of the standard maps and slices.

storage/memory/log_storage.go
```diff
@@ -342,6 +342,22 @@ func (t *logTreeTX) GetActiveLogIDs(ctx context.Context) ([]int64, error) {
 	return t.getActiveLogIDs(ctx)
 }
 
+func (t *readOnlyLogTX) GetUnsequencedCounts(ctx context.Context) (map[int64]int64, error) {
+	t.ms.mu.RLock()
+	defer t.ms.mu.RUnlock()
+
+	ret := make(map[int64]int64)
+	for id, tree := range t.ms.trees {
+		tree.mu.RLock()
```
codingllama: nit: Use `tree.RLock()` (or remove it altogether?). Ditto for the other `tree.*Lock()` methods.

(pre-existing) nit: I would expect the global `t.ms.mu` lock to be enough for an in-memory impl. If it's not, a comment explaining why would be helpful.

storage/mysql/log_storage.go
```diff
@@ -51,8 +51,9 @@ const (
 			VALUES(?,0,?,?,?)`
 	insertSequencedLeafSQL = `INSERT INTO SequencedLeafData(TreeId,LeafIdentityHash,MerkleLeafHash,SequenceNumber)
 			VALUES(?,?,?,?)`
-	selectSequencedLeafCountSQL  = "SELECT COUNT(*) FROM SequencedLeafData WHERE TreeId=?"
-	selectLatestSignedLogRootSQL = `SELECT TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature
+	selectSequencedLeafCountSQL   = "SELECT COUNT(*) FROM SequencedLeafData WHERE TreeId=?"
```
codingllama: MySQL/InnoDB doesn't keep row counts. Both "count(*) from SequencedLeafData" and "count(*) from Unsequenced" may take a looong time to run.

What do you need those counts for? Would an estimated\* global count be good enough? Is whatever it is that's planning to call this OK with an eventual 20/30\*\* min delay while the query takes its merry way?

Btw, `LogMetadata.GetUnsequencedCounts` probably needs a huge disclaimer stating that it may take forever and should be avoided.

\* aka, "almost completely inaccurate"
\*\* YMMV

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+	for i := 0; i < numLogs; i++ {
+		logIDs = append(logIDs, createLogForTests(DB))
+	}
+	s := NewLogStorage(DB, nil)
+
+	ctx := context.Background()
+	expectedCount := int64(0)
+
+	for i := int64(1); i < 10; i++ {
+		// Put some leaves in the queue of each of the logs
```
codingllama: All logs have the same number of leaves throughout the iterations. That would hide a bug where we apply the count of a certain log to all others. I'd rather do a single pass, but having different numbers of leaves for each log (possibly with an empty log too).

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+		for _, logID := range logIDs {
+			tx := beginLogTx(s, logID, t)
+			defer tx.Close()
```
codingllama: defer-in-loop (as explained above).

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+		tx, err := s.Snapshot(context.Background())
+		if err != nil {
+			t.Fatalf("Snapshot() = (_, %v), want no error", err)
+		}
+
+		counts, err := tx.GetUnsequencedCounts(ctx)
+		if err != nil {
+			t.Fatalf("GetUnsequencedCounts() = %v, want no error", err)
+		}
+		if got, want := len(counts), numLogs; got != want {
```
codingllama: Suggestion: You could assemble a `wantMap` and `pretty.Compare(got, want)`

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+
+		// Now check what we get back from GetUnsequencedCounts matches
+		tx, err := s.Snapshot(context.Background())
+		if err != nil {
+			t.Fatalf("Snapshot() = (_, %v), want no error", err)
+		}
+
+		counts, err := tx.GetUnsequencedCounts(ctx)
+		if err != nil {
+			t.Fatalf("GetUnsequencedCounts() = %v, want no error", err)
```
codingllama: nit: Use Errorf/continue for this error. Fatalf is OK for things we're not actively testing, IMO, but this is the method we care about.

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+		}
+		if got, want := len(counts), numLogs; got != want {
+			t.Fatalf("GetUnsequencedCounts returns map with %d entries, want %d", got, want)
```
codingllama: nit: Errorf/continue.

storage/mysql/log_storage_test.go
```diff
@@ -967,6 +967,52 @@ func TestGetActiveLogIDsEmpty(t *testing.T) {
...
+		}
+		for id, c := range counts {
+			if got, want := c, expectedCount; got != want {
+				t.Fatalf("GetUnsequencedCounts: LogID %v has %d unsequenced, want %d", id, got, want)
```
codingllama: nit: Errorf/continue.

storage/log_storage.go
```diff
@@ -136,4 +136,8 @@ type LogRootWriter interface {
 type LogMetadata interface {
 	// GetActiveLogs returns a list of the IDs of all the logs that are configured in storage
 	GetActiveLogIDs(ctx context.Context) ([]int64, error)
+
+	// GetUnsequencedCounts returns a map of the number of unsequenced entries
+	// by log ID.
+	GetUnsequencedCounts(ctx context.Context) (map[int64]int64, error)
```
codingllama: Would it be useful to have `GetUnsequencedCount(ctx, logID)`, either as a replacement or in addition to this method? (Not that I think it would be any better performance-wise.)

AlCutter: Hmm, but that kinda sounds like it's a method which should take a `LogID` parameter?

AlCutter: The particular use-case for this is to background export this map of logID->count as monitoring metrics, so not particularly useful to pull them out individually at the moment...

storage/memory/log_storage.go
```diff
@@ -342,6 +342,22 @@ func (t *logTreeTX) GetActiveLogIDs(ctx context.Context) ([]int64, error) {
...
+		defer tree.mu.RUnlock()
```
AlCutter: Yup, I know defers are func-level, this just locks all the trees it touches until it returns when they all become unlocked - I figured that was fine for a test impl, and led to fewer lines of code.

storage/memory/log_storage.go
AlCutter: Ah - thanks - I forgot about those methods - changed.

`ms.mu` really only protects the map of trees, I guess in theory it _could_ protect all the trees too, but I think there are probably places where all you have is your local `*tree`, so you have no way of getting to `ms.mu`.

storage/memory/log_storage.go
AlCutter: "Queue" => q :)

Added some comments around to help explain how it all works.

storage/mysql/log_storage.go
AlCutter: Estimated would probably be fine too, but this will be alright for now.
Added a comment about the dangers of calling this method :)

storage/mysql/log_storage_test.go
AlCutter: same comment as above :)

storage/mysql/log_storage_test.go
AlCutter: done

storage/mysql/log_storage_test.go
AlCutter: Done

storage/mysql/log_storage_test.go
AlCutter: done

storage/mysql/log_storage_test.go
AlCutter: done

storage/mysql/log_storage_test.go
AlCutter: done, as it happens I now have exactly that map from making each log have different numbers of leaves :)

storage/log_storage.go
codingllama: GetLogsUnsequencedCount, maybe?

storage/memory/log_storage.go
codingllama: Could you spell it out, please? I don't think q -> queue is easy to infer form this context.

storage/memory/log_storage.go
codingllama: Maybe add a comment, as it's difficult to guess whether the defer-in-loop is OK or a slip? It doesn't matter much in the test, as you said, but we might as well foster good practices throughout the entire codebase.

```go
defer  tree.mu.RUnlock() // OK to hold until the end of the method
```

storage/mysql/log_storage.go
codingllama: I'm a bit concerned about using this for monitoring, as I've no idea how you're planning to use this. Could you expand a bit on "this will be alright for now"?

storage/mysql/log_storage_test.go
```diff
@@ -977,38 +977,34 @@ func TestGetUnsequencedCounts(t *testing.T) {
...
 		}
-		if got, want := len(counts), numLogs; got != want {
-			t.Fatalf("GetUnsequencedCounts returns map with %d entries, want %d", got, want)
-		}
-		for id, c := range counts {
-			if got, want := c, expectedCount; got != want {
-				t.Fatalf("GetUnsequencedCounts: LogID %v has %d unsequenced, want %d", id, got, want)
-			}
+		if diff := pretty.Compare(expectedCount, got); diff != "" {
+			t.Errorf("GetUnsequencedCounts() = diff -want +got: %s", diff)
```
codingllama: nit: Add a line break before the diff?

t.Errorf("GetUnsequencedCounts() = diff -want +got:**\n**%s", diff)

storage/mysql/log_storage_test.go
```diff
@@ -977,38 +977,34 @@ func TestGetUnsequencedCounts(t *testing.T) {
...
 			if _, err := tx.QueueLeaves(ctx, leaves, fakeDequeueCutoffTime); err != nil {
 				t.Fatalf("Failed to queue leaves: %v", err)
 			}
 			commit(tx, t)
+			expectedCount[logID] += numToAdd
 		}
-		expectedCount += i
 
 		// Now check what we get back from GetUnsequencedCounts matches
 		tx, err := s.Snapshot(context.Background())
```
codingllama: nit: Use `ctx` instead of a new `context.Background()`.

storage/mysql/log_storage_test.go
```diff
@@ -977,38 +977,34 @@ func TestGetUnsequencedCounts(t *testing.T) {
...
 		if err != nil {
 			t.Fatalf("Snapshot() = (_, %v), want no error", err)
 		}
```
codingllama: defer-close the snapshot too?

storage/mysql/log_storage_test.go
```diff
@@ -977,38 +977,34 @@ func TestGetUnsequencedCounts(t *testing.T) {
 	s := NewLogStorage(DB, nil)
 
 	ctx := context.Background()
-	expectedCount := int64(0)
+	expectedCount := make(map[int64]int64)
 
 	for i := int64(1); i < 10; i++ {
```
codingllama: Suggestion: I would prefer this either as a one-shot test, with a fixed (but easier to read count), or a full table-driven thingy.  The outer loop isn't buying us much.

Eg:


```go
wantCount := map[int64]int64 {
  logs[0]: 0,
  logs[1]: 10,  // All arbitrary
  logs[2]: 5,
  logs[3]: 35,
}
for log, count := range wantCount {
  // queue leaves
}
// Snapshot, query, diff
```

AlCutter: That doesn't test that the `GetSequencedCounts()` will actually notice ongoing changes in the queue size - this is the point of the outer loop.

storage/mysql/log_storage_test.go
AlCutter: done, thanks!

storage/mysql/log_storage_test.go
AlCutter: done

AlCutter:
PTAL

storage/mysql/log_storage.go
AlCutter: if there are 20B entries in the queue, then I'd want to know - even if that data was 20mins out of date.
We can build quicker proxy metrics (e.g. `Rate(Sum(ctfe.num_added)) - Rate(Sum(signer.sequenced)))`) too, but ground-truth is the number of rows in the table, and it'd be nice to have that on a graph.

If this murders the DBs then we can revisit (this is really what I meant by "alright for now").

storage/memory/log_storage.go
```diff
@@ -342,6 +351,22 @@ func (t *logTreeTX) GetActiveLogIDs(ctx context.Context) ([]int64, error) {
...
+	t.ms.mu.RLock()
+	defer t.ms.mu.RUnlock()
+
+	ret := make(map[int64]int64)
+	for id, tree := range t.ms.trees {
+		tree.RLock()
+		defer tree.RUnlock() // OK to hold until method returns.
+
+		k := unseqKey(id)
+		queue := tree.store.Get(k).(*kv).v.(*list.List)
```
phad: LOL.

PR #746, opened 2017-07-17
Replace makeSuffixKey with NewNodeID().Split() 

gdbelvin:
In order for Trillian to retrieve the nodes it stores, `makeSuffixKey` must produce exactly the same output as`splitNodeID`. Rather than trying to keep multiple functions in-sync, this PR opts to generate all suffixKeys through `splitNodeID`.  As a result, the code becomes self-documenting. The code declares the exact location of the node that it is operating on, and the rest is handled automatically. 

`NodeID`s are created in several different ways throughout the codebase.  This PR gathers the various methods for node creation in a central place with several new `NewNodeID` functions:
- `NewNodeIDFromPrefix` creates `NodeIDs` from within `PopulateLogSubtreeNodes`
- `NewNodeIDFromRelativeBigInt` creates `NodeIDs` from within `PopulateMapSubtreeNodes`
- `NewNodeIDFromBigInt` creates `NodeIDs` from a future version of `PopulateMapSubtreeNodes`

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -58,6 +58,8 @@ const (
 	depthQuantum = 8
 	// logStrataDepth is the strata that must be used for all log subtrees.
 	logStrataDepth = 8
+	// maxLogDepth is the number of bits in a log path.
+	maxLogDepth = 64
 )
 
 // SubtreeCache provides a caching access to Subtree storage. Currently there are assumptions
@@ -130,23 +132,9 @@ func (s *SubtreeCache) stratumInfoForPrefixLength(numBits int) stratumInfo {
 
 // splitNodeID breaks a NodeID out into its prefix and suffix parts.
 // unless ID is 0 bits long, Suffix must always contain at least one bit.
-func (s *SubtreeCache) splitNodeID(id storage.NodeID) ([]byte, Suffix) {
-	if id.PrefixLenBits == 0 {
-		return []byte{}, Suffix{bits: 0, path: []byte{0}}
-	}
-	a := make([]byte, len(id.Path))
-	copy(a, id.Path)
+func (s *SubtreeCache) splitNodeID(id storage.NodeID) ([]byte, storage.Suffix) {
 	sInfo := s.stratumInfoForPrefixLength(id.PrefixLenBits - 1)
-	prefixSplit := sInfo.prefixBytes
-	sfx := Suffix{
-		bits: byte((id.PrefixLenBits-1)%sInfo.depth) + 1,
-		path: a[prefixSplit : prefixSplit+sInfo.depth/depthQuantum],
-	}
-	maskIndex := int((sfx.bits - 1) / depthQuantum)
-	maskLowBits := (sfx.bits-1)%depthQuantum + 1
-	sfx.path[maskIndex] &= ((0x01 << maskLowBits) - 1) << uint(depthQuantum-maskLowBits)
-
-	return a[:prefixSplit], sfx
+	return id.Split(sInfo.prefixBytes, sInfo.depth)
 }
 
 // preload calculates the set of subtrees required to know the hashes of the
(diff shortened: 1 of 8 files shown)


AlCutter:
Travis seems quite unhappy with borked hashes - I'll wait until you sort that before I take a look at the PR.

gdbelvin:
Unborked the hashes :-)

storage/suffix_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"bytes"
+	"encoding/base64"
+	"math/big"
+	"testing"
+)
+
+const (
+	logStrataDepth = 8
```
Martin2112: These seem to be duplicate consts that could get out of step?

storage/cache/subtree_cache.go
```diff
@@ -58,6 +58,8 @@ const (
 	depthQuantum = 8
 	// logStrataDepth is the strata that must be used for all log subtrees.
 	logStrataDepth = 8
+	// maxLogDepth is the number of bits in a log path.
```
gdbelvin: I'll add an issue for it. #754
At present, many of the log processing algorithms us `int64` for the leaf index, which basically caps the log at 64 levels anyways.  In an upcoming PR, I plan to create a consistent interface for `NodeID` which could help with this. Switching onto it would be an optional way to get around this constraint.

storage/types_test.go
```diff
@@ -230,3 +359,29 @@ func TestCoordString(t *testing.T) {
 		}
 	}
 }
+
+// h26 converts a hex string into an uint64.
+func h26(h string) uint64 {
```
gdbelvin: Fixed by a rebase.

storage/suffix.go
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+package storage
+
+import (
+	"encoding/base64"
+	"fmt"
+)
+
+// Suffix represents the tail of a NodeID, indexing into the Subtree which
```
gdbelvin: done.

storage/suffix.go
```diff
@@ -0,0 +1,53 @@
...
+// corresponds to the prefix of the NodeID.
+type Suffix struct {
+	// bits is the number of bits in the node ID suffix.
+	// TODO(gdbelvin): make bits an integer.
+	Bits byte
+	// path is the suffix itself.
+	Path []byte
+}
+
+// Serialize returns a base64 encoding of a byte array with the following format:
```
gdbelvin: It's used to make map keys. Perhaps we should just call it `String`?

storage/suffix_test.go
gdbelvin: Indeed they are duplicates of constants in `subtree/cache`. 
Since I was trying to test / document some assumptions of how this piece of code worked, it seemed like a good idea to use constants, but didn't seem worth the cost (and circular dependency) of linking these constants to the real ones.  It's also worth noting that these constants are baked into the assumptions of code correctness. We can't change them without significantly changing the related algorithms.

storage/suffix_test.go
```diff
@@ -0,0 +1,125 @@
...
+func TestSuffixKey(t *testing.T) {
+	for _, tc := range []struct {
+		depth   int
+		index   int64
+		want    []byte
+		wantErr bool
+	}{
+		{depth: 0, index: 0x00, want: h2b("0000"), wantErr: false},
+		{depth: 8, index: 0x00, want: h2b("0800"), wantErr: false},
+		// TODO(gdbelvin): want: "0fabcd"?
```
gdbelvin: Kindof...  

`makeSuffixKey` has been deprecated in this PR in favor of `NodeID.Split()`.  To prove to myself that I was keeping the exact same behavior, however, I moved `makeSuffixKey` to `suffix_test.go` and added this test to document its behavior. Because `makeSuffixKey` assumed a maximum subtree depth of 8, it caused problems for the map, which has deep subtrees. 

All these issues have been resolved with the migration to `NodeID.Split()`, which produces the correct output for both log and map sub protos -- and maintains the exact same byte format for the Log's cases.

storage/cache/subtree_cache.go
Martin2112: Going past 64 bit depths would be a major upheaval, which is not likely to happen. But we should remove the need for users to hardcode the magic number '64'.

storage/suffix.go
Martin2112: It doesn't produce a readable version though so it's arguable whether to use String().

storage/suffix_test.go
Martin2112: I know about the constraints, I wrote a lot of that code. If we've got constants available that can't be changed without breakage then it might be better to use them. I'll leave that decision to you.

storage/suffix_test.go
Martin2112: OK so are we removing the TODOs?

storage/suffix_test.go
gdbelvin: I agree. I'll add a TODO to remove the duplicate constants when the import cycle from `cache` to `merkle` that depends on `storage.NodeID` is resolved.

storage/suffix_test.go
gdbelvin: Yes, I'll removed `makeSuffixKey` as it's been properly deprecated.

storage/cache/subtree_cache.go
gdbelvin: Agreed

PR #752, opened 2017-07-19
Record quota and interceptor related metrics

codingllama:
The following metrics are now recorded:

Quotas
* Tokens acquired (per spec)
* Tokens replenished (per spec)
* Tokens returned (per spec)

The distinction between replenished and returned is that the former is sequencing-related, whereas the latter indicates tokens were incorrectly charged (errors requests, duplicate leaves queued, etc)

Interceptor
* Total number of intercepted requests
* Denied requests (per type of denial)

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -381,15 +381,17 @@ func (s Sequencer) SequenceBatch(ctx context.Context, logID int64, limit int, gu
 	// where the tokens come from.
 	if numLeaves > 0 {
 		tokens := int(float64(numLeaves) * quotaIncreaseFactor())
-		glog.V(2).Infof("Replenishing %v tokens for tree %v (numLeaves = %v)", tokens, logID, leaves)
-		if err := s.qm.PutTokens(ctx, tokens, []quota.Spec{
+		specs := []quota.Spec{
 			{Group: quota.Tree, Kind: quota.Read, TreeID: logID},
 			{Group: quota.Tree, Kind: quota.Write, TreeID: logID},
 			{Group: quota.Global, Kind: quota.Read},
 			{Group: quota.Global, Kind: quota.Write},
-		}); err != nil {
+		}
+		glog.V(2).Infof("Replenishing %v tokens for tree %v (numLeaves = %v)", tokens, logID, leaves)
+		if err := s.qm.PutTokens(ctx, tokens, specs); err != nil {
 			glog.Warningf("Failed to replenish %v tokens for tree %v: %v", tokens, logID, err)
 		}
+		quota.Metrics.IncReplenished(tokens, specs)
 	}
 
 	seqCounter.Add(float64(numLeaves), label)
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -388,10 +388,11 @@ func (s Sequencer) SequenceBatch(ctx context.Context, logID int64, limit int, gu
 			{Group: quota.Global, Kind: quota.Write},
 		}
 		glog.V(2).Infof("Replenishing %v tokens for tree %v (numLeaves = %v)", tokens, logID, leaves)
-		if err := s.qm.PutTokens(ctx, tokens, specs); err != nil {
+		err := s.qm.PutTokens(ctx, tokens, specs)
+		if err != nil {
 			glog.Warningf("Failed to replenish %v tokens for tree %v: %v", tokens, logID, err)
 		}
-		quota.Metrics.IncReplenished(tokens, specs)
+		quota.Metrics.IncReplenished(tokens, specs, err != nil)
 	}
 
 	seqCounter.Add(float64(numLeaves), label)
--- a/quota/metrics.go
+++ b/quota/metrics.go
@@ -14,7 +14,11 @@
 
 package quota
 
-import "github.com/google/trillian/monitoring"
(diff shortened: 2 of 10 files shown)


log/sequencer.go
```diff
@@ -381,15 +381,17 @@ func (s Sequencer) SequenceBatch(ctx context.Context, logID int64, limit int, gu
...
 			{Group: quota.Tree, Kind: quota.Write, TreeID: logID},
 			{Group: quota.Global, Kind: quota.Read},
 			{Group: quota.Global, Kind: quota.Write},
-		}); err != nil {
+		}
+		glog.V(2).Infof("Replenishing %v tokens for tree %v (numLeaves = %v)", tokens, logID, leaves)
+		if err := s.qm.PutTokens(ctx, tokens, specs); err != nil {
 			glog.Warningf("Failed to replenish %v tokens for tree %v: %v", tokens, logID, err)
 		}
+		quota.Metrics.IncReplenished(tokens, specs)
```
AlCutter: This increments even if the `PutTokens()` fails?

quota/quota.go
```diff
@@ -67,6 +69,34 @@ type Spec struct {
 	User string
 }
 
+// Name returns a textual representation of the Spec. Names are constant and may be relied not to
```
AlCutter: "relied upon"

quota/quota.go
```diff
@@ -67,6 +69,34 @@ type Spec struct {
...
+		return fmt.Sprintf("%v/%v", group, kind)
+	}
+	var user string
+	switch s.Group {
+	case Tree:
+		user = fmt.Sprint(s.TreeID)
+	case User:
+		user = s.User
+	}
+	return fmt.Sprintf("%vs/%v/%v", group, user, kind)
```
AlCutter: %vs intended?

quota/quota.go
```diff
@@ -67,6 +69,34 @@ type Spec struct {
...
+}
+
+// String returns a description of Spec.
+func (s Spec) String() string {
```
AlCutter: Why not rename Name() -> String() and ding this?

server/interceptor/interceptor.go
```diff
@@ -84,31 +111,36 @@ type trillianProcessor struct {
...
 		}
 		ctx = trees.NewContext(ctx, tree)
 		// TODO(codingllama): Add auth interception
 	}
 
 	if len(info.specs) > 0 && info.tokens > 0 {
+		quota.Metrics.IncAcquired(info.tokens, info.specs)
 		if err := tp.parent.QuotaManager.GetTokens(ctx, info.tokens, info.specs); err != nil {
 			if !tp.parent.QuotaDryRun {
+				increment(requestDeniedCounter, insufficientTokensLabel)
```
AlCutter: I'd like to know which tree/user this is affecting, too...

quota/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package quota
+
+import "github.com/google/trillian/monitoring"
+
+// Metrics groups all quota-related metrics.
+// The metrics represented here are not meant to be maintained by the quota subsystem
+// implementation.  Instead, they're meant to be updated by the quota's callers, in order to record
```
AlCutter: Could this be achieved by an indirection layer? A pass-through QuotaManager shim which handles the metrics and then calls on to the actual registered quota manager impl?

codingllama: QuotaManager can't tell the difference between PutTokens/Replenish and PutTokens/Return. We could change the interface, although I'm not sure that's the best route.

quota/quota.go
codingllama: Done.

quota/quota.go
codingllama: Yes. See the examples / tests.

quota/quota.go
codingllama: It's a personal nit: a spec name is very specific thing, whereas its String() form isn't necessarily the same (it could be more descriptive, for example).

log/sequencer.go
codingllama: Yes. My reasoning for not differentiating errors is that this metric, as the other quota-related metrics in the PR, records the server's interactions with the quota system. The server did ask for replenishment, so not recording it makes it seem like it's misbehaving, when in fact it was the quota system that failed. The exact picture of how the quota system is doing has to come from within itself, as we can't account for some things externally (time-based replenishment, cfg changes, disabled quotas, etc). Plus, if we use the outside metrics to record the intended interactions, we can see how well the count "closes" from the server's POV (ie, if tokens spent - tokens replenished = 0).

Let me know if that makes sense.

server/interceptor/interceptor.go

codingllama:
PTAL?

I've moved interceptor package metrics to their own file and fixed labels, which was a bit bonkers before.

codingllama:
Ping?

log/sequencer.go
codingllama: Discussed offline: I've added a "success" label to quota metrics, as suggested by @AlCutter. That should cover both cases.

codingllama:
PTAL?

PR #753, opened 2017-07-20
Fix MapperMetadata storage error

gdbelvin:
The Get/Set operations for `SignedMapRoot`s are not equal, resulting in data loss and incorrect signatures on the `SignedMapRoot`. 

Store: 
```go
&trillian.SignedMapRoot{ Metadata: &trillian.MapperMetadata{ HighestFullyCompletedSeq: 0 }}  
```

Upon retrieval, we expect to get the same object back, but instead we get: 
```go 
&trillian.SignedMapRoot{ Metadata: nil } 
```

This is because the `MapperMetadata` above only contains default values, resulting in a serialized protobuf zero bytes long. Previously this was treated the same way as the sub proto not being there at all, which is semantically different, and produces a different ObjectHash. 

This PR fixes this (a 1/2 line change), and adds a bunch of unit tests to catch this kind of error in the future. 

Fixes google/keytransparency#710


crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// toFromBytes serializes obj using proto.Marshal, and returns the unmarshaled copy.
+func toFromBytes(obj proto.Message) (proto.Message, error) {
+	b, err := proto.Marshal(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	obj2 := obj
```
codingllama: obj has to be a pointer, otherwise `Reset()` wouldn't work, so `obj2 := obj; obj2.Reset()` ends up being the same as `obj.Reset()`.

You should also note in the docs that this method modifies the input, which I wouldn't expect from the name.

crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/keys"
+	"github.com/google/trillian/testonly"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/kylelemons/godebug/pretty"
+)
+
+// toFromBytes serializes obj using proto.Marshal, and returns the unmarshaled copy.
+func toFromBytes(obj proto.Message) (proto.Message, error) {
```
codingllama: What's the purpose of this method?  Would you ever expect the returned value to differ from obj?

crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+	obj2 := obj
+	obj2.Reset()
+	if err := proto.Unmarshal(b, obj2); err != nil {
+		return nil, err
+	}
+	return obj2, nil
+}
+
+// toFromJSON serializes obj using protopb.Marshal, and returns the unmarshaled copy.
+func toFromJSON(obj proto.Message) (proto.Message, error) {
```
codingllama: Comments above apply here too.

crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+			t.Errorf("SignObject(%#v): %v", tc.obj, err)
+			continue
+		}
+
+		objFromBytes, err := toFromBytes(tc.obj)
+		t.Logf("fromBytes: %v", pretty.Sprint(objFromBytes))
+		if err != nil {
+			t.Errorf("round trip to bytes failed for obj: %#v: %v", tc.obj, err)
+		} else {
+			if err := VerifyObject(key.Public(), objFromBytes, sig); err != nil {
```
codingllama: nit: Combine the if with the else above?

crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+				t.Errorf("fromBytes: VerifyObject(%#v): %v", objFromBytes, err)
+			}
+		}
+		objFromJSON, err := toFromJSON(tc.obj)
+		t.Logf("fromJSON: %v", pretty.Sprint(objFromJSON))
+		if err != nil {
+			t.Errorf("round trip to bytes failed for obj: %#v: %v", tc.obj, err)
+		} else {
+			if err := VerifyObject(key.Public(), objFromJSON, sig); err != nil {
```
codingllama: nit: Ditto (combine the if).

crypto/object.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
Martin2112: 2017

crypto/verifier.go
```diff
@@ -37,17 +35,6 @@ var (
 	}
 )
 
-// VerifyObject verifies the output of Signer.SignObject.
-func VerifyObject(pub crypto.PublicKey, obj interface{}, sig *sigpb.DigitallySigned) error {
```
codingllama: Suggestion: I'd probably leave this function here too, removing the newly-created object.go.

crypto/object_test.go
Martin2112: I think this needs a different name. roundtripJSON?

crypto/object_test.go
```diff
@@ -0,0 +1,152 @@
...
+		objFromBytes, err := toFromBytes(tc.obj)
+		t.Logf("fromBytes: %v", pretty.Sprint(objFromBytes))
+		if err != nil {
+			t.Errorf("round trip to bytes failed for obj: %#v: %v", tc.obj, err)
+		} else {
+			if err := VerifyObject(key.Public(), objFromBytes, sig); err != nil {
+				t.Errorf("fromBytes: VerifyObject(%#v): %v", objFromBytes, err)
+			}
+		}
+		objFromJSON, err := toFromJSON(tc.obj)
```
codingllama: VerifyObject appears to be previously untested, so I'm all in favour of testing it, but I'm failing to see what the marshal-to-proto and marshal-to-json logic is adding to this test.  It seems like we included a step that tests whether marshaling works, which I don't think belongs in our codebase.

storage/mysql/map_storage_test.go
```diff
@@ -143,43 +144,69 @@ func TestMapRootUpdate(t *testing.T) {
...
+				t.Fatalf("Failed to store signed map root: %v", err)
+			}
+			if err := tx.Commit(); err != nil {
+				t.Fatalf("Failed to commit new map roots: %v", err)
+			}
+		}
+		{
+			tx := beginMapTx(ctx, s, mapID, t)
+			defer tx.Close()
+			root3, err := tx.LatestSignedMapRoot(ctx)
```
codingllama: nit: Rename `root3` to `root` ?

storage/mysql/map_storage_test.go
```diff
@@ -143,43 +144,69 @@ func TestMapRootUpdate(t *testing.T) {
 	s := NewMapStorage(DB)
 
 	ctx := context.Background()
-	tx := beginMapTx(ctx, s, mapID, t)
```
codingllama: Delete the comment at line 141? Or, if the intent of the test was to ensure we can't override a newer root with an older one, then we should add such a test to the table. (The comment seems to suggest that.)

storage/mysql/map_storage_test.go
```diff
@@ -143,43 +144,69 @@ func TestMapRootUpdate(t *testing.T) {
...
-	tx = beginMapTx(ctx, s, mapID, t)
-	defer tx.Close()
-	root3, err := tx.LatestSignedMapRoot(ctx)
-	if err != nil {
-		t.Fatalf("Failed to read back new map root: %v", err)
-	}
-	if !proto.Equal(&root2, &root3) {
-		t.Fatalf("Root round trip failed: <%v> and: <%v>", root, root2)
+	for _, tc := range []struct {
+		root trillian.SignedMapRoot
```
codingllama: Please add a `desc string` for each test and prefix the error messages with it. It's a bit more friendly than comparing roots.

storage/mysql/map_storage_test.go
```diff
@@ -143,43 +144,69 @@ func TestMapRootUpdate(t *testing.T) {
...
+		}
+		{
+			tx := beginMapTx(ctx, s, mapID, t)
+			defer tx.Close()
+			root3, err := tx.LatestSignedMapRoot(ctx)
+			if err != nil {
+				t.Fatalf("Failed to read back new map root: %v", err)
+			}
+			if got, want := &root3, &tc.root; !proto.Equal(got, want) {
+				t.Fatalf("LatestSignedMapRoot(): %v, want %v",
```
codingllama: Could you print a diff here instead?

```go
diff := pretty.Compare(got, want)
t.Errorf("%v: post-LatestSignedMapRoot() diff (-got +want):\n%v", tc.desc, diff)
```

storage/mysql/map_storage_test.go
```diff
@@ -143,43 +144,69 @@ func TestMapRootUpdate(t *testing.T) {
...
+			Metadata: &trillian.MapperMetadata{
+				HighestFullyCompletedSeq: 1,
+			},
+		}},
+	} {
+		{
+			tx := beginMapTx(ctx, s, mapID, t)
+			defer tx.Close()
+			if err := tx.StoreSignedMapRoot(ctx, tc.root); err != nil {
+				t.Fatalf("Failed to store signed map root: %v", err)
```
codingllama: Are all those `Fatalf`s meant to be fatal? I guess the ones related to the method in under test (StoreSignedMapRoot) should be `Errorf / continue` instead.

gdbelvin:
Updated per reviewer comments. PTAL

storage/mysql/map_storage_test.go
```diff
@@ -137,49 +138,83 @@ type rootReaderMapTX interface {
...
+		}
+		{
+			tx := beginMapTx(ctx, s, mapID, t)
+			defer tx.Close()
+			root, err := tx.LatestSignedMapRoot(ctx)
+			if err != nil {
+				t.Fatalf("%v: Failed to read back new map root: %v", tc.desc, err)
+			}
+			if got, want := &root, &tc.root; !proto.Equal(got, want) {
+				t.Fatalf("%v: LatestSignedMapRoot(): %v, diff(-got, +want) %v", tc.desc,
```
codingllama: nit: Add a line break before printing the diff?

*t.Fatalf("... diff(-got, +want)**\n**%v", ...)*

crypto/verifier_test.go
```diff
@@ -81,3 +82,59 @@ func TestSignVerify(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to open test key, err=%v", err)
+	}
+	signer := NewSHA256Signer(key)
+
+	type subfield struct {
+		c int
+	}
+
+	for _, tc := range []struct {
```
codingllama: Suggestion: You could use a []interface{} directly, given that the struct has only one field.

There's an argument for keeping the struct as well, so feel free to ignore.

`for _, tc := range []interface{...} { ... }`

storage/mysql/map_storage_test.go
gdbelvin: done

crypto/verifier_test.go
gdbelvin: I'll keep the struct around in case we want to extend the test later.

PR #755, opened 2017-07-21
Disallow user quotas to use sequencing-based replenishment

codingllama:
There's no relationship between sequencing and users, therefore the combination
doesn't make sense.


codingllama:
Ping?

PR #758, opened 2017-07-26
Etcd-based quota.Manager implementation

codingllama:


codingllama:
Issue #490.

codingllama:
Ping?

quota/etcd/etcdqm/etcdqm.go
```diff
@@ -0,0 +1,129 @@
...
+	qs *storage.QuotaStorage
+}
+
+// New returns a new etcd-based quota.Manager.
+func New(client *clientv3.Client) quota.Manager {
+	return &manager{qs: &storage.QuotaStorage{Client: client}}
+}
+
+func (m *manager) GetUser(ctx context.Context, req interface{}) string {
+	return "" // Unused
```
Martin2112: Why do we have this lever?

quota/etcd/etcdqm/etcdqm.go
```diff
@@ -0,0 +1,129 @@
...
+		group = "trees"
+		user = fmt.Sprint(spec.TreeID)
+	case quota.User:
+		group = "users"
+		user = spec.User
+	default:
+		return "", fmt.Errorf("unexpected group: %v", spec.Group)
+	}
+
+	var kind string
```
Martin2112: How about moving the kind validation up above group. Then each group case can completely handle it and return. As it is something could go wrong if Users and Trees ever diverged in how they should be handled.

quota/etcd/etcdqm/etcdqm.go
codingllama: That's the lever we can use for root-based quotas, for example. It's unused now, but it could easily come to life in the future.

quota/etcd/etcdqm/etcdqm.go
codingllama: Actually, we've now merged the PR I needed to address the TODO at line 99, so I've simplified this quite a bit.

codingllama:
Comments addressed, PTAL?

quota/etcd/etcdqm/etcdqm.go
Martin2112: Might have been better to add it when required but 'user' seems a reasonable concept for this code so we'll leave it like this.

quota/etcd/etcdqm/etcdqm.go
Martin2112: That looks better now.

PR #761, opened 2017-07-27
Fix MapBatchTest

gdbelvin:
Fixed the following problems:

- NewNodeIDFromRelativeBitInt
  - Was placing subtree indexes in a left justified rather than right
    justified manner.

- NodeID.Split
  - Modular arithmetic for suffix.Bits assumed equal strata levels


integration/maptest/map_test.go
```diff
@@ -366,3 +365,12 @@ func TestNonExistentLeaf(t *testing.T) {
 		}
 	}
 }
+
+// h2b converts a hex string into []byte.
```
AlCutter: There's `testonly.MustHexDecode` which does the same thing, BTW.

storage/types.go
```diff
@@ -299,11 +304,19 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, Suffix) {
...
 	copy(a, n.Path)
 
+	bits := n.PrefixLenBits - prefixBytes*8
+	if bits > suffixBits {
+		panic(fmt.Sprintf("storage Split: %x(n.PrefixLenBits: %v - prefixBytes: %v *8) > %v", n.Path, n.PrefixLenBits, prefixBytes, suffixBits))
+	}
+	if bits == 0 {
+		panic(fmt.Sprintf("storage Split: %x(n.PrefixLenBits: %v - prefixBytes: %v *8) == 0", n.Path, n.PrefixLenBits, prefixBytes))
+	}
+	suffixBytes := (bits + 7) / 8
```
AlCutter: you can use `bytesForBits` for this :)

Martin2112: He might have removed that in a previous PR?

AlCutter: It lives on (at least currently): https://github.com/google/trillian/blob/d9f0b5187c3cca04fa30a9fdc63fb3afc595fda4/storage/types.go#L71

integration/maptest/map_test.go
```diff
@@ -284,7 +283,7 @@ func RunMapBatchTest(ctx context.Context, env *integration.MapEnv, tree *trillia
 
 		if err := verifyGetMapLeavesResponse(getResp, indexes, int64(numBatches),
 			pubKey, hasher, tree.TreeId); err != nil {
-			return err
+			return fmt.Errorf("batch: %v,  %v", i, err)
```
gdbelvin: done

integration/maptest/map_test.go
gdbelvin: Shorter names are nice?

storage/types.go
gdbelvin: done

integration/maptest/map_test.go
AlCutter: So is deduplication, how about this:
```golang
var h2b = testonly.MustHexDecode
```

gdbelvin: ohhhh, I like!

gdbelvin: done

PR #762, opened 2017-07-27
Add --generate-key flag to createtree command

RJPercival:
Makes the Trillian admin server generate the private key, rather than requiring the user to provide one. This may be more secure, depending on how the server stores the generated key.

It is hard-coded to generate a key using the default parameters. Users that require more control over the key specification should communicate with the admin server directly.

### DIFF
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -154,10 +154,19 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
 		}
 
 		// If no key flags were provided at all, get Trillian to generate a key.
-		ctr.KeySpec = &keyspb.Specification{
-			Params: &keyspb.Specification_EcdsaParams{
+		ctr.KeySpec = &keyspb.Specification{}
+
+		switch sigpb.DigitallySigned_SignatureAlgorithm(sa) {
+		case sigpb.DigitallySigned_ECDSA:
+			ctr.KeySpec.Params = &keyspb.Specification_EcdsaParams{
 				EcdsaParams: &keyspb.Specification_ECDSA{},
-			},
+			}
+		case sigpb.DigitallySigned_RSA:
+			ctr.KeySpec.Params = &keyspb.Specification_RsaParams{
+				RsaParams: &keyspb.Specification_RSA{},
+			}
+		default:
+			return nil, fmt.Errorf("unsupported signature algorithm: %v", sa)
 		}
 	}
 
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -16,10 +16,7 @@
 // command.
 //
 // Example usage:
-// $ ./createtree \
-//     --admin_server=host:port \
-//     --pem_key_path=/path/to/pem/file \
-//     --pem_key_password=mypassword
+// $ ./createtree --admin_server=host:port
 //
 // The command outputs the tree ID of the created tree to stdout, or an error to
 // stderr in case of failure. The output is minimal to allow for easy usage in
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -65,7 +65,7 @@ var (
 	description        = flag.String("description", "", "Description of the new tree")
 	maxRootDuration    = flag.Duration("max_root_duration", 0, "Interval after which a new signed root is produced despite no submissions; zero means never")
 
-	privateKeyFormat = flag.String("private_key_format", "PrivateKey", "Type of private key to be used (PrivateKey, PEMKeyFile, or PKCS11ConfigFile)")
+	privateKeyFormat = flag.String("private_key_format", "", "Type of protobuf message to send the key as (PrivateKey, PEMKeyFile, or PKCS11ConfigFile). If empty, a key will be generated for you by Trillian.")
 	pemKeyPath       = flag.String("pem_key_path", "", "Path to the private key PEM file")
 	pemKeyPassword   = flag.String("pem_key_password", "", "Password of the private key PEM file")
 	pkcs11ConfigPath = flag.String("pkcs11_config_path", "", "Path to the PKCS #11 key configuration file")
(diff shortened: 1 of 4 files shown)


cmd/createtree/main.go
```diff
@@ -65,7 +65,7 @@ var (
 	description        = flag.String("description", "", "Description of the new tree")
 	maxRootDuration    = flag.Duration("max_root_duration", 0, "Interval after which a new signed root is produced despite no submissions; zero means never")
 
-	privateKeyFormat = flag.String("private_key_format", "PrivateKey", "Type of private key to be used (PrivateKey, PEMKeyFile, or PKCS11ConfigFile)")
+	privateKeyFormat = flag.String("private_key_format", "", "Type of protobuf message to send the key as (PrivateKey, PEMKeyFile, or PKCS11ConfigFile). If empty, a key will be generated for you by Trillian.")
```
gdbelvin: If empty, the list of valid choices will be printed out?

cmd/createtree/main.go
```diff
@@ -144,9 +139,31 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
 		SignatureAlgorithm: sigpb.DigitallySigned_SignatureAlgorithm(sa),
 		DisplayName:        opts.displayName,
 		Description:        opts.description,
-		PrivateKey:         pk,
 		MaxRootDuration:    ptypes.DurationProto(opts.maxRootDuration),
 	}}
+
+	if opts.privateKeyType != "" {
```
gdbelvin: Would it be possible to just set this value to a sensible default and let the rest of the code take care of the rest?

cmd/createtree/main_test.go
```diff
@@ -178,17 +187,22 @@ xToc6NWBri0N3VVsswIDAQAB
 			continue
 		}
 
-		if diff := pretty.Compare(tree, test.wantTree); diff != "" {
-			t.Errorf("%v: post-createTree diff:\n%v", test.desc, diff)
+		if !proto.Equal(tree, test.wantTree) {
```
gdbelvin: good change. consider using the full `got, want` format

gdbelvin:
Love the shortness of this PR.  We seem to have some extra flag processing that would be nice to simplify.

cmd/createtree/main_test.go
```diff
@@ -218,6 +232,9 @@ func (s *fakeAdminServer) CreateTree(ctx context.Context, req *trillian.CreateTr
 		return nil, s.err
 	}
 	resp := *req.Tree
+	if req.KeySpec != nil && s.generatedKey != nil {
```
codingllama: Is it worth to error `if req.KeySpec != nil && s.generatedKey == nil` ?

cmd/createtree/main.go
RJPercival: An earlier version of this PR had such a flag, but I removed it because this reduces the number of mutually-exclusive flags and leads to more reasonable default behaviour. I want to encourage letting Trillian generate the key, so I'd like that to be the default. This would require the `--generateKey` flag defaulting to true. However, that would mean you'd need to pass `--generate_key=false` when providing your own key else you'd get an error. Without this flag, it automatically won't generate a key if you provide the flags that specify one - this seemed cleaner.

Of course, as another of your comments seemed to suggest, this will clash with PR #734, which makes an empty --private_key_format flag print out the accepted formats. I could add a --list_private_key_formats flag for that, or just not provide that feature (users should probably have researched the protobuf message type they want to use, rather than blindly pick from a list).

cmd/createtree/main_test.go
RJPercival: In the past, I've gotten the impression that you'd rather avoid having test code that simply detects whether the test is misconfigured. However, I'm happy to add this if you'd prefer in this case.

cmd/createtree/main.go
gdbelvin: I'm happy if we don't implement the list formats feature

cmd/createtree/main.go
```diff
@@ -144,9 +139,31 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
...
+		pk, err := newPK(opts)
+		if err != nil {
+			return nil, err
+		}
+		ctr.Tree.PrivateKey = pk
+	} else {
+		// Cannot continue if options specifying a key were provided but
```
gdbelvin: This is a little confusing. If we want to generate a key, doesn't that take precedence over the other parameters?  Perhaps issue a warning that these options will be ignored?

cmd/createtree/main.go
```diff
@@ -144,9 +139,31 @@ func newRequest(opts *createOpts) (*trillian.CreateTreeRequest, error) {
...
+		// privateKeyType is not set, as there's no way to know what protobuf
+		// message type was intended.
+		if opts.pemKeyPath != "" || opts.pemKeyPass != "" || opts.pkcs11ConfigPath != "" {
```
gdbelvin: This is a little confusing. If we want to generate a key, doesn't that take precedence over the other parameters?  Perhaps issue a warning that these options will be ignored?

cmd/createtree/main_test.go
codingllama: True, but in this case I think it's more about mimicking AdminServer and less about the test checking itself. This is a suggestion that doesn't change much in practice right now, so feel free to ignore it.

cmd/createtree/main.go
RJPercival: If the user specified these options, I think it indicates that they don't want to generate a key, but have inadvertently omitted the key format flag. Therefore, printing an error and aborting seems better, rather than continuing on the (probably incorrect) assumption that a key should be generated.

cmd/createtree/main_test.go

cmd/createtree/main_test.go
RJPercival: I've made it panic in this case. While I know panic is to be avoided, it seems the best option here as I don't want tests to accidentally interpret this mistake in test setup as an expected error (in those test cases where `fakeAdminServer.err != nil`).

RJPercival:
Travis will never be happy with this PR, as a corresponding change to certificate-transparency-go is required (google/certificate-transparency-go#55).

RJPercival:
Travis is now happy.

PR #765, opened 2017-07-28
Use absolute addressing in HStar2

### LINKED ISSUE #670: Map: CONIKS map hasher

gdbelvin:
Secure versions of `HashEmpty` take the node's address into account. See #670.
This PR modifies `HStar2` and its callers to use `NodeID`s for absolute addressing in order to supply full paths to the hashing functions. 

**HStar2 Changes**

`HStar2` in its current form only uses subtree relative addressing. 
This PR modifies the `HStar2` algorithm from operating in terms of subtree height to absolute depth. 

from 
`hStar2b(n int, values []HStar2LeafHash, offset *big.Int)`
to
`hStar2b(depth, maxDepth int, values []HStar2LeafHash, offset *big.Int)`

And it also provides `HStar2Nodes` with the prefix of the subtree it is operating on.

**NodeID Changes**

In order to provide correct conversions to/from `Hstar2LeafHash`, `NodeID`, `Prefix` and `Suffix` the following functions are added and tested: 
- `ParseSuffix`
- `NodeID.BigInt`
- `NewNodeIDFromBigInt`
- `NewNodeIDFromPrefixSuffix`

**Verifier Changes**

On the verification side, the verification algorithm is also modified to provide a correct path to `HashEmpty`. 

Closes #670 
Rebased on #761 and #772


integration/maptest/map_test.go
```diff
@@ -184,17 +200,17 @@ func TestInclusion(t *testing.T) {
...
 		desc                  string
 		HashStrategy          trillian.HashStrategy
 		batchSize, numBatches int
 	}{
+
 		{
 			desc:         "maphasher batch",
 			HashStrategy: trillian.HashStrategy_TEST_MAP_HASHER,
-			batchSize:    64, numBatches: 32,
+			batchSize:    1, numBatches: 1,
```
gdbelvin: revert

merkle/hstar2.go
```diff
@@ -76,78 +72,88 @@ type SparseSetNodeFunc func(depth int, index *big.Int, hash []byte) error
...
+		glog.Infof("HStar2Nodes(%x, %v, %v)", prefix, subtreeDepth, len(values))
 		for _, v := range values {
 			glog.Infof("  %x: %x", v.Index.Bytes(), v.LeafHash)
 		}
 	}
-	if treeLevelOffset < 0 {
+	depth := len(prefix) * 8
+	totalDepth := depth + subtreeDepth
+	if totalDepth > s.hasher.BitLen() {
 		return nil, ErrNegativeTreeLevelOffset
```
gdbelvin: change error

merkle/hstar2.go
```diff
@@ -76,78 +72,88 @@ type SparseSetNodeFunc func(depth int, index *big.Int, hash []byte) error
...
-			if h != nil {
-				return h, nil
-			}
-			// otherwise just return the null hash for this level
-			return s.hasher.HashEmpty(s.treeID, PaddedBytes(index, s.hasher.Size()), depth+treeLevelOffset), nil
-		},
-		func(depth int, index *big.Int, hash []byte) error {
-			return set(treeDepth-depth, index, hash)
-		})
+	offset := new(big.Int).SetBytes(prefix)
```
gdbelvin: Use NewFromPreifx

merkle/hstar2.go
```diff
@@ -76,78 +72,88 @@ type SparseSetNodeFunc func(depth int, index *big.Int, hash []byte) error
...
+		h, err := getter(depth, index)
+		if err != nil {
+			return nil, err
+		}
+		// if we got a value then we'll use that
+		if h != nil {
+			return h, nil
+		}
+	}
+	height := s.hasher.BitLen() - depth
```
gdbelvin: TODO: convert hashers back to accepting depth arguments

merkle/hstar2_test.go
```diff
@@ -127,20 +127,27 @@ func TestHStar2GetSet(t *testing.T) {
...
-		lh[i].Index.SetBytes(b[prefixSize/8:])
-		root, err := s.HStar2Root(s.hasher.BitLen()-prefixSize, []HStar2LeafHash{lh[i]})
+		prefix = prefix[:prefixSize/8] // We only want the first prefixSize bytes.
+		root, err := s.HStar2Nodes(prefix, subtreeDepth, []HStar2LeafHash{lh[i]}, nil, nil)
 		if err != nil {
 			t.Fatalf("Failed to calculate root %v", err)
 		}
-		ret = append(ret, HStar2LeafHash{prefix, root})
+
+		index := new(big.Int).SetBytes(prefix)
```
gdbelvin: Use NewNode Functions

merkle/sparse_merkle_tree.go
```diff
@@ -209,49 +211,51 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context) {
 			s.root <- rootHashOrError{hash: nil, err: err}
 			return
 		}
-		leaves = append(leaves, HStar2LeafHash{Index: new(big.Int).SetBytes(ih.index), LeafHash: ih.hash})
+		node := storage.NewNodeIDFromPrefixSuffix(ih.index, storage.Suffix{}, s.hasher.BitLen())
```
gdbelvin: s/node/nodeID/

merkle/sparse_merkle_tree.go
```diff
@@ -209,49 +211,51 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context) {
...
+
+		leaves = append(leaves, HStar2LeafHash{
+			Index:    node.BigInt(),
+			LeafHash: ih.hash,
+		})
 		nodesToStore = append(nodesToStore,
 			storage.Node{
-				NodeID:       storage.NewNodeIDFromHash(bytes.Join([][]byte{s.prefix, ih.index}, []byte{})),
+				NodeID:       storage.NewNodeIDFromHash(ih.index),
```
gdbelvin: XXX: should this be the above node?

integration/maptest/map_test.go
gdbelvin: done

merkle/hstar2.go
gdbelvin: done

merkle/hstar2.go
gdbelvin: done

merkle/hstar2.go
gdbelvin: done

merkle/hstar2_test.go
gdbelvin: done

merkle/sparse_merkle_tree.go
gdbelvin: done

merkle/sparse_merkle_tree.go
gdbelvin: done

Martin2112:
Can we drop the test timeout change from the PR?

gdbelvin:
Now that we're running `BatchTest` as part of the integration tests on travis, we're going to need more time. We have a couple of options:

- Set the timeout to 9 minutes, so it's less than Travis's timeout.
- Annotate tests with short / long and only run the short tests on travis. (use the `go test -short` flag) 
- Increase the Travis timeout to be > 10 min.  

WDYT?

Martin2112:
Can we get a rough breakdown of where the time is being spent? I'm reluctant to have people waiting > 10 mins for test runs but if it's necessary to ensure correctness then we might have to.

gdbelvin:
Here's the output of a run on my local machine (which is a bit faster than travis). It's clear that the large batch test is taking up almost all of the time. 

=== RUN   TestInclusion
--- PASS: TestInclusion (0.11s)
=== RUN   TestInclusionBatch
--- PASS: TestInclusionBatch (30.80s)
=== RUN   TestNonExistentLeaf
--- PASS: TestNonExistentLeaf (0.02s)
PASS

Martin2112:
Do they run at this speed on Travis? I think 30 seconds isn't unreasonable. But we keep adding tests the cumulatively take a long time to run.

gdbelvin:
They run much slower on travis.  I've set the timeout back to 5 minutes and set travis to only run the short version of tests. 

I should note here that the full integration tests also get run without a timeout through `./integration/integration_test.sh`

Martin2112:
I think this PR needs a rebase / merge now?

gdbelvin:
Rebased and ready for review

merkle/hstar2_test.go
```diff
@@ -130,20 +131,25 @@ func TestHStar2GetSet(t *testing.T) {
...
+	s := NewHStar2(treeID, hasher)
 	for i := range lh {
-		prefix := new(big.Int).Rsh(lh[i].Index, uint(s.hasher.BitLen()-prefixSize))
-		b := lh[i].Index.Bytes()
-		// ensure we've got any chopped of leading zero bytes
-		for len(b) < 32 {
-			b = append([]byte{0}, b...)
+		subtreeDepth := s.hasher.BitLen() - prefixSize
+		prefix := lh[i].Index.Bytes()
+		// ensure we've got any chopped off leading zero bytes
```
Martin2112: I don't think that comment is fully clear. I know it was there before but how about "left pad prefix with any zero bytes that were previously removed".

storage/suffix_test.go
```diff
@@ -29,6 +29,33 @@ const (
...
 )
 
+func TestParseSuffix(t *testing.T) {
+	for _, tc := range []struct {
+		prefix    []byte
+		leafIndex int64
+		want      []byte
+	}{
+		{h2b(""), 1, h2b("0801")},
+		{h2b("00"), 1, h2b("0801")},
```
Martin2112: Not testing any invalid cases?

storage/suffix.go
```diff
@@ -37,3 +37,16 @@ func (s Suffix) String() string {
 	r = append(r, s.Path...)
 	return base64.StdEncoding.EncodeToString(r)
 }
+
+// ParseSuffix converts a suffix string back into a Suffix.
+func ParseSuffix(s string) (Suffix, error) {
+	b, err := base64.StdEncoding.DecodeString(s)
+	if err != nil {
+		return Suffix{}, err
```
gdbelvin: `nil` doesn't work here because the return argument is `Suffix` rather than `*Suffix`.
Fear not, I have more pull requests if you so desire :-)

integration/maptest/map_test.go
```diff
@@ -140,6 +140,22 @@ func TestInclusion(t *testing.T) {
...
+				{Index: h2b("0000000000000180000000000000000000000000000000000000000000000000"), LeafValue: []byte("Z")},
+			},
+		},
+		{
+			desc:         "CONIKS multi",
+			HashStrategy: trillian.HashStrategy_CONIKS_SHA512_256,
+			leaves: []*trillian.MapLeaf{
+				{Index: h2b("0000000000000000000000000000000000000000000000000000000000000000"), LeafValue: []byte("A")},
+				{Index: h2b("0000000000000000000000000000000000000000000000000000000000000001"), LeafValue: []byte("B")},
+				{Index: h2b("0000000000000000000000000000000000000000000000000000000000000002"), LeafValue: []byte("C")},
```
gdbelvin: The empty leaf case is both in the just merged `TestLeafHistory` and in `TestNonexistantLeaf`.  I can add here as well.

gdbelvin: added

merkle/hstar2_test.go
gdbelvin: done

storage/suffix_test.go
gdbelvin: added

gdbelvin:
PTAL

storage/suffix.go
Martin2112: Yep, I can't wait!

PR #769, opened 2017-07-31
Remove depth parameter from `MapHasher.HashLeaf`

gdbelvin:
In Maps, leaves are always stored at `hasher.BitLen()`  
This PR removes the depth parameter from `HashLeaf` for clarity.


server/map_rpc_server.go
```diff
@@ -166,8 +166,7 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 			return nil, status.Errorf(codes.InvalidArgument,
 				"len(%x): %v, want %v", l.Index, got, want)
 		}
-		// TODO(gbelvin) use LeafHash rather than computing here. #423
```
Martin2112: That todo could still be relevant?

gdbelvin: readded

PR #772, opened 2017-08-01
Compute Neighbor during Map Inclusion Verification

gdbelvin:
Have the Map verification algorithm compute the proper neighbor nodes. This is a requirement of verification correctness. 

This updates and simplifies the `Sibling()` logic in `storage.NodeID`. 
The only functional change is masking off the unused bits.


merkle/common.go
```diff
@@ -32,3 +32,37 @@ func bit(index []byte, i int) uint {
...
+func flipBit(index []byte, i int) []byte {
+	r := make([]byte, len(index))
+	copy(r, index)
+	IndexBits := len(index) * 8
+	bIndex := (IndexBits - i - 1) / 8
+	r[bIndex] ^= 1 << uint(i%8)
+	return r
+}
+
+// Neighbor returns index with only the left i bits set and the i'th bit flipped.
```
Martin2112: Can you give a bit more details on this? In particular we have a Sibling() function for logs, which could lead to confusion.

integration/maptest/map_test.go
```diff
@@ -88,7 +88,7 @@ func verifyGetMapLeavesResponse(getResp *trillian.GetMapLeavesResponse, indexes
 		leafHash := incl.GetLeaf().GetLeafHash()
 		proof := incl.GetInclusion()
 
-		if got, want := leafHash, hasher.HashLeaf(treeID, index, hasher.BitLen(), leaf); !bytes.Equal(got, want) {
+		if got, want := leafHash, hasher.HashLeaf(treeID, index, 0, leaf); !bytes.Equal(got, want) {
```
Martin2112: Do we need tests at non zero depth?

gdbelvin: The map doesn't support leaves at non-zero depths, so I think not.

Martin2112: Do we need that parameter then?

merkle/common.go
gdbelvin: I'll move this functionality to `Sibling`

integration/maptest/map_test.go
gdbelvin: Right now, strictly speaking, we don't. 
If we were to support leaves at non-zero levels in the future, we would.

Martin2112: Supporting that would be a fairly big change. I could argue for removing this parameter until it's used.

storage/types.go
```diff
@@ -282,20 +282,59 @@ func (n *NodeID) CoordString() string {
...
+		r[depthBytes-1] = r[depthBytes-1] & leftmask[depth%8]
+	}
+	if depth < n.PrefixLenBits {
+		n.PrefixLenBits = depth
+	}
+	n.Path = r
+	return n
+}
+
+// Neighbor sets this node to be it's own neighbor.
```
Martin2112: Still not sure if this means the same as Sibling(), or is it equivalent to Sibling() for maps. Can you expand the documentation?

integration/maptest/map_test.go
gdbelvin: Sounds good

gdbelvin: Here's the PR to remove it: #769

storage/types.go
gdbelvin: This is doing exactly the same thing as the previous version of `Sibling`. The steps have just been broken up for clarity. 

`Neighbor` is doing the following snippet from the previous: 

```
bi = n.PathLenBits() - n.PrefixLenBits + i ==
height = n.PathLenBits() - (n.PrefixLenBits - i)
r[i].SetBit(bi, n.Bit(bi)^1)
```

Mask Left is doing this snippet
```
r[i].PrefixLenBits =  n.PrefixLenBits - i
```

Martin2112: OK. it looks like this is only used by maps / SMT. In the C++ version we had Sibling() which was used by logs but that wasn't carried over to the new code. I wanted to make sure that there wasn't going to be API confusion.

gdbelvin: Sounds good.  In either case, the original functionality has been maintained.

Martin2112:
Is this all merged now via other PRs?

gdbelvin:
This is not merged yet.  Several other PRs have been rebased on it though.

PR #775, opened 2017-08-03
Save MapRoots when no leaves are changed

gdbelvin:
*Observed behavior*
When `SetMapLeaves` is called with no leaves in order to advance the revision number without changing the map, the map root is reset to the empty root hash. 

*Desired behavior* 
When `SetMapLeaves` is called with no leaves in order to advance the revision number without changing the map, the map root remains the same as it was in the previous revision. 

This adds a unit test to detect this behavior, and a small change to enforce it.

### DIFF
--- a/storage/cache/map_subtree_cache.go
+++ b/storage/cache/map_subtree_cache.go
@@ -61,7 +61,7 @@ func populateMapSubtreeNodes(treeID int64, hasher hashers.MapHasher) storage.Pop
 				return nil, nil
 			},
 			func(depth int, index *big.Int, h []byte) error {
-				if depth == 0 {
+				if depth == 0 && len(st.Prefix) > 0 {
 					// no space for the root in the node cache
 					return nil
 				}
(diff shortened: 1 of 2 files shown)


storage/cache/map_subtree_cache.go
```diff
@@ -61,7 +61,7 @@ func populateMapSubtreeNodes(treeID int64, hasher hashers.MapHasher) storage.Pop
 				return nil, nil
 			},
 			func(depth int, index *big.Int, h []byte) error {
-				if depth == 0 {
+				if depth == 0 && len(st.Prefix) > 0 {
```
Martin2112: This looks like an analogue to the log case. You should test this change with a map large enough to span several subtrees as the unit tests won't be enough. Possibly the map integration test is enough but I don't know how robust it is.

gdbelvin: The batch test runs with 32 * 64 leaf nodes, which I think should be enough to exercise this case.  

The tests that I've added are a nice verification that this is correct: they fail without this change and pass post change.

Martin2112: Yep but the log case broke when the tree spanned multiple subtrees so the unit tests were a false sense of security. I think the batch test is probably enough.

PR #776, opened 2017-08-04
Refactor: NodeID.Equivalent -> storge.Equal

gdbelvin:
This is a simple style change from 

`NodeID.Equivilant(other)`
to
`storage.Equal(a, b *NodeID)`

This better matches `bytes.Equal(a, b []byte)`


Martin2112:
Not sure about this. []byte is a built in type so equality has to be done in that way. I don't see any compelling reason to make the change for NodeID.

PR #777, opened 2017-08-04
Concurrent update test for etcd quota server

codingllama:
~~Branched off https://github.com/google/trillian/pull/774, commit 143a9cb is the only new one. (Better to review from that commit onwards only.)~~

Rebased and good to go.


codingllama:
Thanks for the quick review!

Issue #778, opened 2017-08-05
nil pointer dereference bug


AMarcedone:
https://github.com/google/trillian/blob/master/client/verifier.go#L49:
```
func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
	consistency [][]byte) error {

	// Verify SignedLogRoot signature.
	hash := tcrypto.HashLogRoot(*newRoot)
	if err := tcrypto.Verify(c.pubKey, hash, newRoot.Signature); err != nil {
		return err
	}
```
If newRoot == nil, this makes the application panic.

PR #779, opened 2017-08-05
Fix panic on logVerifier.VerifyRoot when newRoot is nil

AMarcedone:
Solves #778

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,5 +8,6 @@
 #
 # Please keep the list sorted.
 
+Antonio Marcedone <a.marcedone@gmail.com>
 Google Inc.
 Internet Security Research Group
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -17,10 +17,10 @@ package client
 import (
 	"bytes"
 	"context"
-	tcrypto "github.com/google/trillian/crypto"
 	"testing"
 
 	"github.com/google/trillian"
+	tcrypto "github.com/google/trillian/crypto"
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/merkle/rfc6962"
 	"github.com/google/trillian/testonly"
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -17,10 +17,13 @@ package client
 import (
 	"bytes"
 	"context"
+	tcrypto "github.com/google/trillian/crypto"
 	"testing"
 
 	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/merkle/rfc6962"
+	"github.com/google/trillian/testonly"
 	"github.com/google/trillian/testonly/integration"
 )
 
@@ -252,5 +255,44 @@ func TestUpdateRoot(t *testing.T) {
 	}
 	if got, want := client.Root().TreeSize, before; got <= want {
 		t.Errorf("Tree size after add Leaf: %v, want > %v", got, want)
+	}
+}
+
+func TestVerifyRootReturnsErrWhenNewRootIsNil(t *testing.T) {
+	logVerifier := NewLogVerifier(nil, nil)
+
+	// This also makes sure that no nil pointer dereference errors occur (as this would cause a panic).
+	if err := logVerifier.VerifyRoot(nil, nil, nil); err == nil {
+		t.Error("Expected an error when verifying a nil tree root, but got nil")
+	}
+}
+
+func TestVerifyRootPassesWhenTrustedIsNil(t *testing.T) {
+	// First, create a fake root with a valid signature.
+	newRoot := trillian.SignedLogRoot{0, nil, 0, nil, 0, 0}
+
+	key, err := keys.NewFromPrivatePEM(testonly.DemoPrivateKey, testonly.DemoPrivateKeyPass)
+	if err != nil {
+		t.Fatalf("Failed to open test key, err=%v", err)
+	}
+	signer := tcrypto.NewSHA256Signer(key)
+	pk, err := keys.NewFromPublicPEM(testonly.DemoPublicKey)
+	if err != nil {
+		t.Fatalf("Failed to load public key, err=%v", err)
+	}
+
+	hash := tcrypto.HashLogRoot(newRoot)
+	signature, err := signer.Sign(hash)
+	if err != nil {
+		t.Fatal("Failed to create test signature")
+	}
+	newRoot.Signature = signature
+
+	logVerifier := NewLogVerifier(rfc6962.DefaultHasher, pk)
+	// This also makes sure that no nil pointer dereference errors occur (as this would cause a panic).
+	if err := logVerifier.VerifyRoot(nil, &newRoot, nil); err != nil {
+		// The verification should pass as trusted is nil and no proof is provided, so we trust the first
+		// root we get.
+		t.Errorf("VerifyRoot threw an unexpected error: %v", err)
 	}
 }
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -20,10 +20,7 @@ import (
 	"testing"
 
 	"github.com/google/trillian"
-	tcrypto "github.com/google/trillian/crypto"
-	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/merkle/rfc6962"
-	"github.com/google/trillian/testonly"
 	"github.com/google/trillian/testonly/integration"
 )
 
@@ -255,44 +252,5 @@ func TestUpdateRoot(t *testing.T) {
 	}
 	if got, want := client.Root().TreeSize, before; got <= want {
 		t.Errorf("Tree size after add Leaf: %v, want > %v", got, want)
-	}
-}
-
-func TestVerifyRootReturnsErrWhenNewRootIsNil(t *testing.T) {
-	logVerifier := NewLogVerifier(nil, nil)
-
-	// This also makes sure that no nil pointer dereference errors occur (as this would cause a panic).
-	if err := logVerifier.VerifyRoot(nil, nil, nil); err == nil {
-		t.Error("Expected an error when verifying a nil tree root, but got nil")
-	}
-}
-
-func TestVerifyRootPassesWhenTrustedIsNil(t *testing.T) {
-	// First, create a fake root with a valid signature.
-	newRoot := trillian.SignedLogRoot{0, nil, 0, nil, 0, 0}
-
-	key, err := keys.NewFromPrivatePEM(testonly.DemoPrivateKey, testonly.DemoPrivateKeyPass)
-	if err != nil {
-		t.Fatalf("Failed to open test key, err=%v", err)
-	}
-	signer := tcrypto.NewSHA256Signer(key)
-	pk, err := keys.NewFromPublicPEM(testonly.DemoPublicKey)
-	if err != nil {
-		t.Fatalf("Failed to load public key, err=%v", err)
-	}
-
-	hash := tcrypto.HashLogRoot(newRoot)
-	signature, err := signer.Sign(hash)
-	if err != nil {
-		t.Fatal("Failed to create test signature")
-	}
-	newRoot.Signature = signature
-
-	logVerifier := NewLogVerifier(rfc6962.DefaultHasher, pk)
-	// This also makes sure that no nil pointer dereference errors occur (as this would cause a panic).
-	if err := logVerifier.VerifyRoot(nil, &newRoot, nil); err != nil {
-		// The verification should pass as trusted is nil and no proof is provided, so we trust the first
-		// root we get.
-		t.Errorf("VerifyRoot threw an unexpected error: %v", err)
 	}
 }
--- a/client/verifier.go
+++ b/client/verifier.go
@@ -22,6 +22,7 @@ import (
 	tcrypto "github.com/google/trillian/crypto"
 	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/merkle/hashers"
+	"fmt"
 )
 
 // logVerifier contains state needed to verify output from Trillian Logs.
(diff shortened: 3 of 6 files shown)


client/verifier.go
```diff
@@ -46,17 +47,21 @@ func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
 	consistency [][]byte) error {
 
 	// Verify SignedLogRoot signature.
-	hash := tcrypto.HashLogRoot(*newRoot)
-	if err := tcrypto.Verify(c.pubKey, hash, newRoot.Signature); err != nil {
-		return err
+	if (newRoot == nil){
+		return fmt.Errorf("VerifyRoot() error: newRoot == nil.")
```
gdbelvin: Error strings should not be capitalized.

client/verifier.go
```diff
@@ -46,17 +47,21 @@ func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
...
+	}else {
```
gdbelvin: This block can be outdented

AUTHORS
```diff
@@ -8,5 +8,6 @@
 #
 # Please keep the list sorted.
 
+Antonio Marcedone <a.marcedone@gmail.com>
```
codingllama: We should add you to contributors, but not authors. It has a different meaning.

client/verifier.go
```diff
@@ -22,6 +22,7 @@ import (
 	tcrypto "github.com/google/trillian/crypto"
 	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/merkle/hashers"
+	"fmt"
```
codingllama: nit: Please move "fmt" to the block of Golang imports (below crypto/sha256)

client/verifier.go
```diff
@@ -46,17 +47,21 @@ func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
 	consistency [][]byte) error {
 
 	// Verify SignedLogRoot signature.
-	hash := tcrypto.HashLogRoot(*newRoot)
-	if err := tcrypto.Verify(c.pubKey, hash, newRoot.Signature); err != nil {
-		return err
+	if (newRoot == nil){
```
codingllama: nit: Parenthesis aren't usually in Go ifs (unless really needed)

client/verifier.go
codingllama: I'd remove the "." at the end too.

client/verifier.go
```diff
@@ -46,17 +47,21 @@ func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
...
+	}else {
+		hash := tcrypto.HashLogRoot(*newRoot)
+		if err := tcrypto.Verify(c.pubKey, hash, newRoot.GetSignature()); err != nil {
+			return err
+		}
 	}
 
 	// Implicitly trust the first root we get.
-	if trusted.TreeSize != 0 {
+	if trusted.GetTreeSize() != 0 {
```
codingllama: Is there a reason for the change on the getters here and below?

crypto/data_formats.go
```diff
@@ -42,9 +42,9 @@ func HashLogRoot(root trillian.SignedLogRoot) []byte {
 	// JSON encoded otherwise (it uses floats). We want to be sure that people
 	// using JSON to verify hashes can build the exact same input to ObjectHash.
 	rootMap := map[string]string{
-		mapKeyRootHash:       base64.StdEncoding.EncodeToString(root.RootHash),
-		mapKeyTimestampNanos: strconv.FormatInt(root.TimestampNanos, 10),
-		mapKeyTreeSize:       strconv.FormatInt(root.TreeSize, 10)}
+		mapKeyRootHash:       base64.StdEncoding.EncodeToString(root.GetRootHash()),
```
codingllama: Is there a reason for the changes here?

codingllama:
Could you edit the PR's title as well, it's too generic and will be ambiguous once in the Git log. I suggest something like "Fix panic on logVerifier.VerifyRoot when newRoot is nil"

AUTHORS
AMarcedone: From the contributors file:
```
# When adding J Random Contributor's name to this file,
# either J's name or J's organization's name should be
# added to the AUTHORS file, depending on whether the
# individual or corporate CLA was used.
```
I signed the individual CLA.

client/verifier.go
AMarcedone: Not sure what you mean. Do you want me to take the code out of the else branch, as the else is redundant since when the if branch is taken the function returns?
I wrote it like this as I think it is easier to see that in the else branch newRoot cannot be null, but happy to switch back

client/verifier.go
AMarcedone: Yes. If trusted == nil, trusted.TreeSize is a nil pointer dereference panic which causes SEGFAULT when tested. The getter prevents this (if you look at the code, it just adds a nil test).

crypto/data_formats.go
AMarcedone: As above, getters prevent nil pointer dereference errors.

client/verifier.go
AMarcedone: @gdbelvin I made the switch. Please let me know if this is what you meant.

AMarcedone:
@gdbelvin and @codingllama thanks a lot for your reviews! 
I made the requested changes. PTAL

AUTHORS
codingllama: OK. What got me is that, apparently, we're missing a bunch of people at AUTHORS, so I understood their absence as a sign that we shouldn't do that.

client/client_test.go
```diff
@@ -254,3 +257,42 @@ func TestUpdateRoot(t *testing.T) {
 		t.Errorf("Tree size after add Leaf: %v, want > %v", got, want)
 	}
 }
+
```
codingllama: Could we move the tests to `verifier_test.go`?

client/client_test.go
```diff
@@ -254,3 +257,42 @@ func TestUpdateRoot(t *testing.T) {
...
+func TestVerifyRootReturnsErrWhenNewRootIsNil(t *testing.T) {
```
codingllama: go style nit: Could we structure both tests as a single, table-driven test?

```go
func TestVerifyRootErrors(t *testing.T) {
  tests := []struct {
    desc string
    trusted, newRoot *trillian.SignedLogRoot
  }{
    {desc: "newRootNil", trusted: someRoot, newRoot: nil},
    {desc: "trustedNil",trusted: nil, newRoot: someRoot},
    // so on...
  }
  for _, test := range tests {
    if err := verifier.VerifyRoot(test.trusted, test.newRoot); err == nil {
      t.Errorf("%v: VerifyRoot() returned err = nil", test.desc)
    }
  }
}
```

https://github.com/golang/go/wiki/TableDrivenTests

client/verifier.go
```diff
@@ -18,6 +18,7 @@ import (
 	"crypto"
 	"crypto/sha256"
 
+	"fmt"
```
codingllama: nit: Move to the block above

client/verifier.go
codingllama: Calling VerifyRoot with trusted = nil sounds like a programming error to me. I think we should error out earlier, just as we do if newRoot == nil.

@gdbelvin, what say you?

client/verifier.go
```diff
@@ -68,14 +72,14 @@ func (c *logVerifier) VerifyRoot(trusted, newRoot *trillian.SignedLogRoot,
 // the currently trusted root. The inclusion proof must be requested for Root().TreeSize.
 func (c *logVerifier) VerifyInclusionAtIndex(trusted *trillian.SignedLogRoot, data []byte, leafIndex int64, proof [][]byte) error {
 	leaf := c.buildLeaf(data)
-	return c.v.VerifyInclusionProof(leafIndex, trusted.TreeSize,
-		proof, trusted.RootHash, leaf.MerkleLeafHash)
+	return c.v.VerifyInclusionProof(leafIndex, trusted.GetTreeSize(),
```
codingllama: Ditto here and at VerifyInclusionByHash. I'd rather error out or panic, silently proceeding makes the `trusted = nil` misstep harder to detect.

Please add a test for VerifyInclusion* methods too, if you decide to touch them.

crypto/data_formats.go
codingllama: root is not a pointer, you can't really pass nil to this function.  (Maybe it *should* be a pointer, but that's another discussion.)

client/verifier.go
codingllama: Not Gary, but it looks good. :)

Just clarifying, it's somewhat of a general Go-style rule that, if you return inside the if, you omit the else clause.

Eg, we prefer:

```go
if something {
  return fmt.Errorf("...")
}
// code continues here
```

instead of

```go
if something {
  return fmt.Errorf("...")
} else {
  // code continues here
}
```

client/verifier.go
codingllama: Actually, sorry for the back and forth here, but since the capitalized string is actually the method name, what you had originally is correct.

We don't capitalize sentences in errors because error chaining is a common practice, so it would look weird, but in this case it works fine.

I'd go with a small change, which is dropping the " error" part: `fmt.Errorf("VerifyRoot(): newRoot == nil")`

client/verifier.go
gdbelvin: I would agree.  Clients should explicitly create an initial trusted root to work with.
This can easily be a static object with TreeSize == 0 and RootHash = empty root

client/verifier.go
AMarcedone: A panic (SEGFAULT) is bad for gobind integration (the android app hangs) and might cause errors too, so I would go for an error instead.

gdbelvin:
That sounds like the right path.  Thanks for taking this on Antonio

AMarcedone:
Updated after the most recent comments. And thanks for your patience, I am just starting to write in go for this project :)
PTAL

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/trillian"
+	tcrypto "github.com/google/trillian/crypto"
+	"github.com/google/trillian/crypto/keys"
+	"github.com/google/trillian/merkle/rfc6962"
+	"github.com/google/trillian/testonly"
+	"strings"
```
codingllama: nit: Move "strings" above "testing", on the Golang libraries block.

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+			t.Errorf("%v: VerifyRoot() error expected, but got nil", test.desc)
+		}
+	}
+}
+
+func TestVerifyInclusionAtIndexErrors(t *testing.T) {
+	logVerifier := NewLogVerifier(nil, nil)
+	err := logVerifier.VerifyInclusionAtIndex(nil, nil, 1, nil)
+	if err == nil {
+		t.Errorf("VerifyInclusionAtIndex() error expected, but none was thrown")
```
codingllama: nit: Technically, Go errors aren't "thrown", that's Java/C++/etc lingo.  ;)

Ditto below.

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if !strings.Contains(err.Error(), "trusted == nil"){
```
codingllama: I'd rather drop the error string check and call VerifyInclusionAtIndex with only trusted = nil, so it's clear what the source of the error is. Error message checking lands too close to [change detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Errorf("VerifyInclusionAtIndex() throws wrong error when trusted == nil: %v", err)
+	}
+}
+
+func TestVerifyInclusionByHashErrors(t *testing.T) {
+	tests := []struct {
+		desc             string
+		trusted *trillian.SignedLogRoot
```
codingllama: You're missing a gofmt run here. Try doing `scripts/presubmit.sh --fix --no-generate`, that should do all the formatting needed.

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+		{desc: "proofNil", trusted: &trillian.SignedLogRoot{}, proof: nil, errorDesc: "proof == nil"},
+	}
+	for _, test := range tests {
+
+		logVerifier := NewLogVerifier(nil, nil)
+		err := logVerifier.VerifyInclusionByHash(test.trusted, nil, test.proof)
+		if err == nil {
+			t.Errorf("%v: VerifyInclusionByHash() error expected, but none was thrown", test.desc)
+		}
+		if !strings.Contains(err.Error(), test.errorDesc){
```
codingllama: nit: As above, I think it's fine to drop this.

client/verifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		t.Fatalf("Failed to open test key, err=%v", err)
+	}
+	signer := tcrypto.NewSHA256Signer(key)
+	pk, err := keys.NewFromPublicPEM(testonly.DemoPublicKey)
+	if err != nil {
+		t.Fatalf("Failed to load public key, err=%v", err)
+	}
+
+	signedRoot := trillian.SignedLogRoot{0, nil, 0, nil, 0, 0}
```
gdbelvin: this could be rewritten as `trillian.SignedLogRoot{}` since all the fields have their default values

client/verifier_test.go
AMarcedone: I see your point. 
I removed the check and added a comment.

AMarcedone:
Updated. PTAL and thanks again :)

codingllama:
Oh, nevermind, I see that it was deferred to me. Merging.

PR #780, opened 2017-08-07
Map Verification: Fix for proofs of leaves in empty branches.

gdbelvin:
The CONIKS hashing algorithm commits to empty branches uniquely for every position in the tree #691. This breaks the previous assumption that `E2 = H(E1, E1)`. The verification function needs to be aware of when it is verifying a node within an empty branch and use the commitment to the whole empty branch.  Thus, `VerifyMapInclusionProof` is modified to accept `LeafValue` rather than `LeafHash` so that it can perform this empty branch detection. 

This PR also adds a check to prevent clients from storing empty leaf nodes in the leaf table.

The testing of empty branches has been incorporated into `TestLeafHistory` in order to test empty branch inclusion proofs at a variety of points - and remove duplicate code.

### DIFF
--- a/integration/maptest/map_test.go
+++ b/integration/maptest/map_test.go
@@ -131,7 +131,7 @@ func TestLeafHistory(t *testing.T) {
 	}{
 		{
 			desc:         "single leaf update",
-			HashStrategy: []trillian.HashStrategy{trillian.HashStrategy_CONIKS_SHA512_256},
+			HashStrategy: []trillian.HashStrategy{trillian.HashStrategy_TEST_MAP_HASHER, trillian.HashStrategy_CONIKS_SHA512_256},
 			set: [][]*trillian.MapLeaf{
 				{}, // Advance revision without changing anything.
 				{
--- a/integration/maptest/map_test.go
+++ b/integration/maptest/map_test.go
@@ -92,7 +92,7 @@ func verifyGetMapLeavesResponse(getResp *trillian.GetMapLeavesResponse, indexes 
 			return fmt.Errorf("HashLeaf(%s): %x, want %x", leaf, got, want)
 		}
 		if err := merkle.VerifyMapInclusionProof(treeID, index,
-			leafHash, rootHash, proof, hasher); err != nil {
+			leaf, rootHash, proof, hasher); err != nil {
 			return fmt.Errorf("VerifyMapInclusionProof(%x): %v", index, err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


Martin2112:
Can you rebase? Looks like just the tests conflict.

gdbelvin:
Rebased

server/map_rpc_server.go
```diff
@@ -166,6 +166,12 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 			return nil, status.Errorf(codes.InvalidArgument,
 				"len(%x): %v, want %v", l.Index, got, want)
 		}
+		if l.LeafValue == nil {
```
Martin2112: I was slightly surprised that this isn't an error but I'm not certain of the map semantics here.

integration/maptest/map_test.go
```diff
@@ -212,83 +215,76 @@ func TestInclusion(t *testing.T) {
...
+			HashStrategy: []trillian.HashStrategy{trillian.HashStrategy_TEST_MAP_HASHER, trillian.HashStrategy_CONIKS_SHA512_256},
 			leaves: []*trillian.MapLeaf{
 				{Index: h2b("0000000000000000000000000000000000000000000000000000000000000000"), LeafValue: []byte("A")},
 			},
 		},
 		{
-			desc:         "maphasher multi",
-			HashStrategy: trillian.HashStrategy_TEST_MAP_HASHER,
+			desc:         "multi",
+			HashStrategy: []trillian.HashStrategy{trillian.HashStrategy_TEST_MAP_HASHER, trillian.HashStrategy_CONIKS_SHA512_256},
```
Martin2112: Do you think these tests are enough to show verification works correctly at multiple levels and leaves? It's hard for me to be sure.

merkle/map_verifier.go
```diff
@@ -42,23 +42,44 @@ func VerifyMapInclusionProof(treeID int64, index, leafHash, expectedRoot []byte,
...
+		runningHash = h.HashLeaf(treeID, index, leaf)
+	}
+
 	nID := storage.NewNodeIDFromHash(index)
+	for height, sib := range nID.Siblings() {
+		pElement := proof[height]
 
-	runningHash := make([]byte, len(leafHash))
-	copy(runningHash, leafHash)
+		// Since HashChildren(e0, e0) is not always equal to the empty
```
gdbelvin: How's this: 
// Since empty values are tied to a location and a 
// HashEmpty(leve1) != HashChildren(E0, 
// Therefore we need to maintain an empty marker along the
// proof path until the first non-empty element so we can call
// HashEmpty once at the top of the empty branch.

server/map_rpc_server.go
gdbelvin: Storing empty leaves in maps is valid, we just don't want to put anything on disk for them.

integration/maptest/map_test.go
gdbelvin: added some comments for clarity.

merkle/map_verifier.go
Martin2112: Basically OK. Looks like there might be a couple of typos "leve1" and the HashChildren line
seems to be truncated.

gdbelvin:
Thank you very much for all your reviews, Martin!

PR #784, opened 2017-08-08
GetActiveLogIDs filters by type and state

codingllama:
Added a state filter and simplified logic overall, as most of the intermediate
functions are leftovers from the time where we had GetActiveLogIDs in more than
one interface.


codingllama:
Friendly ping?

codingllama:
Thanks!

PR #788, opened 2017-08-10
Refactor createtree command

RJPercival:
This simplifies PR #734, which requires putting the PKCS#11 flag used by createtree behind a build tag.

### DIFF
--- a/cmd/createtree/main_test.go
+++ b/cmd/createtree/main_test.go
@@ -15,10 +15,9 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"flag"
-	"fmt"
-	"net"
 	"testing"
 	"time"
 
@@ -27,12 +26,10 @@ import (
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/trillian"
-	"github.com/google/trillian/crypto/keyspb"
+	"github.com/google/trillian/cmd/createtree/testonly"
 	"github.com/google/trillian/crypto/sigpb"
 	"github.com/google/trillian/util/flagsaver"
 	"github.com/kylelemons/godebug/pretty"
-	"golang.org/x/net/context"
-	"google.golang.org/grpc"
 )
 
 // defaultTree reflects all flag defaults with the addition of a valid private key.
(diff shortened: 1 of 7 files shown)


cmd/createtree/main_test.go
```diff
@@ -27,173 +27,132 @@ import (
...
 )
 
-func marshalAny(p proto.Message) *any.Any {
+// defaultTree reflects all flag defaults with the addition of a valid private key.
+var defaultTree = &trillian.Tree{
+	TreeState:          trillian.TreeState_ACTIVE,
+	TreeType:           trillian.TreeType_LOG,
+	HashStrategy:       trillian.HashStrategy_RFC6962_SHA256,
+	HashAlgorithm:      sigpb.DigitallySigned_SHA256,
+	SignatureAlgorithm: sigpb.DigitallySigned_RSA,
```
codingllama: Suggestion: default to ECDSA? Isn't that what the cool kids use?  ;)

Ps: don't forget about `nonDefaultTree` below.

cmd/createtree/pem_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian/crypto/keys"
+	"github.com/google/trillian/crypto/keyspb"
+)
+
+func TestWithPEMKeyFile(t *testing.T) {
+	pemPath, pemPassword := "../../testdata/log-rpc-server.privkey.pem", "towel"
+
+	wantTree := *defaultTree
+	keyProto, err := ptypes.MarshalAny(&keyspb.PEMKeyFile{
```
codingllama: Suggestion: `wantTree.PrivateKey = mustMarshalAny(...)`

cmd/createtree/pem_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	wantTree := *defaultTree
+	keyProto, err := ptypes.MarshalAny(&keyspb.PrivateKey{
+		Der: keyDER,
+	})
+	if err != nil {
+		t.Fatalf("MarshalAny(PrivateKey): %v", err)
+	}
+	wantTree.PrivateKey = keyProto
```
codingllama: Ditto (mustMarshalAny).

So on for other tests.

util/flagsaver/flagsaver_test.go
```diff
@@ -0,0 +1,107 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package flagsaver
+
+import (
+	"testing"
+	"time"
+
```
codingllama: nit: Remove blank line and gofmt.

cmd/createtree/main_test.go
RJPercival: I agree that ECDSA should be the default, but is it reasonable to change the default given that this flag has existed for a while? It could surprise some people if they're currently assuming that they'll get RSA trees if they don't provide this flag.

codingllama: If we would like to do it, better now than later. I expect that, if you care about the key type, you'll state it explicitly.

It's only a suggestion, though. Don't worry too much about it.

PR #794, opened 2017-08-12
Refactor keys package

RJPercival:
In the future, this will make it easier to exclude support for particular key types from the build through the use of build tags. For example, PKCS#11 key support currently requires a C module that causes problems for some users - this could be excluded from the build by default, so that only those people that want it need to install the dependency.


gdbelvin:
Please rebase and then ping me

RJPercival:
Rebased - PTAL.

RJPercival:
Travis will never be happy, as this breaks certificate-transparency-go. PR google/certificate-transparency-go#58 fixes it.

PR #795, opened 2017-08-14
Return appropriate error codes for bad quota requests

codingllama:
Expose IsNameValid and use trillian/errors for storage errors that are likely to
be forwarded as user-visible errors (such as invalid quota Configs).

Regexes used by IsNameValid have been improved, as has the checking of tree IDs.

### DIFF
--- a/quota/etcd/quotaapi/quota_server.go
+++ b/quota/etcd/quotaapi/quota_server.go
@@ -39,11 +39,11 @@ func NewServer(client *clientv3.Client) *Server {
 
 // CreateConfig implements quotapb.QuotaServer.CreateConfig.
 func (s *Server) CreateConfig(ctx context.Context, req *quotapb.CreateConfigRequest) (*quotapb.Config, error) {
-	switch {
-	case req.Name == "":
-		return nil, status.Errorf(codes.InvalidArgument, "name is required")
-	case req.Config == nil:
+	if req.Config == nil {
 		return nil, status.Errorf(codes.InvalidArgument, "config is required")
+	}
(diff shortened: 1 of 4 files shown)


codingllama:
Issue #490

codingllama:
Note: I didn't switch *all* errors in https://github.com/google/trillian/blob/master/quota/etcd/storage/quota_storage.go to trillian/error, I targeted mostly the ones that flow from UpdateConfigs, as that's the method underlying most API calls (with Configs being the other).

quota/etcd/quotaapi/quota_server.go
```diff
@@ -200,3 +201,13 @@ func findByName(name string, cfgs *storagepb.Configs) (*storagepb.Config, bool)
 	}
 	return nil, false
 }
+
+func validateName(name string) error {
+	switch {
+	case name == "":
+		return status.Errorf(codes.InvalidArgument, "name is required")
+	case !storage.IsNameValid(name):
```
Martin2112: Seemed odd to me that naming is a storage responsibility. Are you happy with this boundary?

codingllama: I've made config names mandated by storage, so it's easier to reason that a bad name can't get in. For convenience, those match the names exposed by the API, akin to having a primary key on Config.name. Since the names are storage-mandated, it seemed OK to expose the validation logic from it.

(I hope that answers your question.)

Martin2112: As it's a restricted domain it's probably OK but we could end up with code duplication.

PR #796, opened 2017-08-14
Common function to produce keypb.PublicKey

gdbelvin:
This PR creates a common function to convert `crypto.PublicKey` into
`keyspb.PublicKey`. This code is currently spread throughout a few
locations.

This also makes it easier for all the `keys.NewFrom*` functions, which
all produce `crypto.PublicKey`, and then convert it back into a
protobuf.

### DIFF
--- a/crypto/keys/der/der.go
+++ b/crypto/keys/der/der.go
@@ -29,6 +29,11 @@ func FromProto(pb *keyspb.PrivateKey) (crypto.Signer, error) {
 	return UnmarshalPrivateKey(pb.GetDer())
 }
 
+// FromPublicProto takes a PublicKey protobuf message and returns the public key contained within.
+func FromPublicProto(pb *keyspb.PublicKey) (crypto.PublicKey, error) {
+	return UnmarshalPublicKey(pb.GetDer())
+}
+
 // UnmarshalPrivateKey reads a DER-encoded private key.
 func UnmarshalPrivateKey(keyDER []byte) (crypto.Signer, error) {
 	key1, err1 := x509.ParseECPrivateKey(keyDER)


server/admin/admin_server.go
```diff
@@ -139,19 +137,19 @@ func (s *Server) CreateTree(ctx context.Context, request *trillian.CreateTreeReq
 
 	// Derive the public key that corresponds to the private key for this tree.
 	// The caller may have provided the public key, but for safety we shouldn't rely on it being correct.
-	publicKeyDER, err := x509.MarshalPKIXPublicKey(signer.Public())
+	publicKey, err := keys.ToPublicKeyPB(signer.Public())
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to marshal public key: %v", err.Error())
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err.Error())
```
codingllama: nit: Use `status.Error(codes.InvalidArgument, err.Error())` instead?

crypto/keys/public_keys.go
```diff
@@ -25,9 +25,19 @@ import (
 	"fmt"
 	"io/ioutil"
 
+	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/crypto/sigpb"
 )
 
+// ToPublicKeyPB returns a keyspb.PublicKey from crypto.PublicKey.
+func ToPublicKeyPB(pubKey crypto.PublicKey) (*keyspb.PublicKey, error) {
```
codingllama: Suggestion: Write a simple test for ToPublicKeyPB, since you plan to depend directly on it?

server/admin/admin_server.go
gdbelvin: I was following the pattern here in the rest of the file, which doesn't use status.Error?

codingllama: We probably missed that on the previous reviews, but we don't need `Errorf` if a printf-like pattern is not being used. Up to you whether to consider this out of scope, in which case I can do the Errorf/Error changes in a separate PR.

gdbelvin:
This was superseeded by the keys refactor

PR #798, opened 2017-08-15
Add To/From PublicProto functions to keys/der

gdbelvin:
Adds the following helper functions to the `keys/der` package
- `FromPublicProto`
- `ToPublicProto`

### DIFF
--- a/crypto/keys/der/der.go
+++ b/crypto/keys/der/der.go
@@ -29,9 +29,19 @@ func FromProto(pb *keyspb.PrivateKey) (crypto.Signer, error) {
 	return UnmarshalPrivateKey(pb.GetDer())
 }
 
-// FromPublicProto takes a PublicKey protobuf message and returns the public key contained within.
+// FromPublicProto takes a PublicKey protobuf message and returns the public
+// key contained within.
 func FromPublicProto(pb *keyspb.PublicKey) (crypto.PublicKey, error) {
 	return UnmarshalPublicKey(pb.GetDer())
+}
+
+// ToPublicProto returns a keyspb.PublicKey
+func ToPublicProto(pubKey crypto.PublicKey) (*keyspb.PublicKey, error) {
+	keyDER, err := MarshalPublicKey(pubKey)
+	if err != nil {
+		return nil, err
+	}
+	return &keyspb.PublicKey{Der: keyDER}, nil
 }
 
 // UnmarshalPrivateKey reads a DER-encoded private key.
@@ -68,6 +78,16 @@ func UnmarshalPublicKey(keyDER []byte) (crypto.PublicKey, error) {
 	}
 
 	return key, nil
+}
+
+// MarshalPublicKey outputs a DER-encoded public key.
+func MarshalPublicKey(pubKey crypto.PublicKey) ([]byte, error) {
+	der, err := x509.MarshalPKIXPublicKey(pubKey)
+	if err != nil {
+		return nil, fmt.Errorf("der: could not marshal public key as PKIX (%v)", err)
+	}
+
+	return der, nil
 }
 
 // MarshalPrivateKey serializes an RSA or ECDSA private key as DER.
--- a/crypto/keys/der/der.go
+++ b/crypto/keys/der/der.go
@@ -35,7 +35,7 @@ func FromPublicProto(pb *keyspb.PublicKey) (crypto.PublicKey, error) {
 	return UnmarshalPublicKey(pb.GetDer())
 }
 
-// ToPublicProto returns a keyspb.PublicKey
+// ToPublicProto returns a keyspb.PublicKey that contains pubKey in DER encoding.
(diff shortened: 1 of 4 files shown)


crypto/keys/der/der_test.go
```diff
@@ -0,0 +1,97 @@
...
+			continue
+		}
+
+		pubKey2, err := FromPublicProto(pubProto)
+		if err != nil {
+			t.Errorf("FromPublicProto(%v): %v", pubKey, err)
+			continue
+		}
+
+		pubDER2, err := MarshalPublicKey(pubKey2)
```
codingllama: Suggestion: skip the last marshal step and compare pubKey to pubKey2 instead?

crypto/keys/der/der.go
```diff
@@ -65,6 +80,16 @@ func UnmarshalPublicKey(keyDER []byte) (crypto.PublicKey, error) {
 	return key, nil
 }
 
+// MarshalPublicKey outputs a DER-encoded public key.
```
RJPercival: s/outputs/returns/

crypto/keys/testonly/keys.go
```diff
@@ -44,6 +45,7 @@ func MustMarshalPublicPEMToDER(keyPEM string) []byte {
 	if err != nil {
 		panic(err)
 	}
+	log.Printf("keyDER: %#v", keyDER)
```
codingllama: Revert this file?

crypto/keys/der/der_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"bytes"
+	"encoding/pem"
+	"testing"
+
+	"github.com/google/trillian/testonly"
+)
+
+func MustMarshalPublicPEMToDER(keyPEM string) []byte {
```
RJPercival: No need to export this func.

RJPercival: In fact, rather than declaring this func, you could use the func declared at https://github.com/google/trillian/blob/master/crypto/keys/testonly/keys.go#L37. You'll just need to change this file's package to "der_test", then dot-import the "der" package, to avoid an import cycle.

gdbelvin: done

crypto/keys/der/der_test.go
gdbelvin: I'm not sure that I can actually compare public keys to one another.  It's an empty interface and the pointers will be pointing to different objects.

crypto/keys/der/der.go
gdbelvin: done

crypto/keys/testonly/keys.go
gdbelvin: done

crypto/keys/der/der_test.go
RJPercival: Did you see my second comment?

crypto/keys/der/der.go
```diff
@@ -29,6 +29,21 @@ func FromProto(pb *keyspb.PrivateKey) (crypto.Signer, error) {
 	return UnmarshalPrivateKey(pb.GetDer())
 }
 
+// FromPublicProto takes a PublicKey protobuf message and returns the public
+// key contained within.
+func FromPublicProto(pb *keyspb.PublicKey) (crypto.PublicKey, error) {
+	return UnmarshalPublicKey(pb.GetDer())
+}
+
+// ToPublicProto returns a keyspb.PublicKey
```
RJPercival: s/returns a keyspb.PublicKey/returns a keyspb.PublicKey that contains pubKey in DER encoding./

crypto/keys/der/der_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return block.Bytes
+}
+
+func TestMarshalUnmarshalPublicKey(t *testing.T) {
+	for _, tc := range []struct {
+		pubPEM string
+	}{
+		{pubPEM: testonly.DemoPublicKey},
+	} {
```
RJPercival: With only one test case, it doesn't seem worthwhile to make this a tabular test.

crypto/keys/der/der.go
gdbelvin: done

crypto/keys/der/der_test.go
RJPercival: I doubt test cases will be added here in the future, but maybe that's a lack of imagination of my part. Seems like a case of YAGNI, and it wouldn't be hard to change into a tabular test if needed in the future. I'm not that bothered though.

RJPercival: I can't say I've observed that or seen that in any readability guidance. Regardless, I'm ok with leaving this as-is.

crypto/keys/der/der_test.go
RJPercival: *ping*

PR #808, opened 2017-08-21
Use golang/protobuf instead of gogo/protobuf

codingllama:


codingllama:
A few mocks / pb.go files seem to be outdated. Fixed by #809.

PR #813, opened 2017-08-24
Stash coverage reports in /tmp

Martin2112:
This should fix report uploads to codecov as it seemed that that this file was getting removed during the rest of the build.


Martin2112:
It's not an ideal solution but it works and with the current speed of Travis finding out what's happening to the output file could be a long job.

PR #817, opened 2017-09-04
Wire etcd quotas on log and map servers and logsigner

codingllama:
Etcd-based quotas may now be used with Log and Map servers.

A few refactors went into the PR, such as:
* Etcd clients are created by main and passed to other packages. This ensures the same client and configurations are used.
* Log integration test (the bash-driven one) uses etcd quotas when etcd is available
* Package quota/mysql renamed to quota/mysqlqm (in order to avoid import aliasing)

A quota-focused integration test for etcd has been added as well.

### DIFF
--- a/integration/functions.sh
+++ b/integration/functions.sh
@@ -8,7 +8,7 @@ declare -a RPC_SERVER_PIDS
 declare -a LOG_SIGNER_PIDS
 declare -a TO_KILL
 declare -a TO_DELETE
-ADMIN_SERVER=''
+HTTP_SERVER_1=''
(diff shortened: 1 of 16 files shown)


codingllama:
Reviewers: feel free to review commit-by-commit if you like. There's some overlap between commits, but they're *mostly* independent.

The last outstanding item on my list is to add a README / howto for etcd quotas. I'll be addressing that in a follow up PR.

Issue #490.

codingllama:
Friendly ping?

codingllama:
Thanks, Martin!

PR #820, opened 2017-09-06
Return current tokens in etcd quota API responses

codingllama:


codingllama:
Issue #490.

quota/etcd/quotaapi/quota_server.go
```diff
@@ -95,17 +96,26 @@ func (s *Server) GetConfig(ctx context.Context, req *quotapb.GetConfigRequest) (
...
+func (s *Server) getConfig(ctx context.Context, name string, cfgs *storagepb.Configs, code codes.Code) (*quotapb.Config, error) {
+	storedCfg, ok := findByName(name, cfgs)
 	if !ok {
 		return nil, status.Errorf(code, "%q not found", name)
 	}
-	return convertToAPI(cfg), nil
+	cfg := convertToAPI(storedCfg)
+
+	tokens, err := s.qs.Peek(ctx, []string{cfg.Name})
+	if err == nil {
```
Martin2112: I think style is to indent the non error case and then return early. I don't like it either.

quota/etcd/quotaapi/quota_server.go
```diff
@@ -129,6 +139,22 @@ func (s *Server) ListConfigs(ctx context.Context, req *quotapb.ListConfigsReques
...
 	}
+
+	// Peek token counts for FULL view
+	if req.View == quotapb.ListConfigsRequest_FULL {
+		names := make([]string, 0, len(resp.Configs))
+		for _, cfg := range resp.Configs {
+			names = append(names, cfg.Name)
+		}
+		tokens, err := s.qs.Peek(ctx, names)
+		if err == nil {
```
Martin2112: Also here.

quota/etcd/quotaapi/quota_server.go
Martin2112: Sorry, I meant indent the error case.

codingllama: I don't think that practice rule applies in this context; this error instance never gets forwarded, nor does it really abort the flow. We can (and do) proceed regardless of it. It's kind of a "bonus".

quota/etcd/quotaapi/quota_server.go
codingllama: Same as above.

quota/etcd/quotaapi/quota_server.go
codingllama: I didn't see the 2nd reply in time, now I'm not sure I understood you in the first place.

Martin2112: It's normally 

if err != nil {
  handle it
 return wibble
}

// non error case

codingllama: Right. I understood you correctly, then. I don't think that applies here, as this isn't the "standard" `return err` scenario: in this case err doesn't stop the flow nor ever gets returned. I could test for the negative if you prefer, although personally I dislike testing the negative when both if branches have work to do.

Martin2112: Why does Peek() return an error if we don't need to handle it? Can it not just return nil tokens?

codingllama: Discussed offline.

I've bumped the log to warning and tweaked the message a bit.

TL;DR: It's OK to discard Peek() errors in this context, as it's likely that the caller can live without current_tokens.

codingllama:
PTAL?

PR #821, opened 2017-09-06
Cache tokens when using etcd quotas

codingllama:


codingllama:
Issue #490.

codingllama:
Friendly ping?

server/quota.go
```diff
@@ -79,7 +84,13 @@ func NewQuotaManager(params *QuotaParams) (quota.Manager, error) {
 	default:
 		return nil, fmt.Errorf("unknown quota system: %v", params.QuotaSystem)
 	}
+	qmType := fmt.Sprintf("%T", qm)
+
+	if params.QuotaSystem == QuotaEtcd && params.MinBatchSize > 0 && params.MaxCacheEntries > 0 {
```
Martin2112: If someone provides negative values does this still work?

codingllama: No, neither of those make sense as negative numbers. There aren't any magical "zero = infinite" or "zero = disabled" values either.

I could change `cache.NewCachedManager` to return an error and validate there too, if you like.

Martin2112: That sounds like the best option rather than doing it here as well.

codingllama: OK. I'll keep both though: `cache.NewCachedManager` to avoid bugs and here so caching could be disabled.

I'll update the flag descriptions too. (I'll ping when done.)

codingllama:
PTAL?

codingllama:
#822 has the Travis fix.

PR #822, opened 2017-09-08
Disable certificate-transparency-go integration tests

codingllama:
Temporary until certificate-transparency-go vendors its dependencies, at
which point Travis should work again


codingllama:
Tests should be green, otherwise the fix failed. :)

Feel free to merge before me if I take too long.

Issue #590, opened 2017-05-05
Support tree deletion / undeletion


codingllama:
The Admin API defines a [DeleteTree RPC](https://github.com/google/trillian/blob/18b70c3b22763498f702bc8ed6d2e30df6ad6f24/trillian_admin_api.proto#L105), but currently it's [not implemented](https://github.com/google/trillian/blob/18b70c3b22763498f702bc8ed6d2e30df6ad6f24/server/admin/admin_server.go#L175).

As deletion is a potentially dangerous operation, DeleteTree is meant to instead soft-delete trees. A background process will be responsible for transitioning soft-deleted trees to hard-delete after a reasonable amount of time passes (say, 2 weeks). While in soft deletion, trees may be restored via UndeleteTree.

A (soft) deleted tree becomes immediately invisible for the purpose of other RPCs, such as the ones in Log / Map API (plus most Admin requests).

Edit: links

codingllama:
Arbitrarily assigned to M10.

codingllama:
This is done now.

PR #823, opened 2017-09-08
Remove tree deletion states and add "deleted" and "delete_time"

codingllama:
deleted and delete_time are the standard way to represent soft deletion
according to Google's API style guide[1]. Furthermore, it has the added
benefit of "remembering" the correct previous state for undeletion.

Storage and tests have been changed accordingly, even though the actual
Delete / Undelete features are outstanding.

The objective of the PR is not to implement deletion, but to prepare the
existing codebase for it by adding the correct concepts to API, storage
and delete-sensitive methods that already exist.

[1] https://cloud.google.com/apis/design/standard_fields


codingllama:
Issue #590

PR #838, opened 2017-09-21
Change SignedMapRoot metadata to google.protobuf.Any

phad:
This leaves the choice of exactly what metadata should be stored in the
signed map head to the map personality.

The previous MapperMetadata proto moves into the examples/ct/ctmapper
as an example of how marshalling should be done.  It is also used in
the existing Trillian map tests as it was before.


examples/ct/ctmapper/ctmapperpb/ct_mapper.proto
```diff
@@ -16,6 +16,12 @@ syntax = "proto3";
 
 package ctmapperpb;
 
+message MapperMetadata {
```
codingllama: nit: Please document proto and fields.

trillian.proto
```diff
@@ -194,20 +194,15 @@ message SignedLogRoot {
...
-  int64 highest_fully_completed_seq = 2;
-  int64 highest_partially_completed_seq = 3;
- }
-
 // SignedMapRoot represents a commitment by a Map to a particular tree.
 message SignedMapRoot {
   int64 timestamp_nanos = 1;
   bytes root_hash = 2;
-  MapperMetadata metadata = 3;
+  reserved 3;
```
codingllama: Maybe `reserved 3;  // MapperMetadata metadata (removed)` ?

trillian.proto
```diff
@@ -194,20 +194,15 @@ message SignedLogRoot {
...
 	// TODO(al): define serialized format for the signature scheme.
   sigpb.DigitallySigned signature = 4;
 
   int64 map_id = 5;
   int64 map_revision = 6;
+  google.protobuf.Any metadata = 7;
```
codingllama: Please document. In particular, what types allowed/expected to be in the Any?

trillian_map_api.proto
```diff
@@ -58,7 +59,8 @@ message GetMapLeavesResponse {
 message SetMapLeavesRequest {
   int64 map_id = 1;
   repeated MapLeaf leaves = 2;
-  MapperMetadata mapper_data = 3;
+  reserved 3;
```
codingllama: Maybe `reserved 3;  // MapperMetadata mapper_data (removed)` ?

trillian_map_api.proto
```diff
@@ -58,7 +59,8 @@ message GetMapLeavesResponse {
...
+  google.protobuf.Any mapper_metadata = 4;
```
codingllama: Please document (as above)

trillian.proto
phad: The intention is that any protocol message could be stored here, via ptypes Marshaling.
In reality the Trillian storage in use will probably impose a size limit on that, but I'm not sure
I can/should document that here other than saying something general about it.

trillian.proto
phad: Done

examples/ct/ctmapper/ctmapperpb/ct_mapper.proto
phad: OK ... in fairness it wasn't documented to start with!  I'll improve on that :)

integration/maptest/map_test.go
```diff
@@ -211,6 +216,16 @@ func TestLeafHistory(t *testing.T) {
 	}
 }
 
+func mustMarshalMetadata(t *testing.T, in proto.Message) *any.Any {
```
AlCutter: factor this out to `testonly/` so can reuse here and in `crypto/verifier_test.go`?

storage/mysql/map_storage.go
```diff
@@ -270,32 +271,31 @@ func (m *mapTreeTX) LatestSignedMapRoot(ctx context.Context) (trillian.SignedMap
...
 		Signature:      &rootSignature,
 		MapId:          m.treeID,
-		Metadata:       mapperMeta,
 	}
 
-	return ret, nil
+	if mapperMetaBytes != nil {
+		var mapperMeta any.Any
+		if err := proto.Unmarshal(mapperMetaBytes, &mapperMeta); err != nil {
+			glog.Warningf("Failed to unmarshal Metadata; %v", err)
```
AlCutter: ";" -> ":"

storage/mysql/map_storage_test.go
```diff
@@ -137,6 +140,16 @@ type rootReaderMapTX interface {
 	storage.MapRootReader
 }
 
+func mustMarshalMetadata(t *testing.T, in proto.Message) *any.Any {
```
AlCutter: another one! :)

trillian_map_api.proto
phad: Will do.

Looking back at this I'm bothered by having 'mapper_metadata' in SetMapLeavesRequest but 'metadata' in SignedMapHead.

I'd prefer them to be the same, and think on balance 'metadata' is fine. wdyt?

trillian_map_api.proto
phad: Done

examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -77,14 +80,19 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
...
+		mapperMetadata = metadataProto.Message.(*ctmapperpb.MapperMetadata)
+	}
 
 	startEntry := int64(0)
 
-	if meta == nil {
-		meta = &trillian.MapperMetadata{}
-	}
-	startEntry = meta.HighestFullyCompletedSeq + 1
+	startEntry = mapperMetadata.HighestFullyCompletedSeq + 1
```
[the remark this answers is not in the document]
phad: oops, done.

storage/mysql/map_storage.go
phad: done

trillian.proto
```diff
@@ -194,20 +194,20 @@ message SignedLogRoot {
...
-  MapperMetadata metadata = 3;
+  reserved 3;  // was MapperMetadata (removed, replaced by 'Any metadata').
 	// TODO(al): define serialized format for the signature scheme.
   sigpb.DigitallySigned signature = 4;
-
   int64 map_id = 5;
   int64 map_revision = 6;
+  // Metadata associated with the Map root.  This is set by the map personality
+  // layer in a SetMapLeavesRequest request.  It could be used to persist state
+  // needed to recreate the Map from an external data source. 'Any' can hold
```
codingllama: That's good, thanks. I think it's fine to drop the "how any works section", you've answered the important questions by saying how data gets here (SetMapLeaves) and who's responsible for it (map personality).

A final q: is it OK to not have any metadata? (I assume so, but you may want to mention whether it's optional.)

trillian_map_api.proto
```diff
@@ -58,7 +59,15 @@ message GetMapLeavesResponse {
 message SetMapLeavesRequest {
   int64 map_id = 1;
   repeated MapLeaf leaves = 2;
-  MapperMetadata mapper_data = 3;
+  reserved 3;  // was MapperMetadata (removed, replaced by mapper_metadata).
+  // Metadata that the Map should associate with the new Map root after
+  // incorporating the changes in leaves.  The metadata will be reflected in
+  // the Map Root returned in SetMapLeavesResponse.
+  // Map personalities should use metadata to persist any state needed later
+  // to continue mapping from an external data source. Any protocol message
```
codingllama: (Ditto for the Any section)

examples/ct/ctmapper/ctmapperpb/ct_mapper.proto
codingllama: Apologies for pushing that to you. I'm glad I did, though, as it's a good improvement.  ;)

trillian_map_api.proto
```diff
@@ -58,7 +59,15 @@ message GetMapLeavesResponse {
...
+  // can be assigned, via Any marshalling (see github.com/golang/protobuf/ptypes).
+  // Size limitations may apply dependent on the Trillian storage type in use.
```
codingllama: I think it's OK to not talk about the size limit at this layer if you prefer. We don't talk specifically about size limits in other places that do have them, plus I think it's fine to omit this error as long as we have good error messages.

trillian.proto
codingllama: My questions were mostly around purpose, lifecycle and content restrictions (if any). You've made those clear.

trillian_map_api.proto
codingllama: I think metadata is OK. I'd definitely encourage using regular names, it makes for more intuitive APIs.

trillian_map_api.proto
```diff
@@ -58,7 +59,15 @@ message GetMapLeavesResponse {
 message SetMapLeavesRequest {
   int64 map_id = 1;
   repeated MapLeaf leaves = 2;
-  MapperMetadata mapper_data = 3;
+  reserved 3;  // was MapperMetadata (removed, replaced by mapper_metadata).
+  // Metadata that the Map should associate with the new Map root after
```
codingllama: There's an assumption of a 1:1 relationship between SetMapLeaves calls and SignedMapRoots. Is that true/a guarantee we'd like to make? If not, what happens in regards to the metadata?

trillian.proto
```diff
@@ -194,20 +194,20 @@ message SignedLogRoot {
...
   bytes root_hash = 2;
-  MapperMetadata metadata = 3;
+  reserved 3;  // was MapperMetadata (removed, replaced by 'Any metadata').
 	// TODO(al): define serialized format for the signature scheme.
   sigpb.DigitallySigned signature = 4;
-
   int64 map_id = 5;
   int64 map_revision = 6;
+  // Metadata associated with the Map root.  This is set by the map personality
+  // layer in a SetMapLeavesRequest request.  It could be used to persist state
```
codingllama: nit: I think it best to document in terms of RPC calls (SetMapLeaves) instead of request messages (SetMapLeavesRequest). This is heavy nit territory though, so feel free to push back.

trillian.proto
phad: Do you think the bit about "the Map personality should know what types it has previously assigned in SetMapLeavesRequest calls" is implied and could be omitted too?

The 'ok to not have metadata' question is something I want to leave for the next PR.  That's going to depend on how (or if) we decide to represent the use of map metadata in the Admin CreateTree API.

trillian.proto
phad: Actually the RPC is just called 'SetLeaves'. it feels a bit general but I'll go with that.

trillian_map_api.proto
phad: OK, done.

trillian_map_api.proto
phad: Done.

trillian_map_api.proto
phad: I believe the Trillian Map generates a new signed map root after every call to SetLeaves.
However this is not documented, and feels to me like something that might end up having to be made tweakable in a practical deployment?

trillian.proto
codingllama: 1. Yes, I think it's fine to imply that.

2. Fair enough.

trillian_map_api.proto
codingllama: If it's tweakable, then what happens to the metadata when the 1:1 relationship is broken? Does it get overwritten? Do we need a `repeated metadata` in SignedMapRoot?

phad: I have misled you :).  It isn't tweakable.  As I understand it there is no sequencing in the Map, so as soon as a SetLeaves RPC has completed there will have been a new signed map head created, which is returned in the SetLeaves response.  So there is a 1:1 relationship, although it isn't explicitly stated in the API.

My intended meaning was to express some doubt as to whether this will be performant for a map under heavy load, but I guess we cross that bridge when we get to it.

AlCutter: 👍 it's intended to both enable mapper personalities to 'pick up' where they left off, as well as provide  a commitment from the system that the given root hash is reproducible at the associated mapper state. The latter allows monitors to verify the correct operation of the mapper & Map.

phad: So do you think a verifying monitor (monitoring the mapper/Map) be likely to iterate through the historic SignedMapHeads and use the metadata at each snapshot to conduct its verification?  Or is this getting a bit application-specific?

phad: I do think the 'verify that a given root hash is reproducible at the associated mapper state' is a rather application (personality)-specific thing.  It assumes that the Map is being used in conjunction with an append-only log, and that the full log can be queried by the verifying party.  That might not always be the case?

storage/mysql/map_storage.go
```diff
@@ -270,32 +271,31 @@ func (m *mapTreeTX) LatestSignedMapRoot(ctx context.Context) (trillian.SignedMap
 
 func (m *mapTreeTX) signedMapRoot(timestamp, mapRevision int64, rootHash, rootSignatureBytes, mapperMetaBytes []byte) (trillian.SignedMapRoot, error) {
```
codingllama: (pre-existing) I think this should return `*trillian.SignedMapRoot`. Protos are usually too large to comfortably pass by value (plus they have a tendency to "fatten" over time).

Maybe add a TODO?

storage/mysql/map_storage.go
```diff
@@ -270,32 +271,31 @@ func (m *mapTreeTX) LatestSignedMapRoot(ctx context.Context) (trillian.SignedMap
...
 		Signature:      &rootSignature,
 		MapId:          m.treeID,
-		Metadata:       mapperMeta,
 	}
 
-	return ret, nil
+	if mapperMetaBytes != nil {
+		var mapperMeta any.Any
+		if err := proto.Unmarshal(mapperMetaBytes, &mapperMeta); err != nil {
+			glog.Warningf("Failed to unmarshal Metadata: %v", err)
```
codingllama: nit: Don't log and return the error; let the caller log, if they want to.

storage/mysql/map_storage_test.go
```diff
@@ -167,28 +170,24 @@ func TestMapRootUpdate(t *testing.T) {
...
 			root: trillian.SignedMapRoot{
 				MapId:          mapID,
 				TimestampNanos: 98768,
 				MapRevision:    7,
 				RootHash:       []byte(dummyHash),
 				Signature:      &spb.DigitallySigned{Signature: []byte("notempty")},
-				Metadata: &trillian.MapperMetadata{
-					HighestFullyCompletedSeq: 0,
-				},
+				Metadata: &any.Any{},
```
codingllama: I find an empty Any a bit unusual. I would expect it to either be nil or be "unmarshable".

testonly/map_metadata.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+)
+
+// MustMarshalMetadata is used in tests to Marshal proto messages used as map
+// metadata into the protobuf.ptypes.any.Any used in the Trillian API and in
+// storage.  Failure to marshal will fail the test.
+func MustMarshalMetadata(t *testing.T, in proto.Message) *any.Any {
```
codingllama: Could you make this more general? Maybe `MustMarshalAny` / `testonly/any.go`?

Bonus points for replacing our [ever-growing army of mustMarshalAny functions](https://github.com/google/trillian/search?utf8=%E2%9C%93&q=mustMarshalAny&type=) (or adding a TODO).

testonly/map_metadata.go
```diff
@@ -0,0 +1,35 @@
...
+	var metaAny *any.Any
+	metaAny, err := ptypes.MarshalAny(in)
+	if err != nil {
+		t.Fatalf("failed to marshal %v as 'any': err %v", in, err)
```
codingllama: FYI, our other must*Something* functions tend to have a single parameter and panic on failures. Both are acceptable, but I've grown to slightly prefer the panic version, FWIW.

testonly/map_metadata.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+)
+
+// MustMarshalMetadata is used in tests to Marshal proto messages used as map
+// metadata into the protobuf.ptypes.any.Any used in the Trillian API and in
+// storage.  Failure to marshal will fail the test.
+func MustMarshalMetadata(t *testing.T, in proto.Message) *any.Any {
+	var metaAny *any.Any
```
codingllama: nit: No need to declare `metaAny` beforehand.

examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -77,14 +80,17 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
...
 		return false, err
 	}
-	meta := getRootResp.MapRoot.Metadata
 
-	startEntry := int64(0)
-
-	if meta == nil {
-		meta = &trillian.MapperMetadata{}
+	mapperMetadata := &ctmapperpb.MapperMetadata{}
+	if getRootResp.MapRoot.Metadata != nil {
```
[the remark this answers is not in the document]
phad: I'd be surprised if a valid RPC response for GetRoot didn't contain a root...
I'll change this to GetMapRoot() anyway as it will not return nil.

examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -184,7 +190,12 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
 		})
 	}
 
-	setReq.MapperData = meta
+	var metaAny *any.Any
```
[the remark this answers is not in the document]
phad: I'm not hugely in favour of doing this as if MarshalAny returned both a *Any and an error we'd have updated setReq in dubious circumstances.

storage/mysql/map_storage.go
phad: Unsure about this.  The entire Log and Map API is expressed in this way. @AlCutter views?

storage/mysql/map_storage.go
```diff
@@ -270,32 +271,31 @@ func (m *mapTreeTX) LatestSignedMapRoot(ctx context.Context) (trillian.SignedMap
...
 		TimestampNanos: timestamp,
 		MapRevision:    mapRevision,
 		Signature:      &rootSignature,
 		MapId:          m.treeID,
-		Metadata:       mapperMeta,
 	}
 
-	return ret, nil
+	if mapperMetaBytes != nil {
+		var mapperMeta any.Any
```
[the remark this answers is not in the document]
phad: I'm pretty sure proto.Unmarshal requires an allocated struct to write into.
So the best I could do is to write:
`
mapperMeta := &any.Any{}
proto.Unmarshal(mapperMetaBytes, mapperMeta)
smr.Metadata = mapperMeta
`
or
`
smr.Metadata = &any.Any{}
proto.Unmarshal(mapperMetaBytes, smr.Metadata)
`
although it rubs my 'what are we left with if an error is returned' thing like above.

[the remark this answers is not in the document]
phad: OK, I read the docs ... proto.Unmarshal() always starts by resetting the passed-in proto. If an error occurs when subsequently reading the marshalled data then hopefully the proto remains in its reset state.  I'm happy enough to trust that that happens.

storage/mysql/map_storage.go
phad: Done.

storage/mysql/map_storage_test.go
phad: I don't think this is a realistic test case, but this test is checking round-trips.  Even if unrealistic this structure should still be round-trippable.  I can remove it if you think it's pointless?
The nil case is the one above.
Perhaps I should devise a test where the Any contains unmarshalable data.

testonly/map_metadata.go
phad: Oh I didn't realise how many times that's already been done! I'll go and fix that.

testonly/map_metadata.go
AlCutter: Won't `panic()`ing here cause the whole test suite to fail?

testonly/map_metadata.go
AlCutter: maybe add a`TODO(phad): t.Helper() when we bump to go 1.9`?

examples/ct/ctmapper/mapper/mapper.go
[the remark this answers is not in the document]
codingllama: I'm OK with that as long as error return follows shortly after, so it's easy to see that we never pass "setReq" forward when it's in an undefined state.

Don't feel obliged to change anything, though.

storage/mysql/map_storage.go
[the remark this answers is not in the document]
codingllama: I would still favor a pointer variable. `mapperMeta := &any.Any{}` LGTM.

storage/mysql/map_storage.go
```diff
@@ -270,32 +271,31 @@ func (m *mapTreeTX) LatestSignedMapRoot(ctx context.Context) (trillian.SignedMap
...
+		if err := proto.Unmarshal(mapperMetaBytes, &mapperMeta); err != nil {
```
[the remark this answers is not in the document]
codingllama: Sorry, this kind of misses the point that Unmarshal needs an allocated Any. We could do `smr.Metadata = &any.Any{}` then Unmarshal, as you mentioned above. Up to you.

testonly/map_metadata.go
codingllama: In the context of a "MustMarshal" method that's fine. The assumption behind MustMarshal calls is that the test is working with canned, known-to-be-valid data. If the marshal breaks something is very wrong. It's "never should happen" territory.

examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -82,7 +82,7 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
 	}
 
 	mapperMetadata := &ctmapperpb.MapperMetadata{}
-	if getRootResp.MapRoot.Metadata != nil {
+	if getRootResp.GetMapRoot().Metadata != nil {
```
codingllama: Just `GetMapRoot()` doesn't buy you much, as panics anyway with a (hypothetically) nil getRootResp.MapRoot. We either go all the way (`getRootResp.GetMapRoot().GetMetadata()`) or don't.

testonly/map_metadata.go
```diff
@@ -26,9 +26,8 @@ import(
 // metadata into the protobuf.ptypes.any.Any used in the Trillian API and in
 // storage.  Failure to marshal will fail the test.
 func MustMarshalMetadata(t *testing.T, in proto.Message) *any.Any {
-	var metaAny *any.Any
 	metaAny, err := ptypes.MarshalAny(in)
-	if err != nil {
+	if err == nil {
```
codingllama: Shouldn't this be `err != nil`, as before?

testonly/map_metadata.go
codingllama: Thanks!  :D

phad:
The amalgamation of all the mustMarshalAny() call sites is enough work that I'll defer it for a future PR.

storage/mysql/map_storage.go
AlCutter: I don't really mind either way - the copying the proto is unlikely to be a significant fraction of the operation.

AlCutter:
Pop a `TODO` in for the `mustUnmarshalAny` refactor?

Build's breaking with this, btw:
`storage/mysql/map_storage_test.go:190: undefined: "github.com/google/trillian/testonly".MustMarshalAny`

testonly/map_metadata.go
AlCutter: I don't agree - if there multiple things wrong then you're masking the others by panic()ing; I think it's better to have a complete picture of what needs to be fixed rather than serially picking your way through multiple failures.

testonly/marshalling.go
```diff
@@ -1,3 +1,4 @@
+
```
AlCutter: spurious linebreak?

phad: Yup

testonly/map_metadata.go
codingllama: I think you're picking the wrong bone. Have a look at mustMarshalAny usages: they all start with hand-written, valid protos. If the proto is wrong, the entire test is broken. You wouldn't get much more out of Fatalf either, if anything.

It bears repeating: this is for hand-written, known-to-be-valid protos used in tests. Even if wanted to, I find it difficult to conjure a hand-written proto that breaks the marshaling.

codingllama:
Re deferring mustMarshalAny: SGTM. I'm happy to pick that up if you prefer; at least one of the existing ones is on me.

storage/mysql/map_storage.go
gdbelvin: FYI, Go style recommends always passing protos by reference

PR #839, opened 2017-09-22
Increase go test timeouts on Travis

codingllama:


codingllama:
I suppose 20m is enough, let's see.

@phad you might want to rebase #838 on this once it's merged.

scripts/presubmit.sh
```diff
@@ -126,7 +127,11 @@ main() {
 
       # Do not run go test in the loop, instead echo it so we can use xargs to
       # add some parallelism.
-      echo go test -short -timeout=5m ${coverflags} ${goflags} "$d"
+      echo go test \
+          -short \
+          -timeout=${GO_TEST_TIMEOUT:=5m} \
```
phad: unsure of this bash syntax.  is the = meant to be there (I assume this marks the default when GO_TEST_TIMEOUT isn't set?)

AlCutter: The `:=` means if `var` is set to null or undeclared then set it to `5m` and use that value.
More frequently you see `:-` which means if... null or undeclared then just use this default value.

codingllama: Yes, `:=` defines a default value in case the var is not set.

I forget those frequently too, so my trick is to lookup "expansion" on [man bash](https://linux.die.net/man/1/bash) (search for "assign default value" for an easier time).

codingllama:
Thanks, Paul. Now let's see if it passes Travis. :)

scripts/presubmit.sh
codingllama: @AlCutter, oh, right, I guess I meant to use `:-` instead of `:=`.

*codingllama silently despairs about the prospect of facing yet another Travis run*

PR #842, opened 2017-09-26
Make trillian.Tree.PrivateKey a mutable field

RJPercival:
Checks are performed to ensure that the underlying key material is not changed; only the protobuf message that communicates how to obtain it. This makes it possible to migrate a tree from one key management system to another.

### DIFF
--- a/trillian.proto
+++ b/trillian.proto
@@ -139,9 +139,9 @@ message Tree {
   // This can be any type of message to accommodate different key management
   // systems, e.g. PEM files, HSMs, etc.
   // Private keys are write-only: they're never returned by RPCs.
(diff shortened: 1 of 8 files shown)


trillian.proto
```diff
@@ -139,9 +139,6 @@ message Tree {
   // This can be any type of message to accommodate different key management
   // systems, e.g. PEM files, HSMs, etc.
   // Private keys are write-only: they're never returned by RPCs.
```
codingllama: Could you document the update logic? Ie, that it's valid to change providers, but the key itself can't change.

RJPercival: Done.

PR #844, opened 2017-09-27
Tree garbage collection

codingllama:
Create a background task that periodically scans for and hard-deletes trees that have been soft-deleted longer than the specified period.

### DIFF
--- a/server/admin/tree_gc.go
+++ b/server/admin/tree_gc.go
@@ -43,7 +43,7 @@ type DeletedTreeGC struct {
 	Admin storage.AdminStorage
 
 	// DeleteThreshold defines the minimum time a tree has to remain in the soft-deleted state
-	// before its eligible for garbage collection.
+	// before it's eligible for garbage collection.
 	DeleteThreshold time.Duration
 
 	// MinRunInterval defines how frequently sweeps for deleted trees are performed.
@@ -51,9 +51,15 @@ type DeletedTreeGC struct {
 	MinRunInterval time.Duration
 }
 
-// Run starts the tree garbage collection process. It never returns.
+// Run starts the tree garbage collection process. It runs until ctx is cancelled.
 func (gc *DeletedTreeGC) Run(ctx context.Context) {
 	for true {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+
 		gc.RunOnce(ctx)
 
 		d := gc.MinRunInterval + time.Duration(rand.Int63n(gc.MinRunInterval.Nanoseconds()))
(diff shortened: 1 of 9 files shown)


server/admin/tree_gc.go
```diff
@@ -0,0 +1,121 @@
...
+// * Hard deletion, which effectively removes all tree data
+//
+// DeletedTreeGC performs the transition from soft to hard deletion. Trees that have been deleted
+// for at least DeletedThreshold are eligible for garbage collection.
+type DeletedTreeGC struct {
+	// Admin is the storage.AdminStorage interface.
+	Admin storage.AdminStorage
+
+	// DeleteThreshold defines the minimum time a tree has to remain in the soft-deleted state
+	// before its eligible for garbage collection.
```
AlCutter: it's

server/admin/tree_gc.go
```diff
@@ -0,0 +1,121 @@
...
+	DeleteThreshold time.Duration
+
+	// MinRunInterval defines how frequently sweeps for deleted trees are performed.
+	// Actual runs happen randomly between [minInterval,2*minInterval).
+	MinRunInterval time.Duration
+}
+
+// Run starts the tree garbage collection process. It never returns.
```
AlCutter: Shouldn't it return if ctx is cancelled?

server/admin/tree_gc.go
```diff
@@ -0,0 +1,121 @@
...
+	// deleted trees are unlikely to change, specially those deleted for a while.
+	tx, err := gc.Admin.Snapshot(ctx)
+	if err != nil {
+		glog.Errorf("DeletedTreeGC.RunOnce: error creating snapshot: %v", err)
+		return
+	}
+	defer tx.Close()
+	trees, err := tx.ListTrees(ctx, true /* includeDeleted */)
+	if err != nil {
+		glog.Errorf("DeletedTreeGC.RunOnce: error listing trees: %v", err)
```
AlCutter: Should we increment a counter on these fails so that the operator can be alerted?

server/admin/tree_gc.go
codingllama: Done.

server/admin/tree_gc.go
codingllama: Done. Good call.

server/admin/tree_gc.go
codingllama: I had that planned for a separate PR, if that's OK.

AlCutter: Sure thing - pop a `TODO` in?

codingllama: Thanks. I'll forgo the TODO as I have it the new PR pretty much done already. I hope that's OK.

PR #849, opened 2017-09-28
Add private_key to AdminServer.UpdateTree FieldMask test

RJPercival:
PR #842 made this a valid addition to the field mask, but did not augment the test to prevent regression.

### DIFF
--- a/server/admin/admin_server_test.go
+++ b/server/admin/admin_server_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/google/trillian"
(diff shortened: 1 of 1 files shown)


server/admin/admin_server_test.go
```diff
@@ -562,7 +562,9 @@ func TestServer_UpdateTree(t *testing.T) {
 		StorageSettings: settings,
 		MaxRootDuration: ptypes.DurationProto(2 * time.Nanosecond),
 	}
-	successMask := &field_mask.FieldMask{Paths: []string{"tree_state", "display_name", "description", "storage_settings", "max_root_duration"}}
+	successMask := &field_mask.FieldMask{
+		Paths: []string{"tree_state", "display_name", "description", "storage_settings", "max_root_duration", "private_key"},
```
codingllama: I think you need a corresponding change in successTree, otherwise I'd expect it to error.

RJPercival: Because private_key is redacted from the response, it doesn't need a change to successTree.

codingllama: Because successTree and successMask are paired together in the update request, I expected successTree to have a non-nil PrivateKey, otherwise applying the mask would just "nil" the old key.

The test is heavily mocked, so I think that's why it works. You might get better value by adding this to the admin integration test instead.

RJPercival: Ah yes, and I see that `successWant.PrivateKey = nil // redacted on responses` is also misleading, since in the test `applyUpdateMask()` is run **after** `UpdateTree()` has returned, which would override the redaction done by `UpdateTree()`. I've fixed the test so that the tree is updated at the appropriate time and the redaction still takes place.

server/admin/admin_server_test.go
```diff
@@ -561,8 +562,17 @@ func TestServer_UpdateTree(t *testing.T) {
 		Description:     "Brand New Tree Desc",
 		StorageSettings: settings,
 		MaxRootDuration: ptypes.DurationProto(2 * time.Nanosecond),
+		PrivateKey: func() *any.Any {
```
codingllama: Suggestion: `PrivateKey: testonly.MustMarshalAny(t, &empty.Empty{}),`

(where `testonly` is [github.com/google/trillian/testonly](https://github.com/google/trillian/blob/master/testonly/marshalling.go#L41))

RJPercival: Done.

PR #852, opened 2017-10-02
travis: add 1.9.x instead 1.8 go version

dvrkps:


codingllama:
Hey Davor, thanks for the contribution. In principle looks good to me, I just want to make sure there are no objections on the team before I merge, as this means a "formal" move to 1.9.

PR #854, opened 2017-10-03
Add a timeout to PutTokens calls made by interceptors

codingllama:


server/interceptor/interceptor.go
```diff
@@ -40,6 +41,11 @@ const (
 )
 
 var (
+	// PutTokensTimeout is the timeout used for PutTokens calls.
+	// PutTokens happens in a separate goroutine and with an independent context, therefore it has
+	// its own timeout, separate from the RPC that causes the calls.
+	PutTokensTimeout = 5 * time.Second
```
gdbelvin: Should this be a flag?

server/interceptor/interceptor_test.go
```diff
@@ -397,12 +397,13 @@ func TestTrillianInterceptor_QuotaInterception_ReturnsTokens(t *testing.T) {
 		},
 	}
 
-	// Use a ctx with deadline so we can check whether PutTokens inherits it.
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	// Use a ctx with a timeout smaller than PutTokensTimeout
+	ctx, cancel := context.WithTimeout(context.Background(), PutTokensTimeout-3*time.Second)
```
gdbelvin: Depending on the value of `PutTokensTimeout`, this might be a negative number. 
Consider using `PutTokensTimeout/2`

server/interceptor/interceptor.go
codingllama: It could be a flag, that's why it's setup this way, but I don't think it warrants a flag right now. We can put one in rather quickly if we need it.

server/interceptor/interceptor_test.go
codingllama: Good call. I've set an explicit value to PutTokensTimeout for the duration of this test, so it's easy to reason about the numbers.

codingllama:
Thanks for the review!

PR #864, opened 2017-10-12
Fix bug introduced in PR#862

phad:
I thought I could factor out the code

  req.Index = [][]byte{s.pickKey(latestCopy)}

(as it seemed to be used in every branch of switch)
but neglected to consider the case where the set of known keys is empty.

This bug trips if by random seed the first operation being hammered is a getLeavesInvalid().

This PR restores the original logic.


phad:
Note this bug will make the Travis run flaky, so ought to be fixed ASAP!

PR #871, opened 2017-10-23
Add tests for GetSignedMapRootByRevision.

phad:


server/map_rpc_server_test.go
```diff
@@ -128,6 +142,84 @@ func TestGetSignedMapRoot(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	ctx := context.Background()
+
+	tests := []struct {
+		desc string
+		req *trillian.GetSignedMapRootByRevisionRequest
+		mapRoot trillian.SignedMapRoot
+		snapShErr, lsmrErr error
+		wantErr bool
```
gdbelvin: is `wantErr` redundant / equal to ` snapShErr != nill || lsmrErr != nil`?

phad: Yes, effectively.  I had it in the test tables explicitly to make it clear what outcome the test expects.

server/map_rpc_server_test.go
```diff
@@ -32,17 +33,30 @@ const (
 	mapID1 = int64(1)
 )
 
+var (
+	signedMapRootId1Rev1 = trillian.SignedMapRoot{
```
AlCutter: should this be `signedMapRootIDRev1`?

PR #874, opened 2017-10-24
Extract the bulk of map_test.go into new map.go

phad:
This allows the map integration tests to be run in
other environments.


phad:
Sorry, github presents the diff very badly.
If you were to clone my repo and examine the commit at the cmdline you'll see that map.go was forked from map_test.go and the diffs should then be clear.

integration/map.go
```diff
@@ -0,0 +1,416 @@
...
+		if err := merkle.VerifyMapInclusionProof(treeID, index,
+			leaf, rootHash, proof, hasher); err != nil {
+			return fmt.Errorf("VerifyMapInclusionProof(%x): %v", index, err)
+		}
+	}
+	return nil
+}
+
+// newTreeWithHasher is a test setup helper for creating new trees with a given hasher.
+func newTreeWithHasher(ctx context.Context, aCli trillian.TrillianAdminClient, hashStrategy trillian.HashStrategy) (*trillian.Tree, hashers.MapHasher, error) {
```
gdbelvin: I tend to call these clients something like `tadmin`.  Up to you if you think that's cleaner

gdbelvin:
I see that the integration tests have been moved to another package. 
Could you give a little more detail around the reasoning for putting the unit tests in a different package? Could the same goals be achieved without the package move?

integration/map.go
phad: Done, in the interest of consistency.  Also 'admin' means more to people than 'a'.
Did the same for mCli -> tmap as well.

phad:
The package move is for consistency with the log integration tests, where the bulk of the test logic lives in ./integration.

For log integration the test driver (log_integration_test.go) is also in ./integration.  I'd like to have placed the map integration driver there as well, but then all log/map integration tests would be in a single package, which will confuse the .sh scripts that run them.

gdbelvin:
What would you think about modifying the log integration tests to look a bit more like the map tests?

phad:
I'd rather not, to be honest.

phad:
A bit more detail on why the package move occured:
1/ I needed to separate the code that establishes the map integration test environment (with the grpc, mysql deps) from the tests themselves, so split map.go away from map_test.go
2/ That could be done with the resulting two files remaining in the same package but it seemed cleaner for them to be done in separate packages, especially when considering the re-use of map.go (the test logic) in a different non-grpc, non-mysql environment.

gdbelvin:
That sounds pretty reasonable.  The plan as-is sounds good.  LGTM once travis is happy.

phad:
Travis is happy;  WDYT?

integration/maptest/map_test.go
```diff
@@ -15,409 +15,54 @@
...
-			}); err != nil {
-				t.Errorf("%v: SetLeaves(): %v", tc.desc, err)
-				continue
-			}
-
-			indexes := [][]byte{}
-			for _, l := range tc.leaves {
-				indexes = append(indexes, l.Index)
+	for _, test := range AllTests {
+		t.Run(test.Name, func(t *testing.T) {
```
gdbelvin: What do you think about passing the `testing.T` object to the subtests? 
That way we preserve the line number where things went wrong

integration/maptest/map_test.go
```diff
@@ -15,409 +15,54 @@
...
-	if err := tcrypto.VerifyObject(pubKey, smr, getResp.GetMapRoot().GetSignature()); err != nil {
-		return fmt.Errorf("VerifyObject(SMR): %v", err)
-	}
-	rootHash := getResp.GetMapRoot().GetRootHash()
-	for _, incl := range getResp.MapLeafInclusion {
-		leaf := incl.GetLeaf().GetLeafValue()
-		index := incl.GetLeaf().GetIndex()
-		leafHash := incl.GetLeaf().GetLeafHash()
-		proof := incl.GetInclusion()
+func TestMain(m *testing.M) {
```
gdbelvin: Could we keep `TestMain` in `main_test.go`?   Would simplify the diff.

gdbelvin: Also, do we still need a TestMain?

integration/maptest/map_test.go
```diff
@@ -15,409 +15,54 @@
...
-					continue
-				}
-			}
-		}
-	}
+	os.Exit(m.Run())
 }
 
-func TestInclusion(t *testing.T) {
+func TestMapIntegration(t *testing.T) {
```
gdbelvin: Perhaps move `TestMapIntegration` to `main_test.go` since it kind of replaces the previous `TestMain` function?

integration/maptest/map_test.go
phad: We do still need TestMain, as it is the only opportunity to install the pre-test hooks for the environment.

integration/maptest/map_test.go
phad: I think this is the correct way to use subtests.

The testing.Run implementation does create a new testing.T to pass into the sub-test.

If you pass in the parent one (by declaring the sub-test as func(*testing.T) omitting the arg name) then errors from the subtest get mis-attributed

integration/maptest/map_test.go
phad: Yes, we could keep TestMain in main_test.go

integration/maptest/map_test.go
phad: Sorry, I don't follow this.

Issue #875, opened 2017-10-24
Logserver crashes with an empty leaf on QueueLeaf


codingllama:
Example:

```bash
# window 1
trillian_log_server -alsologtostderr

# window 2
createtree --admin_server localhost:8090 --display_name Test
> 620855126134985004

curl -d '{}' localhost:8091/v1beta1/logs/620855126134985004/leaves
curl: (52) Empty reply from server
```

Stack trace (edited):

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x91cf2f]

goroutine 105 [running]:
(...)
github.com/google/trillian/server.(*TrillianLogRPCServer).QueueLeaves(0xc420234630, 0xe534a0, 0xc4203229c0, 0xc42039b568, 0x0, 0x0, 0x0)
  	github.com/google/trillian/server/log_rpc_server.go:100 +0x19f
github.com/google/trillian/server.(*TrillianLogRPCServer).QueueLeaf(0xc420234630, 0xe534a0, 0xc420322690, 0xc42035c160, 0xc420234630, 0xc42039b501, 0xc42035c1b0)
  	github.com/google/trillian/server/log_rpc_server.go:73 +0xef
(...)
```

While we're at it, the error message for empty `leaf_identity_hash` could be better. It currently reads:

```
queued leaf must have a leaf ID hash of length 32
```

A quick check on other logserver RPCs won't hurt either.

PR #877, opened 2017-10-26
Additional request validations for TrillianLogRPCServer

codingllama:
Added additional validations to a few TrillianLogRPCServer methods, most
notably QueueLeaf/QueueLeaves.

Validation tests have been refactored to go through RPC calls, so we can
test both correct wiring and the effectively returned error code. As a
consequence, validation_test.go was removed.

There's a small functionality change: leaf_identity_hash is not a
required parameter anymore. If it's not present, it's assumed to be the
same as merkle_leaf_hash.


codingllama:
Fixes #875.

server/log_rpc_server.go
```diff
@@ -96,8 +100,11 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
 	}
 	ctx = trees.NewContext(ctx, tree)
 
-	for i := range req.Leaves {
-		req.Leaves[i].MerkleLeafHash = hasher.HashLeaf(req.Leaves[i].LeafValue)
+	for _, leaf := range req.Leaves {
+		leaf.MerkleLeafHash = hasher.HashLeaf(leaf.LeafValue)
+		if len(leaf.LeafIdentityHash) == 0 {
```
gdbelvin: Is there a way we could document this behavior somewhere? Perhaps a comment on the proto or rpc definition? 

In the [log client](https://github.com/google/trillian/blob/master/client/verifier.go#L100), we're using SHA256 to fill `LeafIdentityHash`.  It would be nice to have these two be consistent.

codingllama: Done and done. I found a few other (now) redundant usages of LeafIdentityHash and removed those too.

codingllama:
PTAL?

server/validate.go
```diff
@@ -38,8 +38,32 @@ func validateGetInclusionProofByHashRequest(req *trillian.GetInclusionProofByHas
 	if req.TreeSize <= 0 {
 		return status.Errorf(codes.InvalidArgument, "GetInclusionProofByHashRequest.TreeSize: %v, want > 0", req.TreeSize)
 	}
+	if !isLeafHashValid(req.LeafHash) {
```
AlCutter: Maybe in-line this `len() > 0` test since the error message is quite specifically about that?
(or make `isLeafHashValid` return an err and use that here?)

server/validate.go
```diff
@@ -70,9 +94,41 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
 	return nil
 }
 
+func validateQueueLeafRequest(req *trillian.QueueLeafRequest) error {
+	if req.Leaf == nil {
+		return status.Error(codes.InvalidArgument, "QueueLeafRequest.Leaf empty")
+	}
+	if err := validateLogLeaf(req.Leaf); err != nil {
+		return status.Errorf(codes.InvalidArgument, "QueueLeafRequest.Leaf.%v", err)
```
AlCutter: Curious about the `.%v` for `err` - why not the more usual `...: %v` style?

storage/memory/log_storage.go
```diff
@@ -299,7 +298,7 @@ func (t *logTreeTX) UpdateSequencedLeaves(ctx context.Context, leaves []*trillia
 	for _, leaf := range leaves {
 		// This should fail on insert but catch it early
 		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
-			return errors.New("Sequenced leaf has incorrect hash size")
+			return errors.New("sequenced leaf has incorrect hash size")
```
AlCutter: Would making this a `fmt.Errorf` with the expected and received sizes be useful for debugging?

server/validate.go
codingllama: I didn't want to change the messages, but I agree with you. Changed `isLeafHash() bool` to `validateLeafHash() error` (and tweaked messages a bit as a result of it).

server/validate.go
codingllama: validateLogLeaf returns a message that chains nicely with this format. Added comments / docs to make it clearer.

storage/memory/log_storage.go
codingllama: Done.

codingllama:
Comments addressed (and rebased on master). PTAL.

PR #878, opened 2017-10-27
Enable keygen in integration tests

gdbelvin:
This allows integration tests to generate keys dynamically. 
This is needed to allow Key Transparency to exercise its new multi-tenant API during integration tests.  google/keytransparency#845

### DIFF
--- a/testonly/integration/logenv.go
+++ b/testonly/integration/logenv.go
@@ -24,8 +24,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/golang/protobuf/ptypes"
 	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/keys/der"
 	"github.com/google/trillian/crypto/keys/pem"
 	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/extension"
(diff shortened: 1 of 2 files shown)


RJPercival:
We should probably do the same thing for logenv.go, but doesn't need to happen in this PR.

testonly/integration/logenv.go
```diff
@@ -85,6 +88,9 @@ func NewLogEnv(ctx context.Context, numSequencers int, testID string) (*LogEnv,
 		AdminStorage: mysql.NewAdminStorage(db),
 		LogStorage:   mysql.NewLogStorage(db, nil),
 		QuotaManager: quota.Noop(),
+		NewKeyProto: func(ctx context.Context, spec *keyspb.Specification) (proto.Message, error) {
+			return der.NewProtoFromSpec(spec)
+		},
```
RJPercival: It's not *quite* this simple for LogEnv. If you look further down the file, you'll notice that it's using a hard-coded key for the trees. That needs changing to generate a key.

gdbelvin: That's actually ok.  `CreateLog` can still use a hardcoded key.  This change just allows _other_ tests that don't use `CreateLog` to use generated keys.  Example: If Key Transparency wanted to use a LogEnv and dynamically provision multiple trees.

PR #883, opened 2017-10-31
Try to fix up root hashing by using map[string]interface{}

Martin2112:
Try to fix up root hashing by using map[string]interface{}.

### DIFF
--- a/crypto/data_formats_test.go
+++ b/crypto/data_formats_test.go
@@ -22,9 +22,8 @@ import (
 )
 
 // It's important that signatures don't change.
-var expectedSigHex = "5e6baba8dc3465de9c01d669059dda590b7ce123d6ccd436bcd898f1c79ff6d9"
-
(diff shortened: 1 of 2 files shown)


crypto/data_formats_test.go
```diff
@@ -15,11 +15,26 @@
...
 
 import (
+	"encoding/hex"
 	"testing"
 
 	"github.com/google/trillian"
 )
 
+// It's important that signatures don't change.
+var expectedSigHex = "5e6baba8dc3465de9c01d669059dda590b7ce123d6ccd436bcd898f1c79ff6d9"
```
codingllama: nit: Move inside the test? It's not used anywhere else.

Martin2112: It looked quite ugly.

Martin2112: Oh move the var, sorry. Misunderstood.

Martin2112: Done.

PR #884, opened 2017-10-31
Make the server interceptor stop if context cancelled / deadline expires

Martin2112:
Previously it would continue e.g. if the get tree info timed out it would carry on. This is probably what was causing confusion in monitoring. I added a count metric for these so we can monitor it.

### DIFF
--- a/server/interceptor/interceptor.go
+++ b/server/interceptor/interceptor.go
@@ -160,7 +160,7 @@ func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (conte
 			incRequestDeniedCounter(badTreeReason, info.treeID, quotaUser)
 			return ctx, err
 		}
-		if err = ctx.Err(); err != nil {
+		if err := ctx.Err(); err != nil {
 			contextErrCounter.Inc(getTreeStage)
 			return ctx, err
 		}
--- a/server/interceptor/interceptor.go
+++ b/server/interceptor/interceptor.go
@@ -38,6 +38,9 @@ const (
 	badInfoReason            = "bad_info"
 	badTreeReason            = "bad_tree"
 	insufficientTokensReason = "insufficient_tokens"
+	getTreeInfoStage         = "get_tree_info"
+	getTreeStage             = "get_tree"
+	getTokensStage           = "get_tokens"
 )
 
 var (
@@ -48,6 +51,7 @@ var (
 
 	requestCounter       monitoring.Counter
 	requestDeniedCounter monitoring.Counter
+	contextErrCounter    monitoring.Counter
 	metricsOnce          sync.Once
 )
 
@@ -102,6 +106,10 @@ func initMetrics(mf monitoring.MetricFactory) {
 		"interceptor_request_denied_count",
 		"Number of requests by denied, labeled according to the reason for denial",
 		"reason", monitoring.TreeIDLabel, "quota_user")
+	contextErrCounter = mf.NewCounter(
+		"context_err_counter",
+		"Total number of times request context has been cancelled or deadline exceeded by stage",
+		"stage")
 }
 
 func incRequestDeniedCounter(reason string, treeID int64, quotaUser string) {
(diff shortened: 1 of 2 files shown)


server/interceptor/interceptor.go
```diff
@@ -102,6 +106,10 @@ func initMetrics(mf monitoring.MetricFactory) {
 		"interceptor_request_denied_count",
 		"Number of requests by denied, labeled according to the reason for denial",
 		"reason", monitoring.TreeIDLabel, "quota_user")
+	contextErrCounter = mf.NewCounter(
+		"context_err_counter",
```
codingllama: "interceptor_context_err_count", so it has a clearer scope and the same suffix as other metrics?

Do you think this is an interesting metric to track? Won't is just show up as cancelled RPCs anyway?

server/interceptor/interceptor.go
```diff
@@ -141,6 +149,10 @@ func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (conte
 		incRequestDeniedCounter(badInfoReason, 0, quotaUser)
 		return ctx, err
 	}
+	if err = ctx.Err(); err != nil {
```
codingllama: FYI, this is all in-memory processing. If the ctx is expired here, chances are it was already expired (or the deadline is way too low)

server/interceptor/interceptor_test.go
```diff
@@ -68,6 +68,7 @@ func TestTrillianInterceptor_TreeInterception(t *testing.T) {
 		handlerErr error
 		wantErr    bool
 		wantTree   *trillian.Tree
+		cancelled  bool
```
codingllama: nit: I think a `ctx` parameter would be better/more flexible than `cancelled`. We can have a pre-created cancelled ctx for this test.

server/interceptor/interceptor_test.go
```diff
@@ -108,13 +109,26 @@ func TestTrillianInterceptor_TreeInterception(t *testing.T) {
...
 	}
 
 	ctx := context.Background()
 	intercept := New(admin, quota.Noop(), false /* quotaDryRun */, nil /* mf */)
 	for _, test := range tests {
 		handler := &fakeHandler{resp: "handler response", err: test.handlerErr}
 
+		if test.cancelled {
+			// Use a context that's already been cancelled
+			newCtx, cf := context.WithTimeout(ctx, time.Second)
```
codingllama: Use `context.WithCancel()`, so you don't need to specify a timeout.

nit: call the func `cancel`.

server/interceptor/interceptor.go
Martin2112: The log message about "read tree info" was confusing. I'll probably remove this check then.

server/interceptor/interceptor.go
Martin2112: Renamed.

At least for the moment I'd like to be able to see which stage is slow and try to correlate it with the existing rpc client metrics.

server/interceptor/interceptor.go
Martin2112: Removed.

server/interceptor/interceptor_test.go
Martin2112: Not sure that's correct. We'd then have to manage not creating / cancelling the other test contexts. If you want to look at refactoring the interceptor code to make it easier to test cancellation then that's another option.

server/interceptor/interceptor_test.go
Martin2112: Renamed.

using WithCancel changes the semantics of existing tests. I don't think it should be done in this PR unless it flakes as a result of respecting deadlines now.

codingllama: It doesn't change the semantics, the end is a cancelled context anyway. Using `WithTimeout()` immediately followed by `cancel()` is the same as using `WithCancel()`, only the former has more arguments.

Martin2112: I've read the code again and you're right.

PR #885, opened 2017-10-31
Docker and Kubernetes config files

gdbelvin:
I didn't see a Docker file for spinning up a mysql server and configuring the database correctly.
Key Transparency has such a docker file, but I think it really belongs here.


Martin2112:
OK we can take another look at the CT part later.

Issue #456, opened 2017-03-21
Proposal: Remove unit test dependency from MySQL


codingllama:
Currently, unit tests require a local MySQL to be running and have a root user without a password. While the presence of MySQL is paramount for integration-like tests, unit tests shouldn't have such requirement. That precludes users from doing a simple `go test ./...` and forces a specific local configuration.

I've had a quick look at sqlite in memory as substitute for unit tests and it looks viable.

A course of action could be:

1. Make tests rely on [GetTestDB](https://github.com/google/trillian/blob/master/testonly/integration/db.go#L35) instead of opening connections directly
1. Replace the DB with `sqlite3/:memory:`
1. ?
1. Profit

From what I've seen, there a few changes we'd have to make for it to work:
* Remove '#' comments from storage.sql
* Declare indexes separately from table definitions
* Remove `SET GLOBAL sql_mode`
* Find a way around enums

A simple solution would be to have separate-but-equivalent schemas. That's not entirely desirable due to duplication, but it does fix a whole category of issues. Another venue is to "reinterpret" the script at `GetTestDB` and remove / change the problematic lines.

There are a few other options, such as writing our own fake storage implementation, but that doesn't solve the issue of storage layer tests.

WDYT?

gdbelvin:
I was _just_ [working](https://github.com/gdbelvin/trillian/tree/test ) on something like this, but abandoned it for the moment. 

In addition to the issues you've identified, I also identified the following:
- No support for SQL ENUMS
- Unique Indexes need to be declared outside tables

This would make it much easier to test Trillian with other projects. +1 

For what it's worth we could supply a command line flag so the travis tests could still run using proper MySQL to make sure we haven't missed anything.

codingllama:
Yep, I was expecting you to pitch along. :)
@pphaneuf may be interested as well, as we were talking about this earlier today.

I have an [experimental branch](https://github.com/google/trillian/compare/master...codingllama:sqlite-experiment) with a bunch of similar changes to yours, although it's nowhere close to a complete solution.

I like the idea of a flag for Travis. We could run unit tests using both in-memory and MySQL to be extra safe. There's also less reason not to do this if we have that.

Martin2112:
I did have a fake storage implementation back at the start of the project but it was too much work to maintain with things in flux.

I'd like to defer this until we've settled the schema a bit more. Note that there are some more issues with parameter ordering that might not affect SQLite but do affect sql variants with positional parameters like Postgres.

[...]

PR #886, opened 2017-11-02
Use sqlite3 as the default driver for unit tests

codingllama:
The testdb package adds utilities to create test databases for different 
drivers. Currently, it supports sqlite3 (default) and mysql. The default driver 
can be specified by the TRILLIAN_SQL_DRIVER environment variable, allowing tests
to transparently execute against different databases/drivers.

Adjustments done to existing code:

* storage.sql declares unique constraints separately from the table declaration
  (functionally equivalent, but more portable)
* A few constants in query strings are now quoted (eg, 'false' instead of 
  false, otherwise SQLite interprets the constant as a column name)
* Duplicate handling (isDuplicateErr) handles both MySQL and SQLite error codes
* map_storage.go no longer distinguishes between empty and nil mapper metadata
  (SQLite writes []byte{} as NULL). (This is a positive change in many ways, as 
  an empty Any proto is not valid due to having no type_url.)

A few test loops that defer-close resources had their scope reduced (using 
anonymous functions), as doing so greatly reduces test duration on SQLite.

Finally, integration.GetTestDB has been completely replaced by 
testdb.NewTrillianDB.


codingllama:
Fixes #456.

.travis.yml
```diff
@@ -80,11 +81,14 @@ script:
       if [[ "${WITH_ETCD}" == "true" ]]; then
         export ETCD_DIR="${GOPATH}/bin"
       fi
-  - ./integration/integration_test.sh
-  - cd "$HOME/gopath/src/github.com/google/certificate-transparency-go"
-  - ./trillian/integration/integration_test.sh
-  - cd $HOME/gopath/src/github.com/google/trillian
-  - HAMMER_OPTS="--operations=50" ./integration/maphammer.sh 3
+  - |
+      if [[ -n "${TRILLIAN_SQL_DRIVER}" && "${TRILLIAN_SQL_DRIVER}" != sqlite3 ]]; then
```
AlCutter: "sqlite3"?

AlCutter: or, just to capture an IRL chat, flip this to check for `mysql` since that what the scripts contained in the block are expecting?

codingllama: Flipped and quoted.

storage/mysql/log_storage_test.go
```diff
@@ -980,31 +981,39 @@ func TestGetUnsequencedCounts(t *testing.T) {
...
-		defer tx.Close()
-
-		got, err := tx.GetUnsequencedCounts(ctx)
-		if err != nil {
-			t.Errorf("GetUnsequencedCounts() = %v, want no error", err)
-		}
-		if diff := pretty.Compare(expectedCount, got); diff != "" {
-			t.Errorf("GetUnsequencedCounts() = diff -want +got:\n%s", diff)
-		}
+		func() {
```
AlCutter: same here?

storage/mysql/log_storage_test.go
```diff
@@ -980,31 +981,39 @@ func TestGetUnsequencedCounts(t *testing.T) {
...
-			leaves := createTestLeaves(numToAdd, expectedCount[logID])
-			if _, err := tx.QueueLeaves(ctx, leaves, fakeDequeueCutoffTime); err != nil {
-				t.Fatalf("Failed to queue leaves: %v", err)
-			}
-			commit(tx, t)
-			expectedCount[logID] += numToAdd
+			j, logID := j, logID
+			func() {
+				tx := beginLogTx(s, logID, t)
+				defer tx.Close()
```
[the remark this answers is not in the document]
AlCutter: One suggestion might be something like a `func runWithTX(s, logID, f func(tx) err) err` and you'd do here:
```go
for ... {
  err := runWithTX(s, logID, func f(tx ...) err {
      numToAdd ..
      ...
      if bad {
         return errors.New(...) 
      }
      ...
      return nil
    })
    if err != nil {
      t.Error...
    }
}
```

and `runWithTX` does the
```go
tx := beginLog...
defer tx.Close()
return f(tx)
```

storage/mysql/storage.sql
```diff
@@ -55,10 +55,11 @@ CREATE TABLE IF NOT EXISTS TreeHead(
   RootSignature        VARBINARY(1024) NOT NULL,
   TreeRevision         BIGINT,
   PRIMARY KEY(TreeId, TreeHeadTimestamp),
-  UNIQUE INDEX TreeRevisionIdx(TreeId, TreeRevision),
   FOREIGN KEY(TreeId) REFERENCES Trees(TreeId) ON DELETE CASCADE
 );
 
+CREATE UNIQUE INDEX TreeHeadRevisionIdx
```
AlCutter: (curiosity) is there a material difference between this and how it was declared before?

storage/testdb/testdb.go
```diff
@@ -0,0 +1,182 @@
...
+// NewTrillianDB creates an empty database with the Trillian schema. The database name is randomly
+// generated.
+// NewTrillianDB is equivalent to Default().NewTrillianDB(ctx).
+func NewTrillianDB(ctx context.Context) (*sql.DB, error) {
+	return Default().NewTrillianDB(ctx)
+}
+
+// Provider is an object capable of creating new test databases.
+type Provider struct {
+	Driver           string
```
AlCutter: Do these need comments too?

storage/testdb/testdb.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (p *Provider) sanitize(script string) string {
+	buf := &bytes.Buffer{}
+	for _, line := range strings.Split(string(script), "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" || line[0] == '#' || strings.Index(line, "--") == 0 {
+			continue // skip empty lines and comments
+		}
+		if p.Driver == sqliteDriver {
+			line = enumRegex.ReplaceAllString(line, "VARCHAR(50)")
```
AlCutter: :sob:

integration/maptest/map_test.go
```diff
@@ -17,9 +17,14 @@ package maptest
 import (
 	"context"
 	"testing"
+
+	"github.com/google/trillian/storage/testdb"
 )
 
 func TestMapIntegration(t *testing.T) {
+	if provider := testdb.Default(); !provider.IsMySQL() {
```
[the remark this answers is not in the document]
codingllama: I think this is OK, it's not that verbose and both the Skip and the message are clearly visible in the test.

storage/mysql/log_storage_test.go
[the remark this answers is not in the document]
codingllama: Yes, the funcs were because of the defer-close. I went with your first suggestion and explicitly closed the tx.

I have been meaning to add `runInTX` methods for a while now, but in a slightly different form. I'll save those to another PR.

storage/mysql/log_storage_test.go
codingllama: Done.

storage/mysql/storage.sql
codingllama: Apart from having an explicit name, I think it's the same.

storage/testdb/testdb.go
codingllama: I'll assume those are happy tears.  ;)

storage/testdb/testdb.go
codingllama: Done.

codingllama:
@AlCutter: all comments addressed, PTAL.

codingllama:
Thanks! Rebased on master, that should fix Travis errors.

m0r3n0cabr0n94 (approved):

Issue #828, opened 2017-09-12
Map: Sign initial map revision


gdbelvin:
Currently requests for map revision 0 fail because the first opportunity to set a signed map head happens at revision 1. 

The admin API should create an initial empty tree head like the log does. This would remove a set of special cases around 0 for clients.

phad:
There are a few inconsistent behaviors related to the state of the map when newly created.

I captured a transcript talking to the map server to show these.
https://gist.github.com/phad/e201686c112928546b1f2bc6065f0074

tl;dr

* Newly created Map is implicitly at revision 0
* Map at revision 0 is never retrievable via GetSignedMapRootByRevision
* Only when Map is newly created can we access its Map Root using GetSignedMapRoot
* Once leaves are set that rev 0 map root will never be seen again.

Orthogonal to this is the API for specifying 'give me the most recent version' when using GetLeaves.
There 'revision' must be set negative.

This conflicts with the behavior a reasonable client would expect, where simply omitting GetLeavesRequest.revision  (and so passing the default value '0') ought to give you the most recent data.

To resolve both of these I would propose:
* Ensure Map Root at rev 0 is never visible; the response to GetSignedMapRoot for a new, empty map should definitely not be 'map_root{}' as it is at present.
* Redefine GetLeaves such that the default value for revision (0) will get you the most recent leaves.

However there's an argument that, at revision 0, the map is empty and GetLeaves should return empty values for all keys; and GetSignedMapRoot should return the root hash of the empty tree.
Maybe that's important in order to make cryptographic proofs that a current (rev > 0) map state can be reached from the empty map?

In that case an alternative proposal is:
* Define Map Root at rev 0, and ensure it is returned by GetMapRoot when the map is new, and GetMapRootByRevision when revision=0 is specified
* Rename the current GetLeaves to GetLeavesByRevision, and make revision <0 illegal
* Define a new GetLeaves which implicity retrieves the most recent version (behaviour equivalent to current API with -ve revision)

gdbelvin:
Thanks for writing this up! 

Being able to query the map from a known-empty starting condition seems like a valuable property to preserve. Therefore, I think the second proposal is best. 

Incidentally, starting with a valid, signed SignedMapRoot at revision 0 would remove a number of edge cases like [this](https://github.com/google/keytransparency/blob/master/core/sequencer/sequencer.go#L137) where I test whether the map root has been signed, and then make an empty SetLeaves request if it isn't.

codingllama:
[...]

PR #890, opened 2017-11-13
Lazy init Map to write rev 0 Map Head on first use

phad:
Addresses issue #828.

Each Map API endpoint checks if the Map is init'd.  If not, it creates the revision 0 Signed Map Head.

Clearly this introduces some overhead into each call.  For now, it's a simple place to start.
It would be easy to add a `map[int64]bool` to `TrillianMapServer` to track which mapIDs have been init'd and then make an immediate exit from the new `func (s *TrillianMapServer) Init(mapID int64) error`

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -60,7 +60,7 @@ func (t *TrillianMapServer) Init(ctx context.Context, mapID int64) error {
 
 	tx, err := t.registry.MapStorage.BeginForTree(ctx, mapID)
 	if err != storage.ErrMapNeedsInit && err != nil {
-			return err
+		return err
 	}
 	defer tx.Close()
 
@@ -75,7 +75,7 @@ func (t *TrillianMapServer) Init(ctx context.Context, mapID int64) error {
 	smtWriter, err := merkle.NewSparseMerkleTreeWriter(
 		ctx,
 		mapID,
-		0 /* write revision */,
+		0, /* write revision */
 		hasher, func() (storage.TreeTX, error) {
 			ttx, err := t.registry.MapStorage.BeginForTree(ctx, mapID)
 			if err == storage.ErrMapNeedsInit {
@@ -267,7 +267,7 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 		return nil, fmt.Errorf("CalculateRoot(): %v", err)
 	}
 
-	newRoot, err := t.makeSignedMapRoot(ctx, tree, time.Now(), rootHash,req.MapId, tx.WriteRevision(), req.Metadata)
+	newRoot, err := t.makeSignedMapRoot(ctx, tree, time.Now(), rootHash, req.MapId, tx.WriteRevision(), req.Metadata)
 	if err != nil {
 		return nil, fmt.Errorf("makeSignedMapRoot(): %v", err)
 	}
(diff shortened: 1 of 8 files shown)


gdbelvin:
The [log](https://github.com/google/trillian/blob/b108c9362db916cfdb4940ef065658a4f388e17c/log/sequencer.go#L267) and the map both lazy init their first SignedTreeHeads. This seems less than ideal. 
Could we come up with a strategy to create logs and maps with an initial, correct state directly from the admin API?  

cc @codingllama @Martin2112

phad:
Are you happy to proceed along the lines of this PR for the map, until such a time as we have an init tool for both logs and maps?

Martin2112:
I thought log provisioning did create the tree head now and that code was obsolete. I could be wrong.

server/map_rpc_server.go
```diff
@@ -49,11 +49,85 @@ func (t *TrillianMapServer) IsHealthy() error {
...
+// Init creates the initial revision 0 SignedMapHead, if one doesn't already exist.
+func (t *TrillianMapServer) Init(ctx context.Context, mapID int64) error {
+	tree, hasher, err := t.getTreeAndHasher(ctx, mapID, false /* readonly */)
+	if err != nil {
+		return err
+	}
+	ctx = trees.NewContext(ctx, tree)
+
+	tx, err := t.registry.MapStorage.BeginForTree(ctx, mapID)
+	if err != nil && err != storage.ErrMapNeedsInit {
```
gdbelvin: does `tx.Close()` need to be called when `err == storage.ErrMapNeedsInit` ?

server/map_rpc_server.go
```diff
@@ -49,11 +49,85 @@ func (t *TrillianMapServer) IsHealthy() error {
...
+	if err != nil {
+		return err
+	}
+
+	rootHash, err := smtWriter.CalculateRoot()
+	if err != nil {
+		return fmt.Errorf("CalculateRoot(): %v", err)
+	}
+	newRoot := trillian.SignedMapRoot{
+		TimestampNanos: time.Now().UnixNano(),  // TODO(phad): should this be the tree create timestamp?
```
gdbelvin: This is correct.  We want to timestamp when each SignedMapRoot is created. 
If we were to create this SignedMapRoot during admin.NewTree, then the timestamps would match.

server/map_rpc_server.go
```diff
@@ -49,11 +49,85 @@ func (t *TrillianMapServer) IsHealthy() error {
...
+		})
+	if err != nil {
+		return err
+	}
+
+	rootHash, err := smtWriter.CalculateRoot()
+	if err != nil {
+		return fmt.Errorf("CalculateRoot(): %v", err)
+	}
+	newRoot := trillian.SignedMapRoot{
```
gdbelvin: We should probably have only one function that creates SignedMapRoots in case we decide to change the format and miss the other one.

gdbelvin:
LGTM.  
The refactor sounds like a good idea. Were you thinking that would be a separate PR? 
Also, it looks like something upstream is causing travis to fail.

server/map_rpc_server.go
phad: It's there mainly for the `err==nil` case.  For the `ErrMapNeedsInit` case it's a no-op when you get as far as the Commit but is useful for the intervening `return err`s.

PR #891, opened 2017-11-14
Strengthen maptest

gdbelvin:
`NewMapEnv` is used as a library in several tests inside and outside of the Trillian repo.

This PR 
 - Adds a precondition check inside `NewMapEnv`.
 - Strengthens the map integration test so that it doesn't panic when invalid data is returned.
 - Removes the global `env` variable and moves it inside the integration test runner.


integration/maptest/map_test.go
```diff
@@ -16,16 +16,44 @@ package maptest
...
+		}
+		env = mapEnv
+		defer env.Close()
+	} else {
+		mapEnv, err := integration.NewMapEnvFromConn(*server)
+		if err != nil {
+			log.Fatalf("failed to get map client: %v", err)
+		}
+		env = mapEnv
+		defer env.Close()
```
phad: Not introduced here, but you could knock defer env.Close() out of the if ... else and avoid the repetition.

phad: Also, since we log.Fatalf() if err != nil in both branches, you might as well assign directly to 'env' and omit mapEnv?

testonly/integration/mapenv.go
```diff
@@ -64,6 +65,10 @@ func NewMapEnvFromConn(addr string) (*MapEnv, error) {
 
 // NewMapEnv creates a fresh DB, map server, and client.
 func NewMapEnv(ctx context.Context) (*MapEnv, error) {
+	if provider := testdb.Default(); !provider.IsMySQL() {
```
phad: Do you need to repeat the !IsMySQL check here, when you've done it in map_test.go?
There, you skip the test, and would not go on to execute this.
Or is mapenv.go used in other tests too?

integration/maptest/map_test.go
gdbelvin: Fixed.

testonly/integration/mapenv.go
gdbelvin: This check makes sense here because it verifies the preconditions necessary to use MapEnv. 
Individual tests may also choose to `t.Skip()`, but in terms of functional correctness, a check here seems sensible.   It also provides a nice error when the preconditions for `MapEnv` are not meet rather than spewing out db locked garbage.

`NewMapEnv` is used by KeyTransparency.

phad: OK, it's fine.

PR #892, opened 2017-11-15
[resetdb.sh] Quote password correctly when invoking mysql.

pavelkalinnikov:


scripts/resetdb.sh
```diff
@@ -34,7 +34,7 @@ collect_vars() {
   [[ ${VERBOSE} = 'true' ]] && echo "- Using MySQL Flags: ${FLAGS[@]}"
 
   # append password if supplied
-  [ -z ${DB_PASSWORD+x} ] || FLAGS+=("-p'${DB_PASSWORD}'")
+  [ -z ${DB_PASSWORD+x} ] || FLAGS+=(-p"${DB_PASSWORD}")
```
RJPercival: I don't think this will work if `$DB_PASSWORD` contains a space.

RJPercival: Having said that, the current way of doing it won't work if `$DB_PASSWORD` contains an apostraphe...

RJPercival: It looks like the correct way to do this is with `${DB_PASSWORD@Q}` (see the bottom of https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html), except that was only added in Bash 4.4 and we're using Bash 4.3 :(

pavelkalinnikov: The thing is, we don't need quotes in this place because below in "${FLAGS[@]}" the parameters get quoted again (which is the reason why the previous version of the script didn't work even for non space-containing passwords). This way, mysql gets '-pMY PASSWORD WITH SPACES' as a single parameter and successfully parses it.

Moreover, yesterday Gary and I observed that the new script works with space-containing passwords.

RJPercival: Ah ok, I'd missed that it re-quotes everything further down. LGTM.

codingllama: nit: Prefer using [[ instead of [

pavelkalinnikov: I can fix this and couple of other lines in a follow-up PR.

Issue #894, opened 2017-11-15
Disallow creation of log-type trees in map only servers (and vice-versa)


codingllama:
The Admin API is agnostic in terms of tree types, but Trillian deployments, in practice, are not. A Trillian server that means to run only one type of tree shouldn't accept create requests for other types.

(cc @gdbelvin)

PR #897, opened 2017-11-16
Disallow creation of trees whose type don't match the server

codingllama:
I.e., creating LOG-type trees in trillian_map_server and MAP-type trees
in trillian_log_server.


codingllama:
Closes #894.

PR #898, opened 2017-11-20
RunInAdmin* methods and common helpers

codingllama:
RunInAdminSnapshot and RunInAdminTX encapsulate the common
Begin()/defer Close()/Commit() pattern for Admin TX objects.

Admin helpers are given a "proper" place under the storage package and
replace the (ever increasing) population of helpers.

Both categories of helpers are added as standalone methods, instead of
AdminStorage members, so that different storage implementations may take
advantage of them without extra effort.


codingllama:
FYI @AlCutter: this addresses a long standing todo of mine and a suggestion you added in previous PR.

codingllama:
Thanks!

PR #900, opened 2017-11-22
Map get leaves by revision

phad:
Addresses #828 - Map API changes detailed at end of issue.
This PR adds a new Map API endpoint: `GetLeavesByRevision`.

* `GetLeavesByRevision` behaves similarly to the current `GetLeaves` endpoint, but does not accept a -ve revision number to indicate 'use latest revision'.

The map integration test and the maphammer are updated to use the new API.

In future work `GetLeavesRequest` will have its 'revision' field removed so it only retrieves the latest revision.


integration/maptest/map.go
```diff
@@ -219,7 +219,7 @@ func RunLeafHistory(ctx context.Context, t *testing.T, tadmin trillian.TrillianA
 		HashStrategy []trillian.HashStrategy
 		set          [][]*trillian.MapLeaf
 		get          []struct {
-			revision  int64
+			revision  uint64
```
gdbelvin: Even though the revision number will always be positive, it may be better to avoid a bunch of casting and just use int64 everywhere. The API should do initial bounds checking.

server/map_rpc_server_test.go.old
```diff
@@ -0,0 +1,197 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
gdbelvin: Is this file supposed to be updated or removed?

phad: It's not supposed to have been put in a commit, that's for sure!

phad:
@gdbelvin  PTAL

integration/maptest/map.go
```diff
@@ -277,25 +279,37 @@ func RunLeafHistory(ctx context.Context, t *testing.T, tadmin trillian.TrillianA
 
 			for _, batch := range tc.get {
 				indexes := [][]byte{batch.Index}
-				getResp, err := tmap.GetLeaves(ctx, &trillian.GetMapLeavesRequest{
+				getResp, err := tmap.GetLeavesByRevision(ctx, &trillian.GetMapLeavesByRevisionRequest{
 					MapId:    tree.TreeId,
 					Index:    indexes,
 					Revision: batch.revision,
 				})
+				switch batch.revision {
```
gdbelvin: I wonder if this should be a separate test really. 
TestHistory is verifying that generic set and get operations work correctly. This new bit is testing an edge case around GetLeaves?

server/map_rpc_server.go
```diff
@@ -112,8 +112,21 @@ func (t *TrillianMapServer) Init(ctx context.Context, mapID int64) error {
 // GetLeaves implements the GetLeaves RPC method.  Each requested index will
 // return an inclusion proof to either the leaf, or nil if the leaf does not
 // exist.
+// TODO(phad): this will soon change to only request the most recent version
+// of the specified leaf indices.
 func (t *TrillianMapServer) GetLeaves(ctx context.Context, req *trillian.GetMapLeavesRequest) (*trillian.GetMapLeavesResponse, error) {
-	mapID := req.MapId
+	return t.getLeavesByRevision(ctx, req.MapId, req.Index, req.Revision)
```
gdbelvin: Should `GetLeaves` call `GetLeavesByRevision` directly?  
It would benefit from the additional input validation in `GetLeavesByRevision`

integration/maptest/map.go
phad: Yep, happy to split it out.

Issue #908, opened 2017-12-04
LogServer: export merge delay metric


AlCutter:
It would be good to know how long the current merge delay is for entries.
We should have a metric which tracks this.

codingllama:
Thanks for raising.

Martin2112:
Isn't this implemented now?

AlCutter:
Yes!

PR #909, opened 2017-12-04
Add timestamps to LogLeaf.

AlCutter:
Store timestamps of when leaves were queued and integrated into the tree.


trillian_log_api.proto
```diff
@@ -55,6 +55,12 @@ message LogLeaf {
     // If leaf_identity_hash is empty, it's assumed to be the same as the
     // merkle_leaf_hash.
     bytes leaf_identity_hash = 5;
+    // queue_timestamp_nanos records the time at which this leaf was first passed to QueueLeaves.
+    // This value will be determined and set by the LogServer.
+    int64 queue_timestamp_nanos = 6;
```
codingllama: Prefer using google.protobuf.Timestamp to represent time.

(`import "google/protobuf/timestamp.proto"` / [link](https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto#L43))

storage/mysql/log_storage.go
```diff
@@ -40,29 +40,30 @@ import (
...
 			AND s.MerkleLeafHash IN (` + placeholderSQL + `) AND l.TreeId = ? AND s.TreeId = l.TreeId`
 	// TODO(drysdale): rework the code so the dummy hash isn't needed (e.g. this assumes hash size is 32)
-	dummyMerkleLeafHash = "00000000000000000000000000000000"
+	dummyMerkleLeafHash     = "00000000000000000000000000000000"
+	dummyIntegrateTimestamp = "0"
 	// This statement returns a dummy Merkle leaf hash value (which must be
 	// of the right size) so that its signature matches that of the other
 	// leaf-selection statements.
-	selectLeavesByLeafIdentityHashSQL = `SELECT '` + dummyMerkleLeafHash + `',l.LeafIdentityHash,l.LeafValue,-1,l.ExtraData
+	selectLeavesByLeafIdentityHashSQL = `SELECT '` + dummyMerkleLeafHash + `',l.LeafIdentityHash,l.LeafValue,-1,l.ExtraData, l.QueueTimestampNanos, ` + dummyIntegrateTimestamp + `
```
codingllama: Why did you use a dummy timestamp here? Could we use the real one (or add a comment)?

storage/mysql/log_storage.go
```diff
@@ -292,9 +293,9 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	for rows.Next() {
 		var leafIDHash []byte
 		var merkleHash []byte
-		var meta dequeueMeta
+		var queueTimestamp int64
```
codingllama: QueueTimestamp is a new column, so I suppose it must be nullable (because backwards compatibility). In this case you should be using sql.NullInt64 as the type.

Incidentally, could you make sure we have tests that dequeue leaves with null timestamps?

storage/mysql/log_storage.go
```diff
@@ -348,6 +350,7 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
 		if len(leaf.LeafIdentityHash) != t.hashSizeBytes {
 			return nil, fmt.Errorf("queued leaf must have a leaf ID hash of length %d", t.hashSizeBytes)
 		}
+		leaf.QueueTimestampNanos = queueTimestamp.UnixNano()
```
codingllama: FYI, you'll probably need ptypes.TimestampProto here. (See comment on proto file below.)

https://godoc.org/github.com/golang/protobuf/ptypes#TimestampProto

storage/mysql/log_storage_test.go
```diff
@@ -60,18 +60,20 @@ const sequenceNumber int64 = 237
 // no locks afterwards.
 
 func createFakeLeaf(ctx context.Context, db *sql.DB, logID int64, rawHash, hash, data, extraData []byte, seq int64, t *testing.T) *trillian.LogLeaf {
-	_, err := db.ExecContext(ctx, "INSERT INTO LeafData(TreeId, LeafIdentityHash, LeafValue, ExtraData) VALUES(?,?,?,?)", logID, rawHash, data, extraData)
-	_, err2 := db.ExecContext(ctx, "INSERT INTO SequencedLeafData(TreeId, SequenceNumber, LeafIdentityHash, MerkleLeafHash) VALUES(?,?,?,?)", logID, seq, rawHash, hash)
+	ts := fakeQueueTime.UnixNano()
```
codingllama: nit: Use a different fake time for queue and integrate, probably with integrate > queue, so it's a bit more realistic.

storage/mysql/log_storage_test.go
```diff
@@ -60,18 +60,20 @@ const sequenceNumber int64 = 237
...
+	_, err := db.ExecContext(ctx, "INSERT INTO LeafData(TreeId, LeafIdentityHash, LeafValue, ExtraData, QueueTimestampNanos) VALUES(?,?,?,?,?)", logID, rawHash, data, extraData, ts)
+	_, err2 := db.ExecContext(ctx, "INSERT INTO SequencedLeafData(TreeId, SequenceNumber, LeafIdentityHash, MerkleLeafHash, IntegrateTimestampNanos) VALUES(?,?,?,?,?)", logID, seq, rawHash, hash, ts)
 
 	if err != nil || err2 != nil {
```
codingllama: nit: This is a bit lazy  :P

storage/mysql/log_storage_test.go
```diff
@@ -399,6 +401,7 @@ func TestDequeueLeaves(t *testing.T) {
 			t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
 		}
 		ensureAllLeavesDistinct(leaves2, t)
+		ensureLeavesHaveQueueTimestamp(t, leaves2, fakeDequeueCutoffTime)
```
codingllama: Suggestion: since this is a new behavior we're adding I think it's worth having a separate test, instead appending the assert to all tests. We can use it to assert that dequeue works both with null and non-null timestamps. If the timestamp logic breaks then all tests here break; usually it's clearer to have a single, focused test break than the entire suite.

storage/mysql/log_storage_test.go
```diff
@@ -1117,6 +1125,15 @@ func ensureAllLeavesDistinct(leaves []*trillian.LogLeaf, t *testing.T) {
 	}
 }
```
codingllama: I see only changes to Dequeue-based tests. We should test that we're getting the correct timestamps from GetLeaves* methods too.

storage/mysql/storage.sql
```diff
@@ -83,6 +83,8 @@ CREATE TABLE IF NOT EXISTS LeafData(
   -- This is extra data that the application can associate with the leaf should it wish to.
   -- This data is not included in signing and hashing.
   ExtraData            LONGBLOB,
+  -- The timestamp from when this leaf data was first queued for inclusion.
+  QueueTimestampNanos  BIGINT NOT NULL,
```
codingllama: The NOT NULL would be ideal, but given the problems it might cause with existing installations I think we should drop it.

storage/mysql/storage.sql
```diff
@@ -103,6 +105,7 @@ CREATE TABLE IF NOT EXISTS SequencedLeafData(
   -- This is a MerkleLeafHash as defined by the treehasher that the log uses. For example for
   -- CT this hash will include the leaf prefix byte as well as the leaf data.
   MerkleLeafHash       VARBINARY(255) NOT NULL,
+  IntegrateTimestampNanos BIGINT NOT NULL,
```
codingllama: Ditto.

storage/mysql/log_storage.go
AlCutter: This query is only over `LeafData`, so doesn't have the `IntegrateTimestampNanos` available.  I could `JOIN` it like the other queries, but this is used by `QueueLeaves` and I'm not sure what the performance implications would be.

storage/mysql/log_storage.go
AlCutter: Actually no, you can create new `NOT NULL` columns, and those cells will take the default value of the type.

storage/mysql/log_storage_test.go
AlCutter: not mine!

storage/mysql/storage.sql
AlCutter: I don't think it causes problems because of the implicit default value...

storage/mysql/storage.sql
AlCutter: ditto :)

trillian_log_api.proto
AlCutter: Could you outline the benefit?
We're storing nanos since epoch, which with 63 bits is fine up until 2262-04-11 23:47:16, but using an embedded timestamp message has the cost of extra complexity if/when we decide we want to include the `QueueTimestampNanos` under the SET signature.

storage/mysql/log_storage.go
AlCutter: Hmm, thinking about it a bit more, maybe I should just take the hit - if we return a dupe leaf then it might be useful to have the timestamps there too, at least for consistency.
We can reconsider if there's a high performance cost.

storage/mysql/log_storage_test.go
AlCutter: Good call, done!

storage/mysql/log_storage_test.go
AlCutter: done

storage/mysql/log_storage_test.go
AlCutter: Done.

trillian_log_api.proto
codingllama: OK, I'll bite. A few benefits are:

* Using a proper type to represent time instead of a primitive, which:
    * Encourages callers to handle it correctly
    * Removes ambiguity in regards to time unit
    * Just all-around good design advice
* Consistency with our own APIs ([link](https://github.com/google/trillian/blob/3f6222aa42face2c2d81a24de5438d95e4f4bb3d/trillian.proto#L161))
* Consistency with the Google API Style ([link](https://cloud.google.com/apis/design/standard_fields))

You can still get the nanos for the SET, so that's not really an issue. If you want to go down the "using a proper type to represent time adds too much complexity" route I won't try to stray you away from it. I can assure you, though, that an API reviewer won't buy this either.

storage/mysql/log_storage.go
codingllama: I'm happy with a comment, but if you want to try the join that sounds good to me too.

storage/mysql/log_storage.go
codingllama: Are you sure about the not null? I recall problems with that, but maybe I have a different DB in mind.

Either way, it may be better to distinguish null from zero, though, so instead of returning QueueTime/IntegrateTime = epoch for "old" leaves we can return nothing.

storage/mysql/log_storage_test.go
codingllama: It is yours. It is mine. It belongs to all of us. We're one.  :P

trillian_log_api.proto
AlCutter: On consistency, I'll point out [this](https://github.com/google/trillian/blob/master/trillian.proto#L187) which existed _long_ before `Tree.CreateTime` :)

> You can still get the nanos for the SET, so that's not really an issue. 

Well, there _is_ one issue: 63 bits of seconds + 31 bits of nanoseconds can represent a much _much_ bigger timestamp than 63 bits of nanos can, so we'll have to be very careful about what we define the serialization format for SET and STH to be if we were ever going to use `Timestamp` anywhere other than as a read-only representation...

codingllama: Were you aware of Timestamp when you defined SignedLogRoot? I'd be happy to update it to use Timestamp, but I figured that might get strong push back. 

Re Timestamp's precision, although the type definition allows for a "larger" representation, the defining factor is still storage, so I don't think that changes much. We don't have to use all the precision it allows for. Plus, nanos-since-epoch should be good until "2262-04-11 23:47:16", right?

storage/mysql/log_storage.go
AlCutter: Seems to be the case (e.g. [this](https://stackoverflow.com/questions/22868345/mysql-add-a-not-null-column))

We can reasonably safely special case when the timestamps are zero (that's impossible given when we started writing Trillian, so we know that it's just old data), but gain the advantage of it being impossible to create more leaves with no timestamps.

AlCutter:
PTAL

server/sequencer_manager_test.go
```diff
@@ -86,6 +89,14 @@ var zeroDuration = 0 * time.Second
 
 const writeRev = int64(24)
 
+func mustToTimestampProto(t time.Time) *tspb.Timestamp {
```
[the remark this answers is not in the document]
AlCutter: done

codingllama:
Issue #908.

PR #910, opened 2017-12-04
No more sudo, install softhsm with apt.

AlCutter:


.travis.yml
```diff
@@ -1,6 +1,11 @@
-sudo: required # required to install softhsm for PKCS 11 tests
+sudo: false
```
codingllama: Is this the same as not declaring "sudo" at all? Could we just remove the line?

AlCutter: But then we couldn't relish the glory of defeating the `sudo` monster each time we looked inside this  file...

PR #913, opened 2017-12-05
Fix nil dereference

gdbelvin:


integration/maptest/map.go
```diff
@@ -434,7 +434,7 @@ func runMapBatchTest(ctx context.Context, t *testing.T, desc string, tmap trilli
 
 	// Ensure we're starting with an empty map
 	if err := isEmptyMap(ctx, tmap, tree); err != nil {
-		t.Errorf("%s: isEmptyMap() err=%v want nil", desc, err)
+		t.Fatalf("%s: isEmptyMap() err=%v want nil", desc, err)
```
phad: Why the Errorf -> Fatalf changes?
Generally one should use Errorf unless the failure is in critical test setup and there's no point continuing.

gdbelvin: The condition being tested here validate assumptions made later on in the test.  If we think we are starting with an empty map, but we are not there is no point in continuing. Errors further on will be spurious.

PR #914, opened 2017-12-05
Try to speed up travis runs.

AlCutter:
- cache test deps by `go install`ing them
- only run tests individually where we require coverage reports.

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -103,6 +103,7 @@ main() {
     echo 'installed test deps'
     go test -i ./...
 
+    if [[ ${coverage} -eq 1 ]]; then
     # Individual package profiles are written to "$profile.out" files under
     # /tmp/trillian_profile.
     # An aggregate profile is created at /tmp/coverage.txt.
@@ -110,19 +111,17 @@ main() {
     rm -f /tmp/trillian_profile/*
 
     for d in $(go list ./...); do
-      # Create a different -coverprofile for each test (if enabled)
+      # Create a different -coverprofile for each test
       local coverflags=
-      if [[ ${coverage} -eq 1 ]]; then
-        # Transform $d to a smaller, valid file name.
-        # For example:
-        # * github.com/google/trillian becomes trillian.out
-        # * github.com/google/trillian/cmd/createtree/keys becomes
-        #   trillian-cmd-createtree-keys.out
-        local profile="${d}.out"
-        profile="${profile#github.com/*/}"
-        profile="${profile//\//-}"
-        coverflags="-covermode=atomic -coverprofile='/tmp/trillian_profile/${profile}'"
-      fi
+      # Transform $d to a smaller, valid file name.
+      # For example:
+      # * github.com/google/trillian becomes trillian.out
+      # * github.com/google/trillian/cmd/createtree/keys becomes
+      #   trillian-cmd-createtree-keys.out
+      local profile="${d}.out"
+      profile="${profile#github.com/*/}"
+      profile="${profile//\//-}"
+      coverflags="-covermode=atomic -coverprofile='/tmp/trillian_profile/${profile}'"
 
       # Do not run go test in the loop, instead echo it so we can use xargs to
       # add some parallelism.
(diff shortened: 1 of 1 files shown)


scripts/presubmit.sh
```diff
@@ -97,6 +97,11 @@ main() {
     go build ${goflags} ./...
 
     echo 'running go test'
+    # Install test deps so that individual test runs below can reuse them.
+    echo 'getting test deps'
+    go get -t
```
RJPercival: Without "./...", this will only do so for the "trillian" package.

AlCutter: Done

AlCutter:
PTAL

scripts/presubmit.sh
```diff
@@ -128,8 +132,14 @@ main() {
           ${goflags} "$d"
     done | xargs -I '{}' -P ${GO_TEST_PARALLELISM:-10} bash -c '{}'
 
-    [[ ${coverage} -eq 1 ]] && \
-      cat /tmp/trillian_profile/*.out > /tmp/coverage.txt
+    cat /tmp/trillian_profile/*.out > /tmp/coverage.txt
+  fi
+  else
+      echo go test \
```
RJPercival: While this will indeed be fast, that's because it won't actually run the tests. Need to delete the "echo" at the start.

AlCutter: hehe, whoops!

scripts/presubmit.sh
```diff
@@ -128,8 +132,14 @@ main() {
           ${goflags} "$d"
     done | xargs -I '{}' -P ${GO_TEST_PARALLELISM:-10} bash -c '{}'
 
-    [[ ${coverage} -eq 1 ]] && \
-      cat /tmp/trillian_profile/*.out > /tmp/coverage.txt
+    cat /tmp/trillian_profile/*.out > /tmp/coverage.txt
+  fi
```
RJPercival: Shouldn't be a `fi` right before an `else`.

scripts/presubmit.sh
```diff
@@ -128,8 +132,14 @@ main() {
...
+  else
+      go test \
+        -short \
+        -timeout=${GO_TEST_TIMEOUT:-5m} \
+        ${coverflags} \
```
RJPercival: Could remove `${coverflags}`, since it'll always be empty.

scripts/presubmit.sh
```diff
@@ -97,27 +97,31 @@ main() {
...
     # Individual package profiles are written to "$profile.out" files under
     # /tmp/trillian_profile.
     # An aggregate profile is created at /tmp/coverage.txt.
     mkdir -p /tmp/trillian_profile
     rm -f /tmp/trillian_profile/*
 
     for d in $(go list ./...); do
-      # Create a different -coverprofile for each test (if enabled)
+      # Create a different -coverprofile for each test
       local coverflags=
```
RJPercival: Can delete this line, and just add `local` to line 124.

scripts/presubmit.sh
```diff
@@ -97,27 +97,31 @@ main() {
     go build ${goflags} ./...
 
     echo 'running go test'
+    # Install test deps so that individual test runs below can reuse them.
+    echo 'getting test deps'
+    go get -t
+    echo 'installed test deps'
+    go test -i ./...
 
+    if [[ ${coverage} -eq 1 ]]; then
```
RJPercival: Indentation is a bit funky (though it does make the diff a bit clearer).

scripts/presubmit.sh
```diff
@@ -97,27 +97,31 @@ main() {
     go build ${goflags} ./...
 
     echo 'running go test'
+    # Install test deps so that individual test runs below can reuse them.
+    echo 'getting test deps'
+    go get -t
```
RJPercival: Need "./..." for `go get -t` as well as `go test -i`.

scripts/presubmit.sh
AlCutter: done

scripts/presubmit.sh
AlCutter: whoops, done

scripts/presubmit.sh
AlCutter: done

scripts/presubmit.sh
AlCutter: I think the go get ./... is actually already done in .travis.yml, so I've removed it.

scripts/presubmit.sh
AlCutter: Fixed, I think

scripts/presubmit.sh
```diff
@@ -97,17 +97,19 @@ main() {
...
-    # An aggregate profile is created at /tmp/coverage.txt.
-    mkdir -p /tmp/trillian_profile
-    rm -f /tmp/trillian_profile/*
-
-    for d in $(go list ./...); do
-      # Create a different -coverprofile for each test (if enabled)
-      local coverflags=
-      if [[ ${coverage} -eq 1 ]]; then
+    # Install test deps so that individual test runs below can reuse them.
+    echo 'installed test deps'
```
RJPercival: s/installed/installing/

AlCutter: done

PR #917, opened 2017-12-06
Install the correct (race/norace) versions of test dependencies

gdbelvin:
Building go packages with -race takes some time.
If we build them correctly we get a big speed up (16 min down from 24)


AlCutter:
LGTM, thanks!

PR #921, opened 2017-12-12
Measure observed merge delay and export to metric

AlCutter:


log/sequencer.go
```diff
@@ -189,10 +191,22 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []*trilli
 		}
 		// The leaf has now been sequenced.
 		leaves[i].LeafIndex = seq
```
codingllama: nit: Use `leaf` instead of `leaves[i]`. Same throughout.

log/sequencer.go
```diff
@@ -189,10 +191,22 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []*trilli
...
+		leaves[i].IntegrateTimestamp, err = ptypes.TimestampProto(integrateTS)
 		if err != nil {
 			return nil, nil, fmt.Errorf("got invalid integrate timestamp: %v", err)
 		}
+
+		// Old leaves might not have a QueueTimestamp, only calculate the merge delay if it does.
+		if leaves[i].QueueTimestamp != nil && leaves[i].QueueTimestamp.Seconds != 0 {
+			queueTS, err := ptypes.Timestamp(leaves[i].QueueTimestamp)
+			if err != nil {
+				return nil, nil, fmt.Errorf("got invalid queue timestamp: %v", queueTS)
```
codingllama: Do you want to log and proceed as if the leaf had a zero QueueTimestamp in this case, so we're a bit more resilient?

(This is a bit hypothetical anyway, so it might make no difference in practice.)

log/sequencer.go
```diff
@@ -189,10 +191,22 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []*trilli
...
+			}
+			mergeDelay := integrateTS.Sub(queueTS)
+			seqMergeDelay.Observe(mergeDelay.Seconds(), label)
+
```
codingllama: nit: vertical whitespace.

codingllama:
Issue #908.

log/sequencer.go
AlCutter: done

log/sequencer.go
AlCutter: Yeah, I went back and forward on this myself...
It's hard to see how this could fail, or if it does, then something is probably corrupt, so maybe it's best to bail and get operator intervention.

log/sequencer.go
AlCutter: done

log/sequencer.go
```diff
@@ -188,11 +190,23 @@ func (s Sequencer) sequenceLeaves(mt *merkle.CompactMerkleTree, leaves []*trilli
 			return nil, nil, err
 		}
 		// The leaf has now been sequenced.
-		leaves[i].LeafIndex = seq
-		leaves[i].IntegrateTimestamp, err = ptypes.TimestampProto(s.timeSource.Now())
+		leaf.LeafIndex = seq
+		integrateTS := s.timeSource.Now()
+		leafIntegrateTimestamp, err = ptypes.TimestampProto(integrateTS)
```
codingllama: This is not being assigned to the leaf anymore.

AlCutter: typo, fixed

Issue #428, opened 2017-03-13
Serve APIs over TLS?


gdbelvin:
`server/trillian_log_server` does not currently support TLS.  
This could be intentional because Trillian is designed to be a backend service. 
If it ever becomes publicly accessible, we should support gRPC over TLS and authentication.

Martin2112:
We should probably support TLS but it wasn't a priority in early development.

Martin2112:
Related to #477.

Martin2112:
I think this was implemented. Reopen if I'm wrong.

PR #927, opened 2017-12-28
Add TLS support to the Trillian log signer

maljub01:
Addresses #428

### DIFF
--- a/server/trillian_log_signer/main.go
+++ b/server/trillian_log_signer/main.go
@@ -48,6 +48,8 @@ import (
 var (
 	mySQLURI                 = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	httpEndpoint             = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP (host:port, empty means disabled)")
+	tlsCertFile              = flag.String("tls_cert_file", "", "Path to the TLS server certificate.")
+	tlsKeyFile               = flag.String("tls_key_file", "", "Path to the TLS server key.")
 	sequencerIntervalFlag    = flag.Duration("sequencer_interval", time.Second*10, "Time between each sequencing pass through all logs")
 	batchSizeFlag            = flag.Int("batch_size", 50, "Max number of leaves to process per batch")
 	numSeqFlag               = flag.Int("num_sequencers", 10, "Number of sequencer workers to run in parallel")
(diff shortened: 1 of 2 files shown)


server/trillian_log_signer/main.go
```diff
@@ -48,6 +48,8 @@ import (
 var (
 	mySQLURI                 = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	httpEndpoint             = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP (host:port, empty means disabled)")
+	tlsCertFile              = flag.String("tls_cert_file", "", "Path to the TLS server certificate.")
```
AlCutter: Perhaps add some text to say not configuring/setting and tlsKeyFile to empty results in a non-TLS server being run?

util/process.go
```diff
@@ -25,14 +25,21 @@ import (
...
-func StartHTTPServer(addr string) error {
+func StartHTTPServer(addr, certFile, keyFile string) error {
 	sock, err := net.Listen("tcp", addr)
 	if err != nil {
 		return err
 	}
 	go func() {
 		glog.Info("HTTP server starting")
-		http.Serve(sock, nil)
+		if certFile != "" || keyFile != "" {
```
AlCutter: Can this work without both bring present, or are you just wanting to interpret either flag being non-empty as operator intent to run TLS but let `http.ServeTLS` 
 handle this potential error case? (maybe add a comment to that effect if that's what you're after)

maljub01: Yes, that was my intention. Added a comment.

server/trillian_log_signer/main.go
maljub01: Done.

server/trillian_log_signer/main.go
```diff
@@ -48,8 +48,8 @@ import (
 var (
 	mySQLURI                 = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	httpEndpoint             = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP (host:port, empty means disabled)")
-	tlsCertFile              = flag.String("tls_cert_file", "", "Path to the TLS server certificate.")
-	tlsKeyFile               = flag.String("tls_key_file", "", "Path to the TLS server key.")
+	tlsCertFile              = flag.String("tls_cert_file", "", "Path to the TLS server certificate. If not set, the server will use unsecured connections.")
```
AlCutter: nit: "not set" -> "unset"

maljub01: Done

PR #928, opened 2017-12-28
Make integration tests log to stderr

maljub01:
This will make test output a little more verbose, but it will be very
useful for understanding what's happening when tests fail.

### DIFF
--- a/integration/log_integration_test.sh
+++ b/integration/log_integration_test.sh
@@ -34,7 +34,7 @@ set +e
 TRILLIAN_SQL_DRIVER=mysql go test ${GOFLAGS} \
   -run ".*LiveLog.*" \
   -timeout=${GO_TEST_TIMEOUT:-5m} \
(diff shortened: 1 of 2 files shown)


integration/log_integration_test.sh
```diff
@@ -34,7 +34,7 @@ set +e
 TRILLIAN_SQL_DRIVER=mysql go test ${GOFLAGS} \
   -run ".*LiveLog.*" \
   -timeout=${GO_TEST_TIMEOUT:-5m} \
-  ./ --log_rpc_server="${RPC_SERVER_1}" --treeid ${TEST_TREE_ID}
+  ./ --log_rpc_server="${RPC_SERVER_1}" --treeid ${TEST_TREE_ID} --alsologtostderr
```
AlCutter: Could you wrap this over multiple lines, one per flag, similar to how you did in the other file?

maljub01: Done.

PR #934, opened 2018-01-03
Warn that an error, which can be ignored, may be seen during shutdown.

benlaurie:


AlCutter:
Closing as I've just merged #937 which makes this redundant.

PR #938, opened 2018-01-03
Kill map integration test flake due to nil head

AlCutter:


phad:
How flaky is this?

AlCutter:
Reasonably flaky, you often get a nil pointer deref here :/

phad:
So the GetSignedMapRoot call returns a non-nil err, but a response r.MapRoot.MapRevision in which r.MapRoot is nil ?

AlCutter:
Or more likely `r` is nil, but yup.

phad:
Feels like a bug more than a race?  (or perhaps both)

AlCutter:
Could be, but given this only happens on Travis I suspect it's perhaps to do with slow IO or something and the crash is not helping with debugging it.

PR #939, opened 2018-01-03
Take it easy on Travis' limited I/O

AlCutter:


Martin2112:
Did we decide this was moot or still needed?

PR #949, opened 2018-01-22
Add inclusion proof test for an item that exists

gdbelvin:


merkle/map_verifier_test.go
```diff
@@ -43,14 +45,22 @@ func TestConiksHasherTestVectors(t *testing.T) {
...
+			desc:  "One item, empty proof",
 			index: []byte{0xb7, 0x57, 0x2d, 0xf6, 0xe1, 0x9, 0x1f, 0xc0, 0x6, 0x9e, 0x4, 0xbf, 0x80, 0x98, 0x75, 0x25, 0xe7, 0x7a, 0xc9, 0xa6, 0xc2, 0x94, 0xd2, 0x8d, 0xb7, 0xf4, 0xe3, 0x60, 0x25, 0x1d, 0x83, 0xbf},
 			proof: append(make([][]byte, 255), []byte{
 				92, 215, 13, 113, 97, 138, 214, 158, 13, 29, 227, 67, 236, 34, 215, 4, 76, 188, 79, 247, 149, 223, 227, 147, 86, 214, 90, 126, 192, 212, 113, 64,
 			}),
 			root:   []byte{0x2c, 0x27, 0x03, 0xe0, 0x34, 0xf4, 0x00, 0x2f, 0x94, 0x1d, 0xfc, 0xea, 0x7a, 0x4e, 0x16, 0x03, 0xee, 0x8b, 0x4e, 0xe3, 0x75, 0xbd, 0xf8, 0x72, 0x5e, 0xb8, 0xaf, 0x04, 0xbf, 0xa3, 0xd1, 0x56},
 			treeID: 2595744899657020594,
 		},
+		{
+			desc:   "One item, exists proof",
```
phad: Are these desc strings transposed?  This looks like an empty proof to me; the one above looks like an 'exists' (existing?) proof.

gdbelvin: The descriptions are correct. 

This one is a proof for a leaf that exists in a tree with only one item. As a result, all the neighbor nodes are nil. 

The one above is a proof for a nil leaf in a tree with only one other item. As a result, that one other item causes the top neighbor to be non-nil.

PR #955, opened 2018-01-29
Set ExtraData field

gdbelvin:
Some schemas have ExtraData as NOT NULL.
As a result, they can't be set back to allowing NULLs.
Fix: set some extra_data even if we don't have any.


Martin2112:
Which storage(s) is/are affected? Can we not change them so it will work for all clients?

gdbelvin:
Our internal storage system is the one affected.

gdbelvin:
Closing as an alternative solution was found

PR #961, opened 2018-01-31
Make Map initialisation be explicit.

AlCutter:
This removes a bunch of complexity and cost during normal operation of the Map, shifting the initialisation to a one-time step at the point of tree creation.


cmd/createtree/main.go
```diff
@@ -81,18 +81,41 @@ func createTree(ctx context.Context) (*trillian.Tree, error) {
...
 	client := trillian.NewTrillianAdminClient(conn)
+	var tree *trillian.Tree
 	for {
-		tree, err := client.CreateTree(ctx, req)
-		if err == nil {
-			return tree, nil
+		glog.Info("CreateTree...")
+		tree, err = client.CreateTree(ctx, req)
+		if err != nil {
+			if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
```
gdbelvin: Should this error be propagated back to the user rather than retrying internally?

cmd/createtree/main.go
```diff
@@ -81,18 +81,41 @@ func createTree(ctx context.Context) (*trillian.Tree, error) {
...
+			return nil, fmt.Errorf("failed to CreateTree(%+v): %T %v", req, err, err)
 		}
-		if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
-			glog.Errorf("Admin server unavailable, trying again: %v", err)
-			time.Sleep(100 * time.Millisecond)
-			continue
+		break
+	}
+
+	if tree.TreeType == trillian.TreeType_MAP {
```
gdbelvin: While we're at it... should we think about doing this for the Log here as well?

trillian_map_api.proto
```diff
@@ -111,4 +118,5 @@ service TrillianMap {
         get: "/v1beta1/maps/{map_id}/roots/{revision}"
       };
   }
+  rpc InitMap(InitMapRequest) returns(InitMapResponse) {}
```
gdbelvin: Maps are initialized during `CreateTree` already. 
Could you expand on the use case in mind?

cmd/createtree/main.go
```diff
@@ -81,18 +81,41 @@ func createTree(ctx context.Context) (*trillian.Tree, error) {
 	defer conn.Close()
 
 	client := trillian.NewTrillianAdminClient(conn)
+	var tree *trillian.Tree
 	for {
-		tree, err := client.CreateTree(ctx, req)
-		if err == nil {
-			return tree, nil
+		glog.Info("CreateTree...")
```
gdbelvin: I think the pattern we want here looks a little more like:
```
for {
   select {
      case <- ctx.Done():
            return ctx.Err()
      case <- time.After(backoff.Time())
              // Retry
   }
}
```

trillian_map_api.proto
AlCutter: That is not true :)
The Map is created by CreateTree, but the map code fails spectacularly if there isn't an initial rev 0 Map head, which is why there were a bunch of calls to `Init()` dotted throughout the RPC server inline with all the incoming requests.  Unfortunately, this leads to multiplication of the number of in-flight transactions, and also makes it impossible (or very hard) to refactor the storage implementation to support other transaction programmatic models (see #954).

cmd/createtree/main.go
AlCutter: Happy to modify the pre-existing pattern to a friendlier backoff like you propose, but I'd rather do that in a follow up to keep this delta as small as possible.

AlCutter:
It's not so much about a use-case for having this separate as reducing complexity and overhead for the 99% path.  A follow-on change which might help would be to move the `Init` stuff into the `AdminServer`, but that requires a bunch more work since it then needs access to a lot of the stuff the `MapServer` has (keys, map writing logic etc.).

`InitMap()` is safe to call multiple times (it'll only init the map if the map needs initialising), and of course Trillian APIs are only intended to be visible to personalities/operators anyway.

cmd/createtree/main.go
```diff
@@ -81,18 +81,41 @@ func createTree(ctx context.Context) (*trillian.Tree, error) {
...
+		break
+	}
+
+	if tree.TreeType == trillian.TreeType_MAP {
+		for {
+			glog.Infof("Initialising Map %x...", tree.TreeId)
+			// Now, if it's a Map, initialise it.
+			mc := trillian.NewTrillianMapClient(conn)
+			req := &trillian.InitMapRequest{MapId: tree.TreeId}
+			_, err := mc.InitMap(ctx, req)
```
gdbelvin: This line and the following can be collaped

AlCutter:
It'd be really nice to have this be part of the Admin API and happen transactionally within `CreateTree()`, but there are some difficulties with that currently since an `AdminTX` != `MapTX`.
I think the best we can do there is have the admin `CreateTree` perform the two TX and return the result, but the rough edge there is that the 2nd TX (the map init) TX fails and you, as the caller, have no way at all to complete the operation; that means you end up with potentially many half-initialised map trees cluttering up your DB with no way to know, or do anything, about them :/

The current implementation has a similar dual TX issue, but you, as the operator, have all the information you need to recover - you know the treeID, and if the init fails you know that and can retry.

I'll refactor the code from the `CreateTree` command out so that it can be reused, though.

cmd/createtree/main.go
```diff
@@ -81,18 +81,41 @@ func createTree(ctx context.Context) (*trillian.Tree, error) {
...
 		}
-		if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
-			glog.Errorf("Admin server unavailable, trying again: %v", err)
-			time.Sleep(100 * time.Millisecond)
-			continue
+		break
+	}
+
+	if tree.TreeType == trillian.TreeType_MAP {
+		for {
```
gdbelvin: Consider supporting cancellation and a max retry count in this block as well.

cmd/createtree/main.go
AlCutter: done

cmd/createtree/main.go
AlCutter: ack.

integration/maptest/map.go
```diff
@@ -132,14 +132,19 @@ func verifyGetMapLeavesResponse(getResp *trillian.GetMapLeavesResponse, indexes
 }
 
 // newTreeWithHasher is a test setup helper for creating new trees with a given hasher.
-func newTreeWithHasher(ctx context.Context, tadmin trillian.TrillianAdminClient, hashStrategy trillian.HashStrategy) (*trillian.Tree, hashers.MapHasher, error) {
+func newTreeWithHasher(ctx context.Context, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient, hashStrategy trillian.HashStrategy) (*trillian.Tree, hashers.MapHasher, error) {
 	treeParams := stestonly.MapTree
 	treeParams.HashStrategy = hashStrategy
 	tree, err := tadmin.CreateTree(ctx, &trillian.CreateTreeRequest{
```
gdbelvin: This combination of `CreateTree`, `InitMap` could be pulled out into a general `CreateAndInitMap` function

integration/maptest/map.go
```diff
@@ -163,24 +168,28 @@ func RunMapRevisionZero(ctx context.Context, t *testing.T, tadmin trillian.Trill
 		},
 	} {
 		for _, hashStrategy := range tc.hashStrategy {
```
gdbelvin: This pattern of testing is slightly easier if we put it in the following form: 
```
t.Run(tc.desc, func (t testing.T) {
    ...
    if err != nil {
         t.Fatalf()
    }
})
```

This has the following nice benefits: 
- Individual test cases can be isolated when running with `go test ./integration/maptest -run RunMapRevisionZero/Desc`
- Test descriptions don't need to be included in every error message
- `t.Fatalf` and `return` can be used without stopping tests later on.

integration/maptest/map.go
```diff
@@ -244,16 +256,18 @@ func RunMapRevisionInvalid(ctx context.Context, t *testing.T, tadmin trillian.Tr
 		},
 	} {
 		for _, hashStrategy := range tc.HashStrategy {
```
gdbelvin: Consider using `t.Run` here. This would remove the need to introduce a Goto location definition.

integration/maptest/map.go
```diff
@@ -273,6 +288,7 @@ func RunMapRevisionInvalid(ctx context.Context, t *testing.T, tadmin trillian.Tr
 
 // RunLeafHistory performs checks on Trillian Map leaf updates under a variety of Hash Strategies.
 func RunLeafHistory(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient) {
+tests:
```
gdbelvin: Same here

integration/maptest/map.go
```diff
@@ -366,6 +385,7 @@ func RunLeafHistory(ctx context.Context, t *testing.T, tadmin trillian.TrillianA
 // RunInclusion performs checks on Trillian Map inclusion proofs after setting and getting leafs,
 // for a variety of hash strategies.
 func RunInclusion(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient) {
+tests:
```
gdbelvin: and here

server/interceptor/interceptor.go
```diff
@@ -334,12 +334,13 @@ func newRPCInfo(req interface{}, quotaUser string) (*rpcInfo, error) {
 		info.treeType = trillian.TreeType_MAP
 
 	// Map / readwrite
-	case *trillian.SetMapLeavesRequest:
+	case *trillian.SetMapLeavesRequest,
```
gdbelvin: wow, this is cool.  Didn't know you could do that!

server/map_rpc_server.go
```diff
@@ -393,3 +379,8 @@ func (t *TrillianMapServer) getTreeAndHasher(ctx context.Context, treeID int64,
 	}
 	return tree, th, nil
 }
+
+// InitMap implements the RPC Method of the same name.
+func (t *TrillianMapServer) InitMap(ctx context.Context, req *trillian.InitMapRequest) (*trillian.InitMapResponse, error) {
+	return &trillian.InitMapResponse{}, t.Init(ctx, req.MapId)
```
gdbelvin: Protobuf.empty might be useful as the response type here

trillian_map_api.proto
gdbelvin: `rpc InitMap(InitMapRequest) returns(protobuf.Empty) {
     post: "/v1beta1/maps/{map_id}:init
}`

cmd/createtree/testonly/fake_server.go
```diff
@@ -92,3 +97,12 @@ func (s *FakeAdminServer) CreateTree(ctx context.Context, req *trillian.CreateTr
...
 	return &resp, nil
 }
+
+// InitMap returns an error if s.InitErr is set, and an empty InitMapResponse
+// struct otherwise.
+func (s *FakeAdminMapServer) InitMap(ctx context.Context, req *trillian.InitMapRequest) (*trillian.InitMapResponse, error) {
+	if s.InitErr != nil {
+		return nil, s.InitErr
+	}
+	return &trillian.InitMapResponse{}, nil
```
gdbelvin: return nil, nil?

cmd/createtree/main.go
gdbelvin: Sounds good

cmd/createtree/testonly/fake_server.go
gdbelvin: Fair enough. Consider using the empty protobuf as a response though?

server/map_rpc_server.go
```diff
@@ -56,62 +56,42 @@ func (t *TrillianMapServer) IsHealthy() error {
...
 	if err != storage.ErrMapNeedsInit && err != nil {
-		return err
+		return nil, err
 	}
 	defer tx.Close()
 
 	if err == nil {
 		// Init() not needed.
-		return nil
+		return nil, nil
```
gdbelvin: When `Init` is called _after_ the log has already been initialized, should it return the SMR of revision 0 rather than nil? 

A description of how clients would use the SMR might help determine what a consistent response from `Init` should be.

server/map_rpc_server.go
```diff
@@ -393,3 +359,14 @@ func (t *TrillianMapServer) getTreeAndHasher(ctx context.Context, treeID int64,
 	}
 	return tree, th, nil
 }
+
+// InitMap implements the RPC Method of the same name.
+func (t *TrillianMapServer) InitMap(ctx context.Context, req *trillian.InitMapRequest) (*trillian.InitMapResponse, error) {
```
gdbelvin: It's not clear why there are two `Init` functions in the map server...

server/map_rpc_server.go
AlCutter: I don't think so, no - this way you can tell if whether or not the Init() did anything.
I'd expect that when a fresh SMR is created, the operator using the `CreateTree` tool might want to inspect/verify it - if the Map's already been initialised, then there's no need to do that again (or they can use the GetSignedMapRootAtRevision() to inspect it if they have some other reason).

server/map_rpc_server.go
AlCutter: For minimal diff, really - `Init()` was pre-existing, but I've collapsed it into the RPC method now.

gdbelvin: Consider either collapsing the two functions -- I don't think the diff size is going to be that bad, or maybe make the pre-existing `Init` function private? It would be nice to have a single point of entry for this function.

server/map_rpc_server.go
gdbelvin: If there's a use case for doing something special then this makes sense.  

It intuitively feels like it doesn't match typical REST API or singleton init semantics which always return a valid thing when called, or error when called twice. I'd worry a bit that callers would try to dereference the response if `err==nil` and result in a crash.

AlCutter: Good call - `InitMap()` now returns an AlreadyExists error if the map was already initialised.

server/map_rpc_server.go
AlCutter: Reread the previous comment :)

cmd/createtree/testonly/fake_server.go
```diff
@@ -26,25 +26,30 @@ import (
 	"google.golang.org/grpc"
 )
 
-// FakeAdminServer implements the TrillianAdminServer CreateTree RPC.
+// FakeAdminMapServer implements the TrillianAdminServer CreateTree, and
+// the TrillianMapServer InitMap RPCs.
 // The remaining RPCs are not implemented.
-type FakeAdminServer struct {
+type FakeAdminMapServer struct {
```
Martin2112: Why was this renamed btw? It's not just used by map tests.

PR #963, opened 2018-01-31
Make WaitForRootUpdate available for clients

gdbelvin:
Sometimes clients need to wait for new roots.

This commit also adds more robust timeout logic to the function as well.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -52,12 +52,6 @@ func New(logID int64, client trillian.TrillianLogClient, hasher hashers.LogHashe
 	}
 }
 
-// Root returns the last valid root seen by UpdateRoot.
-// Returns an empty SignedLogRoot if UpdateRoot has not been called.
-func (c *LogClient) Root() trillian.SignedLogRoot {
-	return c.root
-}
-
 // AddLeaf adds leaf to the append only log.
 // Blocks until it gets a verifiable response.
 func (c *LogClient) AddLeaf(ctx context.Context, data []byte) error {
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -109,9 +109,9 @@ func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*tri
 	return resp.Leaves, nil
 }
 
-// waitForRootUpdate repeatedly fetches the Root until the TreeSize changes
+// WaitForRootUpdate repeatedly fetches the Root until the TreeSize changes
 // or until ctx times out.
-func (c *LogClient) waitForRootUpdate(ctx context.Context) error {
+func (c *LogClient) WaitForRootUpdate(ctx context.Context) error {
```
AlCutter: Would it be handy for some callers to have the new root returned?

client/client.go
```diff
@@ -120,18 +120,24 @@ func (c *LogClient) waitForRootUpdate(ctx context.Context) error {
...
-		if err := ctx.Err(); err != nil {
+
+		select {
+		case <-ctx.Done():
 			return status.Errorf(codes.DeadlineExceeded,
-				"%v. TreeSize: %v, want > %v. Tried %v times.",
-				err, c.root.TreeSize, startTreeSize, i+1)
+				"%v. TreeSize: %v, want > %v. Tried %v times: %v",
+				err, c.root.TreeSize, startTreeSize, i+1, ctx.Err())
+		case <-time.After(b.Duration()):
```
AlCutter: Brain melting - what does this actually do? Do we get a new channel created when we enter the select{}, even if `<-ctx.Done()` is selected?

Might be clearer to write it with a `time.Ticker`:
```golang
tick := time.NewTicker(b.Duration())
...
select {
  ...
  case <-tick.C
}
```

gdbelvin: Fair question :-) 

`time.After` is basically on a one time ticker. It's super useful because it's a channel and we can put it into that select block.  The select block allows us to terminate early if the context is canceled or times out first rather than needing to wait for the full duration only to discover that we had timed out :-) 

Yes, the channel gets created when we enter the select.  This is the use case it's designed for. See https://golang.org/pkg/time/#After. https://golang.org/ref/spec#Select_statements

AlCutter: Okaaaay, but it's evil witchcraft to me :)

client/client.go
gdbelvin: Indeed it would!

gdbelvin:
Updated with better API interfaces (ones that return SignedLogRoot)

client/client.go
```diff
@@ -109,71 +103,86 @@ func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*tri
...
-		time.Sleep(b.Duration())
 	}
 }
 
-// UpdateRoot retrieves the current SignedLogRoot.
-// Verifies the signature, and the consistency proof if this is not the first root this client has seen.
-func (c *LogClient) UpdateRoot(ctx context.Context) error {
+// GetLatestRoot fetches and verifies the latest root against a trusted root, seen in the past.
+// Pass nil for trusted if this is the first time querying this log.
+func (c *LogClient) GetLatestRoot(ctx context.Context, trusted *trillian.SignedLogRoot) (*trillian.SignedLogRoot, error) {
```
AlCutter: The naming seems a little reversed here - it feels like:

```golang
func (c) GetLatestRoot(ctx) (*trillian.SignedLogRoot, error) {
  untrusted, err := fetchLatestAndVerifySigs()
  ...
  if err := c.UpdateRoot(untrusted); err != nil {
   return nil, err
  }
  return untrusted, nil
}
 
func (c) UpdateRoot(..., untrusted *trillian.SignedLogRoot) error {
   proof, err := fetchConsistencyProof()
   ...
   if err := verifyConsistencyProof(c.trusted, untrusted); err != nil {
     return err
   }
   if untrusted.timestamp > c.trusted.timestamp && untrusted.size >= c.trusted.size {
      c.trusted = untrusted
   }
   return nil
}

```

is a bit more natural?

That'd allow you to pass in your own set of untrusted roots to `UpdateRoot()` if you wanted too (I'm guessing that's why you've added the param?)

client/client.go
```diff
@@ -109,71 +103,86 @@ func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*tri
...
-		c.root = *resp.SignedLogRoot
 	}
-	return nil
+	return resp.SignedLogRoot, nil
+}
+
+// UpdateRoot retrieves the current SignedLogRoot, verifying it against roots this client has
+// seen in the past, and updating the currently trusted root if the new root verifies.
+func (c *LogClient) UpdateRoot(ctx context.Context) (*trillian.SignedLogRoot, error) {
+	newTrusted, err := c.GetLatestRoot(ctx, &c.root)
```
AlCutter: Thinking of clock-skew, caching, etc.; `newTrusted` could be older/smaller than the currently trusted root?

gdbelvin: Hmm. Good thinking. Yeah, especially if we happen to talk to a different cluster.  I'll add a check

client/client.go
gdbelvin: Hmm.. 
`GetLatestRoot` fetches and compares a new root to a previously trusted root. This is why you pass in a "currently" trusted root or nil to validate the new root against. 

`UpdateRoot` fetches and verifies the `LogClient`'s internal trusted root, verifying consistency with the last root the `LogClient` has seen.  `NewLogClient` supports supplying an initial trusted root, but from there on, we want to maintain a single chain of consistency proofs between the roots that the `LogClient` trusts.

client/client.go
```diff
@@ -177,12 +177,17 @@ func (c *LogClient) GetLatestRoot(ctx context.Context, trusted *trillian.SignedL
...
 	if err != nil {
 		return nil, err
 	}
-	c.root = *newTrusted
-	return newTrusted, nil
+	if newTrusted.TimestampNanos > currentlyTrusted.TimestampNanos &&
+		newTrusted.TreeSize >= currentlyTrusted.TreeSize {
+		c.root = *newTrusted
+	}
+	ret := c.root // copy the internal trusted root in order to not leak it.
```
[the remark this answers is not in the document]
gdbelvin: Not allow clients to modify

client/client.go
gdbelvin: Making `getLatestRoot` private sounds good to me.

gdbelvin: `Update` has been renamed to `UpdateRoot`

gdbelvin:
Unexported `GetLatestRoot`

PR #964, opened 2018-01-31
Introduce new TreeType for pre-ordered logs

pavelkalinnikov:
- The new type is not supported, as verified in admin_server_test.
- Run "go generate" to update the APIs.


storage/testonly/admin_storage_tester.go
```diff
@@ -113,6 +113,24 @@ var (
...
 	}()
 
+	// PreorderedLogTree is a valid, PREORDERED_LOG-type trillian.Tree for tests.
+	PreorderedLogTree = &trillian.Tree{
+		TreeState:          trillian.TreeState_ACTIVE,
+		TreeType:           trillian.TreeType_PREORDERED_LOG,
+		HashStrategy:       trillian.HashStrategy_RFC6962_SHA256,
+		HashAlgorithm:      spb.DigitallySigned_SHA256,
+		SignatureAlgorithm: spb.DigitallySigned_ECDSA,
+		DisplayName:        "Llamas Preordered Log",
```
phad: I think you can put your name in here if you want, now that Llama (Alan) has moved on from Trillian

pavelkalinnikov: Done.

PR #966, opened 2018-02-01
Improve description of ctmapper --source flag

RJPercival:
Make it clear in the description that this flag requires a URL.

### DIFF
--- a/examples/ct/ctmapper/mapper/mapper.go
+++ b/examples/ct/ctmapper/mapper/mapper.go
@@ -36,7 +36,7 @@ import (
 	ct "github.com/google/certificate-transparency-go"
 )
 
-var sourceLog = flag.String("source", "https://ct.googleapis.com/submariner", "Source CT Log")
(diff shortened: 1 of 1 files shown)


examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -36,7 +36,7 @@ import (
 	ct "github.com/google/certificate-transparency-go"
 )
 
-var sourceLog = flag.String("source", "https://ct.googleapis.com/submariner", "Source CT Log")
+var sourceLog = flag.String("source", "https://ct.googleapis.com/submariner", "URL of the CT Log")
```
Martin2112: should that still say "source CT log"?

RJPercival: There's no "destination" CT log, and the flag itself is called "source", so it seemed superfluous to have that in the description too.

Martin2112: OK.

Issue #968, opened 2018-02-01
Logs cannot be safely frozen at the moment


Martin2112:
Setting the frozen state blocks both queuing and signing leaving anything queued unable to be integrated. The proposed solution is:

* Expose metrics on integration rate / errors to allow inspection that queue is empty for a tree
* Add a new tree state that will allow signing but block queueing
* Various new tests and stuff that it works correctly

Will need a minor MySQL schema change to the tree states ENUM but this does not need to be done immediately.

Martin2112:
Metrics added in #957.

Martin2112:
This cannot be cleanly implemented without changes to the storage API. This is being worked on now.

Martin2112:
Though it involves api changes #1007 will be the basis of a cleaner implementation.

Martin2112:
Doc is in #1045. This should now be fixed.

Martin2112:
Now this is all merged I'll fix up the superfluous readonly flag in #1047.

PR #970, opened 2018-02-02
Add DRAINING state for trees. Add table driven operation type / state / tree type checks.

Martin2112:
Reject queue leaves in draining state but allow signing to continue. Add table driven validation rules to check operation type access by tree state and type. Update tests.


Martin2112:
Towards fixing #968.

Martin2112:
I think this is ready to review. Please check if I missed anything. I have some WIP to extend the integration tests.

server/log_rpc_server.go
```diff
@@ -98,6 +98,14 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
 	if err != nil {
 		return nil, err
 	}
+
+	// If we're draining the queue reject attempts to add new leaves. This is
+	// not the ideal place to enforce this restriction but it's currently
+	// the only really practical one.
```
daviddrysdale: Why not enforce it in the same place as `TreeState_FROZEN`?
 - server/interceptor/interceptor.go marks a `QueueLeaves` request as `readonly=false`
 - change `trees.GetTree` to then check for `DRAINING` as well as `FROZEN` -- either is effectively readonly from a gRPC API perspective.

trillian.proto
```diff
@@ -64,14 +64,19 @@ enum TreeState {
   ACTIVE = 1;
 
   // Frozen trees are only able to respond to read requests, writing to a frozen
-  // tree is forbidden.
+  // tree is forbidden. Trees should not be frozen when there are entries
+  // in the queue that have not yet been integrated.
```
daviddrysdale: maybe add a pointer to the other state: "... that have not yet been integrated (the DRAINING state covers this case)"

server/log_rpc_server.go
Martin2112: I tried this first and it doesn't work because the storage API only has a readonly flag and you don't have an options struct for GetTree. Fixing this up would be an intrusive change to public APIs.

daviddrysdale: How about this then:
 - add an extra `Internal bool // Tree will only be used for internal (non-API-driven) purposes` field to `trees.GetOpts`
 - set `Internal: true` in `SequencerManager.ExecutePass` (only)
 - in `trees.GetTree` allow `DRAINING` trees to be returned if `!opts.Readonly || opts.Internal`
Would that work?

Martin2112: I don't think so. GetTrees is called 3 times - by the interceptor, the code in the RPC server and via storage. The last of these is the problem because you've lost the context of what is being done. You only know if it's a read or write transaction.  This wasn't previously a problem because there was nothing selective about write permission as FROZEN denies everything.

daviddrysdale: OK, so the storage/*/*_storage.go invocations of `trees.GetTree` would all have to be `Internal: true` as well, and we'd be relying on the interceptor and RPC server policing the external boundaries correctly.

Martin2112: I think that's more or less it. GetOpts doesn't make it into storage. It makes a new one and stuffs the readonly flag in it so it can call back to GetTrees. There's no easy way to set internal for one operation only. We should probably have had an options struct on the storage code but it's a breaking change to add it now.

Martin2112:
Might be better to take the API hit now and do it properly. I'm working on doing this in #993.

trillian.proto
Martin2112: Done.

server/log_rpc_server.go
Martin2112: This has now been completely redone after the API changes were merged. All the code that handles state / access type validation is now in one place in trees.go and is much cleaner.

Martin2112:
PTAL.

trees/trees.go
```diff
@@ -34,6 +34,86 @@ import (
...
 type treeKey struct{}
 
+type accessRule struct {
+	// Tree states are accepted if there is a 'true' value for them in this map.
+	okStates map[trillian.TreeState]bool
+	// Allows the error code to be specified for specific rejected states.
+	rejectCodes map[trillian.TreeState]codes.Code
+	// Tree types are accepted if there is a 'true' value for them in this map.
+	okTypes map[trillian.TreeType]bool
+	//okWrite is true if write access should be allowed.
```
daviddrysdale: (nit) Missing space

trees/trees.go
```diff
@@ -34,6 +34,86 @@ import (
...
+	okWrite bool
+}
+
+// These rules define the permissible combinations of tree state and type
+// for each operation type.
```
daviddrysdale: This is neat -- all of the rules about what's allowed in a single table.

daviddrysdale:
I'm guessing you're planning this, but in case you're not...

It would be nice to have a follow-up PR that changes the script-driven integration test(s) to move logs to draining then frozen using the tool from your #956, so this stuff gets exercised regularly.

crypto/keyspb/keyspb.pb.go
```diff
@@ -76,9 +76,7 @@ func (m *Specification) String() string            { return proto.CompactTextStr
 func (*Specification) ProtoMessage()               {}
 func (*Specification) Descriptor() ([]byte, []int) { return fileDescriptor0, []int{0} }
 
-type isSpecification_Params interface {
-	isSpecification_Params()
-}
+type isSpecification_Params interface{ isSpecification_Params() }
```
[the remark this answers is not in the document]
Martin2112: Yes I just regenerated the protos. I'll back it out of this PR to avoid confusion.

Martin2112:
That's basically where I started before I found out freezing didn't really work properly. Once all this is merged I'm going to document the freeze process for log operators and then go back to testing.

storage/mysql/tree_storage.go
```diff
@@ -36,7 +36,7 @@ const (
 	insertSubtreeMultiSQL = `INSERT INTO Subtree(TreeId, SubtreeId, Nodes, SubtreeRevision) ` + placeholderSQL
 	insertTreeHeadSQL     = `INSERT INTO TreeHead(TreeId,TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature)
 		 VALUES(?,?,?,?,?,?)`
-	selectNonDeletedTreeIDByTypeAndStateSQL = "SELECT TreeId FROM Trees WHERE TreeType = ? AND TreeState = ? AND (Deleted IS NULL OR Deleted = 'false')"
+	selectNonDeletedTreeIDByTypeAndStateSQL = "SELECT TreeId FROM Trees WHERE TreeType = ? AND TreeState IN(?,?) AND (Deleted IS NULL OR Deleted = 'false')"
```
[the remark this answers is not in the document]
Martin2112: Moved it to the more logical place. I have a feeling we thought it might be reused by maps.

trees/trees.go
Martin2112: Sorry I can 't see it.

daviddrysdale: `// okWrite...` not `//okWrite..`

Martin2112: Aha, fixed.

trees/trees.go
```diff
@@ -34,6 +34,86 @@ import (
...
+	QueueLog: {
+		okStates: map[trillian.TreeState]bool{
+			trillian.TreeState_ACTIVE: true,
+		},
+		rejectCodes: map[trillian.TreeState]codes.Code{
+			trillian.TreeState_DRAINING: codes.PermissionDenied,
+		},
+		okTypes: map[trillian.TreeType]bool{
+			trillian.TreeType_LOG:            true,
+			trillian.TreeType_PREORDERED_LOG: true,
```
pavelkalinnikov: Isn't `QueueLog` only for `TreeType_LOG`?

trees/trees.go
```diff
@@ -34,6 +34,86 @@ import (
...
+		},
+		okWrite: true,
+	},
+	SequenceLog: {
+		okStates: map[trillian.TreeState]bool{
+			trillian.TreeState_ACTIVE:   true,
+			trillian.TreeState_DRAINING: true,
```
pavelkalinnikov: Does `DRAINING` make sense for `PREORDERED_LOG`? There is no queue in that mode.

However, I can imagine `DRAINING` mean "still has sequenced entries not integrated into the tree, hence the sequencer is still working". The caveat though is that there might be gaps in sequenced entries, so in this sense the `DRAINING` state will last forever.

Martin2112:
I wasn't sure exactly how preordered logs would end up so left these in
place for the moment. Could have called QueueLog something like SubmitWork
or AddLeaves but the internal APIs are all Queue<Something>.

If someone wants to TODO them that's fine.

Martin

PR #972, opened 2018-02-02
Explicit log initialisation via RPC

AlCutter:


client/admin.go
```diff
@@ -59,31 +65,91 @@ func CreateAndInitTree(ctx context.Context, req *trillian.CreateTreeRequest, adm
...
-					glog.Errorf("Map server unavailable: %v", err)
-					return err
-				case ok && s.Code() == codes.AlreadyExists:
-					glog.Warningf("Bizarrely, the just-created Map (%x) is already initialised!: %v", tree.TreeId, err)
-					return err
-				}
-				return fmt.Errorf("failed to InitMap(%+v): %T %v", req, err, err)
+	switch tree.TreeType {
+	case trillian.TreeType_MAP:
+		err = InitMap(ctx, tree, mapClient)
```
Martin2112: what about making these if err = InitMap() etc. and then losing the if below?

client/admin.go
```diff
@@ -59,31 +65,91 @@ func CreateAndInitTree(ctx context.Context, req *trillian.CreateTreeRequest, adm
...
+	case trillian.TreeType_LOG:
+		err = InitLog(ctx, tree, logClient)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return tree, nil
```
Martin2112: Should we guard against unknown tree types here or is doing nothing OK?

client/admin.go
```diff
@@ -59,31 +65,91 @@ func CreateAndInitTree(ctx context.Context, req *trillian.CreateTreeRequest, adm
...
+
+// InitMap initialises a freshly created Map tree.
+func InitMap(ctx context.Context, tree *trillian.Tree, mapClient trillian.TrillianMapClient) error {
+	if tree.TreeType != trillian.TreeType_MAP {
+		return fmt.Errorf("InitMap called with tree of type %v", tree.TreeType)
+	}
+
+	b := &backoff.Backoff{
+		Min:    100 * time.Millisecond,
+		Max:    10 * time.Second,
```
Martin2112: Should max wait be configurable?

client/admin.go
```diff
@@ -59,31 +65,91 @@ func CreateAndInitTree(ctx context.Context, req *trillian.CreateTreeRequest, adm
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// InitLog initialises a freshly created Log tree.
+func InitLog(ctx context.Context, tree *trillian.Tree, logClient trillian.TrillianLogClient) error {
```
Martin2112: InitLog and InitMap are very similar.  Would be good if they could share code but that's probably not easy because of different client types etc.

server/log_rpc_server.go
```diff
@@ -515,3 +515,66 @@ func (t *TrillianLogRPCServer) getTreeAndHasher(ctx context.Context, treeID int6
...
+	}
+
+	tx, err := t.registry.LogStorage.BeginForTree(ctx, logID)
+	if err != nil && err != storage.ErrLogNeedsInit {
+		return nil, err
+	}
+	defer tx.Close()
+	if err == nil {
+		// all good, no need to init.
+		return nil, status.New(codes.AlreadyExists, "log already intialised.").Err()
```
Martin2112: Typo + same one further down.

testonly/integration/logenv.go
```diff
@@ -243,7 +243,10 @@ func (env *LogEnv) CreateLog() (int64, error) {
 	if err := tx.Commit(); err != nil {
 		return 0, err
 	}
-	// Sign the first empty tree head.
-	env.Sequencer.OperationSingle(ctx)
+
+	_, err := env.logServer.InitLog(ctx, &trillian.InitLogRequest{LogId: tree.TreeId})
```
Martin2112: Doesn't compile - no new variables.

log/sequencer.go
```diff
@@ -267,18 +267,9 @@ func (s Sequencer) SequenceBatch(ctx context.Context, logID int64, limit int, gu
...
 		glog.Warningf("%v: Fresh log - no previous TreeHeads exist.", logID)
-		// SignRoot starts a new transaction, and we've got one open here until
-		// this function returns.
-		// This explicit Close() is a work-around for the in-memory storage which
-		// locks the tree for each TX.
-		// TODO(al): Producing the first signed root for a new tree should be
-		// handled by the provisioning, move it there.
-		tx.Close()
-		return 0, s.SignRoot(ctx, logID)
+		return 0, storage.ErrLogNeedsInit
```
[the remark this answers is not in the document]
AlCutter: L247 has a `defer tx.Close()`, so I think so.

testonly/integration/logenv.go
AlCutter: Soz, fixed

client/admin.go
AlCutter: ok, done.

client/admin.go
AlCutter: Added an error for unknown tree types

client/admin.go
AlCutter: Don't think it really matters, does it?
This is just the largest interval between retries rather than when it'll give up.

client/admin.go
AlCutter: Yeah - I thought of that but it'll end up just looking quite like `Backoff` with a large function being passed in, or some hair ball of strings and bits... didn't seem worth it.

server/log_rpc_server.go
AlCutter: thanks, done!

client/admin.go
Martin2112: Does it ever give up though?

AlCutter: Yes, when the context is cancelled.

Martin2112: I think m ax wait is badly named though I suspect it will turn out I named it.

PR #973, opened 2018-02-02
Improve map tests

gdbelvin:
Make the map test more robust by not referencing nil pointers

### DIFF
--- a/integration/maptest/map.go
+++ b/integration/maptest/map.go
@@ -150,7 +150,7 @@ func newTreeWithHasher(ctx context.Context, tadmin trillian.TrillianAdminClient,
 
 	hasher, err := hashers.NewMapHasher(tree.HashStrategy)
 	if err != nil {
-		return nil, nil, nil, nil
+		return nil, nil, nil, err
 	}
 	return tree, hasher, initResp.Created, nil
 }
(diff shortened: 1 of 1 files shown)


integration/maptest/map.go
```diff
@@ -177,55 +177,58 @@ func RunMapRevisionZero(ctx context.Context, t *testing.T, tadmin trillian.Trill
...
+					MapId: tree.TreeId,
+				})
+				if err != nil {
+					t.Fatalf("GetSignedMapRoot(): %v", err)
+				}
+				if err := verifyGetSignedMapRootResponse(getSmrResp.GetMapRoot(), tc.wantRev,
+					pubKey, hasher, tree.TreeId); err != nil {
+					t.Errorf("verifyGetSignedMapRootResponse(rev %v): %v", tc.wantRev, err)
+				}
+				//
```
AlCutter: What are these blank comment lines for, would a blank line suffice?

gdbelvin: Not sure. I'll remove it. I think it was pre-existing.

gdbelvin:
PTAL

integration/maptest/map.go
```diff
@@ -177,55 +177,57 @@ func RunMapRevisionZero(ctx context.Context, t *testing.T, tadmin trillian.Trill
...
-				t.Errorf("%v: newTreeWithHasher() returned unexpected root hash %x, want %x", tc.desc, got, want)
-			}
-			pubKey, err := der.UnmarshalPublicKey(tree.GetPublicKey().GetDer())
-			if err != nil {
-				t.Errorf("%v: UnmarshalPublicKey(%v): %v", tc.desc, hsr.hashStrategy, err)
-			}
+			t.Run(fmt.Sprintf("%v/%v", tc.desc, hsr.hashStrategy), func(t *testing.T) {
+				tree, hasher, smr, err := newTreeWithHasher(ctx, tadmin, tmap, hsr.hashStrategy)
+				if err != nil {
+					t.Fatalf("newTreeWithHasher(%v): %v", hsr.hashStrategy, err)
```
phad: Does t.Fatalf() kill the entire test suite or will it be recovered by test.T so the next sub-test can run?

gdbelvin: Fatalf only kills the test currently inside `t.Run` which is nice.

PR #975, opened 2018-02-05
Factor out LeafSequencer from LeafDequeuer

pavelkalinnikov:


pavelkalinnikov:
Postponing this until #977 gets merged.

pavelkalinnikov:
The net effect of this change after #977 would be a TODO comment. Droping it for now.

PR #978, opened 2018-02-05
Wait for first root to become available during {Log,Map}Init

gdbelvin:
`InitMap` is essentially an asynchronous call. When it returns, Map revision 0 may not be available via the `GetMapRevision` APIs.

This change causes the client to wait until the first map root does become available. 
Should this waiting operation be a client or server side operation?


AlCutter:
Looks like you might need to update the fake server too, though.

gdbelvin:
Added fake Log and Map implementations. 
These were files that were in KeyTransparency, but perhaps this is a better home for them.

cmd/createtree/main_test.go
```diff
@@ -105,27 +114,34 @@ func TestCreateTree(t *testing.T) {
 			wantErr:  true,
 		},
 		{
-			desc:      "createErr",
-			createErr: errors.New("create tree failed"),
-			wantErr:   true,
+			desc:         "createErr",
+			expectCreate: 4,
```
AlCutter: This seems quite brittle - if the logic of the backoff or the context timeout changes this is likely to start failing in unexpected ways.
So, how about if an `err` is set for the corresponding function, we implicitly set up the expect as `MinTimes(1)` or similar?

gdbelvin: sounds good

gdbelvin:
Added a function to compute the number of times we should expect a function to be called. PTAL

cmd/createtree/main_test.go
```diff
@@ -195,6 +186,25 @@ func runTest(t *testing.T, tests []*testCase) {
...
+	// If a function prior to this function errored,
+	// we do not expect this function to be called.
+	for _, e := range prevErr {
+		if e != nil {
+			return 0
+		}
+	}
+	// If this function errors, it will be retried multiple times.
+	if err != nil {
+		return 3
```
AlCutter: Should this just be 1 (or perhaps 2 if we're testing that it does retry)? and let the `MinTimes` take care of it?

cmd/createtree/main_test.go
```diff
@@ -195,6 +186,25 @@ func runTest(t *testing.T, tests []*testCase) {
 	}
 }
 
+// expectCalls returns the minimum number of times a function is expected to be called
+// given the return error for the function (err), and all previous errors in the function's
+// code path.
+func expectCalls(err error, prevErr ...error) int {
```
AlCutter: another option might be to have this take a `*gomock.Call` rather than returning an `int`, then it could choose to use `.AnyTimes`, `.MinTimes`, or .`Times` as appropriate?

e.g.
```golang
call := server.MockTrillianMapServer.EXPECT().GetSignedMapRootByRevision(gomock.Any(),  gomock.Any()).Return(&trillian.GetSignedMapRootResponse{}, nil)
expectCalls(call, retErr error, prevErr ...error)
```
wdyt?

cmd/createtree/main_test.go
gdbelvin: Yep. I'm still using this in `MinTimes` I can make the number 2

cmd/createtree/main_test.go
gdbelvin: That would be slightly more precise wouldn't it?

gdbelvin:
Done

PR #979, opened 2018-02-06
Add preordered log type to MySQL enum

Martin2112:
I think this was omitted from #964?


Martin2112:
This is a safe change to the schema so we can do it early. I'd be more cautious if it could break things.

PR #982, opened 2018-02-07
Fix/ Add NotFound test to WaitForInclusion

gdbelvin:
`client.WaitForInclusion` was erroring when `TreeSize == 0` and several other edge cases. 

- Added test
- Simplified logic


client/client.go
```diff
@@ -224,36 +224,22 @@ func (c *LogClient) WaitForInclusion(ctx context.Context, data []byte) error {
 	}
 
 	// Fetch the current Root to improve our chances at a valid inclusion proof.
-	root, err := c.UpdateRoot(ctx)
+	// It is illegal to ask for an inclusion proof with TreeSize = 0.
```
Martin2112: Not so much that it's illegal, it doesn't make sense because the tree has no entries.

gdbelvin: Well, if one does try to do that, the server will respond with an error :-)

PR #986, opened 2018-02-08
ErrLogNeedsInit -> ErrTreeNeedsInit

gdbelvin:
`ErrLogNeedsInit` occurs for both log and map servers


storage/log_storage.go
```diff
@@ -22,8 +22,8 @@ import (
 	"github.com/google/trillian/errors"
 )
 
-// ErrLogNeedsInit is returned when calling methods on an uninitialised Log.
-var ErrLogNeedsInit = errors.New(errors.FailedPrecondition, "log needs initialising")
+// ErrTreeNeedsInit is returned when calling methods on an uninitialised Log.
```
AlCutter: This should be in storage/tree_storage.go I think (I'm having a sense of déjà vu!)

gdbelvin:
This change (with requested edits) was included in #990

PR #987, opened 2018-02-08
Fix MapInit

gdbelvin:
`MapInit` needs the same fix as #976 

`BeginForTree` may not return an error when the map is uninitialized on all storage implementations.

### DIFF
--- a/storage/log_storage.go
+++ b/storage/log_storage.go
@@ -22,8 +22,8 @@ import (
 	"github.com/google/trillian/errors"
 )
 
-// ErrLogNeedsInit is returned when calling methods on an uninitialised Log.
-var ErrLogNeedsInit = errors.New(errors.FailedPrecondition, "log needs initialising")
+// ErrTreeNeedsInit is returned when calling methods on an uninitialised Log.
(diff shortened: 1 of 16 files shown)


storage/log_storage.go
```diff
@@ -22,8 +22,8 @@ import (
 	"github.com/google/trillian/errors"
 )
 
-// ErrLogNeedsInit is returned when calling methods on an uninitialised Log.
-var ErrLogNeedsInit = errors.New(errors.FailedPrecondition, "log needs initialising")
+// ErrTreeNeedsInit is returned when calling methods on an uninitialised Log.
```
AlCutter: This should be in `tree_storage.go` now, I'd imagine?

gdbelvin: yep!

storage/tree_storage.go
```diff
@@ -18,6 +18,9 @@ import (
 	"context"
 )
 
+// ErrTreeNeedsInit is returned when calling methods on an uninitialised Log.
```
AlCutter: "on an uninitialised tree."

gdbelvin: done

Issue #985, opened 2018-02-08
Follow CloudPlatform API guidelines


pavelkalinnikov:
It appears that Trillian APIs diverge from best practices outlined in CloudPlatform [guidelines](https://cloud.google.com/apis/design/). It would be great to revisit the APIs and try to follow the advice.

In particular, we can improve [commenting](https://cloud.google.com/apis/design/documentation), reorganize [file structure](https://cloud.google.com/apis/design/file_structure) and (potentially) follow some [design patterns](https://cloud.google.com/apis/design/design_patterns) like "Representing Ranges", "Pagination" and "Output Fields" (the last one is part of commenting bit as well).

pavelkalinnikov:
I will send a PR with some improvements to the Log API shortly.

gdbelvin:
Would you mind also taking a look at the APIs in Key Transparency? 
https://github.com/google/keytransparency/tree/master/core/api

pavelkalinnikov:
@gdbelvin Sure, happy to help with that. Can you please add an issue for me in KT? Will address it a bit later.

PR #988, opened 2018-02-08
Format Log API according to Cloud guidelines

pavelkalinnikov:
This addresses #985 with regards to formatting/commenting.
I recommend reviewing commit by commit to see the effects of individual guideline points.


trillian_log_api.proto
```diff
@@ -219,60 +235,83 @@ message GetLeavesByHashRequest {
...
+    //  - `google.rpc.FAILED_PRECONDITION`: A conflicting entry is already
+    //    present in the log, e.g., same `leaf_index` but different `leaf_data`.
     google.rpc.Status status = 2;
 }
 
+// A leaf of the log's Merkle tree, corresponds to a single log entry. Each leaf
+// has a unique `leaf_index` in the scope of this tree.
 message LogLeaf {
-    // merkle_leaf_hash is over leaf data and optional extra_data.
+    // Output only. The hash over `leaf_data`.
```
pavelkalinnikov: Apparently, `extra_data` is [not included](https://github.com/google/trillian/blob/7185394727abf33444773cc1bacd6a5d7c98ab11/server/log_rpc_server.go#L105) in the hash. Should it be?

trillian_log_api.proto
```diff
@@ -219,60 +235,83 @@ message GetLeavesByHashRequest {
 }
 
 message GetLeavesByHashResponse {
-    // TODO(gbelvin) reply with error codes.
+    // TODO(gbelvin) reply with error codes. Reuse QueueLogLeaf?
```
pavelkalinnikov: selfnit: `QueuedLogLeaf`.

pavelkalinnikov:
@daviddrysdale Thanks for your review, will address it shortly. Correct, no code changes.

gdbelvin:
This looks good to me.  Thanks for cleaning this up!

At some point - perhaps on the next API revision bump - change the directory structure as well

trillian_log_api.proto
pavelkalinnikov: Done.

pavelkalinnikov:
ping

trillian_log_api.proto
AlCutter: No, this is just a way of associating informational metadata with the entry at submit time. If a personality wanted to commit to that data then, as you say, they'd make it part of the `leaf_value`.

trillian_log_api.proto
```diff
@@ -14,75 +14,106 @@
...
+option java_package = "com.google.trillian.proto";
 
-import "trillian.proto";
-import "google/rpc/status.proto";
 import "google/api/annotations.proto";
 import "google/protobuf/timestamp.proto";
+import "google/rpc/status.proto";
+import "trillian.proto";
+
+// Provides access to a Verifiable Log as defined in the [Verifiable Data
```
Martin2112: Can you tidy this comment while you're there. It says "provides access" and "provides direct access". Could be shortened.

Martin2112:
This is another PR where GitHub does a terrible job of change tracking.

trillian_log_api.proto
pavelkalinnikov: Done.

PR #990, opened 2018-02-09
Clean/logclient

gdbelvin:
`LogEnv` changes:
- `PublicKey` is a property of trees.  Remove it from `LogEnv`. 
- Use `CreateAndInitLog` rather than a duplicate `env.CreateLog`

Client changes:
- Move `NewLogVerifierFromTree` to the `LogVerifier`. 
   - This makes the `LogVerifier` more reusable by other libraries. 
- Add `GetAndVerify*` to a few methods for clarity.


client/log_client.go
```diff
@@ -36,43 +32,27 @@ import (
 type LogClient struct {
 	LogID  int64
 	client trillian.TrillianLogClient
-	*logVerifier
+	LogVerifier
```
phad: Is it usual for struct embeds to be placed down amongst the fields like this?
I'd expect to see it first in the definition.

phad: Or actually name it. Any reason for this not to be
```
type LogClient struct {
  ...
  verifier LogVerifier
  ...
}
```

client/log_verifier.go
```diff
@@ -40,6 +41,27 @@ func NewLogVerifier(hasher hashers.LogHasher, pubKey crypto.PublicKey) LogVerifi
 	}
 }
 
+// NewLogVerifierFromTree creates a new LogVerifier
```
phad: ... using the algorithm set specified by config.  ?

client/log_verifier.go
```diff
@@ -40,6 +41,27 @@ func NewLogVerifier(hasher hashers.LogHasher, pubKey crypto.PublicKey) LogVerifi
...
+func NewLogVerifierFromTree(config *trillian.Tree) (LogVerifier, error) {
+	if got, want := config.TreeType, trillian.TreeType_LOG; got != want {
+		return nil, fmt.Errorf("client: NewFromTree(): TreeType: %v, want %v", got, want)
```
phad: errors still reference NewFromTree() here and the two below.

client/log_client.go
```diff
@@ -189,7 +169,7 @@ func (c *LogClient) getLatestRoot(ctx context.Context, trusted *trillian.SignedL
 	}
 	// Verify root update if the tree / the latest signed log root isn't empty.
 	if resp.GetSignedLogRoot().GetTreeSize() > 0 {
-		if err := c.logVerifier.VerifyRoot(trusted, resp.GetSignedLogRoot(),
+		if err := c.LogVerifier.VerifyRoot(trusted, resp.GetSignedLogRoot(),
```
phad: If I've understood the point of struct embedding correctly, you no longer need to include `LogVerifier` in this expression, and `c.VerifyRoot` should do fine.

phad: Which you do for the `BuildLeaf` calls below.

client/log_client.go
phad: OK I see that later on you are taking advantage of the embed to shorten the 'dotted path' to fields and methods.
I still think you should move the embed up to the top of this struct though, and have a separating empty line, as it gets a bit lost as it is here.

client/log_client_test.go
```diff
@@ -54,12 +55,18 @@ func TestGetByIndex(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer env.Close()
-	logID, err := env.CreateLog()
+
+	tree, err := CreateAndInitTree(ctx, &trillian.CreateTreeRequest{
+		Tree: stestonly.LogTree,
+	}, env.Admin, nil, env.Log)
```
phad: I'm finding this a bit hard to read because of the  `}` being outdented as far as this.  Presumably gofmt has done that.
Would it be nicer to read like this:
```
tree, err := CreateAndInitTree(ctx, &trillian.CreateTreeRequest{Tree: stestonly.LogTree},
        env.Admin, nil, env.Log)
```

phad: Or like this:
```
tree, err := CreateAndInitTree(ctx,
        &trillian.CreateTreeRequest{Tree: stestonly.LogTree},
        env.Admin, nil, env.Log)
```

phad: I'd only split and indent it the way you did if there were several fields being declared in `CreateTreeRequest`

client/log_client_test.go
```diff
@@ -54,12 +55,18 @@ func TestGetByIndex(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Failed to create log: %v", err)
 	}
 
-	client := New(logID, env.Log, rfc6962.DefaultHasher, env.PublicKey)
+	client, err := NewFromTree(env.Log, tree)
```
phad: I can't easily tell where `rfc6962.DefaultHasher, env.PublicKey` ended up after your refactor in this PR?

client/log_client.go
gdbelvin: Embeds should indeed come first.

client/log_client_test.go
gdbelvin: Thanks, the second version seems much clearer.

client/log_client_test.go
gdbelvin: Good question.  They are inside the `*trillian.Tree` object which is returned from `CreateAndInitTree`. 

The hasher and public key are unique to each tree rather than the server.

client/log_verifier.go
gdbelvin: yep. Added comment.

client/log_verifier.go
gdbelvin: done

gdbelvin:
Updated. PTAL

client/log_client_test.go
phad: OK, makes sense.

client/log_client_test.go
```diff
@@ -90,12 +97,17 @@ func TestListByIndex(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer env.Close()
-	logID, err := env.CreateLog()
+	tree, err := CreateAndInitTree(ctx, &trillian.CreateTreeRequest{
```
phad: Can you repeat the same indent and keep-together changes in other places in this file as you did above? ( keeping `&trillian.CreateTreeRequest{Tree: stestonly.LogTree}` online line and reflowing around it)

gdbelvin:
Turns out this PR is dependent on certificate-transparency-go#148

client/log_client_test.go
gdbelvin: yep. Done

PR #992, opened 2018-02-09
QueueLeaves on LogStorage

AlCutter:
Queuing leaves is currently part of the `LogTreeTX` API, but that implies that a batch of leaves should be entirely rejected if there's a duplicate entry within the batch. In reality, current storage implementations quietly ignore this and work to submit the non-duplicate entries.

This PR adds a `LogStorage.QueueLeaves` func which delegates to the original TX-based implementation, and switches `LogRPCServer` over to use it.

We should consider whether there may be a use case for `LogTreeTX.QueueLeaves` to remain, enforcing the strict no-dupes-or-the-whole-batch-fails semantics one would expect; `QueueLeavesRequest` could be extended indicate the requested behaviour and the RPC server could select the appropriate method.

### DIFF
--- a/storage/log_storage.go
+++ b/storage/log_storage.go
@@ -115,6 +115,15 @@ type LogStorage interface {
 	// If f fails and returns an error, the storage implementation may optionally
 	// retry with a new transaction, and f MUST NOT keep state across calls.
 	ReadWriteTransaction(ctx context.Context, treeID int64, f LogTXFunc) error
+
+	// QueueLeaves enqueues leaves for later integration into the tree.
+	// If error is nil, the returned slice of leaves will be the same size as the
+	// input, and each entry will hold:
+	//  - the existing leaf entry if a duplicate has been submitted
+	//  - nil otherwise.
+	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
+	// considered duplicate if their leaf.LeafIdentityHash matches.
+	QueueLeaves(ctx context.Context, treeID int64, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.LogLeaf, error)
 }
 
 // CountByLogID is a map of total number of items keyed by log ID.
--- a/storage/log_storage.go
+++ b/storage/log_storage.go
@@ -123,7 +123,7 @@ type LogStorage interface {
 	//  - nil otherwise.
 	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
 	// considered duplicate if their leaf.LeafIdentityHash matches.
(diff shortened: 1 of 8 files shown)


pavelkalinnikov:
I am about to introduce `AddSequencedLeaves` storage API. Whatever semantics/parameters we end up with for `QueueLeaves`, I think `AddSequencedLeaves` should be consistent with it.

**Caveat**. In MySQL storage `QueueLeaves` and `AddSequencedLeaves` both need to put the entries to `LeafData`. If we do it in separate transactions from updating the corresponding `Unsequenced` / `Sequenced` tables, we might end up with "dangling" entries in `LeafData` if something goes wrong. Consequently, this may result in Trillian reporting false duplicates afterwards.

Martin2112:
Yes it definitely needs to work for both cases and storage should be left in a consistent state if something goes wrong.

AlCutter:
@pavelkalinnikov, yup: inserting _each_ leaf **must** be done transactionally if it takes more than 1 atomic operation to do so (e.g. insert to LeafData and Unsequenced/Sequenced).

Martin2112:
Yeah sorry I was trying to get my dodgy code working before looking at your errr... PR.

Martin2112:
Can you perhaps rename the PR so it doesn't look like we're doing work without a transaction.

storage/log_storage.go
```diff
@@ -116,6 +116,15 @@ type LogStorage interface {
...
 	BeginForTree(ctx context.Context, treeID int64) (LogTreeTX, error)
+
+	// QueueLeaves enqueues leaves for later integration into the tree.
+	// If error is nil, the returned slice of leaves will be the same size as the
+	// input, and each entry will hold:
+	//  - the existing leaf entry if a duplicate has been submitted
+	//  - nil otherwise.
+	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
+	// considered duplicate if their leaf.LeafIdentityHash matches.
+	QueueLeaves(ctx context.Context, treeID int64, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.LogLeaf, error)
```
pavelkalinnikov: Since this method is now allowed to fail queuing some of the entries, don't we need to return some indication of that? E.g., return `QueuedLogLeaf` instead of `LogLeaf` to associate more elaborate statuses with the entries.

AlCutter: You might be right, but I'll leave that for another PR to keep this as small as possible.

AlCutter: Yes - done

AlCutter:
You might be right about not echoing the request back, but I'll leave that for another PR to keep this as small as possible.  I will explicitly return an `OK` status though, rather than the `nil` status we've been returning previously.

PR #993, opened 2018-02-12
First stab at refactor - pass GetOpts through to storage

Martin2112:
Also breaks the dependency link from trees -> storage packages. Can be used with anything that can return a tree.

This should make it cleaner to implement the tree readonly / freezing feature.

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -65,8 +65,8 @@ var (
 	QuotaIncreaseFactor = 1.1
 
 	// TODO(Martin2112): Update to add options for sequencing / signing when they exist
-	sequencerOpts = trees.GetOpts{TreeType:trillian.TreeType_LOG}
-	signerOpts = trees.GetOpts{TreeType:trillian.TreeType_LOG}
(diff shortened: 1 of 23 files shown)


Martin2112:
It looks worse than it is, mostly moving stuff around.

Martin2112:
Probably no point reviewing until Al finishes submitting his TX changes as it will need rework.

Martin2112:
Closing this and reworking it because of major changes.

pavelkalinnikov:
@Martin2112 I need to make more changes that might interfere with your work (e.g., enable all read requests from storage for both `LOG` and `PREORDERED_LOG`). This will help me to test the new API along the way. What are your plans w.r.t. the reworking?

Martin2112:
I have mostly done it but it's not finished yet.

pavelkalinnikov:
Ok, I think it makes sense for me to follow up on your change. Please add me to review that one.

Martin2112:
You might regret that.

pavelkalinnikov:
It's better to regret what you have done than what you haven't

Martin2112:
Yeah I suppose so. I'm hoping to have a new PR tomorrow.

PR #994, opened 2018-02-12
Create and use a MapVerifier

### LINKED ISSUE #870: Map Verifier
Clients shouldn't need to reconstruct the full set of verification steps themselves. 
This logic should be packaged up, much like the log client.

gdbelvin:
Create a helper object for verifying responses from the Trillian Map. 

Closes #870

### DIFF
--- a/client/map_verifier.go
+++ b/client/map_verifier.go
@@ -49,10 +49,7 @@ func NewMapVerifierFromTree(config *trillian.Tree) (*MapVerifier, error) {
 		return nil, fmt.Errorf("Failed parsing Map public key: %v", err)
 	}
 
-	return &MapVerifier{
-		Hasher: mapHasher,
-		PubKey: mapPubKey,
-	}, nil
+	return &MapVerifier{Hasher: mapHasher, PubKey: mapPubKey}, nil
 }
 
 // VerifyMapLeafInclusion verifies a MapLeafInclusion response.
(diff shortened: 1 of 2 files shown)


client/map_verifier.go
```diff
@@ -0,0 +1,76 @@
...
+
+// MapVerifier verifies protos produced by the Trillian Map.
+type MapVerifier struct {
+	Hasher hashers.MapHasher
+	PubKey crypto.PublicKey
+}
+
+// NewMapVerifierFromTree creates a new MapVerifier.
+func NewMapVerifierFromTree(config *trillian.Tree) (*MapVerifier, error) {
+	if got, want := config.TreeType, trillian.TreeType_MAP; got != want {
```
phad: I thought this got, want style was to be restricted to test code, where it is idiomatic.
In production code it just seems to obscure the intent a little.

client/map_verifier.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, fmt.Errorf("Failed creating MapHasher: %v", err)
+	}
+
+	// Map Key
+	mapPubKey, err := der.UnmarshalPublicKey(config.GetPublicKey().GetDer())
+	if err != nil {
+		return nil, fmt.Errorf("Failed parsing Map public key: %v", err)
+	}
+
+	return &MapVerifier{
```
phad: Personally I'd make this struct initialization a one-liner as there's one two short fields being written.

client/map_verifier.go
gdbelvin: In production code the got, want style can be helpful if the got and want values are reused inside the if block

client/map_verifier.go
gdbelvin: done

client/map_verifier.go
phad: ok

PR #995, opened 2018-02-12
Example of dealing with server skew.

benlaurie:
If acceptable, presumably should be extended to most other Get* calls.


gdbelvin:
You may want to investigate a few methods in `client/log_client.go` Specifically `WaitForSTH`, which will wait for the server to catch up to a deterministic point in time. 

I've found it necessary to add various check points with `WaitForSTH` in integration test suites in order to maintain determinism.

benlaurie:
@gdbelvin there is no client/log_client.go, nor a WaitForSTH.

In any case, seems to me that doesn't actually solve the problem, even if I'm prepared to wait. I might wait for the STH but then still end up getting leaves from a different server, which has not seen that STH yet...

Martin2112:
I think this might be doing some extra work that we can avoid. Also I need to think about naming and API structure a bit more.

Let me play with it a bit.

trillian_log_api.proto
```diff
@@ -240,7 +261,7 @@ service TrillianLog {
       };
     }
 
-    // Corresponds to the LeafReader API
+    // DO NOT USE - FOR DEBUGGING ONLY (FIXME: remove or move to a debugging interface)
```
Martin2112: This is for integration tests not debugging really. It requires a full scan of distributed storage and is very slow.

Martin2112:
I think this doesn't work as it is because it will return an error from storage that gets propagated. I will do more thinking.

trillian_log_api.proto
```diff
@@ -175,6 +178,7 @@ message GetInclusionProofByHashResponse {
     // for this to return multiple proofs.
     // TODO(gbelvin) only return one proof.
     repeated Proof proof = 2;
+    SignedLogRoot signed_log_root = 3;
 }
 
 message GetConsistencyProofRequest {
```
AlCutter: Will you add the latest root here too?

benlaurie: Yes, there are quite a few more to do!

client/log_client.go
```diff
@@ -234,7 +232,14 @@ func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
 	if err != nil {
 		return fmt.Errorf("UpdateRoot(): %v", err)
 	}
-	return c.getAndVerifyInclusionProof(ctx, leaf.MerkleLeafHash, root)
+	ok, err := c.getAndVerifyInclusionProof(ctx, leaf.MerkleLeafHash, root)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return fmt.Errorf("No proof")
```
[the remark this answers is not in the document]
AlCutter: ...?

client/log_client.go
```diff
@@ -255,25 +260,25 @@ func (c *LogClient) GetAndVerifyInclusionAtIndex(ctx context.Context, data []byt
...
 			LeafHash: leafHash,
 			TreeSize: sth.TreeSize,
 		})
 	if err != nil {
-		return err
+		return false, err
 	}
 	if len(resp.Proof) < 1 {
-		return errors.New("no inclusion proof supplied")
+		return false, nil
```
AlCutter: You seem to have moved the creation of an error for this condition up a level to the caller (L239 is effectively `if err == nil && !ok`, why not just return `false, errors.New("no inclusion...")` as before?

integration/log.go
```diff
@@ -473,9 +473,14 @@ func checkConsistencyProof(consistParams consistencyProofParams, treeID int64, t
 		return fmt.Errorf("GetConsistencyProof(%v) = %v %v", consistParams, err, resp)
 	}
 
+	if req.SecondTreeSize > resp.SignedLogRoot.TreeSize {
+		return fmt.Errorf("Requested tree size %d > available tree size %d", req.SecondTreeSize, resp.SignedLogRoot.TreeSize)
```
AlCutter: Requested -> requested

integration/log.go
```diff
@@ -473,9 +473,14 @@ func checkConsistencyProof(consistParams consistencyProofParams, treeID int64, t
...
+	}
+
 	verifier := merkle.NewLogVerifier(rfc6962.DefaultHasher)
 	root1 := tree.RootAtSnapshot(req.FirstTreeSize).Hash()
 	root2 := tree.RootAtSnapshot(req.SecondTreeSize).Hash()
+	fmt.Printf("req: %v\nresp: %v\n", req, resp)
```
AlCutter: delete this?

server/log_rpc_server.go
```diff
@@ -506,31 +506,35 @@ func (t *TrillianLogRPCServer) GetEntryAndProof(ctx context.Context, req *trilli
 	if err != nil {
 		return nil, err
 	}
+	r := &trillian.GetEntryAndProofResponse{SignedLogRoot: &root}
 
-	proof, err := getInclusionProofForLeafIndex(ctx, tx, hasher, req.TreeSize, req.LeafIndex, root.TreeSize)
-	if err != nil {
-		return nil, err
-	}
+	// FIXME: should we return a proof to the available tree size if req.LeafIndex < root.TreeSize?
```
AlCutter: No, I think that's best covered by the `SuperDoAllTheThings` RPC (that admittedly) doesn't yet exist); the client may well want the exact tree-size they asked for because they have a subset of particular STHs they want to use (and have separately confirmed to be consistent).

server/log_rpc_server_test.go
```diff
@@ -1180,6 +1180,38 @@ func TestGetEntryAndProof(t *testing.T) {
...
+	}
+	server := NewTrillianLogRPCServer(registry, fakeTimeSource)
+
+	response, err := server.GetEntryAndProof(context.Background(), &getEntryAndProofRequest17)
+	if err != nil {
+		t.Fatalf("get entry and proof should have succeeded but we got: %v", err)
+	}
+
+	if response.Proof != nil {
+		t.Fatalf("expected nil proof but got: %v", response.Proof)
```
AlCutter: `Errorf` would let the next check happen too rather than potentially hiding the fact that `Leaf` is nil too.

server/log_rpc_server.go
```diff
@@ -508,7 +508,11 @@ func (t *TrillianLogRPCServer) GetEntryAndProof(ctx context.Context, req *trilli
 	}
 	r := &trillian.GetEntryAndProofResponse{SignedLogRoot: &root}
 
-	// FIXME: should we return a proof to the available tree size if req.LeafIndex < root.TreeSize?
+	if req.TreeSize > root.TreeSize && req.LeafIndex < root.TreeSize {
```
AlCutter: Ah, this does not do what I understood the `FIXME` comment to mean!
> FIXME: should we return a proof to the available tree size if req.LeafIndex < root.TreeSize?

I understood that to say if the leaf is within *our* [implicitly larger] tree, then return a proof to that rather than the treeSize the client asked for.

But the this change is not that and is ok :)

client/log_client.go
benlaurie: An error here breaks WaitForInclusion().

server/log_rpc_server_test.go
```diff
@@ -1206,15 +1206,15 @@ func TestGetEntryAndProofSkewNoProof(t *testing.T) {
 
 	response, err := server.GetEntryAndProof(context.Background(), &getEntryAndProofRequest17_11)
 	if err != nil {
-		t.Fatalf("get entry and proof should have succeeded but we got: %v", err)
+		t.Errorf("get entry and proof should have succeeded but we got: %v", err)
```
AlCutter: That one probably should still be `Fatal` because `response` is likely to be nil :)

integration/log.go
AlCutter: ...?

AlCutter:
just a couple of nits, LGTM otherwise

server/log_rpc_server_test.go
benlaurie: If response is nil, this will crash :-)

benlaurie: but ok

Martin2112:
The api changes have been merged. Please rebase this against master. I'll do the CTFE change next then the rest should merge without breaking tests.

server/log_rpc_server.go
```diff
@@ -504,31 +539,39 @@ func (t *TrillianLogRPCServer) GetEntryAndProof(ctx context.Context, req *trilli
 	if err != nil {
 		return nil, err
 	}
+	r := &trillian.GetEntryAndProofResponse{SignedLogRoot: &root}
 
-	proof, err := getInclusionProofForLeafIndex(ctx, tx, hasher, req.TreeSize, req.LeafIndex, root.TreeSize)
-	if err != nil {
-		return nil, err
+	if req.TreeSize > root.TreeSize && req.LeafIndex < root.TreeSize {
+		// return latest proof we can manage
```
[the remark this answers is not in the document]
benlaurie: A client in this situation will either wait and try again, or request a proof against the tree head the server does know.

In the latter case, we save them at least one round trip. In the former, we cost them a little extra traffic.

Martin2112:
Please don't merge this until I've completed the CTFE changes!

benlaurie:
Rebased. Hope I got it right, it was a bit confusing.

Martin2112:
CTFE changes are merged.

Martin2112:
The tests pass now, which is encouraging.

server/log_rpc_server.go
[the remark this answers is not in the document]
Martin2112: Yes but I think it shouldn't return the proof in this case, just the STH and let the client decide what to do.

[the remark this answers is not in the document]
benlaurie: Well, I disagree. If it returns the STH, there's a risk that the next server is even more skewed and it will fail again. I realise that's pretty unlikely with current storage, but I can easily imagine other types of storage layers where it is far more likely.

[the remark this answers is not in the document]
Martin2112: Not sure how this helps the client though. Maybe I'm not understanding what you mean? If client asks for proof at size 100 and gets one at size 50 for example. If 50 would have been enough then it could have been requested.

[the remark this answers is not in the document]
benlaurie: That seems inefficient - client should request against an STH it already has, to reduce the number of STH consistency proofs it has to fetch...

Martin2112:
I'm still doubtful of the utility of returning a proof at a lower tree size but it doesn't break anything so I don't object to it being merged.

AlCutter:
Cool, let's do that and we can argue/remove the inclusion proof bit as appropriate.
@benlaurie:  Looks like it needs another rebase before we can commit though.

benlaurie:
If you want me to do the rebase, then that's unlikely to happen til end of the week at the earliest. Kinda feel like its someone else's turn!

PR #997, opened 2018-02-12
Add LeafIdentityHash to LogLeaf

gdbelvin:


client/log_verifier.go
```diff
@@ -133,7 +133,8 @@ func (c *logVerifier) BuildLeaf(data []byte) (*trillian.LogLeaf, error) {
 		return nil, err
 	}
 	return &trillian.LogLeaf{
-		LeafValue:      data,
-		MerkleLeafHash: leafHash,
+		LeafValue:        data,
+		MerkleLeafHash:   leafHash,
+		LeafIdentityHash: leafHash,
```
AlCutter: This isn't necessarily the case; it's only equivalent to `MerkleLeafHash` iff the personality hasn't defined it to be otherwise.

gdbelvin: Would `sha256` of data be better?
Maybe in the future we can add hooks into the client that will allow personalities to customize it. At the moment though, that doesn't seem to be needed. 

The `LeafIdentityHash` is required to be present in prod. This change is to just have _something_ in this field.

client/log_verifier.go
```diff
@@ -132,9 +133,11 @@ func (c *logVerifier) BuildLeaf(data []byte) (*trillian.LogLeaf, error) {
 	if err != nil {
 		return nil, err
 	}
+	h := sha256.New()
+	h.Write(data)
 	return &trillian.LogLeaf{
 		LeafValue:        data,
 		MerkleLeafHash:   leafHash,
-		LeafIdentityHash: leafHash,
+		LeafIdentityHash: h.Sum(nil),
```
AlCutter: Since this is completely bogus, rather than putting something which looks plausible on the surface could we have a const which we assign here, preferably a string like "INVALID" or something?

gdbelvin: Well it's not completely bogus... If two leafs have the same identity hash, the second one won't get stored.  The hash of the leaf data is a perfectly sane choice. 

The only thing that seems to be missing here is the ability to swap out the sha256 algorithm for another one which I think we should do as soon as we have a use case for it. wdyt?

AlCutter: Blindly setting the `LeafIdentityHash` to the hash of the leaf data is not a one-size-fits-all choice: CT, for example, explicitly doesn't do that (it uses `H(cert)`, not `H(leaf)`) and you'd open the logs to spam if the CTFE personality uses this library.

`LeafIdentityHash` is only be used by the log when submitting entries, and since Trillian is the layer below the personality you need a way of having the personality which is using this library to tell it what the ID hash of a particular leaf is.

gdbelvin:
I've pulled the` LeafIdentityAlgo` out as a separate function which clients can set, but left sha256 as the default.  Is that ok?

client/log_verifier.go
```diff
@@ -28,20 +29,32 @@ import (
...
+		v:                merkle.NewLogVerifier(hasher),
+		LeafIdentityAlgo: Sha256,
 	}
 }
 
+// LeafIdentity computes the identity value for a leaf.
+type LeafIdentity func(leaf []byte) ([]byte, error)
+
+// Sha256 computes the sha256 of the input.
+func Sha256(data []byte) ([]byte, error) {
```
AlCutter: You don't even know that a personality is using SHA256 at this stage so this is still dangerous :)

But, I believe that `TrillianLogServer` will set `LeafIdentityHash` to `H(leaf_value)` on your behalf (with the correct hasher for the tree) iff you don't fill in an explicit `LeafIdentityHash` - does that not work for your usecase?

gdbelvin:
Woops. This was the wrong solution to the wrong problem.

PR #998, opened 2018-02-13
Remove explicit calls to Commit when using ReadWriteTransaction

AlCutter:
The `ReadWriteTransaction` method will always call `Commit()` if no error is returned by the passed-in function, so this PR removes the explicit calls to `Commit` from the passed-in function implementations.

### DIFF
--- a/storage/memory/admin_storage.go
+++ b/storage/memory/admin_storage.go
@@ -52,7 +52,7 @@ func (s *memoryAdminStorage) ReadWriteTransaction(ctx context.Context, f storage
 	}
 	defer tx.Close()
 	if err := f(ctx, tx); err != nil {
-		tx.Close()
+		return err
 	}
 	return tx.Commit()
 }
--- a/storage/tools/dump_tree/dumplib/dumplib.go
+++ b/storage/tools/dump_tree/dumplib/dumplib.go
@@ -211,15 +211,14 @@ func createTree(as storage.AdminStorage, ls storage.LogStorage) (*trillian.Tree,
 		glog.Fatalf("SignLogRoot: %v", err)
 	}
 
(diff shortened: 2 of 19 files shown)


storage/memory/admin_storage.go
```diff
@@ -52,7 +52,7 @@ func (s *memoryAdminStorage) ReadWriteTransaction(ctx context.Context, f storage
 	}
 	defer tx.Close()
 	if err := f(ctx, tx); err != nil {
-		tx.Close()
+		return err
```
Martin2112: Why did I miss that before?

storage/tools/dump_tree/dumplib/dumplib.go
```diff
@@ -256,23 +255,22 @@ func Main(args Options) string {
 	sequenceLeaves(ls, seq, tree.TreeId, args.TreeSize, args.BatchSize, args.LeafFormat)
 
 	// Read the latest STH back
-	tx, err := ls.BeginForTree(context.TODO(), tree.TreeId)
-	if err != nil {
-		glog.Fatalf("BeginForTree got: %v, want: no err", err)
-	}
+	var sth trillian.SignedLogRoot
+	err := ls.ReadWriteTransaction(context.TODO(), tree.TreeId, func(ctx context.Context, tx storage.LogTreeTX) error {
```
Martin2112: SnapshotForTree? Doesn't really matter as this is test code really.

AlCutter: Yeah, I just converted it as-was.

storage/memory/admin_storage.go
AlCutter: easy to miss in the noise :)

PR #999, opened 2018-02-13
Add option to get leaves to return zero leaves if start > sth.treesize

Martin2112:
Also returns the current tree size if set.


server/log_rpc_server.go
```diff
@@ -370,6 +370,17 @@ func (t *TrillianLogRPCServer) GetLeavesByRange(ctx context.Context, req *trilli
...
 	defer tx.Close()
 
+	if req.ZeroLeavesOk {
+		sth, err := tx.LatestSignedLogRoot(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if req.StartIndex >= sth.TreeSize {
+			// We can't serve any leaves from this range as it starts outside our tree.
+			return &trillian.GetLeavesByRangeResponse{Leaves: []*trillian.LogLeaf{}, TreeSize: sth.TreeSize}, nil
```
benlaurie: In this case you can just not set Leaves.

benlaurie:
Its not clear to me why the client would ever want an error in this case, so why is there an option on the request?

Also, we need to require that there is no other circumstance in which zero leaves could be returned (or have an explicit flag for this condition [note that TreeSize could be zero]).

I _think_ I also want the option to get an STH rather than just a TreeSize on failure. Not sure its necessary, but its definitely easier to understand in terms of verification of behaviour.

server/log_rpc_server.go
Martin2112: It makes the test simpler if it's not nil as nil != []Wibble{}.

Martin2112:
There's an option because it guarantees that the error handling behaviour CT sees is unchanged.

benlaurie:
That doesn't seem like a great reason to force all sensible clients to have to send extra stuff forever...

Also, what about the other point?

Martin2112:
I thought your proposal require clients to send extra stuff? Not breaking compatibility for CT is a good reason to require clients that want different behaviour to explicitly request it.

If you don't set the option you should never get zero leaves because it's converted into an error. If you do set the option then getting zero leaves means you're starting out of range, in which case you get the server tree size back.  I don't think this is a problem?

Also you told me yesterday that you only needed the tree size so I may have assumed that meant you only needed the tree size -:)

benlaurie:
No, my proposal had clients optionally sending extra stuff - if they wanted to be told a change in tree size, then they could send their own tree size.

I can be persuaded that's functionality that's not really needed, in which case my proposal did not require anything to be sent.

"If you don't set the option you should never get zero leaves because it's converted into an error" - you mean the current implementation converts it into an error? My concern is that you need to make this part of the (currently nonexistent) contract with the storage layer.

My solution, OTOH, doesn't require any particular contract with the storage layer.

benlaurie:
On STH vs. size, I haven't managed to figure out how a server could do anything interesting by lying about tree size, but on the general principle that it should be possible to verify cryptographically all claims from servers, it seems like a good idea to enable that (clearly, asking for an STH after might legitimately not yield that proof).

Martin2112:
That check should probably be enforced here rather than relying on storage.

benlaurie:
Not sure what check you mean?

Martin2112:
I mean not relying on storage to do that enforcement of zero leaves being returned -> error.

Martin2112:
Closing this pending creating an issue and someone implementing what we finally agreed.

PR #1000, opened 2018-02-13
Remove use of DoAndReturn

AlCutter:
Replacing calls to DoAndReturn on Storage mocks with a bunch of Fakes, to better follow guidance around the careful use of golang mocks.

### DIFF
--- a/log/sequencer_test.go
+++ b/log/sequencer_test.go
@@ -152,7 +152,7 @@ type testParameters struct {
 // Tests get their own mock context so they can be run in parallel safely
 type testContext struct {
 	mockTx      *storage.MockLogTreeTX
-	mockStorage storage.LogStorage
+	fakeStorage storage.LogStorage
 	signer      *crypto.Signer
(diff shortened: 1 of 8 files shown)


log/sequencer_test.go
```diff
@@ -201,15 +201,15 @@ func newSignerWithErr(signErr error) (gocrypto.Signer, error) {
 }
 
 func createTestContext(ctrl *gomock.Controller, params testParameters) (testContext, context.Context) {
-	mockStorage := storage.NewMockLogStorage(ctrl)
+	mockStorage := &stestonly.FakeLogStorage{}
```
Martin2112: So we can't use mocks for this any more? But we're still calling them mocks. Might be better to rename the variables though I know it's fiddly.

AlCutter: Yeah :/
I left the names deliberately to reduce diff count, but I've updated them now - PTAL?

PR #1005, opened 2018-02-15
Enable creating PREORDERED_LOG

pavelkalinnikov:


storage/testonly/admin_storage_tester.go
```diff
@@ -120,7 +120,7 @@ var (
 		HashStrategy:       trillian.HashStrategy_RFC6962_SHA256,
 		HashAlgorithm:      spb.DigitallySigned_SHA256,
 		SignatureAlgorithm: spb.DigitallySigned_ECDSA,
-		DisplayName:        "Pavel's Pre-ordered Log",
+		DisplayName:        "Pre-ordered Log",
```
pavelkalinnikov: This line was longer than 20 chars, which was against the MySQL schema.

Martin2112: We could increase that number, not sure why I made it so short.

pavelkalinnikov: Meh, I'm fine not having my name in the code :)

Martin2112: This is indeed wise.

PR #1006, opened 2018-02-15
Move mocks to their own package

gdbelvin:
Mocks that share the same package as their protobufs messes up other protobuf BUILD targets. 
This PR moves the mocks into their own package.


testonly/tmock/fake_server.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package fake
+package tmock
```
AlCutter: Is there a godoc package description for this somewhere which needs updating too?

gdbelvin: `Package tmock is a generated GoMock package.`  Would you recommend putting the fake server in it's own package?

PR #1009, opened 2018-02-17
Move list/slack to top, add slack invitation

benlaurie:


Martin2112:
Just wondering why these branches are being created in the main repo and not a fork?

benlaurie:
Could be done in a fork if its a problem.

Martin2112:
Don't think it's a real problem, was just curious. I thought you could
preview via the "View" button in the PR.

Martin

benlaurie:
On 21 February 2018 at 10:52, Martin <notifications@github.com> wrote:

> Don't think it's a real problem, was just curious. I thought you could
> preview via the "View" button in the PR.
>

Possibly so, but then you have to commit and push, if you edit on github
you can preview before committing.


>
> Martin
>

PR #1010, opened 2018-02-19
Channel AddSequencedLeaves to storage

pavelkalinnikov:
This change adds `AddSequencedLeaves` to the storage API and implements the corresponding public API on top of it. This also adds method stubs to concrete storage implementations.

### DIFF
--- a/server/log_rpc_server.go
+++ b/server/log_rpc_server.go
@@ -87,6 +87,20 @@ func (t *TrillianLogRPCServer) QueueLeaf(ctx context.Context, req *trillian.Queu
 	return &trillian.QueueLeafResponse{QueuedLeaf: queueRsp.QueuedLeaves[0]}, nil
 }
 
+func hashLeaves(leaves []*trillian.LogLeaf, hasher hashers.LogHasher) error {
+	for _, leaf := range leaves {
+		var err error
+		leaf.MerkleLeafHash, err = hasher.HashLeaf(leaf.LeafValue)
+		if err != nil {
+			return err
+		}
+		if len(leaf.LeafIdentityHash) == 0 {
+			leaf.LeafIdentityHash = leaf.MerkleLeafHash
+		}
+	}
+	return nil
+}
+
 // QueueLeaves submits a batch of leaves to the log for later integration into the underlying tree.
 func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.QueueLeavesRequest) (*trillian.QueueLeavesResponse, error) {
 	if err := validateQueueLeavesRequest(req); err != nil {
@@ -100,15 +114,8 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
 	}
 	ctx = trees.NewContext(ctx, tree)
 
-	for _, leaf := range req.Leaves {
-		var err error
-		leaf.MerkleLeafHash, err = hasher.HashLeaf(leaf.LeafValue)
-		if err != nil {
-			return nil, err
-		}
-		if len(leaf.LeafIdentityHash) == 0 {
-			leaf.LeafIdentityHash = leaf.MerkleLeafHash
-		}
+	if err := hashLeaves(req.Leaves, hasher); err != nil {
+		return nil, err
 	}
 
 	ret, err := t.registry.LogStorage.QueueLeaves(ctx, logID, req.Leaves, t.timeSource.Now())
@@ -130,13 +137,56 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
 
 // AddSequencedLeaf submits one sequenced leaf to the storage.
 func (t *TrillianLogRPCServer) AddSequencedLeaf(ctx context.Context, req *trillian.AddSequencedLeafRequest) (*trillian.AddSequencedLeafResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "AddSequencedLeaf is not implemented")
+	if err := validateAddSequencedLeafRequest(req); err != nil {
+		return nil, err
+	}
+
+	batchReq := &trillian.AddSequencedLeavesRequest{
+		LogId:  req.LogId,
+		Leaves: []*trillian.LogLeaf{req.Leaf},
+	}
+	rsp, err := t.AddSequencedLeaves(ctx, batchReq)
+	if err != nil {
+		return nil, err
+	}
+	if rsp == nil {
+		return nil, status.Errorf(codes.Internal, "missing response")
+	}
+	if ln := len(rsp.Results); ln != 1 {
+		return nil, status.Errorf(codes.Internal, "unexpected count of sequenced leaves %d", ln)
+	}
+	return &trillian.AddSequencedLeafResponse{Result: rsp.Results[0]}, nil
 }
 
 // AddSequencedLeaves submits a batch of sequenced leaves to a pre-ordered log
 // for later integration into its underlying tree.
 func (t *TrillianLogRPCServer) AddSequencedLeaves(ctx context.Context, req *trillian.AddSequencedLeavesRequest) (*trillian.AddSequencedLeavesResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "AddSequencedLeaves is not implemented")
+	if err := validateAddSequencedLeavesRequest(req); err != nil {
+		return nil, err
+	}
+
+	tree, hasher, err := t.getTreeAndHasher(ctx, req.LogId, false /* readonly */)
+	if err != nil {
+		return nil, err
+	}
+	if want := trillian.TreeType_PREORDERED_LOG; tree.TreeType != want {
+		return nil, status.Errorf(codes.FailedPrecondition, "adding sequenced leaves available only for %v trees", want)
+	}
+
+	if err := hashLeaves(req.Leaves, hasher); err != nil {
+		return nil, err
+	}
+
+	ctx = trees.NewContext(ctx, tree)
+	leaves, err := t.registry.LogStorage.AddSequencedLeaves(ctx, tree.TreeId, req.Leaves)
+	if err != nil {
+		return nil, err
+	}
+	if got, want := len(leaves), len(req.Leaves); got != want {
+		return nil, status.Errorf(codes.Internal, "AddSequencedLeaves returned %v codes, want: %d", got, want)
+	}
+
+	return &trillian.AddSequencedLeavesResponse{Results: leaves}, nil
 }
 
 // GetInclusionProof obtains the proof of inclusion in the tree for a leaf that has been sequenced.
--- a/server/log_rpc_server.go
+++ b/server/log_rpc_server.go
@@ -72,7 +72,7 @@ func (t *TrillianLogRPCServer) IsHealthy() error {
 
 // QueueLeaf submits one leaf to the queue.
 func (t *TrillianLogRPCServer) QueueLeaf(ctx context.Context, req *trillian.QueueLeafRequest) (*trillian.QueueLeafResponse, error) {
-	if err := validateQueueLeafRequest(req); err != nil {
+	if err := validateLogLeaf(req.Leaf, "QueueLeafRequest.Leaf"); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


server/log_rpc_server.go
```diff
@@ -130,13 +137,56 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
...
+		Leaves: []*trillian.LogLeaf{req.Leaf},
+	}
+	rsp, err := t.AddSequencedLeaves(ctx, batchReq)
+	if err != nil {
+		return nil, err
+	}
+	if rsp == nil {
+		return nil, status.Errorf(codes.Internal, "missing response")
+	}
+	if ln := len(rsp.Results); ln != 1 {
```
Martin2112: Maybe do got, want here.

server/log_rpc_server.go
```diff
@@ -130,13 +137,56 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
...
+	if err := validateAddSequencedLeavesRequest(req); err != nil {
+		return nil, err
+	}
+
+	tree, hasher, err := t.getTreeAndHasher(ctx, req.LogId, false /* readonly */)
+	if err != nil {
+		return nil, err
+	}
+	if want := trillian.TreeType_PREORDERED_LOG; tree.TreeType != want {
+		return nil, status.Errorf(codes.FailedPrecondition, "adding sequenced leaves available only for %v trees", want)
```
Martin2112: Error is a bit wordy and consider adding what state the tree is in to help the caller.

server/validate.go
```diff
@@ -106,28 +106,33 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
 	}
 	return nil
 }
 
 func validateQueueLeavesRequest(req *trillian.QueueLeavesRequest) error {
-	if len(req.Leaves) == 0 {
-		return status.Error(codes.InvalidArgument, "QueueLeavesRequest.Leaves empty")
+	if err := validateLogLeaves(req.Leaves); err != nil {
+		// validateLogLeaves errors chain nicely with "QueueLeavesRequest.".
+		return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.", err)
```
Martin2112: Errorf() with nothing in the format string. Is something missing here?

storage/log_storage.go
```diff
@@ -124,6 +125,30 @@ type LogStorage interface {
 	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
 	// considered duplicate if their leaf.LeafIdentityHash matches.
 	QueueLeaves(ctx context.Context, treeID int64, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.QueuedLogLeaf, error)
+
+	// AddSequencedLeaves stores the `leaves` and associates them with the log
+	// positions according to their `LeafIndex` field. The indices must be unique
+	// and sorted in ascending order.
```
Martin2112: Where is the ordering constraint enforced?

server/log_rpc_server.go
```diff
@@ -130,13 +137,56 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
...
 
 // AddSequencedLeaves submits a batch of sequenced leaves to a pre-ordered log
 // for later integration into its underlying tree.
 func (t *TrillianLogRPCServer) AddSequencedLeaves(ctx context.Context, req *trillian.AddSequencedLeavesRequest) (*trillian.AddSequencedLeavesResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "AddSequencedLeaves is not implemented")
+	if err := validateAddSequencedLeavesRequest(req); err != nil {
+		return nil, err
+	}
+
+	tree, hasher, err := t.getTreeAndHasher(ctx, req.LogId, false /* readonly */)
```
pavelkalinnikov: This line will not get the tree currently, as `getTreeAndHasher` uses `GetOpts` with `LOG` type. Need to refactor it a bit.

server/validate.go
```diff
@@ -106,28 +106,33 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
 }
 
 func validateQueueLeafRequest(req *trillian.QueueLeafRequest) error {
-	if req.Leaf == nil {
-		return status.Error(codes.InvalidArgument, "QueueLeafRequest.Leaf empty")
+	if err := validateLogLeaf(req.Leaf); err != nil {
```
Martin2112: Could we have a validateLeaf(leaf, "QueueLeaf") function and get rid of some of the duplication? They're all a bit similar.

server/log_rpc_server.go
pavelkalinnikov: Time to add some tests...

server/validate.go
```diff
@@ -106,28 +106,33 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
-			// validateLogLeaf errors are meant to chain nicely with "Leaves."
-			return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.Leaves[%v].%v", i, err)
-		}
+	return nil
+}
+
+func validateAddSequencedLeavesRequest(req *trillian.AddSequencedLeavesRequest) error {
+	if err := validateLogLeaves(req.Leaves); err != nil {
+		// validateLogLeaves errors chain nicely with "AddSequencedLeavesRequest.".
+		return status.Errorf(codes.InvalidArgument, "AddSequencedLeavesRequest.", err)
```
Martin2112: Similar here, Errorf() with no formatting.

server/validate.go
```diff
@@ -106,28 +106,33 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
+		return status.Errorf(codes.InvalidArgument, "AddSequencedLeafRequest.Leaf%v", err)
 	}
 	return nil
 }
 
 func validateQueueLeavesRequest(req *trillian.QueueLeavesRequest) error {
-	if len(req.Leaves) == 0 {
-		return status.Error(codes.InvalidArgument, "QueueLeavesRequest.Leaves empty")
+	if err := validateLogLeaves(req.Leaves); err != nil {
+		// validateLogLeaves errors chain nicely with "QueueLeavesRequest.".
```
Martin2112: Can I just check that what I understand by "chain" is what you mean?

server/validate.go
pavelkalinnikov: Done. Also did the same in `validateLogLeaves`.

pavelkalinnikov:
`server/log_rpc_server_test.go` file is marked as generated for some reason. Please expand and review it as well.

server/validate.go
```diff
@@ -106,47 +106,61 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
+
+	lastIndex := int64(-1)
 	for i, leaf := range req.Leaves {
-		if leaf == nil {
-			return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.Leaves[%v] empty", i)
-		}
-		if err := validateLogLeaf(leaf); err != nil {
-			// validateLogLeaf errors are meant to chain nicely with "Leaves."
-			return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.Leaves[%v].%v", i, err)
+		if leaf.LeafIndex <= lastIndex {
```
AlCutter: Do we want to ensure the leaves are contiguous?

storage/log_storage.go
```diff
@@ -124,6 +125,30 @@ type LogStorage interface {
...
+	//
+	// If error is nil, the returned slice is the same size as the input, entries
+	// correspond to the `leaves` in the same order. Each entry describes the
+	// result of adding the corresponding leaf.
+	//
+	// Possible `QueuedLogLeaf.status` values with their semantics:
+	//  - OK: The leaf has been successfully stored.
+	//  - AlreadyExists: The storage has another leaf with the same `LeafIndex`.
```
AlCutter: "the storage already contains an identical leaf at the specified `LeafIndex`."?

server/validate.go
pavelkalinnikov: Initially I wanted do so, but I was not sure.

Suppose you submit entries 0-99, but get a response that only entries 0,2,4,...,98 were saved. You can further submit entries 1,3,...,99 in a single batch if non-contiguous is allowed. If not, you can do one of:

extreme1: 50 one-element ranges (kills the advantage of batching)
extreme2: all 100 entries (50 entries are re-submitted)
trade-off between the two (e.g., [0-3] [5-8] ...)
I am not sure how much the above example is realistic.

Perhaps we can start with an API with tighter requirements (that is, submit only contiguous ranges). If it ends up too strict, the requirements can be loosened. WDYT?

storage/log_storage.go
pavelkalinnikov: Done.

server/validate.go
pavelkalinnikov: @gdbelvin In Key Transparency, do you ever need to submit arbitrary non-contiguous batches? Or single entry / contiguous batches would be sufficient?

pavelkalinnikov: @AlCutter I propose to fine-tune this API in follow-up PRs, as I get more insight into what we can do with the batches in MySQL.

pavelkalinnikov:
@Martin2112 Can you take another look today?

Martin2112:
I'll try to make time to look at it again.

server/log_rpc_server.go
```diff
@@ -504,7 +557,7 @@ func (t *TrillianLogRPCServer) getTreeAndHasher(ctx context.Context, treeID int6
 // InitLog initialises a freshly created Log by creating the first STH with size 0.
 func (t *TrillianLogRPCServer) InitLog(ctx context.Context, req *trillian.InitLogRequest) (*trillian.InitLogResponse, error) {
 	logID := req.LogId
-	tree, hasher, err := t.getTreeAndHasher(ctx, logID, false /* readonly */)
+	tree, hasher, err := t.getTreeAndHasher(ctx, logID, optsLogRead)
```
Martin2112: write?

server/validate.go
```diff
@@ -106,47 +106,62 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
+}
+
+func validateAddSequencedLeafRequest(req *trillian.AddSequencedLeafRequest) error {
+	return validateLogLeaf(req.Leaf, "AddSequencedLeafRequest.Leaf")
 }
 
 func validateQueueLeavesRequest(req *trillian.QueueLeavesRequest) error {
-	if len(req.Leaves) == 0 {
-		return status.Error(codes.InvalidArgument, "QueueLeavesRequest.Leaves empty")
+	return validateLogLeaves(req.Leaves, "QueueLeavesRequest")
```
Martin2112: Can we not just remove this and the 2 functions above and call the validateLogXXX() functions directly?

server/validate.go
```diff
@@ -106,47 +106,62 @@ func validateGetEntryAndProofRequest(req *trillian.GetEntryAndProofRequest) erro
...
+	nextIndex := req.Leaves[0].LeafIndex
 	for i, leaf := range req.Leaves {
-		if leaf == nil {
-			return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.Leaves[%v] empty", i)
-		}
-		if err := validateLogLeaf(leaf); err != nil {
-			// validateLogLeaf errors are meant to chain nicely with "Leaves."
-			return status.Errorf(codes.InvalidArgument, "QueueLeavesRequest.Leaves[%v].%v", i, err)
+		if leaf.LeafIndex != nextIndex {
+			return status.Errorf(codes.OutOfRange, "%v.Leaves[%v].LeafIndex=%v, want %v", prefix, i, leaf.LeafIndex, nextIndex)
```
Martin2112: This is checking for sequential index? Are we using OutOfRange to be able to distinguish it because it could be FailedPrecondition?

server/log_rpc_server.go
pavelkalinnikov: Yes, fixed.

Actually, `InitLog` should accept `PREORDERED_LOG` in this line as well, but this requires `LogStorage.ReadWriteTransaction` refactoring similar to #1007. Left a TODO.

server/validate.go
pavelkalinnikov: Done.

server/validate.go
pavelkalinnikov: Right, let's make it a `FailedPrecondition`.

pavelkalinnikov:
@AlCutter ping

server/log_rpc_server.go
```diff
@@ -130,13 +143,53 @@ func (t *TrillianLogRPCServer) QueueLeaves(ctx context.Context, req *trillian.Qu
...
+		return nil, err
+	}
+
+	ctx = trees.NewContext(ctx, tree)
+	leaves, err := t.registry.LogStorage.AddSequencedLeaves(ctx, tree.TreeId, req.Leaves)
+	if err != nil {
+		return nil, err
+	}
+	if got, want := len(leaves), len(req.Leaves); got != want {
+		return nil, status.Errorf(codes.Internal, "AddSequencedLeaves returned %d codes, want: %d", got, want)
```
AlCutter: codes->leaves?

pavelkalinnikov: Done.

Issue #950, opened 2018-01-24
Cloud Spanner support


tarcieri:
My understanding is that Google runs Trillian internally on top of Spanner.

Would it make sense for Trillian to support Cloud Spanner for use on GCP?

AlCutter:
Hi Tony,

we'd love to have GCP Cloud Spanner supported by Trillian, but so far we've not had the resource to commit to implementing it within the team. 
You're right that we run on top of Spanner internally, but unfortunately it's not quite as simple as simply shipping out that code and have it work with Cloud Spanner.

If you're interested in helping to build our Cloud Spanner support we'd very much welcome the contributions!

Martin2112:
See #954.

AlCutter:
#1015 is the initial implementation

AlCutter:
#1054 fixes some performance issues with it

tarcieri:
Nice!

Martin2112:
Our CloudSpanner test log has 6.8 billion entries and has shown linear performance since the most recent changes. I think we can close this for now? There is no map implementation but it's possibly too early to write that.

Martin2112:
Tested to 15B entries. Performance seems to be holding up well.

```
curl -S http://ct-spanner.ci.certificate-transparency.org/test/ct/v1/get-sth
{"tree_size":12810013038,"timestamp":1533825412397,"sha256_root_hash":"2lZsmm5zKV4GddO270+HWQUEJitzqJXf3DB15UxqRks=","tree_head_signature":"BAMASDBGAiEAwe+KmAw+q0NyAVKrChKtlZt6mogdEPpAZwm7NBZDUjECIQCNs45tCeMNCNh1szDE3A06klybp/e5xLKJ6sMzXdq5oA=="}
```

PR #1012, opened 2018-02-21
Pass a DB Close func into server/Main instead

AlCutter:
Paving the way for #950


AlCutter:
Yes - it currently means you can only support storage implementations which are accessed through golang's `sql` package :/

PR #1013, opened 2018-02-21
Replace ad-hoc check with DB.Ping

pavelkalinnikov:


storage/mysql/admin_storage.go
```diff
@@ -98,7 +98,7 @@ func (s *mysqlAdminStorage) ReadWriteTransaction(ctx context.Context, f storage.
 }
 
 func (s *mysqlAdminStorage) CheckDatabaseAccessible(ctx context.Context) error {
-	return checkDatabaseAccessible(ctx, s.db)
+	return s.db.Ping()
```
pavelkalinnikov: Probably should use PingContext(ctx) instead. Same below.

pavelkalinnikov:
@Martin2112 Any particular reason why `Ping` was not used before?

pavelkalinnikov:
Oh, I see, you might be checking not only availability but also that the DB has a schema?

Martin2112:
Ping had some odd behaviour in early versions of Go. And yes we were checking that the schema existed though I think I scaled that back.

storage/mysql/admin_storage.go
Martin2112: If we're going to do this it should be PingContext()

pavelkalinnikov:
PTAL.

PR #1032, opened 2018-03-01
Fix docker-compose deployment

pavelkalinnikov:
- Fix hostname in Dockerfiles.
- Grant db access beyond `localhost` as it is accessed from another container.
- Use `mariadb:10.1` image to sync to what is installed by `apt-get` in `seed` container.

### DIFF
--- a/scripts/resetdb.sh
+++ b/scripts/resetdb.sh
@@ -53,7 +53,7 @@ main() {
       echo "Resetting DB..."
       mysql "${FLAGS[@]}" -e "DROP DATABASE IF EXISTS ${DB_NAME};"
       mysql "${FLAGS[@]}" -e "CREATE DATABASE ${DB_NAME};"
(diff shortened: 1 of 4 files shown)


scripts/resetdb.sh
```diff
@@ -53,7 +53,7 @@ main() {
       echo "Resetting DB..."
       mysql "${FLAGS[@]}" -e "DROP DATABASE IF EXISTS ${DB_NAME};"
       mysql "${FLAGS[@]}" -e "CREATE DATABASE ${DB_NAME};"
-      mysql "${FLAGS[@]}" -e "GRANT ALL ON ${DB_NAME}.* TO '${DB_NAME}'@'localhost' IDENTIFIED BY 'zaphod';"
```
pavelkalinnikov: Probably could specify a net-mask here to make it a bit narrower.

Martin2112: This script was originally written for running tests so wasn't particularly strict.

pavelkalinnikov: SGTM, leaving as is.

PR #1033, opened 2018-03-01
Allow InitLog operation on PREORDERED_LOG

pavelkalinnikov:


server/log_rpc_server.go
```diff
@@ -37,6 +37,7 @@ const proofMaxBitLen = 64
 
 var (
 	optsLogRead            = trees.NewGetOpts(trees.Query, true, trillian.TreeType_LOG, trillian.TreeType_PREORDERED_LOG)
+	optsLogInit            = trees.NewGetOpts(trees.Queue, false, trillian.TreeType_LOG, trillian.TreeType_PREORDERED_LOG)
```
Martin2112: trees.Admin perhaps?

pavelkalinnikov: Yup, also fixed `optsPreorderedLogWrite` while I'm here.
I wonder if there is any difference between them apart from naming, seems to work both ways.

Martin2112:
Can you also make the admin change to map_rpc_server if required?

These operation types aren't checked at the moment but they will be soon.

pavelkalinnikov:
@Martin2112 done.

PR #1036, opened 2018-03-01
Implement AddSequencedLeaves in MySQL storage

pavelkalinnikov:
This is a stub which does not yet fully implement the API contract, in particular it doesn't return duplicates if any.


storage/mysql/log_storage.go
```diff
@@ -212,7 +214,7 @@ func (t *readOnlyLogTX) GetActiveLogIDs(ctx context.Context) ([]int64, error) {
 	return ids, rows.Err()
 }
 
-func (m *mySQLLogStorage) beginInternal(ctx context.Context, tree *trillian.Tree) (storage.LogTreeTX, error) {
+func (m *mySQLLogStorage) beginInternal(ctx context.Context, tree *trillian.Tree) (*logTreeTX, error) {
```
Martin2112: Not sure why this changed.

storage/mysql/log_storage.go
```diff
@@ -259,15 +261,66 @@ func (m *mySQLLogStorage) ReadWriteTransaction(ctx context.Context, tree *trilli
 }
 
 func (m *mySQLLogStorage) AddSequencedLeaves(ctx context.Context, tree *trillian.Tree, leaves []*trillian.LogLeaf) ([]*trillian.QueuedLogLeaf, error) {
-	return nil, status.Errorf(codes.Unimplemented, "AddSequencedLeaves is not implemented")
+	tx, err := m.beginInternal(ctx, tree)
+	if err != nil {
+		return nil, err
+	}
+
+	ok := status.New(codes.OK, "OK").Proto()
```
Martin2112: Not sure these really help that much?

storage/mysql/log_storage.go
```diff
@@ -259,15 +261,66 @@ func (m *mySQLLogStorage) ReadWriteTransaction(ctx context.Context, tree *trilli
...
+
+	return res, nil
 }
 
 func (m *mySQLLogStorage) SnapshotForTree(ctx context.Context, tree *trillian.Tree) (storage.ReadOnlyLogTreeTX, error) {
 	tx, err := m.beginInternal(ctx, tree)
 	if err != nil && err != storage.ErrTreeNeedsInit {
 		return nil, err
 	}
-	return tx.(storage.ReadOnlyLogTreeTX), err
```
Martin2112: Does removing this cast change any behaviour such as losing the readonly?

pavelkalinnikov: Now `tx` is again `storage.LogTreeTX`, but I think casting is unnecessary.

storage/mysql/log_storage.go
```diff
@@ -489,6 +502,52 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
 	return existingLeaves, nil
 }
 
+func (t *logTreeTX) AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) ([]*trillian.QueuedLogLeaf, error) {
+	res := make([]*trillian.QueuedLogLeaf, len(leaves))
+	ok := status.New(codes.OK, "OK").Proto()
+
+	// Note: Leaves are sorted by LeafIndex, so no reordering is necessary.
```
pavelkalinnikov: selfnit: ... no **deterministic** reordering is necessary.

pavelkalinnikov:
@AlCutter @daviddrysdale Martin is OOO, could you please take a look?

storage/mysql/log_storage.go
```diff
@@ -259,15 +261,26 @@ func (m *mySQLLogStorage) ReadWriteTransaction(ctx context.Context, tree *trilli
 }
 
 func (m *mySQLLogStorage) AddSequencedLeaves(ctx context.Context, tree *trillian.Tree, leaves []*trillian.LogLeaf) ([]*trillian.QueuedLogLeaf, error) {
```
AlCutter: Is this func (on LogStorage impl) needed, or is this a `LogTreeTX` thing?

pavelkalinnikov: It is both, similarly to `QueueLeaves`. Currently both are transactional, but we discussed earlier that we wanted to allow the `LogStorage` ones treat entries independently. WDYT?

AlCutter: SGTM

storage/mysql/log_storage.go
```diff
@@ -489,6 +502,52 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
...
+		if got, want := len(leaf.LeafIdentityHash), t.hashSizeBytes; got != want {
+			return nil, status.Errorf(codes.FailedPrecondition, "leaves[%d] has incorrect hash size %d, want %d", i, got, want)
+		}
+
+		res[i] = &trillian.QueuedLogLeaf{Status: ok}
+
+		// TODO(pavelkalinnikov): Measure latencies.
+		_, err := t.tx.ExecContext(ctx, insertLeafDataSQL,
+			t.treeID, leaf.LeafIdentityHash, leaf.LeafValue, leaf.ExtraData, 0)
+		// Note: QueueTimestamp == 0 because the entry bypasses the queue.
```
AlCutter: Will that mess with the integration latency metrics? (The signer sets integration_timestamp and computes the difference for each leaf, which it then add to a histogram - _might_ be useful for seeing how the mirroring etc. is going on fetching leaves vs. integrating them?)

pavelkalinnikov: Also, I think we should split the metric in two: one for the `LOG`'s merge delay, another for `PREORDERED_LOG`'s integration delay.

storage/mysql/log_storage.go
```diff
@@ -500,6 +513,52 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
...
+		} else if err != nil {
+			glog.Warningf("Error inserting leaves[%d] into LeafData: %s", i, err)
+			return nil, err
+		}
+
+		_, err = t.tx.ExecContext(ctx, insertSequencedLeafSQL,
+			t.treeID, leaf.LeafIdentityHash, leaf.MerkleLeafHash, leaf.LeafIndex, 0)
+		// TODO(pavelkalinnikov): Update IntegrateTimestamp on integrating the leaf.
+
+		if isDuplicateErr(err) {
```
pavelkalinnikov: Started manually testing, found a bug. Suppose we are trying to erroneously insert a new unique identity to an occupied leaf index. The `INSERT` above stores the identity, but the second `INSERT` fails because the index is occupied. Still, the `tx` gets committed, and the side-effect remains. Now this identity can't be inserted anymore.

I think I should add a test for this, and couple of simpler scenarios.

pavelkalinnikov: Wow, this actually complicates things more than I thought it would. Essentially, we need to do 2 `INSERT`s, each can fail independently of the other (it can be conflicting identity and/or conflicting leaf index). Thus, if one fails, we should cancel the effect of the other to leave the db in the old state. We can do it by either making separate transactions, which is slow, or by manually deleting inserted keys which doesn't look nice.

@AlCutter @Martin2112 Any thoughts?

pavelkalinnikov: The third option is reading before doing the inserts.

Martin2112: Let's look at this again next week. Might be able to use savepoints for this at least in MySQL.

pavelkalinnikov: Yes, this works. Will update the PR shortly.

storage/mysql/log_storage.go
pavelkalinnikov: @AlCutter ping

pavelkalinnikov:
@Martin2112 PTAL.

Martin2112:
Maybe alcutter@ knows but how will we implement this on CloudSpanner? I don't think it has an equivalent of savepoints.

pavelkalinnikov:
I think we can at least do a ReadWrite transaction that first tries to read the identity/entry and does the two inserts only if there is no conflict. Anyway, this would be in a separate PR.

Martin2112:
Yes I'm not suggesting we write the code now. Just checking we have a plan.

storage/mysql/log_storage.go
AlCutter: Sorry, I think splitting the metrics is a good idea.
I guess I was suggesting overloading the `QueueTimestamp`, it's still kinda true in that it's queued to be properly integrated into the tree right?

storage/mysql/log_storage.go
```diff
@@ -501,6 +513,78 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
 	return existingLeaves, nil
 }
 
+func (t *logTreeTX) AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) ([]*trillian.QueuedLogLeaf, error) {
+	res := make([]*trillian.QueuedLogLeaf, len(leaves))
+	ok := status.New(codes.OK, "OK").Proto()
+
+	// Leaves in this transaction are inserted in two tables. For each leaf, if
+	// one of the two inserts fails, we remove the side effect by rolling back to
+	// a savepoint installed before the first insert.
```
Martin2112: I thought the savepoint would be updated each pass through the loop so only the ones that fail are rolled back? Is that not how it's meant to work?

storage/mysql/log_storage.go
```diff
@@ -501,6 +513,78 @@ func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf,
...
+	const savepoint = "SAVEPOINT AddSequencedLeaves"
+	if _, err := t.tx.ExecContext(ctx, savepoint); err != nil {
+		glog.Errorf("Error adding savepoint: %s", err)
+		return nil, err
+	}
+	// TODO(pavelkalinnikov): Consider performance implication of executing this
```
Martin2112: I think the overhead of creating a savepoint is low but we should test this theory.

storage/mysql/log_storage.go
Martin2112: OK I now realize we're still in an all or nothing situation. Ignore that comment.

pavelkalinnikov: Oh, actually we are not in all-or-nothing. Your initial interpretation was right: only the failed-to-insert entries are rolled back. This is why I update the savepoint on each loop iteration below.

Martin2112: OK. It's hard to read GitHub diffs. I'll have another look.

pavelkalinnikov: The reason why I create this savepoint upfront is making sure the "RELEASE SAVEPOINT ..." after the loop has something to delete and doesn't return an error.

Martin2112:
Yep. It's probably better to clean it up like this rather than wait for
commit or rollback.

pavelkalinnikov:
I wonder why the 4th builder in Travis consistently complains about SQL syntax, while others don't.

Martin2112:
That's a different queue implementation for Lets Encrypt with a build flag
to enable it. It's probably a real error.

Martin2112:
I've looked at it again and I think it's doing the right thing. Let's investigate the batched_queue error travis error now.

pavelkalinnikov:
I think the reason is that `insertSequencedLeafSQL` variable gets overridden in `batched_queue` mode by the one in `queue_batching.go` file. Will see how I can fix this.

storage/mysql/log_storage.go
pavelkalinnikov: Done timestamp saving, TODOed the new metric bit.
Yeah, kinda, but not queued in the sense we are used to, rather added/stored. Leaving it as is for now though.

storage/mysql/log_storage.go
pavelkalinnikov: Yes, will get to it later.

pavelkalinnikov:
@Martin2112 I fixed the errors, PTAL. For the performance (with or without savepoints), I will probably add a [Benchmark](https://golang.org/pkg/testing/#hdr-Benchmarks) test in a follow-up.

PR #1038, opened 2018-03-02
GetActiveLogs returns PREORDERED_LOG trees as well

pavelkalinnikov:
The PR fixes this in MySQL and CloudSpanner storages.


storage/log_storage.go
```diff
@@ -156,7 +156,8 @@ type CountByLogID map[int64]int64
 
 // LogMetadata provides access to information about the logs in storage
 type LogMetadata interface {
-	// GetActiveLogs returns a list of the IDs of all the logs that are configured in storage
+	// GetActiveLogs returns a list of the IDs of all the logs that are
```
pavelkalinnikov: nit: GetActiveLogIDs

pavelkalinnikov:
Will see if I can amend some existing tests to verify the behavior.

Martin2112:
Can you do this for cloud spanner as well? We need to keep them both up to date.

storage/log_storage.go
pavelkalinnikov: Done.

storage/mysql/tree_storage.go
```diff
@@ -36,7 +36,7 @@ const (
 	insertSubtreeMultiSQL = `INSERT INTO Subtree(TreeId, SubtreeId, Nodes, SubtreeRevision) ` + placeholderSQL
 	insertTreeHeadSQL     = `INSERT INTO TreeHead(TreeId,TreeHeadTimestamp,TreeSize,RootHash,TreeRevision,RootSignature)
 		 VALUES(?,?,?,?,?,?)`
-	selectNonDeletedTreeIDByTypeAndStateSQL = "SELECT TreeId FROM Trees WHERE TreeType = ? AND TreeState = ? AND (Deleted IS NULL OR Deleted = 'false')"
+	selectNonDeletedTreeIDByTypeAndStateSQL = "SELECT TreeId FROM Trees WHERE TreeType IN (" + placeholderSQL + ") AND TreeState = ? AND (Deleted IS NULL OR Deleted = 'false')"
```
pavelkalinnikov: @Martin2112 Since this query is used only with fixed number of parameters (2), do you think we should just write "IN (?, ?)" instead of `placeholderSQL`? Similarly to [how]( https://github.com/google/trillian/pull/970/files#diff-fda46990892de1b2549108efd05f30e2R45) you did in another RP with `TreeState`.

pavelkalinnikov:
@Martin2112 I will do it in a follow-up PR today.

pavelkalinnikov:
Actually, the cloudspanner fix is quite short, so I'm adding it here.

pavelkalinnikov:
Added test in MySQL storage. PTAL.

storage/mysql/tree_storage.go
pavelkalinnikov: Did it.

storage/log_storage.go
```diff
@@ -156,7 +156,8 @@ type CountByLogID map[int64]int64
 
 // LogMetadata provides access to information about the logs in storage
 type LogMetadata interface {
-	// GetActiveLogs returns a list of the IDs of all the logs that are configured in storage
+	// GetActiveLogIDs returns a list of the IDs of all the logs that are
+	// configured in storage and are in the ACTIVE or DRAINING state.
```
Martin2112: Maybe change this to something like "eligible to have entries sequenced" so it won't get out of date.

Martin2112: LGTM apart from that.

pavelkalinnikov: Done.

PR #1039, opened 2018-03-02
Add log signer to docker-compose deployment

pavelkalinnikov:
This makes it possible to run a fully functional test instance of Trillian with docker-compose.

### DIFF
--- a/examples/deployment/docker-compose.yml
+++ b/examples/deployment/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       - DB_USER=root
       - DB_PASSWORD=$DB_PASSWORD
     command: ./wait-for-it.sh -t 0 mysql:3306 -- ./scripts/resetdb.sh --verbose --force -h mysql
-  trillian-server:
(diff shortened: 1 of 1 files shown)


examples/deployment/docker-compose.yml
```diff
@@ -25,4 +25,15 @@ services:
...
+      context: ../..
+      dockerfile: examples/deployment/docker/log_signer/Dockerfile
+    restart: always # retry while mysql is starting up
+    ports:
+      - "8092:8091"
+    depends_on:
+      - mysql
+    environment:
+      - DB_USER=root
+      - DB_PASSWORD=$DB_PASSWORD
```
gdbelvin: Are you sure this environment variable is being picked up in the way you expect? 
I seem to recall that env variable replacement was a missing feature.

pavelkalinnikov: I think it does. I was able to run this locally. I also tried to manually connect to the db within the mysql container, and it indeed had the correct user/password.

PR #1040, opened 2018-03-02
Integrate sequenced leaves in PREORDERED_LOG

pavelkalinnikov:


log/sequencer.go
```diff
@@ -282,8 +280,25 @@ func (s logSequencingTask) update(ctx context.Context, leaves []*trillian.LogLea
 	return nil
 }
 
+type preorderedLogSequencingTask logSequencingTask
+
+func (s *preorderedLogSequencingTask) fetch(ctx context.Context, limit int, cutoff time.Time) ([]*trillian.LogLeaf, error) {
+	// TODO(pavelkalinnikov): Measure latency.
+	leaves, err := s.tx.GetLeavesByRange(ctx, s.treeSize, int64(limit))
```
pavelkalinnikov: Should probably reword the comment next to `GetLeavesByRange` because it doesn't say returning entries past the tree size is normal for `PREORDERED_LOG`.

log/sequencer.go
```diff
@@ -308,12 +323,19 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 			return storage.ErrTreeNeedsInit
 		}
 
-		var st sequencingTask = logSequencingTask{
+		taskData := &logSequencingTask{
```
AlCutter: This section reads kinda weirdly due to the type definitions and casting below!

log/sequencer.go

log/sequencer.go
```diff
@@ -245,16 +245,16 @@ type sequencingTask interface {
 // logSequencingTask is a sequencingTask implementation for "normal" Log mode,
```
pavelkalinnikov: Oups, should fix the comment.

pavelkalinnikov:
@Martin2112 PTAL.

PR #1043, opened 2018-03-07
Remove LogVerifier interface

gdbelvin:
This follows the general advice to not maintain interfaces just-in-case
or for the purposes of mocking. This interface was previously used to
facilitate a fake.LogVerifier implementation. However, that was bad
practice and has been removed.


gdbelvin:
Key Transparency implemented it. Follow on: google/keytransparency#947

Martin2112:
Yep, I meant other than by us, which I presumed not.

PR #1047, opened 2018-03-08
Remove readonly from getopts as it's implicit in the operation.

Martin2112:
Also rename the test cases so they all follow the same pattern and clean them up a bit.


server/interceptor/interceptor.go
```diff
@@ -155,7 +155,7 @@ func (tp *trillianProcessor) Before(ctx context.Context, req interface{}) (conte
 
 	if info.getTree {
 		tree, err := trees.GetTree(
-			ctx, tp.parent.admin, info.treeID, trees.NewGetOpts(trees.Admin, info.readonly, info.treeTypes...))
+			ctx, tp.parent.admin, info.treeID, trees.NewGetOpts(trees.Admin, info.treeTypes...))
```
pavelkalinnikov: Is the `readonly` flag in `info` needed then? Maybe remove it if not.

Martin2112: Yes, because it's passed to the quota API.

Martin2112:
I think Admin should be read / write and used sparingly.

server/interceptor/interceptor.go
pavelkalinnikov: Ah, ok.

PR #1048, opened 2018-03-09
Use LogRootV1 with SignLogRoot / VerifyLogRoot

gdbelvin:
This PR continues in the theme of #958 by using the `LogRootV1` defined in #1037 to create `*trillian.SignedLogRoot` objects via `SignLogRoot`.  Verifying a `*trillian.SignedLogRoot` will also extract and return the `LogRootV1` object. 

Local references to `trillian.SignedLogRoot` that access it's fields will slowly be replaced with `LogRootV1` objects. 

Note that `LogID` is no longer a part of `LogRootV1`.  In future PRs, `LogID` will be more related to the `signer` than the `LogRootV1` object.

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/client/backoff"
+	"github.com/google/trillian/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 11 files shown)


gdbelvin:
You might know better than me, but it seems to me like it's hitting the right number of tests. 
Many of our tests mock out the signing bits don't they?

Martin2112:
Yes there's a lot of mock signer usage and nothing much is using the signatures at the moment. It's probably OK.

gdbelvin:
LGTM?

PR #1050, opened 2018-03-12
Set MapRoot.Metadata type to bytes

gdbelvin:
This PR uses bytes as the storage format for metadata. 

This allows the MapRoot to have a deterministic serialization which is necessary to implement proper digital signatures -  #958. Bytes also aligns better with the underlying storage formats.


examples/ct/ctmapper/mapper/mapper.go
```diff
@@ -194,8 +199,12 @@ func (m *CTMapper) oneMapperRun(ctx context.Context) (bool, error) {
 	if metaAny, err = ptypes.MarshalAny(mapperMetadata); err != nil {
 		return false, fmt.Errorf("failed to marshal mapper metadata as 'any': err %v", err)
 	}
+	metaBytes, err := proto.Marshal(metaAny)
```
AlCutter: Why the "double wrapping" (i.e. meta->any->[]bytes rather than just `proto.Marshal(meta)`)?

testonly/marshalling.go
```diff
@@ -27,12 +27,16 @@ import (
 // protobuf.ptypes.any.Any used throughout the Trillian API and in
 // storage.  Use if testing.T not available. Failure to marshal will
 // fail the test suite.
-func MustMarshalAnyNoT(in proto.Message) *any.Any {
+func MustMarshalAnyNoT(in proto.Message) []byte {
```
AlCutter: Same here, could this simply be `proto.Marshal(in)`?

examples/ct/ctmapper/mapper/mapper.go
gdbelvin: proto.Marshal would work.  The double wrapping preserves a bit of type checking.

testonly/marshalling.go
gdbelvin: done

gdbelvin:
PTAL

PR #1054, opened 2018-03-15
Limit data read during operation

AlCutter:
Spanner seems to not take an error being returned by a RowIterator func as a reason to cancel/abandon the underlying read, and the encapsulating transaction will not commit until the read has finished and so this causes poor performance, particularly during sequencing.

This change uses `ReadOptions.Limit` to specify the number of rows to read when dequeueing leaves, and switches to SQL for fetching subtree data.

#950

### DIFF
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -578,7 +578,7 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 			break
 		}
 	}
-	if drained > 0 {
+	if drain > 0 {
 		glog.Warningf("Drained %d", drain)
 	}
 	return ret, nil
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -537,12 +537,10 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 	}
 	ret := make([]*trillian.LogLeaf, 0, limit)
 	readOpts := &spanner.ReadOptions{Limit: limit}
-	drain := 0
 	for _, rs := range rows {
 		rows := tx.stx.ReadWithOptions(ctx, unseqTable, rs, []string{colBucket, colQueueTimestampNanos, colMerkleLeafHash, colLeafIdentityHash}, readOpts)
 		if err := rows.Do(func(r *spanner.Row) error {
 			if len(ret) >= limit {
-				drain++
 				return nil
 			}
 			var l trillian.LogLeaf
@@ -577,9 +575,6 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 		if len(ret) >= limit {
 			break
 		}
-	}
-	if drain > 0 {
-		glog.Warningf("Drained %d", drain)
 	}
 	return ret, nil
 }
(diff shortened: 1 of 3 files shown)


storage/cloudspanner/log_storage.go
```diff
@@ -536,11 +536,14 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 		rows = genDequeueRowsOld(tx.getLogStorageConfig(), tx.treeID, now.Unix(), rand.Int31n(numByteValues))
 	}
 	ret := make([]*trillian.LogLeaf, 0, limit)
+	readOpts := &spanner.ReadOptions{Limit: limit}
+	drain := 0
```
Martin2112: 'discarded' maybe? Wasn't immediately clear what it meant.

storage/cloudspanner/log_storage.go
```diff
@@ -572,6 +578,9 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 			break
 		}
 	}
+	if drain > 0 {
+		glog.Warningf("Drained %d", drain)
```
Martin2112: Is this bad? Can you expand on the message if necessary.

storage/cloudspanner/spanner.sdl
```diff
@@ -29,10 +29,7 @@ CREATE TABLE TreeHeads(
   RootSignature           BYTES(1024) NOT NULL,
   TreeRevision            INT64 NOT NULL,
   TreeMetadata            BYTES(2097152),
-) PRIMARY KEY(TreeID, TimestampNanos DESC);
-
-CREATE UNIQUE INDEX TreeRevisionIndex
-  ON TreeHeads(TreeID, TreeRevision DESC);
+) PRIMARY KEY(TreeID, TreeRevision DESC);
```
Martin2112: Just for the release notes this looks like an incompatible change?

storage/cloudspanner/tree_storage.go
```diff
@@ -391,12 +408,10 @@ func (t *treeTX) getSubtree(ctx context.Context, rev int64, id storage.NodeID) (
...
 		}
-		// We've got what we want, tell spanner to stop reading by returning
-		// not-really-an-error:
-		return errFinished
+		return nil
 	})
-	if err == errFinished {
-		err = nil
+	if drain > 0 {
+		glog.Warningf("getSubtree drained %d", drain)
```
Martin2112: Not sure how you can get this given the `LIMIT 1` per subtree?

storage/cloudspanner/log_storage.go
AlCutter: I'll remove it - I was just verifying that it did what I expected.

storage/cloudspanner/log_storage.go
AlCutter: removed

storage/cloudspanner/spanner.sdl
AlCutter: Technically yes, but presumably nobody other than me has an instance of it :)

storage/cloudspanner/tree_storage.go
AlCutter: removed - I was just verifying that.

AlCutter:
PTAL

storage/cloudspanner/spanner.sdl
Martin2112: Added to draft release notes in case.

PR #1059, opened 2018-03-19
Remove unused vmap toy

gdbelvin:
`vmap/toy` and `vmap/rpctoy` seem unused and are acquiring bit rot.
If they have any value, they should be moved to the integration tests.


Martin2112:
Another option is to move them to the examples repo, if they're still useful.

gdbelvin:
It wasn't super clear to me if the vmap stuff was still useful.  @AlCutter ?

Martin2112:
It was a very early demo thing.

AlCutter:
Probably less useful these days, would be great to have a command-line tool for getting/setting/verifying leaves etc. though.

PR #1061, opened 2018-03-19
[storage/mysql] Tolerate gaps in GetLeavesByRange

pavelkalinnikov:
`PREORDERED_LOG` uses `GetLeavesByRange` to load sequenced entries beyond the tree size. It is a normal situation when those entries are not contiguous (`AddSequencedLeaves` does not enforce the order). This PR makes `GetLeavesByRange` return a shorter range in such scenario, instead of an error.

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -234,7 +234,7 @@ func (m *mySQLLogStorage) beginInternal(ctx context.Context, tree *trillian.Tree
 	}
 
 	stCache := cache.NewLogSubtreeCache(defaultLogStrata, hasher)
-	ttx, err := m.beginTreeTx(ctx, tree.TreeId, hasher.Size(), stCache)
+	ttx, err := m.beginTreeTx(ctx, tree, hasher.Size(), stCache)
 	if err != nil && err != storage.ErrTreeNeedsInit {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


storage/mysql/log_storage.go
```diff
@@ -566,6 +566,7 @@ func (t *logTreeTX) GetLeavesByIndex(ctx context.Context, leaves []int64) ([]*tr
 }
 
 func (t *logTreeTX) GetLeavesByRange(ctx context.Context, start, count int64) ([]*trillian.LogLeaf, error) {
+	// TODO(pavelkalinnikov): Clip `count`, for example to max(count, 64k).
```
pavelkalinnikov: nit: min(count, 64k)

pavelkalinnikov: Done.

storage/mysql/log_storage.go
```diff
@@ -609,11 +609,10 @@ func (t *logTreeTX) GetLeavesByRange(ctx context.Context, start, count int64) ([
 			return nil, fmt.Errorf("got invalid integrate timestamp: %v", err)
 		}
 		ret = append(ret, leaf)
-		wantIndex++
 	}
 
 	if len(ret) == 0 {
-		return nil, status.Errorf(codes.InvalidArgument, "no leaves found in range [%d, %d+%d)", start, start, count)
+		return nil, status.Errorf(codes.InvalidArgument, "no contiguous prefix of leaves found in [%d, %d+%d)", start, start, count)
```
pavelkalinnikov: I have an impression this shouldn't be an error. For example, if the request is `(start=treeSize, count=100)`, looks reasonable to return an empty slice?

pavelkalinnikov: Also, this it probably not `InvalidArgument` anymore.

Martin2112:
For safety can we limit this change to PREORDERED_LOG trees? It will probably need a small tweak to retain the tree type in the tx.

pavelkalinnikov:
I am not sure what you mean by safety in this case? I think we should rather make the API consistent regardless of the tree type. How about the following:
- If some entry **within** the tree size is missing, this should be an error regardless of the tree type.
- Any gaps **beyond** the size are tolerable regardless of the tree type.

Since normally `LOG` have neither gaps, nor entries beyond the tree size, the above should be equivalent to what we have currently, modulo returning empty slice instead of an error if the range is too far to the right.

Martin2112:
If there are gaps in a LOG tree somehow then it's probably too late to save it but the server should detect it. Alternatively, we could enforce that a log tree can never request a range outside of the tree and the not worry about gaps.

Possibly the CTFE checks on the response are sufficient but we should try to exclude the possibility of bad data propagating if we can.

pavelkalinnikov:
Maybe also cap `count` to `treeSize-start` for `LOG` trees for safety, if that is what you mean. This may, however, involve reading from `Trees` table just for getting the tree type...

Martin2112:
We should get the tree type via the trillian.Tree when the transaction starts.

pavelkalinnikov:
@Martin2112 Improved safety a bit, PTAL.

storage/mysql/log_storage.go
```diff
@@ -567,11 +567,26 @@ func (t *logTreeTX) GetLeavesByIndex(ctx context.Context, leaves []int64) ([]*tr
 
 func (t *logTreeTX) GetLeavesByRange(ctx context.Context, start, count int64) ([]*trillian.LogLeaf, error) {
 	if count <= 0 {
-		return nil, fmt.Errorf("invalid count %d", count)
+		return nil, fmt.Errorf("count=%d,want>0", count)
```
Martin2112: Not sure about changing these. I think they might get returned in server responses but not 100% certain.

pavelkalinnikov: Yes, it gets returned in servers responses. Should I leave it intact? I thought a bit clarified messages would be better.

Martin2112:
I don't really mind. This change is more Go style but users might be using
another language to make RPCs.

storage/mysql/log_storage.go
pavelkalinnikov: Made it slightly more human readable.

PR #1063, opened 2018-03-20
Refactor createLogForTests

pavelkalinnikov:
- returns `trillian.Tree` directly
- accepts tree-to-be-created parameter
- signed log root is created independently
- `createMapForTests` is no longer needed

This change allows conveniently creating `PREORDERED_LOG` trees for testing by simply substituting `testonly.PreorderedLogTree` instead of `testonly.LogTree` to the `createLogForTests` function (renamed to `createTreeOrPanic`).

Additionally, now most of the tests use the `Tree` returned by `CreateTree` API instead of artificially constructed ones (see `logTree` and `mapTree` functions) when passing in to functions like `ReadWriteTransaction`.

### DIFF
--- a/storage/mysql/storage_test.go
+++ b/storage/mysql/storage_test.go
@@ -40,13 +40,8 @@ import (
 
 func TestNodeRoundTrip(t *testing.T) {
 	cleanTestDB(DB)
-	logID := createLogForTests(DB)
+	tree := createLogForTests(DB, storageto.LogTree)
 	s := NewLogStorage(DB, nil)
-	tree := &trillian.Tree{
-		TreeId:       logID,
-		TreeType:     trillian.TreeType_LOG,
-		HashStrategy: trillian.HashStrategy_RFC6962_SHA256,
-	}
 
 	const writeRevision = int64(100)
 	nodesToStore := createSomeNodes()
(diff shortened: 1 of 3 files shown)


storage/mysql/storage_test.go
```diff
@@ -292,6 +263,14 @@ func createTree(db *sql.DB, tree *trillian.Tree) (*trillian.Tree, error) {
 	return tree, nil
 }
 
+func createTreeOrPanic(db *sql.DB, create *trillian.Tree) *trillian.Tree {
+	tree, err := createTree(db, create)
+	if err != nil {
+		panic(fmt.Sprintf("Error creating tree: %v", err))
```
pavelkalinnikov: Might make more sense to pass in `*testing.T` and calling `t.Fatalf` here. Same for other panicking function in this file.

pavelkalinnikov: Not in the scope of this PR though.

Martin2112: Removing panics would be better. For the moment this should probably be called mustCreateTree() or similar.

pavelkalinnikov: Not sure I follow. By removing panics do you mean the following?
```
if err != nil {
	fmt.Sprintf("Error creating tree: %v", err)
}
```
... or removing panics and adding `t.Fatalf` as I suggested?

PR #1066, opened 2018-03-22
Allow switching TreeType from PREORDERED_LOG to LOG

pavelkalinnikov:
This change enables switching from the `Tree` metadata perspective. Additional support in `storage.UpdateTree` is required to avoid the situation when the type is changed while there are pending `SequencedLeafData` entries.

### DIFF
--- a/storage/mysql/admin_storage.go
+++ b/storage/mysql/admin_storage.go
@@ -61,6 +61,10 @@ const (
 		FROM Trees`
 	selectNonDeletedTrees = selectTrees + nonDeletedWhere
 	selectTreeByID        = selectTrees + " WHERE TreeId = ?"
+
+	updateTreeSQL = `UPDATE Trees
+		SET TreeState = ?, TreeType = ?, DisplayName = ?, Description = ?, UpdateTimeMillis = ?, MaxRootDurationMillis = ?, PrivateKey = ?
+		WHERE TreeId = ?`
 )
 
 // NewAdminStorage returns a MySQL storage.AdminStorage implementation backed by DB.
@@ -472,6 +476,10 @@ func (t *adminTX) UpdateTree(ctx context.Context, treeID int64, updateFunc func(
 		return nil, err
 	}
 
+	// TODO(pavelkalinnikov): When switching TreeState to FROZEN for a
+	// PREORDERED_LOG, ensure there are no gaps in SequencedLeafData, in order to
+	// allow smoothly changing TreeType to LOG.
+
 	// Use the time truncated-to-millis throughout, as that's what's stored.
 	nowMillis := toMillisSinceEpoch(time.Now())
 	now := fromMillisSinceEpoch(nowMillis)
@@ -489,11 +497,7 @@ func (t *adminTX) UpdateTree(ctx context.Context, treeID int64, updateFunc func(
 		return nil, fmt.Errorf("could not marshal PrivateKey: %v", err)
 	}
 
-	stmt, err := t.tx.PrepareContext(
-		ctx,
-		`UPDATE Trees
-		SET TreeState = ?, DisplayName = ?, Description = ?, UpdateTimeMillis = ?, MaxRootDurationMillis = ?, PrivateKey = ?
-		WHERE TreeId = ?`)
+	stmt, err := t.tx.PrepareContext(ctx, updateTreeSQL)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


storage/mysql/admin_storage.go
```diff
@@ -472,6 +476,10 @@ func (t *adminTX) UpdateTree(ctx context.Context, treeID int64, updateFunc func(
 		return nil, err
 	}
 
+	// TODO(pavelkalinnikov): When switching TreeState to FROZEN for a
```
pavelkalinnikov: @Martin2112 Does storage check `Unsequenced` table emptiness during a `DRAINING`->`FROZEN` transition, or it just trusts the caller of the API?

Martin2112: No. Checking this can be prohibitively expensive on a large MySQL database. Distributed databases have other problems with this type of query as it hits all the shards.

pavelkalinnikov: I assume you are talking about the counting query, like `"SELECT COUNT(*) FROM Unsequenced/SequencedLeafData"`?

However, I think that we can afford even heavy requests in the scenario when we try to switch from `DRAINING` to `FROZEN` with caution. The reason is - the tree is inactive (if called at the right moment). Checking would be just a consistency protection mechanism.

pavelkalinnikov: Hm, on the other hand I don't see why a tree can't be FROZEN even if it has some extra entries in `Unsequenced/SequencedLeafData`. Probably this check should be only in the case of changing the log type.

Martin2112: I think you should create a large database of say 600 million log entries and test queries like that while the log is serving requests. It's possible that an EXISTS query might work but I'd not bet on it without trying.

storage/tree_validation.go
```diff
@@ -77,7 +77,13 @@ func ValidateTreeForUpdate(ctx context.Context, storedTree, newTree *trillian.Tr
 	case storedTree.TreeId != newTree.TreeId:
 		return status.Error(codes.InvalidArgument, "readonly field changed: tree_id")
 	case storedTree.TreeType != newTree.TreeType:
-		return status.Error(codes.InvalidArgument, "readonly field changed: tree_type")
+		if got, want := storedTree.TreeType, trillian.TreeType_PREORDERED_LOG; got != want {
+			return status.Errorf(codes.InvalidArgument, "can't change tree_type from %v, only %v", got, want)
+		} else if got, want := newTree.TreeType, trillian.TreeType_LOG; got != want {
+			return status.Errorf(codes.InvalidArgument, "can't change tree_type to %v, only %v", got, want)
+		} else if got, want := storedTree.TreeState, trillian.TreeState_FROZEN; got != want {
+			return status.Errorf(codes.InvalidArgument, "can't change tree_type: tree_state=%v, want %v", got, want)
```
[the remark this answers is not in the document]
pavelkalinnikov: I isolated this code in a function. Then I decided to restrict the transition rule: `TreeType` can be changed if the `Tree` is and **remains** in `FROZEN` state. That eliminates `from_state` from the transition table you mentioned.

I tried table approach locally, it didn't seem worthwhile to me in this case where we have only 1 transition.

storage/tree_validation.go
```diff
@@ -64,6 +64,21 @@ func ValidateTreeForCreation(ctx context.Context, tree *trillian.Tree) error {
 	return validateMutableTreeFields(ctx, tree)
 }
 
+func validateTreeTypeUpdate(oldTree, newTree *trillian.Tree) error {
+	const prefix = "changing tree_type"
```
Martin2112: I still find it hard to look at this code and decide if it's doing the right thing. I think the documentation of what it's validating should be included here.

storage/tree_validation.go
```diff
@@ -64,6 +64,21 @@ func ValidateTreeForCreation(ctx context.Context, tree *trillian.Tree) error {
...
+	const wantState = trillian.TreeState_FROZEN
+	if oldState := oldTree.TreeState; oldState != wantState {
```
Martin2112: Maybe the first two ifs could be combined? Then it might be clearer that the frozen state can't changed, with the following if that allows a preordered_log -> log type change.

storage/tree_validation.go
pavelkalinnikov: Done.

storage/tree_validation.go
pavelkalinnikov: Done to some extent. I deliberately do 2 ifs for the TreeState checks to make error reporting detailed.

pavelkalinnikov:
ping

PR #1068, opened 2018-03-26
Check for CTFE imports into generic repo.

Martin2112:
As discussed earlier. A proposal to ensure we don't accidentally bring CT related code into the generic repo.

### DIFF
--- a/scripts/check_unforked.sh
+++ b/scripts/check_unforked.sh
@@ -14,6 +14,16 @@ check_import() {
   fi
 }
 
+check_blacklist() {
+  local path="$1"
+
+  local result=$(grep -Hne '\("github.com/google/certificate-transparency-go/trillian/ctfe.*"\)' "$path")
+  if [[ ! -z "${result}" ]]; then
+    echo "$result - importing CTFE into generic code is not allowed"
+    return 1
+  fi
+}
+
 main() {
   if [[ $# -lt 1 ]]; then
     echo "Usage: $0 <path>"
@@ -29,9 +39,11 @@ main() {
           continue  # Empty glob
         fi
         check_import "$f" || code=1
+        check_blacklist "$f" || code=1
       done
     else
       check_import "$path" || code=1
+      check_blacklist "$path" || code=1
     fi
     shift
   done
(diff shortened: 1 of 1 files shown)


scripts/check_unforked.sh
```diff
@@ -14,6 +14,16 @@ check_import() {
   fi
 }
 
+check_blacklist() {
+  local path="$1"
+
+  local result=$(grep -Hne '\("github.com/google/certificate-transparency-go/trillian/ctfe.*"\)' "$path")
+  if [[ ! -z "${result}" ]]; then
+    echo "$result - importing CTFE into generic code is not allowed"
```
daviddrysdale: maybe a smidgeon more specific: "importing CT-specific code into generic Trillian code is not allowed"

scripts/check_unforked.sh
```diff
@@ -14,6 +14,16 @@ check_import() {
   fi
 }
 
+check_blacklist() {
+  local path="$1"
+
+  local result=$(grep -Hne '\("github.com/google/certificate-transparency-go/trillian/ctfe.*"\)' "$path")
```
daviddrysdale: maybe disallow anything under `c-t-go/trillian/` ?

scripts/check_unforked.sh
Martin2112: Done.

scripts/check_unforked.sh
Martin2112: Done.

Martin2112:
Sorry, I saw this about 30 seconds after merging it. I think a whitelist is
harder as the blacklist just does an egrep for lines that match.

Martin2112:
Ah, no I merged the other one. Too many PRs. I'll hold off for a bit.

daviddrysdale:
If we shift to a whitelist with `tls` it would also need `asn1` at some point.  And until the `examples/ct/` directory migrates to trillian-examples there would be other things needed too.  So I think a blacklist is OK for now.

Martin2112:
OK I think I'll merge this then and we can revisit it if necessary.

PR #1072, opened 2018-03-28
Fix cloud spanner breakage

AlCutter:
CloudSpanner doesn't have `uint64` types so need to cast to `int64` before trying to store these values into `INT64` columns.


AlCutter:
Thanks for merging!

PR #1075, opened 2018-04-03
Use SQL for dequeuing leaves in cloudspanner

AlCutter:


storage/cloudspanner/log_storage.go
```diff
@@ -519,55 +457,51 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 	// moment, FEs queueing entries will be adding them to different buckets
 	// than we're dequeuing from here.
 	now := time.Now().UTC()
-	var rows []spanner.KeySet
+	cfg := tx.getLogStorageConfig()
+	startBucket := int64((now.Unix()+cfg.NumUnseqBuckets/2)%cfg.NumUnseqBuckets) << 8
```
Martin2112: Maybe comment this to describe the layout of the bits, which would explain the shifting and masking.

AlCutter: ta, done.

PR #1076, opened 2018-04-03
Flags to make signing a bit faster

AlCutter:


Martin2112:
Didn't know you could speed things up by adding flags :-)

PR #1102, opened 2018-04-23
[LogVerifier] Introduce PrefixHashFromInclusionProof

pavelkalinnikov:
The new method calculates root hash of a smaller tree which is a prefix
of a bigger tree based on an inclusion proof of the boundary entry to
the bigger tree. If the passed in root hash of the bigger tree is
trusted, then the result of this function is likewise trusted.

This function can be used, for example, in monitoring applications to
verify intermediate root hashes of a log that is being copied. If, say,
a log publishes its root hash for 1 million entries, it is not necessary
to download the lot to verify only a prefix of 100 entries. Instead,
PrefixHashFromInclusionProof can be utilized.

### DIFF
--- a/merkle/log_verifier.go
+++ b/merkle/log_verifier.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math/bits"
 
 	"github.com/google/trillian/merkle/hashers"
 )
@@ -211,4 +212,35 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
 	}
 
 	return nil // Proof OK.
+}
+
+// PrefixHashFromInclusionProof calculates a root hash over leaves [0..index),
+// based on the inclusion |proof| and |leafHash| for a leaf at the specified
+// |index| in a tree of the specified |size| with the passed in |root| hash.
+// Accepts 0 <= index < size.
+// Returns an error if the proof verification fails.
+func (v LogVerifier) PrefixHashFromInclusionProof(
+	index, size int64,
+	proof [][]byte, root []byte, leafHash []byte,
+) ([]byte, error) {
+	if err := v.VerifyInclusionProof(index, size, proof, root, leafHash); err != nil {
+		return nil, err
+	}
+	if index == 0 {
+		return v.hasher.EmptyRoot(), nil
+	}
+
+	var res []byte
+	tail := bits.Len64(uint64(index ^ (size - 1)))
+	for i, h := range proof {
+		if i < tail && (index>>uint(i))&1 == 0 {
+			continue
+		}
+		if res == nil {
+			res = h
+		} else {
+			res = v.hasher.HashChildren(h, res)
+		}
+	}
+	return res, nil
 }
--- a/merkle/log_verifier_test.go
+++ b/merkle/log_verifier_test.go
@@ -125,7 +125,7 @@ func verifierCheck(v *LogVerifier, leafIndex, treeSize int64, proof [][]byte, ro
 		return err
 	}
 	if want := root; !bytes.Equal(got, want) {
-		return fmt.Errorf("got root:\n%x\nexpected:\n%x", got, want)
+		return fmt.Errorf("got root:\n%s\nexpected:\n%s", shortHash(got), shortHash(want))
 	}
 	if err := v.VerifyInclusionProof(leafIndex, treeSize, proof, root, leafHash); err != nil {
 		return err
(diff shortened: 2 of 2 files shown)


merkle/log_verifier_test.go
```diff
@@ -416,10 +416,60 @@ func TestVerifyConsistencyProof(t *testing.T) {
 	}
 }
 
+func TestPrefixHashFromInclusionProofGenerated(t *testing.T) {
+	sizes := make([]int64, 0, 17)
+	for s := 2; s <= 17; s++ {
```
pavelkalinnikov: nit: Can start this from 1.

merkle/log_verifier.go
```diff
@@ -212,3 +213,34 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
 
 	return nil // Proof OK.
 }
+
+// PrefixHashFromInclusionProof calculates a root hash over leaves [0..index),
+// based on the inclusion |proof| and |leafHash| for a leaf at the specified
+// |index| in a tree of the specified |size| with the passed in |root| hash.
+// Accepts 0 <= index < size.
+// Returns an error if the proof verification fails.
+func (v LogVerifier) PrefixHashFromInclusionProof(
```
Martin2112: The name should probably include Verify for consistency. Or take the verify step out and make it the callers responsibility.

merkle/log_verifier.go
```diff
@@ -212,3 +213,34 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
...
+) ([]byte, error) {
+	if err := v.VerifyInclusionProof(index, size, proof, root, leafHash); err != nil {
+		return nil, err
+	}
+	if index == 0 {
+		return v.hasher.EmptyRoot(), nil
+	}
+
+	var res []byte
+	tail := bits.Len64(uint64(index ^ (size - 1)))
```
Martin2112: I think the go style is not to have spaces between operators.

merkle/log_verifier_test.go
Martin2112: Might need some tests for invalid inputs.

pavelkalinnikov: Done.

merkle/log_verifier.go
AlCutter: Yeah, gofmt has a bunch of heuristics around when/how to apply spaces around operators to do with grouping to show precedence and perhaps complexity of the surrounding context.

merkle/log_verifier.go
```diff
@@ -212,3 +213,35 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
...
 }
+
+// VerifiedPrefixHashFromInclusionProof calculates a root hash over leaves
+// [0..index), based on the inclusion |proof| and |leafHash| for a leaf at the
+// specified |index| in a tree of the specified |size| with the passed in
+// |root| hash.
+// Returns an error if the |proof| verification fails. The resulting smaller
+// tree's root hash is trusted iff the bigger tree's |root| hash is trusted.
+func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
+	index, size int64,
```
AlCutter: Would it be clearer to talk always in terms of sizes rather than size and index since we're really dealing with a smaller tree being a prefix of a larger tree?

merkle/log_verifier.go
```diff
@@ -212,3 +213,35 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
...
+	proof [][]byte, root []byte, leafHash []byte,
+) ([]byte, error) {
+	if err := v.VerifyInclusionProof(index, size, proof, root, leafHash); err != nil {
+		return nil, err
+	}
+	if index == 0 {
```
AlCutter: Do you want this above the call to `VerifyInclusionProof`?

merkle/log_verifier.go
```diff
@@ -212,3 +213,35 @@ func (v LogVerifier) VerifyConsistencyProof(snapshot1, snapshot2 int64, root1, r
...
+		return v.hasher.EmptyRoot(), nil
+	}
+
+	var res []byte
+	tail := bits.Len64(uint64(index ^ (size - 1)))
```
AlCutter: Perhaps add a comment to explain why/what this value is?

merkle/log_verifier_test.go
```diff
@@ -416,10 +416,102 @@ func TestVerifyConsistencyProof(t *testing.T) {
...
+	proof := make([][]byte, len(desc))
+	for i, d := range desc {
+		proof[i] = d.Value.Hash()
+	}
+	leafHash := tree.LeafHash(index + 1)
+	return leafHash, proof
+}
+
+func TestPrefixHashFromInclusionProofGenerated(t *testing.T) {
+	sizes := make([]int64, 0, 18)
```
AlCutter: I'd be tempted to run this up against a much larger number - we've all been bitten by Merkle tree bugs which only appeared at unexpectedly large tree sizes :)

pavelkalinnikov:
@Martin2112 Any more concerns here?

merkle/log_verifier_test.go
pavelkalinnikov: Done.

merkle/log_verifier.go
pavelkalinnikov: Also `VerifyInclusionProof` conveniently does some sanity checking of `index` and `size` params.

pavelkalinnikov:
ping

merkle/log_verifier.go
Martin2112: The danger of relying on other code for sanity checking is if it gets refactored the checks can get dropped without realizing it. Don't think that's a big issue here but it can be.

PR #1105, opened 2018-04-26
[LogVerifier] Factor out hash chaining

pavelkalinnikov:
This PR simplifies proof verification methods in `LogVerifier` by introducing `hashChainer` - the type used for combining proof hashes in various ways depending on how the corresponding (sub-)path is related to the tree or query.
This PR also adds exhaustive testing of inclusion and consistency proof verification on generated in-memory Merkle Trees.

### DIFF
--- a/merkle/log_verifier_test.go
+++ b/merkle/log_verifier_test.go
@@ -345,6 +345,28 @@ func TestVerifyInclusionProof(t *testing.T) {
 
 }
 
+func TestVerifyInclusionProofGenerated(t *testing.T) {
+	var sizes []int64
+	for s := 1; s <= 70; s++ {
+		sizes = append(sizes, int64(s))
+	}
+	sizes = append(sizes, []int64{1024, 5050}...)
+
+	tree, v := createTree(0)
+	for _, size := range sizes {
+		growTree(tree, size)
+		root := tree.CurrentRoot().Hash()
+		for i := int64(0); i < size; i++ {
+			t.Run(fmt.Sprintf("size:%d:index:%d", size, i), func(t *testing.T) {
+				leaf, proof := getLeafAndProof(tree, i)
+				if err := verifierCheck(&v, i, size, proof, root, leaf); err != nil {
+					t.Errorf("verifierCheck(): %v", err)
+				}
+			})
+		}
+	}
+}
+
 func TestVerifyConsistencyProof(t *testing.T) {
 	v := NewLogVerifier(rfc6962.DefaultHasher)
 
--- a/merkle/log_verifier_test.go
+++ b/merkle/log_verifier_test.go
@@ -433,7 +433,31 @@ func TestVerifyConsistencyProof(t *testing.T) {
 			roots[snapshot1-1].h,
 			roots[snapshot2-1].h, proof)
 		if err != nil {
-			t.Fatalf("Failed to verify known good proof for i=%d: %s", i, err)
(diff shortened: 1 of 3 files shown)


merkle/log_verifier_test.go
```diff
@@ -411,7 +433,31 @@ func TestVerifyConsistencyProof(t *testing.T) {
...
+		sizes = append(sizes, int64(s))
+	}
+	sizes = append(sizes, 74)
+
+	for _, size := range sizes {
+		tree, v := createTree(size)
+		for i := int64(0); i <= size; i++ {
+			for j := i; j <= size; j++ {
+				t.Run(fmt.Sprintf("size:%d:consistency:%d-%d", size, i, j), func(t *testing.T) {
+					proof := rawProof(tree.SnapshotConsistency(i, j))
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

merkle/log_verifier_test.go
```diff
@@ -411,7 +433,27 @@ func TestVerifyConsistencyProof(t *testing.T) {
...
+	tree, v := createTree(size)
+	roots := make([][]byte, size+1)
+	for i := int64(0); i <= size; i++ {
+		roots[i] = tree.RootAtSnapshot(i).Hash()
+	}
+
+	for i := int64(0); i <= size; i++ {
+		for j := i; j <= size; j++ {
+			proof := rawProof(tree.SnapshotConsistency(i, j))
+			t.Run(fmt.Sprintf("size:%d:consistency:%d-%d", size, i, j), func(t *testing.T) {
```
pavelkalinnikov: Should we worry about 8515 sub-tests here? If something breaks, this will probably generate quite a big log.

pavelkalinnikov:
@Martin2112 PTAL.

merkle/log_verifier_test.go
Martin2112: It might be an issue, it's many more than we usually have. I don't know TBH.

Martin2112:
Not sure about this change. The original structure looked somewhat like the C++ code so people could compare it. Not sure if we actually need that though.

Martin2112:
This is a good general question for @AlCutter. In the older implementations there were some attempts to go from the RFC definitions of proof paths to code and we've carried some of this over. Maybe we're beyond that now. What do you think?

pavelkalinnikov:
@AlCutter  I may be repeating Martin's words, but:
1. Do we actually need to compare Go code with legacy C++ code which many people may even not be aware of? Especially non-CT users.
2. With regards to RFC, [6962](https://tools.ietf.org/html/rfc6962) does not list any implementation details, I can only see recursive mathematical definitions of proofs. [6962-bis](https://datatracker.ietf.org/doc/draft-ietf-trans-rfc6962-bis/?include_text=1) does indeed have some reference pseudo-code for proof verification, but marked as "... the following algorithm **may** be used ...". Moreover, the key idea in this PR is the same - some bit-wise mojo, only I factored out hash chaining in nice abstractions.

Since Merkle Trees are well known and quite simple, I don't think there is a need to standardize its implementation. Moreover, it has some useful properties that the old RFC has not foreseen. I am preparing a dependent PR on top of this one which introduces "range inclusion proofs" based on 2 boundary inclusion proofs. I can try to untangle that PRs though.

AlCutter:
Could you TODO (and later to) a Merkle tree doc with pics so that people coming along can grok the various path components these new methods provide?

PR #1106, opened 2018-04-26
Use http.DefaultServeMux as the top-level HTTP handler

RJPercival:
`http.DefaultServeMux` may already have some handlers registered, e.g. by `net/http/pprof`, and we don't want to lose those. We can register the grpc-gateway mux as a default handler underneath it, and the Prometheus handler for "/metrics".


Martin2112:
That's a fair point. I think as long as we haven't broken metrics and health checks we can address the other stuff later.

PR #1108, opened 2018-04-27
Range inclusion verification based on 2 inclusion proofs

pavelkalinnikov:
This is a proof-of-concept implementation of range inclusion verification.
Such functionality will be particularly useful in mirroring to Trillian's `PREORDERED_LOG`. With the current Trillian and CT APIs it is only possible to do one of the following:
1. Use `GetEntryAndProof` to obtain and verify individual entries.
2. Use regular `GetLeavesBy*` batch APIs for obtaining multiple unverified entries, put them in Trillian and hope the final STH will match the public one (after we reach at least its tree size which can be millions).

After this PR it is possible to verify **batches** having only 2 boundary entries' inclusion proofs. If, after such batch-level verification, Trillian's STH does not match the public one, we know w.h.p. that the problem is on Trillian's side.

### DIFF
--- a/merkle/hash_stack.go
+++ b/merkle/hash_stack.go
@@ -1,0 +1,42 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package merkle
+
+import "github.com/google/trillian/merkle/hashers"
+
+type hashStack struct {
+	hasher hashers.LogHasher
+	hashes [][]byte
+	begin  int64
+	index  int64
+}
+
+func newHashStack(hasher hashers.LogHasher, begin int64) hashStack {
+	return hashStack{hasher: hasher, begin: begin, index: begin}
+}
+
+func (s *hashStack) pushAndChain(h []byte) {
+	pop, index, ln := 0, s.index, len(s.hashes)
+	for one := int64(1); index&one != 0; one <<= 1 {
+		index ^= one
+		if index < s.begin || pop >= ln {
+			break
+		}
+		pop++
+		h = s.hasher.HashChildren(s.hashes[ln-pop], h)
+	}
+	s.hashes = append(s.hashes[:ln-pop], h)
+	s.index++
+}
--- a/merkle/log_verifier.go
+++ b/merkle/log_verifier.go
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
 	}
 	return res, nil
 }
+
+// VerifyRangeInclusion verifies that [begin, end) range, consisting of the
+// |leaves| with the specified hashes, is included into the tree with the
+// specified |size| and |root| hash, based on inclusion proofs for entries
+// number |begin| and |end-1| respectively.
+//
+// TODO(pavelkalinnikov): Current implementation has some level of redundancy,
+// as the combination of 2 inclusion proofs is a superset of the minimal range
+// inclusion proof. Consider adding proper range inclusion proof notion to
+// Trillian's API.
+func (v LogVerifier) VerifyRangeInclusion(
+	begin, end, size int64,
+	proof1, proof2 [][]byte,
+	root []byte, leaves [][]byte,
+) error {
+	if end <= begin {
+		return fmt.Errorf("empty range [%d,%d)", begin, end)
+	} else if got, want := int64(len(leaves)), end-begin; got != want {
+		return fmt.Errorf("wrong number of leaves %d, want %d", got, want)
+	}
+	if err := v.VerifyInclusionProof(begin, size, proof1, root, leaves[0]); err != nil {
+		return err
+	} else if err := v.VerifyInclusionProof(end-1, size, proof2, root, leaves[len(leaves)-1]); err != nil {
+		return err
+	}
+	if end-begin <= int64(2) {
+		return nil // Inclusion proofs verification above is sufficient.
+	}
+
+	inner1, border1 := decompInclProof(begin, size)
+	inner2, border2 := decompInclProof(end-1, size)
+	if inner2 >= inner1 { // Either inner2 == inner1 and border2 == border2 ...
+		inner1 = innerProofSize(begin, end)
+		inner2, border2 = inner1, 0
+	} else { // ... or inner2 < inner1 and border2 > border1.
+		border2 -= border1
+	}
+	inner1--
+	if border2 == 0 {
+		inner2--
+	} else {
+		border2--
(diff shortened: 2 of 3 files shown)


merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+	} else if err := v.VerifyInclusionProof(end-1, size, proof2, root, leaves[len(leaves)-1]); err != nil {
+		return err
+	}
+	if end-begin <= int64(2) {
+		return nil // Inclusion proofs verification above is sufficient.
+	}
+
+	inner1, border1 := decompInclProof(begin, size)
+	inner2, border2 := decompInclProof(end-1, size)
+	if inner2 >= inner1 { // Either inner2 == inner1 and border2 == border2 ...
```
pavelkalinnikov: border2 == border1

merkle/hash_stack.go
```diff
@@ -0,0 +1,42 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package merkle
+
+import "github.com/google/trillian/merkle/hashers"
+
+type hashStack struct {
```
gdbelvin: Please add documentation for types and methods, even if they are not public

merkle/hash_stack.go
```diff
@@ -0,0 +1,42 @@
...
+	begin  int64
+	index  int64
+}
+
+func newHashStack(hasher hashers.LogHasher, begin int64) hashStack {
+	return hashStack{hasher: hasher, begin: begin, index: begin}
+}
+
+func (s *hashStack) pushAndChain(h []byte) {
+	pop, index, ln := 0, s.index, len(s.hashes)
```
gdbelvin: is pop == level?
is ln == number of levels?

merkle/hash_stack.go
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package merkle
+
+import "github.com/google/trillian/merkle/hashers"
+
+type hashStack struct {
+	hasher hashers.LogHasher
+	hashes [][]byte
```
gdbelvin: Which hashes are these?

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+// |leaves| with the specified hashes, is included into the tree with the
+// specified |size| and |root| hash, based on inclusion proofs for entries
+// number |begin| and |end-1| respectively.
+//
+// TODO(pavelkalinnikov): Current implementation has some level of redundancy,
+// as the combination of 2 inclusion proofs is a superset of the minimal range
+// inclusion proof. Consider adding proper range inclusion proof notion to
+// Trillian's API.
+func (v LogVerifier) VerifyRangeInclusion(
+	begin, end, size int64,
```
gdbelvin: is size tree_size?

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+	proof1, proof2 [][]byte,
+	root []byte, leaves [][]byte,
```
gdbelvin: It might be simper to supply leaves via `map[int64][]byte`. It would remove 2 arguments from the function signature, make the method more general, and avoid off-by one errors lining up the indexes between begin,end, and the leaves slice.

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+		return fmt.Errorf("empty range [%d,%d)", begin, end)
+	} else if got, want := int64(len(leaves)), end-begin; got != want {
+		return fmt.Errorf("wrong number of leaves %d, want %d", got, want)
+	}
+	if err := v.VerifyInclusionProof(begin, size, proof1, root, leaves[0]); err != nil {
+		return err
+	} else if err := v.VerifyInclusionProof(end-1, size, proof2, root, leaves[len(leaves)-1]); err != nil {
+		return err
+	}
+	if end-begin <= int64(2) {
```
gdbelvin: `2` might not need the typecast

merkle/log_verifier_test.go
```diff
@@ -482,6 +482,76 @@ func TestPrefixHashFromInclusionProofErrors(t *testing.T) {
...
+	tree, v := createTree(size)
+	root := tree.CurrentRoot().Hash()
+	leaves := make([][]byte, end-begin)
+	for i := begin; i < end; i++ {
+		leaves[i-begin] = tree.LeafHash(i + 1)
+	}
+	_, proof1 := getLeafAndProof(tree, begin)
+	_, proof2 := getLeafAndProof(tree, end-1)
+
+	if err := v.VerifyRangeInclusion(begin, end, size, proof1, proof2, root, leaves); err != nil {
```
gdbelvin: Could this test be rewritten in table test format?

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+// inclusion proof. Consider adding proper range inclusion proof notion to
+// Trillian's API.
+func (v LogVerifier) VerifyRangeInclusion(
+	begin, end, size int64,
+	proof1, proof2 [][]byte,
+	root []byte, leaves [][]byte,
+) error {
+	if end <= begin {
+		return fmt.Errorf("empty range [%d,%d)", begin, end)
+	} else if got, want := int64(len(leaves)), end-begin; got != want {
```
AlCutter: Can remove the `else` here

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+		return fmt.Errorf("wrong number of leaves %d, want %d", got, want)
+	}
+	if err := v.VerifyInclusionProof(begin, size, proof1, root, leaves[0]); err != nil {
+		return err
+	} else if err := v.VerifyInclusionProof(end-1, size, proof2, root, leaves[len(leaves)-1]); err != nil {
```
AlCutter: this `else` can go too (I think it's less dense and easier to read without, do you agree?)

merkle/log_verifier.go
AlCutter: Should we be asserting these comments with code?

merkle/log_verifier.go
```diff
@@ -248,3 +248,101 @@ func (v LogVerifier) VerifiedPrefixHashFromInclusionProof(
...
+
+	inner1, border1 := decompInclProof(begin, size)
+	inner2, border2 := decompInclProof(end-1, size)
+	if inner2 >= inner1 { // Either inner2 == inner1 and border2 == border2 ...
+		inner1 = innerProofSize(begin, end)
+		inner2, border2 = inner1, 0
+	} else { // ... or inner2 < inner1 and border2 > border1.
+		border2 -= border1
+	}
+	inner1--
```
AlCutter: I'm gonna need pictures again, my brain is hurting :)

gdbelvin:
Is this ready for a second look yet?

pavelkalinnikov:
No, have been doing other stuff. Will ping you when this is ready of another review.

PR #1110, opened 2018-04-27
Idempotent subtree@revision writes

AlCutter:
This PR short circuits writes to subtree data when the data being written is identical to the data already present.  It goes some way to dealing with #1109.

Due to the nature of Trillian's storage design, a write to any given node is always preceded with a read to fetch the previous revision of the subtree which contains that node.  This allows the `subtreeCache` an opportunity to ignore the write when the application is simply attempting to re-write the previous value.

This should not cause any issues, since when reading nodes, Trillian will always use the most recent revision subtree, which will contain the same data.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -149,8 +149,7 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
 		pxKey := string(px)
 		// TODO(al): fix for non-uniform strata
 		id.PrefixLenBits = len(px) * depthQuantum
-		_, ok := s.subtrees[pxKey]
-		if !ok {
+		if _, ok := s.subtrees[pxKey]; !ok {
 			want[pxKey] = &id
 		}
 	}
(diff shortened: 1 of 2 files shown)


storage/cache/subtree_cache.go
```diff
@@ -147,9 +147,9 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
 		id := id
 		px, _ := s.splitNodeID(id)
 		pxKey := string(px)
-		_, ok := s.subtrees[pxKey]
 		// TODO(al): fix for non-uniform strata
 		id.PrefixLenBits = len(px) * depthQuantum
+		_, ok := s.subtrees[pxKey]
```
Martin2112: if blah; !ok {
}

storage/cache/subtree_cache.go
```diff
@@ -326,15 +326,23 @@ func (s *SubtreeCache) SetNodeHash(id storage.NodeID, h []byte, getSubtree GetSu
 	if c.Prefix == nil {
 		return fmt.Errorf("nil prefix for %v (key %v)", id.String(), prefixKey)
 	}
-	s.dirtyPrefixes[prefixKey] = true
 	// Determine whether we're being asked to store a leaf node, or an internal
 	// node, and store it accordingly.
 	sfxKey := sx.String()
 	if int32(sx.Bits) == c.Depth {
+		// Short-circuit write if we're simply overwriting an identical value.
```
Martin2112: Suggest rewording this for morer clarity. How about "Cache state is unchanged if we're overwriting an identical value" or summat?

Martin2112:
The title of the PR makes it sounds like writes were previously not idempotent. But wasn't it just doing more work than necessary or is it more subtle?

storage/cache/subtree_cache.go
AlCutter: done

storage/cache/subtree_cache.go
AlCutter: That's a bit too opaque in my view, that makes it sound a bit like the only change is in the cache (and there's a silent implication, _if_ you know the code well enough, that no writes will be performed).

AlCutter:
Well, they previously weren't idempotent in the case were map updates were only partially applied - this is really the reason for this PR :)

storage/cache/subtree_cache.go
Martin2112: OK I just thought "short circuit" could be ambiguous.

Martin2112:
Perhaps a tweak to the PR title?

PR #1111, opened 2018-04-30
Use close-ended semantics in VerifiedPrefixHashFromInclusionProof

pavelkalinnikov:
This removes the corner case for empty tree, plus simplifies the hash calculation loop.
Will also allow for some simplifications in #1105.


Martin2112:
Looks ok.

PR #1112, opened 2018-04-30
Add log/map- server health check

AlCutter:


Martin2112:
Great minds think alike: https://github.com/Martin2112/trillian/tree/rpc_server_healthcheck. You've done the kubernetes config as well though.

server/main.go
```diff
@@ -80,6 +85,18 @@ type Main struct {
 	TreeDeleteMinInterval time.Duration
 }
 
+func (m *Main) healthz(rw http.ResponseWriter, req *http.Request) {
+	if m.IsHealthy != nil {
+		if err := m.IsHealthy(req.Context()); err != nil {
```
Martin2112: I think (could be wrong) that the request context defaults to context.Background() so might need an explicit deadline here.

examples/deployment/kubernetes/trillian-map-deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             command:
             - curl
             - -f
-            - http://localhost:8091/metrics
+            - http://localhost:8091/healthz
           failureThreshold: 3
           periodSeconds: 30
           timeoutSeconds: 30
```
Martin2112: 30 seconds is prob. too long given we only have 10? sec for live requests.

server/main.go
```diff
@@ -80,6 +85,18 @@ type Main struct {
...
+			rw.WriteHeader(http.StatusServiceUnavailable)
+			rw.Write([]byte(err.Error()))
+			return
+		}
+	}
+	rw.WriteHeader(http.StatusOK)
```
Martin2112: Pretty sure there's an implicit 200 status if not overridden.

examples/deployment/kubernetes/trillian-map-deployment.yaml
AlCutter: done

server/main.go
AlCutter: done

server/main.go
AlCutter: Done

PR #1123, opened 2018-05-03
[LogVerifier] Check hash size, add regression test

pavelkalinnikov:
Previously verification could succeed if `root` and `leafHash` were both empty (or, more generally, incorrect size).


Martin2112:
Just to be completely clear. This changes the behaviour if the supplied hashes are the wrong length and does not affect cases where proofs are empty?

pavelkalinnikov:
Correct.

PR #1135, opened 2018-05-10
Refactor backoff algorithm

pavelkalinnikov:


client/backoff/backoff.go
```diff
@@ -17,42 +17,44 @@ package backoff
...
-// 0 and factor^x * min, capped by Backoff.Max.
+// Duration returns the time to wait on current retry iteration.
+// Every time Duration is called, the returned value will exponentially
+// increase by Factor until Backoff.Max. If Jitter is enabled, will wait an
+// additional random value between 0 and Factor^x * Min, capped by Backoff.Max.
 func (b *Backoff) Duration() time.Duration {
-	// min( min * factor ^ x , max)
-	minNanos := float64(b.Min.Nanoseconds())
-	maxNanos := float64(b.Max.Nanoseconds())
-	nanos := minNanos * math.Pow(b.Factor, float64(b.x))
```
pavelkalinnikov: This line was unwanted for 2 reasons:
- `nanos` could grow uncontrollably, and eventually lose precision
- `math.Pow` wasn't necessary, as the next iteration was one multiplication away from the previous.

pavelkalinnikov:
PTAL.

client/backoff/backoff.go
```diff
@@ -22,37 +22,55 @@ import (
...
+// if 0 < Min <= Max <= 2^62 (nanosec), and Factor >= 1.
 type Backoff struct {
-	Min    time.Duration
-	Max    time.Duration
-	Factor float64
-	Jitter bool
-	x      int
+	Min    time.Duration // Duration of the first pause.
+	Max    time.Duration // Max duration of a pause.
+	Factor float64       // The factor of duration increase between iterations.
```
gdbelvin: Note that factor must be positive

client/backoff/backoff.go
```diff
@@ -22,37 +22,55 @@ import (
...
+// additional random value between 0 and Factor^x * Min, capped by Backoff.Max.
 func (b *Backoff) Duration() time.Duration {
-	// min( min * factor ^ x , max)
-	minNanos := float64(b.Min.Nanoseconds())
-	maxNanos := float64(b.Max.Nanoseconds())
-	nanos := minNanos * math.Pow(b.Factor, float64(b.x))
-	nanos = math.Min(nanos, maxNanos)
+	pause := b.Min + b.delta
+
+	newPause := time.Duration(float64(pause) * b.Factor)
```
gdbelvin: Add a comment that this is computing the *next* pause, not the current one.

gdbelvin: Perhaps try to create some visual separation in this function between the current pause and the next one

client/backoff/backoff.go
```diff
@@ -22,37 +22,55 @@ import (
...
 func (b *Backoff) Reset() {
-	b.x = 0
+	b.delta = 0
+}
+
+// Recover reduces retry pause duration by the given power of Backoff.Factor.
+// This might be useful when calling Retry multiple times, to gradually adapt
+// to the rate of errors: in some use-cases a successfully returned Retry could
+// indicate that a bigger flow of requests can now be handled.
+func (b *Backoff) Recover(factors int) {
```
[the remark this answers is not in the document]
gdbelvin: This feels like the TCP congestion algorithm...  :-)

I suggest allowing `Recover` to not take any arguments. That way it can be called once per successful return. 
Each call of `Recover` should reduce the current latency by a linear factor, not an exponential one.  The desired result is to exponentially backoff during errors, and linearly speed up to find the right performance ceiling.

client/backoff/backoff_test.go
```diff
@@ -51,6 +51,49 @@ func TestBackoff(t *testing.T) {
...
+		Min:    time.Duration(1),
+		Max:    time.Duration(100),
+		Factor: 2,
+	}
+	for _, test := range []struct {
+		times    int
+		recTimes int
+		want     time.Duration
+	}{
+		{1, 0, 2},
```
gdbelvin: Please always use keys in go structs

client/backoff/backoff.go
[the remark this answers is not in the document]
pavelkalinnikov: Looks like this has to be thought through, let me clean this PR to be only a refactoring. I will create a separate PR after some testing.

PR #1137, opened 2018-05-14
Allow creating LogVerifier for PREORDERED_LOG trees

pavelkalinnikov:
Plus some commenting.

### DIFF
--- a/client/log_verifier.go
+++ b/client/log_verifier.go
@@ -28,7 +28,8 @@ import (
 	tcrypto "github.com/google/trillian/crypto"
 )
 
-// LogVerifier contains state needed to verify output from Trillian Logs.
+// LogVerifier contains state needed to verify output from Trillian Logs
+// (regular and pre-ordered ones).
 type LogVerifier struct {
 	// Hasher is the hash strategy used to compute nodes in the Merkle tree.
 	Hasher hashers.LogHasher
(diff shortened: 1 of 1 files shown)


client/log_verifier.go
```diff
@@ -52,8 +53,10 @@ func NewLogVerifier(hasher hashers.LogHasher, pubKey crypto.PublicKey, sigHash c
 // NewLogVerifierFromTree creates a new LogVerifier using the algorithms
 // specified by *trillian.Tree.
 func NewLogVerifierFromTree(config *trillian.Tree) (*LogVerifier, error) {
-	if got, want := config.TreeType, trillian.TreeType_LOG; got != want {
-		return nil, fmt.Errorf("client: NewLogVerifierFromTree(): TreeType: %v, want %v", got, want)
+	got := config.TreeType
+	want1, want2 := trillian.TreeType_LOG, trillian.TreeType_PREORDERED_LOG
```
Martin2112: Not sure the want1, want2 thing helps. You could make a map here and do one test maybe.

pavelkalinnikov: Renamed `want`s to `log` and `pLog`, hope this is now more readable. W.r.t. map, I'd prefer not to because of unnecessary overhead it would create (or maybe not, I'm not sure how well Go optimizes constant maps).

PR #1138, opened 2018-05-14
Refuse to sign out of order log roots

rolandshoemaker:
When using multiple log_signers (and probably in other strange circumstances) it is possible that one node has a clock that has drifted significantly out of range of the other nodes. This can cause signing of log roots that have out of order timestamps which breaks trillian in interesting ways (mainly because the timestamp is used to determine which is the most recent log root).

This change refuses to sign a log root that has a timestamp that is earlier than the timestamp of the current log root and adds a test for this behavior.

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -430,6 +430,12 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 			Revision:       uint64(newVersion),
 		}
 
+		if currentRoot.RootHash != nil && newLogRoot.TimestampNanos < currentRoot.TimestampNanos {
+			err := fmt.Errorf("refusing to sign root with timestamp earlier than previous root (%d < %d)", newLogRoot.TimestampNanos, currentRoot.TimestampNanos)
+			glog.Warningf("%v: %s", tree.TreeId, err)
+			return err
+		}
+
 		newSLR, err = s.signer.SignLogRoot(newLogRoot)
 		if err != nil {
 			glog.Warningf("%v: signer failed to sign root: %v", tree.TreeId, err)
--- a/log/sequencer_test.go
+++ b/log/sequencer_test.go
@@ -68,6 +68,16 @@ var (
 			RootHash:       testRoot16.RootHash,
 		})
 
+	testRoot17 = &types.LogRootV1{
+		TreeSize: 16,
+		Revision: 5,
+		// RootHash can't be nil because that's how the sequencer currently
+		// detects that there was no stored tree head.
+		RootHash:       []byte{},
+		TimestampNanos: uint64(fakeTimeForTest.Add(10 * time.Millisecond).UnixNano()),
+	}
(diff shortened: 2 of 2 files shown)


rolandshoemaker:
Thinking about this the check could probably be done much earlier in the function, I just stuck it where it made some sense initially, ¯\_(ツ)_/¯.

Martin2112:
This is presumably in response to the issue you saw with the test log? I'm surprised clocks can be out of sync on production machines but this is something we tend to take for granted at Google.

If something went wrong with replication rather than time sync this might not be able to fix it. However, it seems like it's worth having.

log/sequencer.go
```diff
@@ -430,6 +430,12 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 			Revision:       uint64(newVersion),
 		}
 
+		if currentRoot.RootHash != nil && newLogRoot.TimestampNanos < currentRoot.TimestampNanos {
```
AlCutter: I think `currentRoot.RootHash` can't be nil here (it gets checked at L334 and the signer will refuse to run a batch until after the operator has manually called `InitLog`).

Might be worth updating the timestamp test to `<=` though; I suspect most storage impls will have timestamp as part of the key and would refuse to store a 2nd STH with identical timestamp, but you never know - and in practice generating two STH with identical nanosecond timestamp should never occur legitimately.

AlCutter:
You _could_ do it earlier, though you'd then want to commit to the timestamp that you'll use when making the STH later on in the func at/just before the check [because if the clock drifts backwards between the check and the creation of the STH you'll day have a bad :)].
You could probably argue it either way in as early out on error vs. keeping vars declared close to usage...
Having it where it is only really incurs the cost of wasted work in off-path situations, though, so it doesn't seem so bad to me.

rolandshoemaker:
> This is presumably in response to the issue you saw with the test log? I'm surprised clocks can be out of sync on production machines but this is something we tend to take for granted at Google.

Yup, this occurred on our test infrastructure on AWS, all of our nodes were synced using their NTP servers. It looks like one of the nodes clock was about ~30 seconds out of step. We're still not 100% of the cause (all of the clocks now appear to be perfectly synced) but wanted to get this fix out.

log/sequencer_test.go
```diff
@@ -487,7 +487,7 @@ func TestIntegrateBatch(t *testing.T) {
 				merkleNodesSet:      &updatedNodes,
 				skipStoreSignedRoot: true,
 			},
-			errStr: "refusing to sign root with timestamp earlier than previous root (1464173705000000000 < 1464173705010000000)",
+			errStr: "refusing to sign root with timestamp earlier than previous root (1464173705000000000 <= 1464173705010000000)",
```
AlCutter: Should we have a test for the `==` case too?

rolandshoemaker: Sure.

AlCutter:
Thanks, Rob!

PR #1147, opened 2018-05-24
`gcloud container node-pools create` requires ${CLUSTER_NAME}

DazWilkin:
Either

As here, set `gcloud config container/cluster`

Or:

My preference (and true for `project` and `compute/zone`) is to set the flag explicitly on each gcloud use:

`gcloud container node-pools create ${NAME} --cluster=${CLUSTER_NAME} --project=${PROJECT_NAME} --zone=${ZONE}

Personal preference.


AlCutter:
I'm good with either approach :)

PR #1154, opened 2018-05-26
Reduce image sizes using multi-stage builds

DazWilkin:
* Docker multi-stage builds helps reduce the image sizes from 1.2GB --> ~40MB
* Swapped `ENTRYPOINT` to the exec form which is preferred; it just runs the default (`/trillian_log_signer`)
* Deleted the arguments provided to `ENTRPOINT` as these appear (!) all provided during the Kubernetes deployment
* Thus deleted the ENV settings in the Dockerfile as redundant

### DIFF
--- a/examples/deployment/docker/log_signer/Dockerfile
+++ b/examples/deployment/docker/log_signer/Dockerfile
@@ -1,34 +1,15 @@
-FROM golang:1.9
-
-ENV DB_FLAG="--mysql_uri=test:zaphod@tcp(mysql:3306)/test" \
-    DB_PROVIDER="mysql"
-
-ENV HOST=0.0.0.0 \
-    HTTP_PORT=8091
-
-ENV SEQUENCER_GUARD_WINDOW=0s \
-    FORCE_MASTER=true \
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Did you make sure the [docker-compose](https://github.com/google/trillian/blob/master/examples/deployment/docker-compose.yml) deployment will also work with the modified docker files?

DazWilkin:
I did not.

It's been slightly challenging for me to understand intent as the code is
basically the documentation.

IIUC the Docker Compose and MySQL came first and has been complemented by
the Kubernetes configs that also include a switch to Spanner?

I think the Dockerfile refactoring is good not only because it
significantly reduces the images size but because it also removes static
config from the images that is now overridden by Kubernetes.

Propose: keep the Docker images minimal as proposed as I'll take a look at
ensuring Compose adds in its config specificity as needed.

Thoughts?

>

DazWilkin:
Ah! [sorry] wait-for-it issues the resetdb.sh script ... so I'm exploring
why that appears not to work.

And -- I think -- simplistically, none of my containers is able to connect
to the db.

>

PR #1160, opened 2018-05-28
Proposed Google Container Builder file

DazWilkin:
Using [Container Builder](https://cloud.google.com/container-builder/) would automate builds of container images (here `trillian_log_server`, `trillian_log_signer`) that could be triggered by commits to (e.g. `master`) on this repo.

Using Container Builder could extend the project's CI/CD and -- more importantly -- could facilitate adoption of Trillian with "known good" container images hosted in an authoritative Container Registry e.g. `gcr.io/trillian/log_server`

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,40 @@
+steps:
+- name: "gcr.io/cloud-builders/go"
+  args: [
+    "get",
+    "-t",
+#    "-u",
+    "-v",
+    "./..."
+  ]
+  env: ["PROJECT_ROOT=github.com/google/trillian"]
+- name: "gcr.io/cloud-builders/go"
+  args: [
+    "get",
+    "./server/trillian_log_server"
+  ]
+  env: ["PROJECT_ROOT=github.com/google/trillian"]
+- name: "gcr.io/cloud-builders/go"
+  args: [
+    "get",
+    "./server/trillian_log_signer"
+  ]
+  env: ["PROJECT_ROOT=github.com/google/trillian"]
+- name: "gcr.io/cloud-builders/docker"
+  args: [
+    "build",
+    "--file=examples/deployment/docker/log_server/Dockerfile",
+    "--tag=us.gcr.io/${PROJECT_ID}/log_server:${COMMIT_SHA}",
+    "."
+  ]
+- name: "gcr.io/cloud-builders/docker"
+  args: [
+    "build",
+    "--file=examples/deployment/docker/log_signer/Dockerfile",
+    "--tag=us.gcr.io/${PROJECT_ID}/log_signer:${COMMIT_SHA}",
+    "."
+  ]
+images: [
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,40 @@
...
+  args: [
+    "get",
+    "./server/trillian_log_server"
+  ]
+  env: ["PROJECT_ROOT=github.com/google/trillian"]
+- name: "gcr.io/cloud-builders/go"
+  args: [
+    "get",
+    "./server/trillian_log_signer"
+  ]
```
RJPercival: Why run `go get` for these two binaries? This config doesn't appear to do anything with them.

DazWilkin: The docker builds in the subsequent steps use these to build container images that are pushed to GCR.

DazWilkin:
Coming from the outside of the project, it's not always entirely clear what all the moving pieces do.

With the Kubernetes and Docker Compose deployments of Trillian, it's redundant for every developer to pull and build images that they could source directly from an authoritative Trillian repository.

The proposal was that you (Trillian) could use Container Builder to help automated your CI/CD and a result could be definitive container images e.g. `gcr.io/trillian/log_server:${TAG}`.

These images could then be sourced from the Docker Compose and Kubernetes scripts. At present, precursive shell scripts do the builds for each deployment.

cloudbuild.yaml
RJPercival: Those Dockerfiles run `go get` for these binaries themselves; it doesn't need to be done in advance.
https://github.com/google/trillian/blob/9c0d0c07e9f5dee99b02ae2db7adef4ce03ac9d4/examples/deployment/docker/log_server/Dockerfile#L6

DazWilkin: Ah, right!

Since you have a CI script, shall I just delete the PR?

RJPercival: This might be a cleaner way of doing most of what the script does, since I see you can [use kubectl from this file](https://cloud.google.com/container-builder/docs/configuring-builds/build-test-deploy-artifacts#deploying_artifacts) as well. Would you be up for trying to translate that script into a cloudbuild.yaml file? If not, feel free to delete the PR and I'll take a look at it myself.

RJPercival:
Yea I can empathise with the pains of figuring out where all the pieces of a new project are and what they're doing; it's easy to miss bits. There's certainly value in providing definitive container images. In fact, I need exactly that for something I'm working on. I'm happy to configure a trigger to create a new image on gcr.io for each tag and point it at this cloudbuild.yaml file.

AlCutter:
Should we push on with this and merge it?

RJPercival:
Not as-is. The top half of the file doesn't really do anything - the compiled binaries are discarded.

RJPercival:
This ended up being done by PRs #1214 and #1352.

PR #1161, opened 2018-05-29
[LogOperation] Introduce and use cancelable Sleep

pavelkalinnikov:
Cancelable `Sleep` reduces time needed for the server to shut down / clean up in cases when some canceled jobs are blocked on sleeping.

### DIFF
--- a/util/time.go
+++ b/util/time.go
@@ -1,0 +1,42 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"context"
+	"time"
+)
+
+// Sleep sleeps for at least the specified duration. Returns false iff done is
+// closed before the timer fires.
+func Sleep(done <-chan struct{}, dur time.Duration) bool {
+	timer := time.NewTimer(dur)
+	select {
+	case <-timer.C:
+		return true
+	case <-done:
+		timer.Stop()
+		return false
+	}
+}
+
+// SleepContext sleeps for at least the specified duration. Returns ctx.Err()
+// iff the context is canceled before the timer fires.
+func SleepContext(ctx context.Context, dur time.Duration) error {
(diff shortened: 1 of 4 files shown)


util/time.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Sleep sleeps for at least the specified duration. Returns false iff done is
+// closed before the timer fires.
+func Sleep(done <-chan struct{}, dur time.Duration) bool {
+	timer := time.NewTimer(dur)
+	select {
+	case <-timer.C:
+		return true
+	case <-done:
+		timer.Stop()
```
taknira: Out of interest, why does the timer get stopped in this case, but not in the other case?  Is the timer going to continue to be used in the other case?  Doesn't look like it can be as it's local to this function...

util/time_test.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, tc := range cases {
+		t.Run(fmt.Sprintf("%v:%v", tc.dur, tc.timeout), func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), tc.timeout)
+			if tc.cancel {
+				cancel()
+			} else {
+				defer cancel()
+			}
+			if got, want := SleepContext(ctx, tc.dur), tc.wantErr; got != want {
+				t.Errorf("Sleep() returned %v, want %v", got, want)
```
taknira: s/Sleep/SleepContext/

util/time.go
pavelkalinnikov: `Stop`, as it says in its doc, prevents the timer from firing. In the `<-timer.C` case it has already fired, so no need to prevent it. However, I might be wrong in thinking that not calling `Stop` is okay in terms of cleaning up timer resources. Couldn't quickly find info on how to do that properly.

util/time_test.go
pavelkalinnikov: Done.

util/time.go
Martin2112: The docs say you need to call stop to release resources. Is there a leak here?

pphaneuf: Alternate implementation idea: use `sleepCtx := context.WithTimeout(ctx, dur)`, and just block on `sleep.Ctx.Done()`? Then you can just return `ctx.Err()`, it will be nil if it was just the `sleepCtx` that finished, which is exactly what you want.

pphaneuf:
Also, check out the recommendation in the "Short, but representative names" section here:

https://rakyll.org/style-packages/

pavelkalinnikov:
@pphaneuf Is there a specific reason you put this link here? Do you mean `util` is a too generic name for a package?

taknira:
util is too generic a name for a package - see https://github.com/golang/go/wiki/CodeReviewComments#package-names.  Sorry, I should have picked up on this during review!

pavelkalinnikov:
I agree. Although, the folder was named `util` before this RP, so the renaming could be done separately.

PR #1166, opened 2018-05-30
Scripted "Next steps" to make it simpler

DazWilkin:


examples/deployment/kubernetes/README.md
```diff
@@ -50,12 +50,39 @@ curl -X POST ${LOG_URL}/v1beta1/logs/${tree_id}:init
...
+```
+kubectl port-forward ${LOG_POD} 8091:8091
+```
+1. Create a Trillian tree:
+```
+RESPONSE=$(curl \
+--silent \
+--request POST \
+http://localhost:8091/v1beta1/trees \
+--data '{ "tree":{ "tree_state":"ACTIVE", "tree_type":"LOG", "hash_strategy":"RFC6962_SHA256", "signature_algorithm":"ECDSA", "max_root_duration":"0", "hash_algorithm":"SHA256" }, "key_spec":{ "ecdsa_params":{ "curve":"P256" } } }')
```
Martin2112: Probably don't want root duration 0 if this tree is going to be used for CT later. Maybe set this to 1 hour.

examples/deployment/kubernetes/README.md
```diff
@@ -50,12 +50,39 @@ curl -X POST ${LOG_URL}/v1beta1/logs/${tree_id}:init
...
+1. Initialize it:
+```
+curl \
+--request POST \
+http://localhost:8091/v1beta1/logs/${TREE_ID}:init
+```
 
-**NOTE: none of the Trillian APIs are exposed to the internet with this config,
-this is intentional since the only access to Trillian should be via a
+**NOTE: none of the Trillian APIs is exposed to the Internet with this config.
```
AlCutter: This scans weirdly now because "APIs" is plural, did you mean to change are -> is/did you want to change APIs -> API too?

AlCutter: None of the cool kids capitalise `internet` [these days](https://en.wikipedia.org/wiki/Capitalization_of_%22Internet%22#Usage_examples) :)

examples/deployment/kubernetes/README.md
DazWilkin: I read then lost this guidance, sorry.

I'll update the post.

I have almost no idea how Trillian works, I'm pretty much resorting to re-engineering stuff that I encounter. So my apologies for my continued ignorance.

When you say "1 hour", what's the unit of `max_root_duration` (seconds? hours?).

examples/deployment/kubernetes/README.md
DazWilkin: Ack... 

Let's revert to "none of the Trillians APIS are"

I'll never yield to "on-premise" as a neologism for "on-premises" but happy (thanks!) to be upgraded on "internet". Does this mean it's not "the internet" anymore either then? I'm so old (school).

examples/deployment/kubernetes/README.md
Martin2112: It's a Duration flag so you include the unit. You should be able to set it to `1h`.

examples/deployment/kubernetes/README.md
AlCutter: Oke doke :)

I'm old school too, and will stand shoulder to shoulder with you against "on-premise" _on the premise_ that it's Plain Wrong.

FWIW, personally I say "the internet" too, although interestingly enough in peninsular Spanish it seems to be just "internet" rather than "el internet". Go figure :)

examples/deployment/kubernetes/README.md
DazWilkin: Ack. Thanks!

mhutchinson:
This PR is super old now. Is there any value in here that should be merged, or time to close it?

AlCutter:
I think the commands in here no-longer work since the HTTP RPC endpoints were removed, so I'll close this :/

PR #1186, opened 2018-06-01
Explicitly glog.Flush() in all binaries

pavelkalinnikov:
From `glog` [documentation](https://github.com/golang/glog/blob/23def4e6c14b4da8ac2ed8007337bc5eb5007998/glog.go#L35): 
```
// Log output is buffered and written periodically using Flush. Programs
// should call Flush before exiting to guarantee all log output is written.
```

### DIFF
--- a/cmd/createtree/main.go
+++ b/cmd/createtree/main.go
@@ -155,6 +155,7 @@ func newRequest() (*trillian.CreateTreeRequest, error) {
 
 func main() {
 	flag.Parse()
+	defer glog.Flush()
 
 	if *configFile != "" {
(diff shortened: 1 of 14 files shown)


pavelkalinnikov:
Sure, sounds like a reasonable thing to do.

pavelkalinnikov:
`git grep` says we only have `os.Exit` in 5 `*_test.go` files, the rest is hidden in `vendor` repos that I couldn't quickly imagine being triggered. I guess this PR is fine then.

Martin2112:
Yes. There used to be more but I think we've removed them then.

PR #1190, opened 2018-06-05
Client.AddSequencedLeafAndWait

gdbelvin:
Adds `AddSequencedLeafAndWait` to the client and uses it in some tests.

### DIFF
--- a/client/log_client_test.go
+++ b/client/log_client_test.go
@@ -36,17 +36,17 @@ func TestAddGetLeaf(t *testing.T) {
 
 // addSequencedLeaves is a temporary stand-in function for tests until the real API gets built.
 func addSequencedLeaves(ctx context.Context, env *integration.LogEnv, client *LogClient, leaves [][]byte) error {
-	// TODO(pavelkalinnikov): Replace with AddSequencedLeaves batch API.
+	if len(leaves) == 0 {
+		return nil
+	}
 	for i, l := range leaves {
 		if err := client.AddSequencedLeaf(ctx, l, int64(i)); err != nil {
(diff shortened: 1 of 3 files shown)


client/log_client_test.go
```diff
@@ -36,13 +37,15 @@ func TestAddGetLeaf(t *testing.T) {
 // addSequencedLeaves is a temporary stand-in function for tests until the real API gets built.
 func addSequencedLeaves(ctx context.Context, env *integration.LogEnv, client *LogClient, leaves [][]byte) error {
 	// TODO(pavelkalinnikov): Replace with AddSequencedLeaves batch API.
```
pavelkalinnikov: You can remove this TODO, I think.

client/log_client_test.go
```diff
@@ -36,13 +37,15 @@ func TestAddGetLeaf(t *testing.T) {
...
-		if err := client.QueueLeaf(ctx, l); err != nil {
-			return err
+	for i, l := range leaves {
+		if err := client.AddSequencedLeaf(ctx, l, int64(i)); err != nil {
+			return fmt.Errorf("AddSequencedLeaf(): %v", err)
 		}
-		env.Sequencer.OperationSingle(ctx)
+	}
+	env.Sequencer.OperationSingle(ctx)
+	for _, l := range leaves {
```
pavelkalinnikov: Would just waiting for N-th entry inclusion work? Looks like it implies other entries are there as well.

client/log_client_test.go
gdbelvin: Done

client/log_client_test.go
gdbelvin: Done

PR #1207, opened 2018-06-15
Use strings in util/election

gdbelvin:
Trillian uses int64 logIDs for unique tree identifiers while Key Transparency uses a domainID string. 

This PR converts `util/election` to use the more general string format with minimal converting to and from strings.  I'm not 100% sure if this is the best approach. Other ideas are welcome.


AlCutter:
I think the premise is good, but the _stringification_ should be limited to the `util/election` package, and `log_operation_manager` should `Sprintf` the treeID.

util/election/runner_test.go
```diff
@@ -141,8 +142,9 @@ func TestElectionRunnerRun(t *testing.T) {
 			fakeTimeSource := util.NewFakeTimeSource(startTime)
 
 			el := test.election
-			tracker := election.NewMasterTracker([]int64{logID}, nil)
-			er := election.NewRunner(logID, &cfg, tracker, nil, el)
+			logString := strconv.FormatInt(logID, 10)
```
AlCutter: Can you just do `const logID = "6962"` above?

server/log_operation_manager_test.go
```diff
@@ -254,13 +255,17 @@ func TestMasterFor(t *testing.T) {
 
 type masterForEvenFactory struct{}
 
-func (m masterForEvenFactory) NewElection(ctx context.Context, treeID int64) (election.MasterElection, error) {
-	isMaster := (treeID % 2) == 0
+func (m masterForEvenFactory) NewElection(ctx context.Context, treeID string) (election.MasterElection, error) {
```
pavelkalinnikov: Does KT's `domainID` identify a tree, more specifically is `itoa(treeID) == domainID`? Even if so, could we rename `treeID` here to something more general, like `resourceID` (and point out somewhere in the comments that `treeID` in Trillian is an example of a master-elected resource)? I think it doesn't matter much on this level of abstraction whether the resource is a tree or something else.

gdbelvin: The answer to your direct question is a qualified yes: a KT `domainID` has a 1:2 relationship with TreeID.  (1 domainID has 1 Trillian Log and 1 Trillian Map).

More generally, I'm hoping to use `util/election` inside KT for the sequencing and batching of mutations.  `resourceID` might indeed be a better name.

util/election/election.go
```diff
@@ -47,12 +47,12 @@ var ErrNoMaster = errors.New("no master")
 
 // Factory encapsulates the creation of a MasterElection instance for a treeID.
```
pavelkalinnikov: Please don't forget to update this comment. Mention generic `resourceID` here and `treeID` as an example.

util/election/tracker.go
```diff
@@ -69,28 +69,28 @@ func (mt *MasterTracker) Count() int {
...
+func (mt *MasterTracker) IDs() []string {
 	mt.mu.RLock()
 	defer mt.mu.RUnlock()
-	ids := make([]int64, 0, len(mt.masterFor))
+	ids := make([]string, 0, len(mt.masterFor))
 	for id := range mt.masterFor {
 		ids = append(ids, id)
 	}
-	sort.Sort(int64arr(ids))
+	sort.Strings(ids)
```
pavelkalinnikov: Are we okay to switch the sorting order from digital to lexicographical? Could it mess things in monitoring etc?

util/election/tracker.go
```diff
@@ -120,21 +119,3 @@ func HeldInfo(held []int64, ids []int64) string {
 	}
 	return result
 }
-
-// Make int64 slice sortable:
-type int64arr []int64
```
pavelkalinnikov: Nice to see this removed.

server/log_operation_manager.go
```diff
@@ -185,31 +185,38 @@ func (l *LogOperationManager) masterFor(ctx context.Context, allIDs []int64) ([]
...
 			}
-			isMaster.Set(val, strconv.FormatInt(id, 10))
+			isMaster.Set(val, id)
 		})
 	}
 
 	// Synchronize the set of configured log IDs with those we are tracking mastership for.
 	for _, logID := range allIDs {
-		knownLogs.Set(1.0, strconv.FormatInt(logID, 10))
+		knownLogs.Set(1, idToString[logID])
```
pavelkalinnikov: Maybe you could avoid having `idToString` map by iterating over `allIDString` instead?

util/election/runner_test.go
gdbelvin: done

util/election/election.go
gdbelvin: done

util/election/tracker.go
gdbelvin: Good catch.  @daviddrysdale ? 
To my knowledge this shouldn't cause any issues.

util/election/tracker.go
gdbelvin: 👍

server/log_operation_manager.go
gdbelvin: later in the function, we use `logID` as an index into `electionRunner`.  In this function, both string and integral formats are needed. It's a bit awkward, but this is the only function that translates between `int64` and `string`

gdbelvin:
Updated according to comments. PTAL

server/log_operation_manager.go
pavelkalinnikov: We could make `electionRunner` map from `string` as well? The overhead seems to be equivalent to (even slightly less than) re-creating and accessing a string-to-int map on the flight.

util/election/tracker_test.go
```diff
@@ -20,73 +20,73 @@ import (
...
 		count int
-		held  []int64
+		held  []string
 		str   string
 	}{
 		{
-			ids:   []int64{1, 2, 3},
-			ops:   []testOperation{{1, true}},
+			ids:   []string{"1", "2", "3"},
+			ops:   []testOperation{{"1", true}},
```
golangci: github.com/google/trillian/util/election.testOperation composite literal uses unkeyed fields

util/election/tracker_test.go
```diff
@@ -20,73 +20,73 @@ import (
...
 			count: 1,
-			held:  []int64{1},
+			held:  []string{"1"},
 			str:   "1 . .",
 		},
 		{
-			ids:   []int64{1, 20000, 30000},
-			ops:   []testOperation{{30000, true}},
+			ids:   []string{"1", "20000", "30000"},
+			ops:   []testOperation{{"30000", true}},
```
golangci: github.com/google/trillian/util/election.testOperation composite literal uses unkeyed fields

util/election/tracker_test.go
```diff
@@ -20,73 +20,73 @@ import (
...
 		},
 		{
-			ids: []int64{1, 2, 3},
+			ids: []string{"1", "2", "3"},
 			ops: []testOperation{
-				{1, true},
-				{2, true},
-				{3, true},
-				{1, false},
+				{"1", true},
```
golangci: github.com/google/trillian/util/election.testOperation composite literal uses unkeyed fields

server/log_operation_manager_test.go
pavelkalinnikov: Please update `treeID` to `resourceID` in this file as well.

pavelkalinnikov:
Looking good, but please address 2 comments: the `resourceID` one and string-to-int map. (I realized it might be difficult to notice in github)

server/log_operation_manager_test.go
gdbelvin: The IDs in this context are `int64`s, so I kept them as logIDs to be consistent.

server/log_operation_manager.go
gdbelvin: done

gdbelvin:
I'm not sure the `resourceID` switch makes sense since the ids in that location are still  `int64`s. 

Removed the `stringToInt` map.

server/log_operation_manager.go
```diff
@@ -185,16 +185,14 @@ func (l *LogOperationManager) masterFor(ctx context.Context, allIDs []int64) ([]
...
 		return allIDs, nil
 	}
-	allIDString := make([]string, 0, len(allIDs))
-	idToString := make(map[int64]string)
+	allStringIDs := make([]string, 0, len(allIDs))
 	for _, id := range allIDs {
 		s := strconv.FormatInt(id, 10)
-		allIDString = append(allIDString, s)
-		idToString[id] = s
+		allStringIDs = append(allStringIDs, s)
```
pavelkalinnikov: nit: Could just write `allStringIDs[i] = strconv.FormatInt(id, 10)` if `allStringIDs` was created like `make([]string, len(allIDs))`.

gdbelvin: Indeed, this is the append vs not append argument :-) 
I'm using append to be consistent with the rest of the trillian code base.

PR #1214, opened 2018-06-18
cloudbuild config for tagged releases

RJPercival:
Used together with a "triggered build" in Google Container Registry to ensure that an image is pushed for every new Git tag.

See:
https://cloud.google.com/container-builder/docs/running-builds/automate-builds
https://cloud.google.com/container-builder/docs/configuring-builds/build-test-deploy-artifacts

### DIFF
--- a/cloudbuild_tag.yaml
+++ b/cloudbuild_tag.yaml
@@ -1,0 +1,20 @@
+steps:
+- id: build_log_server
+  name: gcr.io/cloud-builders/docker
+  args:
+  - build
+  - --file=examples/deployment/docker/log_server/Dockerfile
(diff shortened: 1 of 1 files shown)


Martin2112:
Looks like this is all done on the GCR end and doesn't need us to give it extra repository permissions?

RJPercival:
It asks for permission to access the repo when the triggered build is configured.

Martin2112:
That's just read access though I assume? Just checking as we're not allowed to give out some permissions.

RJPercival:
We can't give out some permissions to third-party services, but Google Container Registry isn't third-party.

PR #1222, opened 2018-07-05
Simplify coverage scripts

gdbelvin:
Post Go 1.10, `-coverageprofile` works with `go test ./...`.


gdbelvin:
The speeds seem equivalent. There was test parallelism before with xargs, there is test parallelism after via go test.

PR #1223, opened 2018-07-06
GetAndVerifyInclusionAtIndex to not update the root.

maljub01:
`GetAndVerifyInclusionAtIndex` is useful for requesting inclusion proofs
after a root update has been performed, similar to
`getAndVerifyInclusionProof`.

However, unlike, `getAndVerifyInclusionProof`,
`GetAndVerifyInclusionAtIndex` currently updates the root every time it
is called. This is a surprising side effect that clients cannot disable.

On the other hand, by removing this update, the two get-and-verify
methods will be more consistent, and clients who do require an update
can still do so by calling `UpdateRoot` explicitly.

### DIFF
--- a/client/log_client_test.go
+++ b/client/log_client_test.go
@@ -201,8 +201,13 @@ func TestVerifyInclusionAtIndex(t *testing.T) {
 		t.Fatalf("Failed to add leaves: %v", err)
 	}
 
+	root, err := client.UpdateRoot(ctx)
+	if err != nil {
+		t.Errorf("UpdateRoot(): %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


client/log_client_test.go
```diff
@@ -201,6 +201,11 @@ func TestVerifyInclusionAtIndex(t *testing.T) {
 		t.Fatalf("Failed to add leaves: %v", err)
 	}
 
+	root, err := c.UpdateRoot(ctx)
```
golangci: ineffectual assignment to `root`

gdbelvin:
This looks reasonable.

PR #1224, opened 2018-07-06
Always require the root parameter when creating new LogClients

maljub01:
Currently, users have no way of setting a starting root for the
LogClient. This means that whenever LogClient is used, the current
latest root returned by Trillian is implicitly trusted.

This change allows users to instead start verification from a known
good state. This is useful for recovering from restarts/downtime without
losing verifiability.

Clients that want the old behavior should call `New` with an empty initial
trusted log root.
    
```
New(..., types.LogRootV1{})
NewFromTree(..., types.LogRootV1{})
```

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -47,6 +47,16 @@ func New(logID int64, client trillian.TrillianLogClient, verifier *LogVerifier) 
 	}
 }
 
+// NewWithRoot returns a new LogClient.
+func NewWithRoot(logID int64, client trillian.TrillianLogClient, verifier *LogVerifier, root types.LogRootV1) *LogClient {
+	return &LogClient{
+		LogVerifier: verifier,
+		LogID:       logID,
+		client:      client,
+		root:        root,
+	}
+}
+
 // NewFromTree creates a new LogClient given a tree config.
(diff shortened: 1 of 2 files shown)


gdbelvin:
`WithRoot` is actually best practice. Would you mind adding it as a required parameter to the existing `New` functions? 

Clients that want the old behavior should call  `New` with an empty initial trusted log root. 
```
New(..., types.LogRootV1{})
```

I realize this will expand the CL, but it should make it much clearer what root is initially being trusted from all the calling locations.

maljub01:
Done. I also had to add the root parameter to `NewFromTree` which uses `New`.

maljub01:
@gdbelvin CI seems to be failing because of gometalinter timing out, but I'm unable to retrigger the tests without pushing more commits, do you mind retriggering them?

gdbelvin:
I've been noticing the deadline issue as well. 
* Retry that particular test in travis. I think this issue occurs when travis is under slightly increased load.
* Try increasing the deadline in `./scripts/presubmit.sh`.
* Have a chat with @Martin2112 about the deadline.

maljub01:
Created a PR that increases the deadline: https://github.com/google/trillian/pull/1226

PR #1225, opened 2018-07-06
Improve the logic around updating the trusted root

maljub01:
- Add a `GetRoot` method that returns a (pointer to a) copy of the latest trusted root.

- Introduce an update lock to prevent root updates from happening concurrently. This is because multiple `UpdateRoot`s running concurrently can result in an invalid update if the Trillian log was forked and each update was given a different branch of the fork.

- Remove the `waitForTreeSize` argument from `WaitForRootUpdate`. This argument makes the function harder to use and adds complexity to it.

  This complexity had introduced the following subtle bug:
  If `WaitForRootUpdate` was called with `waitForTreeSize >= current size + 2`, then the root might be updated (ex. to `current size + 1`) and then an error encountered during the next update. When this happens, the method returns with an error and with `nil` for the new root (incorrectly indicating no root update).

- Stop `VerifyInclusion` and `GetAndVerifyInclusionAtIndex` from updating the trusted root under the hood. Users who do require an update can do so by calling `UpdateRoot` or `WaitForRootUpdate` explicitly before calling those methods.

- Update the documentation for `AddSequencedLeafAndWait` and `AddLeaf` to clarify that they update the trusted root under the hood.

- Rename `getLatestRoot` to `getAndVerifyLatestRoot`.

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -120,23 +120,25 @@ func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*tri
 	return resp.Leaves, nil
 }
 
-// WaitForRootUpdate repeatedly fetches the Root until the fetched tree size >=
-// waitForTreeSize or until ctx times out.
-func (c *LogClient) WaitForRootUpdate(ctx context.Context, waitForTreeSize uint64) (*types.LogRootV1, error) {
+// WaitForRootUpdate repeatedly fetches the latest root until there is an
+// update, which it then applies, or until ctx times out.
+func (c *LogClient) WaitForRootUpdate(ctx context.Context) (*types.LogRootV1, error) {
 	b := &backoff.Backoff{
 		Min:    100 * time.Millisecond,
 		Max:    10 * time.Second,
 		Factor: 2,
 		Jitter: true,
 	}
-	for i := 0; ; i++ {
-		root, err := c.UpdateRoot(ctx)
-		switch x := status.Code(err); x {
+
+	for {
+		newTrusted, err := c.getLatestRoot(ctx, c.GetRoot())
+		switch status.Code(err) {
 		case codes.OK:
-			if root.TreeSize >= waitForTreeSize {
-				return root, nil
+			if c.updateRootIfNewer(newTrusted) {
+				return newTrusted, nil
 			}
-		case codes.Unavailable, codes.NotFound, codes.FailedPrecondition: // Retry.
+		case codes.Unavailable, codes.NotFound, codes.FailedPrecondition:
+			// Retry.
 		default:
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


gdbelvin:
Would you mind expanding on the use case for this particular function?
Is there a time when you'd want to know the current root when you wouldn't want to update it?

maljub01:
The reason I wanted to add this is that there are a lot of methods that will update the root, but without returning the new one (ex. `AddLeaf`, `WaitForInclusion`, `VerifyInclusion`).

The client of the library will at that point have no way of finding out what the new root is.

gdbelvin:
I suppose the bigger idea is that the updating or not updating of the root should be controlled by the caller rather than happen automatically, and giving the caller access to the current root allows them to specify what root to use?

maljub01:
Agreed, but I still think it is useful to have a `GetRoot` function, which can be useful in combination with methods like `WaitForInclusion`

wdyt?

client/log_client.go
```diff
@@ -231,24 +257,25 @@ func (c *LogClient) WaitForInclusion(ctx context.Context, data []byte) error {
...
+				return err
+			} else if ok {
+				return nil
+			}
 		}
-		// Wait for TreeSize to update.
-		if root, err = c.WaitForRootUpdate(ctx, root.TreeSize+1); err != nil {
+
+		// If not found or tree is empty, wait for a root update before retrying again.
+		if root, err = c.WaitForRootUpdate(ctx); err != nil {
```
golangci: this value of root is never used

gdbelvin:
Ping me when the build is green

maljub01:
@gdbelvin done. PTAL.

client/log_client.go
```diff
@@ -120,23 +120,25 @@ func (c *LogClient) ListByIndex(ctx context.Context, start, count int64) ([]*tri
...
-		root, err := c.UpdateRoot(ctx)
-		switch x := status.Code(err); x {
+
+	for {
+		newTrusted, err := c.getLatestRoot(ctx, c.GetRoot())
+		switch status.Code(err) {
 		case codes.OK:
-			if root.TreeSize >= waitForTreeSize {
-				return root, nil
+			if c.updateRootIfNewer(newTrusted) {
```
gdbelvin: I like this clearer logic 👍

client/log_client.go
```diff
@@ -202,67 +204,94 @@ func (c *LogClient) getLatestRoot(ctx context.Context, trusted *types.LogRootV1)
...
+	c.rootLock.Lock()
+	defer c.rootLock.Unlock()
+
+	// Copy the internal trusted root in order to prevent clients from modifying it.
+	ret := c.root
+	return &ret
+}
+
+// updateRootIfNewer updates the current trusted root if the new root is newer
+// than the currently trusted root and its tree size is larger.
```
gdbelvin: Please add a big warning that this method should 
* Not be public
* Only be called from successful `getLatestRoot` which should really be called `getAndVerifyLatestRoot`

client/log_client.go
```diff
@@ -202,67 +204,94 @@ func (c *LogClient) getLatestRoot(ctx context.Context, trusted *types.LogRootV1)
...
 func (c *LogClient) VerifyInclusion(ctx context.Context, data []byte) error {
 	leaf, err := c.BuildLeaf(data)
 	if err != nil {
 		return err
 	}
-	root, err := c.UpdateRoot(ctx)
-	if err != nil {
-		return fmt.Errorf("UpdateRoot(): %v", err)
-	}
+	root := c.GetRoot()
```
gdbelvin: 👍

client/log_client.go
maljub01: Good point. I renamed `getLatestRoot` to `getAndVerifyLatestRoot`. For `updateRootIfNewer`, after looking more carefully at where it is used, I figured I could just get rid of the method.

client/log_client.go
```diff
@@ -215,34 +215,26 @@ func (c *LogClient) GetRoot() *types.LogRootV1 {
...
+	newTrusted, err := c.getAndVerifyLatestRoot(ctx, currentlyTrusted)
 	if err != nil {
 		return nil, err
 	}
 
-	if c.updateRootIfNewer(newTrusted) {
+	c.rootLock.Lock()
+	defer c.rootLock.Unlock()
+
+	// Get the root again, since it might have been updated by a different thread
```
gdbelvin: I like your thinking in delaying the lock, however, let me elaborate on what needs to be verified. 

In addition to signature checks, `getAndVerifyLatestRoot` also verifies that there is a consistency proof between `currentlyTrusted` and `newTrusted` to form a continuous chain of checks for each state of `c.root`. 

Consider the following example of a fork where `B'` indicates a forked state:  
Both the transitions `A -> B'` and `A->C` have valid transition proofs.  `B' -> C` is invalid.
 
```
A -    - > C
  \       
    B'  
```

If a concurrent `UpdateRoot` verifies and sets `A -> B'` while we are verifying `A -> C`, `c.root` will go through an invalid state transition `A->B'->C`.  

We could either lock `UpdateRoot` so that it can't be called concurrently, or upon detecting that `currentlyTrusted` has changed, fetch a new proof for the new `currentlyTrusted`.  What do you think?

maljub01: Makes sense. Thank you for the explanation. I added an update lock to prevent updates from happening concurrently.

gdbelvin:
Looks good, thanks!

PR #1232, opened 2018-07-20
Use grpc_middleware.ChainUnaryServer

gdbelvin:
Replace interceptor.Chain with grpc_middlware.ChainUnaryServer.
interceptor is duplicating what should be standard grpc code.

By switching we both reduce code we need to maintain as well as
improve readability by using standard libraries.

### DIFF
--- a/integration/admin/admin_integration_test.go
+++ b/integration/admin/admin_integration_test.go
@@ -36,6 +36,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	sa "github.com/google/trillian/server/admin"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 )
 
 func TestAdminServer_CreateTree(t *testing.T) {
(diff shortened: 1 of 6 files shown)


integration/admin/admin_integration_test.go
```diff
@@ -36,6 +36,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	sa "github.com/google/trillian/server/admin"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
phad: if you're going to alias this import might it as well become something shorter?

integration/quota/quota_test.go
```diff
@@ -225,10 +227,13 @@ func (s *testServer) serve() {
 func newTestServer(registry extension.Registry) (*testServer, error) {
 	s := &testServer{}
 
-	intercept := interceptor.New(
-		registry.AdminStorage, registry.QuotaManager, false /* quotaDryRun */, registry.MetricFactory)
-	netInterceptor := interceptor.Combine(interceptor.ErrorWrapper, intercept.UnaryInterceptor)
-	s.server = grpc.NewServer(grpc.UnaryInterceptor(netInterceptor))
+	ti := interceptor.New(registry.AdminStorage, registry.QuotaManager, false /* quotaDryRun */, registry.MetricFactory)
```
phad: what's `ti`?  'trillian interceptor'?

integration/admin/admin_integration_test.go
gdbelvin: The import is aliased to the name of the go package that `go-grpc-middleware` declares. 
I could rename it but `grpc_middleware` seems pretty clear and standard.

integration/quota/quota_test.go
gdbelvin: yep

PR #1235, opened 2018-07-24
Support MySQL8

gdbelvin:
MySQL8 deprecates changing user passwords as part of the `GRANT` command.
However, Trillian doesn't set the db password directly any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,12 @@ language: go
 
 addons:
   apt:
+    sources:
+      - mysql-5.7-trusty
     packages:
       - softhsm
+      - mysql-server
+      - mysql-client
 
 cache:
   directories:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,8 +4,12 @@ language: go
 
 addons:
   apt:
+    sources:
+      - mysql-5.7-trusty
     packages:
       - softhsm
+      - mysql-server
+      - mysql-client
```
RJPercival: If we're installing MySQL 5.7 manually, should we be removing "mysql" from the "services" list further down? That appears to install and configure the default MySQL version (5.6).
https://github.com/google/trillian/blob/b16d627d84485e8351a2c87f8a387909089600d8/.travis.yml#L32-L33

RJPercival: Also, based on the [Travis docs](https://docs.travis-ci.com/user/database-setup/#MySQL-5.7), it looks like you need to add something like the following to this file:
```yaml
before_install:
  - sudo mysql -e "use mysql; update user set authentication_string=PASSWORD('new_password') where User='root'; update user set plugin='mysql_native_password';FLUSH PRIVILEGES;"
  - sudo mysql_upgrade
  - sudo service mysql restart
```

RJPercival:
The "Support MySQL8" commit LGTM, but not so sure about the "Update travis to MySQL 5.7" commit.

.travis.yml
RJPercival: It *also* appears you need to change the following code that's further down this file:
https://github.com/google/trillian/blob/b16d627d84485e8351a2c87f8a387909089600d8/.travis.yml#L65-L73

I think you need to change `sudo start/stop mysql` to `sudo services mysql start/stop`.

gdbelvin: done

gdbelvin:
You're right. `IF NOT EXISTS` wasn't added till 5.7.

RJPercival:
The fact that we're dropping support for MySQL 5.6 (at least as far as the test scripts go) should probably get mentioned in CHANGELOG.md by the way.

PR #1236, opened 2018-07-25
Implement Interception Filtering

gdbelvin:
The Trillian Interceptor fails all requests that it doesn't recognize. Unfortunately this doesn't play well when other services are hosted on the same server. 

This PR restricts the Trillian interceptor to operating on Trillian services.

### DIFF
--- a/server/interceptor/interceptor.go
+++ b/server/interceptor/interceptor.go
@@ -18,6 +18,7 @@ package interceptor
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -54,6 +55,11 @@ var (
 	requestDeniedCounter monitoring.Counter
 	contextErrCounter    monitoring.Counter
 	metricsOnce          sync.Once
+	enabledServices      = map[string]bool{
+		"trillian.TrillianLog":   true,
+		"trillian.TrillianMap":   true,
+		"trillian.TrillianAdmin": true,
+	}
 )
 
 // RequestProcessor encapsulates the logic to intercept a request, split into separate stages:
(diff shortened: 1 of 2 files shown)


server/interceptor/interceptor.go
```diff
@@ -258,6 +274,15 @@ func isLeafOK(leaf *trillian.QueuedLogLeaf) bool {
 	return leaf == nil || leaf.Status == nil || leaf.Status.Code == int32(codes.OK)
 }
 
+// serviceName returns "some.package.service" for
+// "/some.package.service/method".
+func serviceName(fullMethod string) string {
+	if !strings.HasPrefix(fullMethod, "/") {
+		return ""
+	}
+	return strings.Split(fullMethod, "/")[1]
```
AlCutter: This should be ...[0], shouldn't it?

server/interceptor/interceptor_test.go
```diff
@@ -538,24 +577,27 @@ func TestTrillianInterceptor_QuotaInterception_ReturnsTokens(t *testing.T) {
...
 func TestTrillianInterceptor_NotIntercepted(t *testing.T) {
 	tests := []struct {
-		req interface{}
+		method string
+		req    interface{}
 	}{
 		// Admin
-		{req: &trillian.CreateTreeRequest{}},
-		{req: &trillian.ListTreesRequest{}},
+		{method: "/trillian.TrillianAdmin/CreateTree", req: &trillian.CreateTreeRequest{}},
```
AlCutter: Consider extending the test coverage to catch bugs where the interceptor is accidentally universally disabled?

server/interceptor/interceptor.go
gdbelvin: [0] gets us "" because strings start with "/"

server/interceptor/interceptor_test.go
gdbelvin: Not sure I follow...
The existing tests verify that the handlers are running for the expected methods. 
If the interceptors get disabled these tests fail when they test for the presence of tree in `ctx`

gdbelvin:
PTAL

server/interceptor/interceptor_test.go
AlCutter: ok, then we're good :)

server/interceptor/interceptor.go
AlCutter: Ah, bad assumption on my part. Carry on!

PR #1244, opened 2018-07-31
Add a command for requesting a Trillian tree's public key

maljub01:


pavelkalinnikov:
Could this tool be more generic? E.g. it could print all the fields of the `Tree` structure, or just the public key if, say, only `--pub_key` flag is active.

maljub01:
My use case here was to use it to write a PEM-formatted public key file (using `./get_tree_public_key --admin_server=host:port --log_id=logid > mytree.pub`), but there's definitely a good case for making this more generic, perhaps converting it into an `inspect_log` command with various options.

PR #1247, opened 2018-08-02
Refactor get_tree_public_key test helper methods

maljub01:
- Move `setFlag` to `setup.SetFlag`.
- Remove `CreateLog` in favor of using the existing `CreateAndInitTree`.


testonly/setup/flags.go
```diff
@@ -0,0 +1,29 @@
...
+
+package setup
+
+import (
+	"flag"
+	"testing"
+)
+
+// SetFlag updates a flag value, failing the test if something goes wrong.
+func SetFlag(t *testing.T, name, value string) {
```
gdbelvin: Do you want to revert the flag after the test is finished?

testonly/setup/trees.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian"
+	"github.com/google/trillian/client"
+	ktestonly "github.com/google/trillian/crypto/keys/testonly"
```
gdbelvin: Please move import renames into a separate section

testonly/setup/trees.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/google/trillian/crypto/sigpb"
+	"github.com/google/trillian/testonly"
+	"github.com/google/trillian/testonly/integration"
+)
+
+// CreateLog creats a Trillian log with for testing purposes
+//
+// It optionally allows specifying an overrides argument to override some
+// properties of the tree to be created.
+func CreateLog(t *testing.T, logEnv *integration.LogEnv, maxRootDuration time.Duration) *trillian.Tree {
```
gdbelvin: Please document maxRootDuration in the comment

testonly/setup/trees.go
```diff
@@ -0,0 +1,72 @@
...
+	t.Helper()
+
+	ctx := context.Background()
```
gdbelvin: ctx should be passed in from the test root

testonly/setup/trees.go
gdbelvin: For easier testing utility, consider passing the client objects used in this method rather than the whole environment.

gdbelvin:
It looks like the test helpers are fairly test specific. 

Consider moving the test helper functions into the test files themselves.  See the `map_test`'s 
 [newTreeWithHasher](https://github.com/google/trillian/blob/master/integration/maptest/map.go#L127)  helper function.

testonly/setup/flags.go
maljub01: Reverting is done in bulk using `defer flagsaver.Save().Restore()`

maljub01:
The reason I'm putting those helper functions in a separate module is that I want to use them in multiple test files. They are currently being used in `cmd/get_tree_public_key/main_test.go`, but I would also like to create a `logmonitor` command and will need those same helpers for its `main_test.go` as well.

testonly/setup/flags.go
gdbelvin: Thanks

testonly/setup/trees.go
```diff
@@ -0,0 +1,71 @@
...
+
+	ktestonly "github.com/google/trillian/crypto/keys/testonly"
+)
+
+// CreateLog creates a Trillian log with for testing purposes
+//
+// It requires the caller to specify a maxRootDuration for the Trillian tree,
+// which will be used as the interval after which a new signed root is produced
+// despite no submissions.
+func CreateLog(ctx context.Context, t *testing.T, logClient trillian.TrillianLogClient, adminClient trillian.TrillianAdminClient, maxRootDuration time.Duration) *trillian.Tree {
```
gdbelvin: It would be nice to avoid an explosion of test helper functions, to set all the various fields of `trillian.Tree`. 

Would it be useful to create a `TreeOption` construct like GRPC's [`With*`](https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L273) methods?

gdbelvin: Existing tests call `CreateAndInitTree` directly.  Please follow that convention in your tests.  
[Example](https://sourcegraph.com/github.com/google/trillian@a7f812df5b55e7d2a04543ba80e75595617d46a5/-/blob/integration/log_integration_test.go#L104) 

It looks like the main utility of this function is a helper for creating a `trillian.Tree`. 
a) There are some convenient, [prepopulated](https://sourcegraph.com/github.com/google/trillian@a7f812df5b55e7d2a04543ba80e75595617d46a5/-/blob/storage/testonly/admin_storage_tester.go#L92) test trees that you can use, copy, and modify in your tests. 
b) If you prefer, this function could be reduced to  `CreateTreeRequest` with a suite of `With*` modifiers. 

What do you think?

gdbelvin:
It looks like there's an [easier way](https://sourcegraph.com/github.com/google/trillian@a7f812df5b55e7d2a04543ba80e75595617d46a5/-/blob/storage/testonly/admin_storage_tester.go#L92) to get test trees setup.  What do you think about using that approach?  We'd prefer to reduce code duplication.

testonly/setup/trees.go
maljub01: I didn't know about `CreateAndInitTree`. I'll switch to using it.

Thank you.

maljub01:
Switched to using `CreateAndInitTree` and the tests are much better now! Thank you!

PR #1248, opened 2018-08-02
Fixed description of a command that was copy/pasted from deletetree

pgporada:


pgporada:
I signed it!

cmd/get_tree_public_key/main.go
```diff
@@ -33,7 +33,7 @@ import (
 
 var (
 	adminServerAddr = flag.String("admin_server", "", "Address of the gRPC Trillian Admin Server (host:port)")
-	logID           = flag.Int64("log_id", 0, "Trillian LogID to delete")
+	logID           = flag.Int64("log_id", 0, "Trillian LogID to retrieve tree public key from")
```
AlCutter: Maybe change "from" to "for"?

AlCutter: (or maybe "Trillian LogID whose public key should be fetched"?)
dunno :)

pgporada: Fixed!

PR #1251, opened 2018-08-07
Stop the MySQL storage provider from losing its mysql.OpenDB error

maljub01:
Currently, if `server.NewStorageProvider("mysql", nil)` is called twice
with a bad MySQL URL, the first time, it will fail properly returning a
non-nil error. However, this error is lost afterwards and the second
time it will silently fail and return a nil StorageProvider.

This change fixes that by making the MySQL storage provider hold on to
any initialization errors it encounters, and returning them for every
new storage provider request.

### DIFF
--- a/server/mysql_storage_provider_test.go
+++ b/server/mysql_storage_provider_test.go
@@ -15,10 +15,18 @@
 package server
 
 import (
+	"flag"
 	"testing"
+
+	"github.com/google/trillian/util/flagsaver"
 )
 
 func TestMySQLStorageProviderErrorPersistence(t *testing.T) {
+	defer flagsaver.Save().Restore()
+	if err := flag.Set("mysql_uri", ""); err != nil {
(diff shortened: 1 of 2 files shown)


server/mysql_storage_provider_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	// Second call: This should fail with the same error.
+	_, err2 := NewStorageProvider("mysql", nil)
+	if err2 == nil {
+		t.Fatalf("Expected second call to 'server.NewStorageProvider' to fail")
+	}
+
+	if err2 != err1 {
+		t.Fatalf("Expected second call to 'server.NewStorageProvider' to fail with %s, instead got: %s", err1, err2)
```
RJPercival: Use `%q` instead of `%s` so that the error messages are quoted.

maljub01:
@RJPercival thanks, I fixed the formatting. PTAL.

PR #1253, opened 2018-08-07
Add a log monitor binary

maljub01:


pavelkalinnikov:
CC @taknira

maljub01:
CC @gdbelvin

cmd/logmonitor/db.go
```diff
@@ -0,0 +1,118 @@
...
+
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/types"
+	"github.com/jinzhu/gorm"
```
gdbelvin: Consider defining a storage interface like we do [here](https://github.com/google/trillian/blob/7b07980fdb5a5b64ab920b59e4a1f9dd3275b455/server/storage_provider.go#L90:6)   This would allow the log monitor to be used by other storage implementations such as in-memory fakes, cloud spanner, and amazon's databases.

cmd/logmonitor/db.go
```diff
@@ -0,0 +1,118 @@
...
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+	_ "github.com/jinzhu/gorm/dialects/sqlite"
+)
+
+// initialEmptyRoot represents an empty trusted root and is used when no
+// information is known about the log being monitored.
+var initialEmptyRoot TrustedLogRoot
+
+func init() {
+	// Set up initialEmptyRoot
```
gdbelvin: Please add a TODO to save and restore the last-verified root.

cmd/logmonitor/main.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	grpcClient := trillian.NewTrillianLogClient(conn)
+	hasher := objhasher.NewLogHasher(rfc6962.New(crypto.SHA256))
+	verifier := client.NewLogVerifier(hasher, publicKey, crypto.SHA256)
+	return client.New(*logID, grpcClient, verifier, startingRoot), nil
+}
+
+func runMonitor(ctx context.Context) error {
+	flag.Parse()
```
gdbelvin: `flag.Parse()` typically lives in `main()`

cmd/logmonitor/main.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := m.Run(ctx); err != nil {
+		return fmt.Errorf("Encountered an error: %v", err)
+	}
+
+	return nil
+}
+
+func main() {
+	defer glog.Flush()
+	err := runMonitor(context.Background())
```
gdbelvin: These two lines can be collapsed

cmd/logmonitor/db.go
maljub01: This is already the case.

cmd/logmonitor/main.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	conn, err := grpc.Dial(*serverAddr, dialOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial %s: %v", *serverAddr, err)
+	}
+
+	grpcClient := trillian.NewTrillianLogClient(conn)
+	hasher := objhasher.NewLogHasher(rfc6962.New(crypto.SHA256))
+	verifier := client.NewLogVerifier(hasher, publicKey, crypto.SHA256)
```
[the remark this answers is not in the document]
maljub01: Yes, but I don't want to rely on Trillian to tell me the tree's properties since if I'm monitoring it, I shouldn't trust what it says. No?

cmd/logmonitor/main.go
maljub01: I'm leaving it here to simplify testing. I basically call `runMonitor` in my tests instead of running main directly.

maljub01: ... and the reason I don't want to call `main` directly in my tests is that it calls `glog.Exit` which would kill my test process.

cmd/logmonitor/db.go
maljub01: I prefer to use a proper ORM rather than have to roll my own.

Gorm is actually pretty good. I didn't have to write any dialect-specific code, and I get support for MySQL, PostgreSQL, SQLite and Microsoft SQL Server (and any potential new dialects that get added to Gorm in the future) for free. Plus, it was very easy to get the binary to also handle setting up and migrating its own database tables.

Gorm's support for PostgreSQL and MySQL means that this monitor already supports CockroachDB and almost all of Amazon RDS's database engines (including Amazon Aurora), with the only exception being the Oracle RDS database engine.

For in-memory fakes, I would argue that using SQLite (either as a temporary on-disk database or an in-memory one) is a much better option than rolling my own in-memory fake for testing. The tests in `main_test.go` are a good example of this, since they are using a real (but temporary) SQLite database.

As for cloud spanner, given gorm's popularity, I suspect a cloud spanner dialect will eventually get added. I have actually been considering adding that support myself since it's not much work to add a new dialect, but I haven't gotten around to it yet.

cmd/logmonitor/db.go
maljub01: I'll update the comment to clarify this point.

cmd/logmonitor/logmonitor.go
```diff
@@ -0,0 +1,85 @@
...
+	logClient            *client.LogClient
+	db                   *gorm.DB
+	waitBetweenDBUpdates time.Duration
+}
+
+// Run starts continuously monitoring the Trillian log specified in the
+// LogMonitor object.
+//
+// It continues running until it encounters an error.
+func (m *LogMonitor) Run(ctx context.Context) error {
```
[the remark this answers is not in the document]
maljub01: I started doing this, but then realized it would add a bit of complexity and boilerplate since the majority of the inner loop logic is about handling breaking out of the infinite loop.

Since a single-shot is the same as running `LogClient.WaitForRootUpdate`, it makes sense to leave single-shot testing to LogClient tests, no?

cmd/logmonitor/logmonitor.go
```diff
@@ -0,0 +1,85 @@
...
+func (m *LogMonitor) periodicallyPersist(ctx context.Context) {
+	latestPersistedRoot := m.logClient.GetRoot()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(m.waitBetweenDBUpdates):
+			latestRoot := m.logClient.GetRoot()
+
+			if latestRoot.TimestampNanos <= latestPersistedRoot.TimestampNanos &&
```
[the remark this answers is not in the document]
maljub01: Done.

cmd/logmonitor/db.go
Martin2112: Our log / map data model isn't complex enough to really benefit from an ORM and we want to control how locks are obtained. Possibly it is for a monitor but it could be arguable. Performance might be an issue as Gorm presumably can't do code generation like Hibernate etc. can do for Java? This might not be an issue if you're monitoring a small number of logs. Hard to say at this point.

Another issue for an ORM is that you're potentially on the hook for supporting configurations you haven't tested. Again, possibly not as vital for a monitor but it was a concern for us.

Testing a server with SQLite is tough and we gave up on it. We had a lot of test flakes. Might not happen with simple unit tests but I think it's worth mentioning.

CloudSpanner isn't just another SQL dialect. The way you interact with transactions changes. If you want to support it then it's worth following their model from the start rather than having to rework it later as we did.

maljub01:
@gdbelvin Travis CI passed but it's not updating the pull request for some reason. Do you mind taking a look?

gdbelvin:
@Martin2112 - did we reach consensus on ORMs?

Martin2112:
I'm not fully sold on ORM but if we're treating this as an application then I see no reason to blanket forbid them.

cmd/logmonitor/db.go
```diff
@@ -0,0 +1,147 @@
...
+	marshalled, err := emptyRoot.MarshalBinary()
+	if err != nil {
+		glog.Exitf("Could not marshal the empty root: %v\n", err)
+	}
+	initialEmptyRoot = TrustedLogRoot{MarshalledRoot: marshalled, TreeSize: 0}
+}
+
+// TrustedLogRoot is the database model used for storing information about
+// trusted log roots.
+type TrustedLogRoot struct {
```
gdbelvin: A trusted Log Root is one that has been verified and it's fields have been split out. 
This currently looks like an unverified log root. Consider storing the split out fields from LogRootV1?

gdbelvin: Consider renaming this `StoredLogRoot` or `ObservedLogRoot`

cmd/logmonitor/db.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if trustedLogRoot.TreeSize == 0 {
+		glog.Warningf("Using an empty root for log %d", logID)
+	} else {
+		glog.Infof("Loaded a root for log %d with tree size = %d", logID, trustedLogRoot.TreeSize)
+	}
+
+	currentRoot := &types.LogRootV1{}
+	if err := currentRoot.UnmarshalBinary(trustedLogRoot.MarshalledRoot); err != nil {
```
gdbelvin: `UnmarshalBinary` is considered a private method.  Please migrate to `client.LogVerifier.VerifyRoot`

cmd/logmonitor/logmonitor.go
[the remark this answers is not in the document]
gdbelvin: By splitting it out into a separate method, I was hoping that we could eliminate a bunch of that complexity.

cmd/logmonitor/main.go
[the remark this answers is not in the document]
gdbelvin: You'll want to store the `trilian.Tree` object yourself and use `NewFromTree`

cmd/logmonitor/db.go
```diff
@@ -0,0 +1,147 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
gdbelvin: Would you mind moving the database operations into their own go package with their own unit tests? 

`cmd` is reserved for `main()` functions which should then use libraries defined elsewhere. 
This helps immensely with portability.

RJPercival:
@gdbelvin, @taknira, @Martin2112: Is this still something we'd like to have?

Martin2112:
I'm not sure how/if this overlaps with Gossip? This might be more general?

Possibly it belongs in the https://github.com/google/trillian-examples repo where the Gossip code lives.

PR #1256, opened 2018-08-14
Fix the build

maljub01:
Currently, building fails with the following error:

```
server/log_operation_manager.go:342: Verbose.Infof format %d has arg
logID of wrong type string
```

This is because `logID` is a string, but `Infof` is being asked to format it
as an integer.


maljub01:
Not sure why this is failing on my machine but not in CI.

maljub01:
@gdbelvin it looks like codecov is stuck.

PR #1268, opened 2018-09-03
Remove exit command from .travis.yml

RJPercival:
https://docs.travis-ci.com/user/customizing-the-build/#how-does-this-work-or-why-you-should-not-use-exit-in-build-steps


RJPercival:
That is actually possible, I think. I'll do that as a separate PR.

PR #1270, opened 2018-09-04
Make `git diff` portion of .travis.yml a bit more concise

RJPercival:
The `git diff` command accepts Git pathspecs, which can be used to
exclude paths by prefixing them with ':!'. This removes the need to pipe
`git status --porcelain` through `grep` and then re-run the Git commands
on failure.


RJPercival:
Is this what you had in mind @pavelkalinnikov?

PR #1274, opened 2018-09-07
Add tests for ExecutePass failing in log_operation_manager_test.go

RJPercival:
Checks that monitoring counters are updated correctly. This improves test coverage, since the tests previously had all calls to `ExecutePass()` return `(0, nil)`, which only exercised a single code path around `ExecutePass()` calls (the one that handled `ExecutePass()` doing nothing).

In order to check the monitoring counters, this commit adds a `CounterSnapshot` struct that can be used to examine the delta between counter values at different points in a test. This is particularly useful since counters are package variables which do not get reset between tests.

### DIFF
--- a/monitoring/testonly/delta.go
+++ b/monitoring/testonly/delta.go
@@ -24,6 +24,8 @@ import (
 // A CounterSnapshot is useful in tests because counters do not reset between
 // test cases, and so their absolute value is not amenable to testing. Instead,
 // the delta between the value at the start and end of the test should be used.
+// This assumes that multiple tests that all affect the counter are not run in
+// parallel.
 type CounterSnapshot struct {
 	c      monitoring.Counter
 	labels []string
--- a/monitoring/testonly/delta.go
+++ b/monitoring/testonly/delta.go
@@ -1,0 +1,44 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testonly
+
+import (
+	"github.com/google/trillian/monitoring"
+)
+
+// CounterSnapshot records the values of a counter at a specific point in time.
+type CounterSnapshot struct {
+	c      monitoring.Counter
+	values map[string]float64
+}
+
+// SnapshotCounter records the current values of a counter.
+// It will only record values associated with the given labels.
+func SnapshotCounter(c monitoring.Counter, labels ...string) CounterSnapshot {
+	s := CounterSnapshot{
+		c:      c,
+		values: make(map[string]float64, len(labels)),
+	}
+	for _, l := range labels {
+		s.values[l] = c.Value(l)
+	}
+	return s
+}
+
+// Delta returns the difference between the current value of a counter and its
+// value when the snapshot was created.
+func (s CounterSnapshot) Delta(label string) float64 {
+	return s.c.Value(label) - s.values[label]
+}
--- a/monitoring/testonly/delta.go
+++ b/monitoring/testonly/delta.go
@@ -15,30 +15,45 @@
 package testonly
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/google/trillian/monitoring"
 )
 
-// CounterSnapshot records the values of a counter at a specific point in time.
+// CounterSnapshot records the value of a counter at a specific point in time.
 type CounterSnapshot struct {
 	c      monitoring.Counter
 	values map[string]float64
 }
 
-// SnapshotCounter records the current values of a counter.
-// It will only record values associated with the given labels.
-func SnapshotCounter(c monitoring.Counter, labels ...string) CounterSnapshot {
+// NewCounterSnapshot creates a CounterSnapshot that can record values from a
+// counter and later report the delta between those recorded values and the
+// current values.
+func NewCounterSnapshot(c monitoring.Counter) CounterSnapshot {
 	s := CounterSnapshot{
 		c:      c,
-		values: make(map[string]float64, len(labels)),
-	}
-	for _, l := range labels {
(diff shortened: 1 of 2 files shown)


monitoring/testonly/delta.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, l := range labels {
+		s.values[l] = c.Value(l)
+	}
+	return s
+}
+
+// Delta returns the difference between the current value of a counter and its
+// value when the snapshot was created.
+func (s CounterSnapshot) Delta(label string) float64 {
+	return s.c.Value(label) - s.values[label]
```
pavelkalinnikov: `Value` accepts a list of labels in a general case. It is a coincidence that Trillian has only 1 label (`treeID`).
Make this function accept `label ...string` instead, in addition to the modification above?

monitoring/testonly/delta.go
```diff
@@ -0,0 +1,44 @@
...
+
+// CounterSnapshot records the values of a counter at a specific point in time.
+type CounterSnapshot struct {
+	c      monitoring.Counter
+	values map[string]float64
+}
+
+// SnapshotCounter records the current values of a counter.
+// It will only record values associated with the given labels.
+func SnapshotCounter(c monitoring.Counter, labels ...string) CounterSnapshot {
```
pavelkalinnikov: It looks a bit confusing when I compare `SnapshotCounter` to a regular `Counter`. Both accept a list of string `labels`, but `Counter` treats them as labels of a single multi-label metric, while `SnapshotCounter` treats them as labels of multiple single-label metrics.

To make it more general and avoid this confusion might be worth changing the signature to:
```
func NewCounterSnapshot(c monitoring.Counter, labels ...[]string) CounterSnapshot
```
... and documenting what `labels` actually encodes.

Note that I also suggest renaming the function.

server/log_operation_manager_test.go
```diff
@@ -171,15 +172,58 @@ func TestLogOperationManagerPassesIDs(t *testing.T) {
...
+		LogStorage:   fakeStorage,
+		AdminStorage: mockAdmin,
+	}
+
+	mockLogOp := NewMockLogOperation(ctrl)
+	infoMatcher := logOpInfoMatcher{50}
+	mockLogOp.EXPECT().ExecutePass(gomock.Any(), logID1, infoMatcher).Return(1, nil)
+	mockLogOp.EXPECT().ExecutePass(gomock.Any(), logID2, infoMatcher).Return(0, errors.New("test error"))
+
+	signingRunsSnapshot := testonly.SnapshotCounter(signingRuns, logID1Label, logID2Label)
```
pavelkalinnikov: ... consequently, these calls will need to pass 2 slices of strings instead of 2 strings.

server/log_operation_manager_test.go
```diff
@@ -200,22 +244,82 @@ func TestLogOperationManagerOperationLoopPassesIDs(t *testing.T) {
...
 		}
+		return 0, nil
 	})
 
 	info := defaultLogOperationInfo(registry)
 	lom := NewLogOperationManager(info, mockLogOp)
 	lom.OperationLoop(ctx)
 }
 
+func TestLogOperationManagerOperationLoopExecutePassError(t *testing.T) {
```
pavelkalinnikov: Is there a way to share some of this code with the other test function above? Seems like the only principal difference between them is `OperationSingle` vs `OperationLoop`?

monitoring/testonly/delta.go
pavelkalinnikov: I don't feel strongly about renaming though, `SnapshotCounter` seems like a good verb.

monitoring/testonly/delta.go
RJPercival: Done.

monitoring/testonly/delta.go
RJPercival: Thanks for the explanation, I misunderstood how those labels were used in the monitoring package. I've fixed this PR to use them as intended.

server/log_operation_manager_test.go
RJPercival: It looks like it should be possible to dedup a fair amount of this code. I'll do this as a follow-up PR.

monitoring/testonly/delta.go
```diff
@@ -15,30 +15,45 @@
...
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/google/trillian/monitoring"
 )
 
-// CounterSnapshot records the values of a counter at a specific point in time.
+// CounterSnapshot records the value of a counter at a specific point in time.
```
pavelkalinnikov: nit: Would `// ... records the value of one or more counters at ...` be more accurate? Or maybe `any number of counters`?

monitoring/testonly/delta.go
```diff
@@ -15,30 +15,45 @@
...
-		values: make(map[string]float64, len(labels)),
-	}
-	for _, l := range labels {
-		s.values[l] = c.Value(l)
+		values: make(map[string]float64),
 	}
 	return s
 }
 
+// Record stores the current value of the counter.
```
pavelkalinnikov: nit: `... the counter identified by its labels.`

server/log_operation_manager_test.go
pavelkalinnikov: SGTM, will be happy to review it.

monitoring/testonly/delta.go
RJPercival: I suppose it depends on whether you think of `Counter` as being one counter with many values, or as a collection of counters.

RJPercival: From reading [the Prometheus docs](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels), I think the correct interpretation is that it's one counter that contains one or more time series, each identified by a set of labels. I've simplified the code and reworded the documentation to reflect this.

monitoring/testonly/delta.go
```diff
@@ -15,45 +15,34 @@
...
-	"strings"
-
 	"github.com/google/trillian/monitoring"
 )
 
-// CounterSnapshot records the value of a counter at a specific point in time.
+// CounterSnapshot records the latest value from a time series in a counter.
+// This value can then be compared with future values. Note that a counter can
+// contain many time series, but a CounterSnapshot will track only one.
+// A CounterSnapshot is useful in tests because counters do not reset between
```
pavelkalinnikov: Is it safe to assume that tests don't run concurrently? That would make counter checks unreliable.

RJPercival: I've added some comments to state that tests using CounterSnapshot should not be run in parallel. Someone would have to add a call to `t.Parallel()` to the tests in order for this to be a problem.

PR #1275, opened 2018-09-07
Link to Daz Wilkin's Medium post about running Trillain on GCP

RJPercival:
While this post is likely to become out-of-date, it is better than a TODO for the time being.

### DIFF
--- a/examples/deployment/README.md
+++ b/examples/deployment/README.md
@@ -51,7 +51,9 @@ a cloud.
 
 ### Run in GCP
 
-TODO
+There is a good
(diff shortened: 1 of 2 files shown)


examples/deployment/README.md
```diff
@@ -51,7 +51,9 @@ a cloud.
 
 ### Run in GCP
 
-TODO
+There is a good
```
AlCutter: While you're here, maybe worth adding a link to [the kubernetes docs](https://github.com/google/trillian/tree/master/examples/deployment/kubernetes) too?

RJPercival: Done. Also moved the AWS instructions into the "aws" directory, for consistency and to avoid people missing them when linked directly to the "aws" directory.

PR #1295, opened 2018-09-18
[storage] Increase log name length limit

pavelkalinnikov:


pavelkalinnikov:
Maybe 32 or 30 to be a bit more conservative?

RJPercival:
I don't think we need to be particularly conservative - I'd be surprised if anyone had so many trees that the size of this field became an issue (before anything else did).

pavelkalinnikov:
At the minimum, I will point out this change in the release notes.

RJPercival:
I've created issue #1297 to track automating schema updates.

Issue #1298, opened 2018-09-19
Support for PostgreSQL


vishalkuo:
Following a conversation with @RJPercival on the trillian slack channel, I think it'd be helpful to add support for a postgres backed data store. I'd like to use this issue for further discussion/tracking progress.

I'll likely borrow heavily from https://github.com/Martin2112/trillian/tree/pgsql/storage/pgsql as a starting point

pavelkalinnikov:
What's the status of this work?

@vishalkuo Is the postgres Trillian storage used anywhere?

vishalkuo:
Apologies - I have no had time to look into this at all. It looks like @bigjimnolan has also started some work on this. It might be worth following up on his PRs.

pavelkalinnikov:
I see only one commit from @bigjimnolan [here](https://github.com/google/trillian/commits?author=bigjimnolan). Jim, could you comment whether this is actually being used / tested / developed?

pavelkalinnikov:
Could you also clarify whether you use the `trillian_log_server` and `trillian_log_signer` binaries that we provide in this repository, or you have your own `main.go` files?

Just to let you know, the PR #1919 disables postgres storage provider in `main.go` files of `trillian_log_server` and `trillian_log_signer` by default given the development status, until we know that it needs to be enabled. Feel free to submit the corresponding "enabling" PR, or ask me/team to do that.

vishalkuo:
Got it. I believe I do use the provided binaries. Apologies for the delay on this and I appreciate the patience/follow up

pavelkalinnikov:
Closing this for now. Feel free to reopen if you actively work on PostgreSQL support.

marcopeereboom:
Boo!

Am I the only one waiting for postgres support? In my experience maria doesn't quite hack it under load.

Would you guys merge a PR if it magically appeared?

Martin2112:
Hi,

It's not that we're opposed to Postgres per se. Our concerns are around the costs of ongoing maintenance, support and CI / build infrastructure, which will multiply by the number of supported storage types.

We did look at Postgres when we started the project but could not find an equivalent to Galera in terms of replication characteristics and the tests with what was available at the time were not very encouraging.

However, it should be possible to use a Postgres storage implementation hosted outside of our repository. It will require the `main,go` files for the binaries to be forked and a storage provider to be written to wrap up the flags but that should be more or less it. If there are obstacles to doing this then we'd probably be willing to look at making smaller changes to support it.

[...]

PR #1299, opened 2018-09-19
Adds sql creation scripts for postgres

vishalkuo:
Referencing https://github.com/google/trillian/issues/1298, this PR adds the SQL creation script for a postgres data store and a stripped down reset script. I've done my best to translate the MySQL <> Postgres types; however they may not be a perfect mapping and it's possible that there'll need to be further changes made but I figured it's a good starting point for now. 

Note that I switched to camel casing since Postgres is case insensitive and would normalize everything to being lowercased which is hard to read without underscores.

### DIFF
--- a/storage/postgres/storage.sql
+++ b/storage/postgres/storage.sql
@@ -60,12 +60,13 @@ CREATE TABLE IF NOT EXISTS tree_head(
   root_hash              BYTEA NOT NULL,
   root_signature         BYTEA NOT NULL,
   tree_revision          BIGINT,
-  PRIMARY KEY(tree_id, tree_head_timestamp),
+  PRIMARY KEY(tree_id, tree_revision),
   FOREIGN KEY(tree_id) REFERENCES trees(tree_id) ON DELETE CASCADE
 );
 
-CREATE UNIQUE INDEX TreeHeadRevisionIdx
-  ON tree_head(tree_id, tree_revision);
+-- TODO(vishal) benchmark this to see if it's a suitable replacement for not
+-- having a DESC scan on the primary key
+CREATE UNIQUE INDEX TreeHeadRevisionIdx ON tree_head(tree_id, tree_revision DESC);
 
 -- ---------------------------------------------
 -- Log specific stuff here
--- a/storage/postgres/storage.sql
+++ b/storage/postgres/storage.sql
@@ -44,7 +44,7 @@ CREATE TABLE IF NOT EXISTS tree_control(
 
 CREATE TABLE IF NOT EXISTS subtree(
   tree_id               BIGINT NOT NULL,
-  subtree_id            BIT VARYING(255) NOT NULL,
+  subtree_id            BYTEA NOT NULL,
   nodes                 BYTEA NOT NULL,
   subtree_revision      INTEGER NOT NULL, 
   PRIMARY KEY(tree_id, subtree_id, subtree_revision),
@@ -57,8 +57,8 @@ CREATE TABLE IF NOT EXISTS tree_head(
   tree_id                BIGINT NOT NULL,
   tree_head_timestamp    BIGINT,
   tree_size              BIGINT,
-  root_hash              BIT VARYING(255) NOT NULL,
-  root_signature         BIT VARYING(1024) NOT NULL,
+  root_hash              BYTEA NOT NULL,
+  root_signature         BYTEA NOT NULL,
   tree_revision          BIGINT,
   PRIMARY KEY(tree_id, tree_head_timestamp),
   FOREIGN KEY(tree_id) REFERENCES trees(tree_id) ON DELETE CASCADE
@@ -79,16 +79,16 @@ CREATE UNIQUE INDEX TreeHeadRevisionIdx
 -- are allowed they will all reference this row.
 CREATE TABLE IF NOT EXISTS leaf_data(
   tree_id               BIGINT NOT NULL,
-  -- This is a personality specific has of some subset of the leaf data.
+  -- This is a personality specific hash of some subset of the leaf data.
   -- It's only purpose is to allow Trillian to identify duplicate entries in
   -- the context of the personality.
-  leaf_identity_hash     BIT VARYING(255) NOT NULL,
+  leaf_identity_hash     BYTEA NOT NULL,
   -- This is the data stored in the leaf for example in CT it contains a DER encoded
   -- X.509 certificate but is application dependent
   leaf_value            BYTEA NOT NULL,
   -- This is extra data that the application can associate with the leaf should it wish to.
   -- This data is not included in signing and hashing.
-  extra_dta            BYTEA,
+  extra_data            BYTEA,
   -- The timestamp from when this leaf data was first queued for inclusion.
   queue_timestamp_nanos  BIGINT NOT NULL,
   PRIMARY KEY(tree_id, leaf_identity_hash),
@@ -107,10 +107,10 @@ CREATE TABLE IF NOT EXISTS sequenced_leaf_data(
   -- This is a personality specific has of some subset of the leaf data.
   -- It's only purpose is to allow Trillian to identify duplicate entries in
   -- the context of the personality.
-  leaf_identity_hash        BIT VARYING(255) NOT NULL,
+  leaf_identity_hash        BYTEA NOT NULL,
   -- This is a MerkleLeafHash as defined by the treehasher that the log uses. For example for
   -- CT this hash will include the leaf prefix byte as well as the leaf data.
-  merkle_leaf_hash          BIT VARYING(255) NOT NULL,
+  merkle_leaf_hash          BYTEA NOT NULL,
   integrate_timestamp_nanos BIGINT NOT NULL,
   PRIMARY KEY(tree_id, sequence_number),
   FOREIGN KEY(tree_id) REFERENCES trees(tree_id) ON DELETE CASCADE,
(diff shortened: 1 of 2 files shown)


storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+
+-- Creating index at same time as table allows some storage engines to better
+-- optimize physical storage layout. Most engines allow multiple nulls in a
+-- unique index but some may not.
+
+-- A leaf that has not been sequenced has a row in this table. If duplicate leaves
+-- are allowed they will all reference this row.
+CREATE TABLE IF NOT EXISTS leaf_data(
+  tree_id               BIGINT NOT NULL,
+  -- This is a personality specific has of some subset of the leaf data.
```
gdbelvin: s/has/hash/

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  -- It's only purpose is to allow Trillian to identify duplicate entries in
+  -- the context of the personality.
+  leaf_identity_hash     BIT VARYING(255) NOT NULL,
```
gdbelvin: should this be `BYTEA` or `BYTEA(32)`?

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+-- on the log parameters and we can't insert into this table until we have the sequence number
+-- which is not available at the time we queue the entry. We need both hashes because the
+-- LeafData table is keyed by the raw data hash.
+CREATE TABLE IF NOT EXISTS sequenced_leaf_data(
+  tree_id                   BIGINT NOT NULL,
+  sequence_number           BIGINT NOT NULL,
+  -- This is a personality specific has of some subset of the leaf data.
+  -- It's only purpose is to allow Trillian to identify duplicate entries in
+  -- the context of the personality.
+  leaf_identity_hash        BIT VARYING(255) NOT NULL,
```
gdbelvin: Same `BYTEA` vs `BIT` question

storage/postgres/storage.sql
AlCutter: In other storage schemas, we've deliberately kept the field size overly large in order to accommodate > 256bit hashes for future expansion.

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  signing_enabled           BOOLEAN NOT NULL,
+  sequencing_enabled        BOOLEAN NOT NULL,
+  sequence_interval_seconds INTEGER NOT NULL,
+  PRIMARY KEY(tree_id),
+  FOREIGN KEY(tree_id) REFERENCES trees(tree_id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS subtree(
+  tree_id               BIGINT NOT NULL,
+  subtree_id            BIT VARYING(255) NOT NULL,
```
AlCutter: Perhaps BYTEA here too (and consider size)

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  FOREIGN KEY(tree_id) REFERENCES Trees(tree_id) ON DELETE CASCADE
+);
+
+-- The TreeRevisionIdx is used to enforce that there is only one STH at any
+-- tree revision
+CREATE TABLE IF NOT EXISTS tree_head(
+  tree_id                BIGINT NOT NULL,
+  tree_head_timestamp    BIGINT,
+  tree_size              BIGINT,
+  root_hash              BIT VARYING(255) NOT NULL,
```
AlCutter: Perhaps BYTEA here too (and consider size)

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  root_signature         BIT VARYING(1024) NOT NULL,
```
AlCutter: Perhaps BYTEA here too (and consider size)

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  -- This is a personality specific has of some subset of the leaf data.
+  -- It's only purpose is to allow Trillian to identify duplicate entries in
+  -- the context of the personality.
+  leaf_identity_hash     BIT VARYING(255) NOT NULL,
+  -- This is the data stored in the leaf for example in CT it contains a DER encoded
+  -- X.509 certificate but is application dependent
+  leaf_value            BYTEA NOT NULL,
+  -- This is extra data that the application can associate with the leaf should it wish to.
+  -- This data is not included in signing and hashing.
+  extra_dta            BYTEA,
```
AlCutter: Consider `extra_data` to mirror naming elsewhere?

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+CREATE TABLE IF NOT EXISTS sequenced_leaf_data(
+  tree_id                   BIGINT NOT NULL,
+  sequence_number           BIGINT NOT NULL,
+  -- This is a personality specific has of some subset of the leaf data.
+  -- It's only purpose is to allow Trillian to identify duplicate entries in
+  -- the context of the personality.
+  leaf_identity_hash        BIT VARYING(255) NOT NULL,
+  -- This is a MerkleLeafHash as defined by the treehasher that the log uses. For example for
+  -- CT this hash will include the leaf prefix byte as well as the leaf data.
+  merkle_leaf_hash          BIT VARYING(255) NOT NULL,
```
AlCutter: Perhaps BYTEA here too (and consider size)

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  -- The bucket field is to allow the use of time based ring bucketed schemes if desired. If
+  -- unused this should be set to zero for all entries.
+  bucket                INTEGER NOT NULL,
+  -- This is a personality specific hash of some subset of the leaf data.
+  -- It's only purpose is to allow Trillian to identify duplicate entries in
+  -- the context of the personality.
+  leaf_identity_hash    BIT VARYING(255) NOT NULL,
+  -- This is a MerkleLeafHash as defined by the treehasher that the log uses. For example for
+  -- CT this hash will include the leaf prefix byte as well as the leaf data.
+  merkle_leaf_hash      BIT VARYING(255) NOT NULL,
```
AlCutter: Perhaps BYTEA here too (and consider size)

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+  queue_timestamp_nanos BIGINT NOT NULL,
+  -- This is a SHA256 hash of the TreeID, LeafIdentityHash and QueueTimestampNanos. It is used
+  -- for batched deletes from the table when trillian_log_server and trillian_log_signer are
+  -- built with the batched_queue tag.
+  queue_id              BIT VARYING(32) DEFAULT NULL UNIQUE,
```
AlCutter: Perhaps BYTEA here too (and consider size)

vishalkuo:
@gdbelvin I've switched from `BIT VARYING` to `BYTEA`; however, it's worth noting that BYTEA doesn't allow one to specify length (although it has a limit of [1GB](https://stackoverflow.com/questions/26635547/postgresql-difference-between-bytea-and-bit-varying-types)). 

Regarding integration tests; are you referring to tests similar to those under the `integration` directory? If that's the case; it seems like I'll need to finish the rest of the pg implementation before being able to provide those. My plan was to use these types as a starting point and build `log_storage.go` and `admin_storage.go`, modifying these sql types as need be, until they reach a satisfactory state. Let me know if this isn't how you intended for this process to go and I can close this and come back with a full PR once I've got a locally working version.

gdbelvin:
Smaller PRs are easier to review.  Let's proceed with this and we can modify as needed when we get some unit tests working.

gdbelvin:
Thanks for pointing out the lack of a size limit on `BYTEA`.  
`BYTEA` is probably still more appropriate for hash outputs than `BIT` which is designed for bit masks.

AlCutter:
Thanks for the updates, I think it's probably ok to have `BYTEA` with no limit, the on-disk size requirements are essentially the same anyway - `len(DATA) + (1 or 4) bytes`, and the personality layer can enforce sensible lower size limits for leaves.

storage/postgres/storage.sql
```diff
@@ -0,0 +1,138 @@
...
+-- The TreeRevisionIdx is used to enforce that there is only one STH at any
+-- tree revision
+CREATE TABLE IF NOT EXISTS tree_head(
+  tree_id                BIGINT NOT NULL,
+  tree_head_timestamp    BIGINT,
+  tree_size              BIGINT,
+  root_hash              BYTEA NOT NULL,
+  root_signature         BYTEA NOT NULL,
+  tree_revision          BIGINT,
+  PRIMARY KEY(tree_id, tree_head_timestamp),
```
AlCutter: I _think_ it would be better to have the PK be `tree_id, tree_revision`, you could then drop the `index` below (I don't think anything needs this key to contain the timestamp - the cloudspanner implementation uses `/treeID/tree_revision` as it's PK on the equivalent table there BTW).

Separately, and I must admit I don't know much about PGSQL, but other storage impls define the PK here as `tree_id, tree_rev DESC`, because by far the most common operation on this table is a scan which  wants to fetch the newest entry.  It looks like PG doesn't support the `DESC` modifier on the PK fields, so I don't know what implications this will have for performance when you come to write the code itself.

Might be worth popping a comment or a `TODO` here as a note to yourself to look at this area again in detail once the code lands too.

vishalkuo:
@AlCutter you're right, it seems like postgres doesn't support scan ordering for primary keys. That being said, the workaround seems to be to create a unique index as well, seen [here](https://stackoverflow.com/questions/45597101/primary-key-with-asc-or-desc-ordering). I left the index below the create table statement and added a `desc` modifier on `tree_revision` to scan the most recent changes first. I also added a TODO to benchmark this change once we have more code in place to make sure this change is acceptable. Let me know if that works

PR #1305, opened 2018-09-25
Adds pg implementations for admin GetTree(id) and CreateTree

vishalkuo:
This PR introduces a partial implementation for admin storage in postgres. Particularly, it implements the `CreateTree` and `GetTree(id)` methods. 

Much of the logic is borrowed heavily from the MySQL methods (including validation and setup) so I tried to pull out common helpers wherever I could. That being said, I think there were even further commonalities I could pull out but I worried that that'd be too big a change for this PR (let me know if you feel otherwise). 

As far as testing, I copied most of the logic in mysql's storage tests but have yet to build our a test driver. I figured I'd leave that to another PR as this one was getting fairly large.

### DIFF
--- a/server/postgres_storage_provider.go
+++ b/server/postgres_storage_provider.go
@@ -1,0 +1,83 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"database/sql"
+	"flag"
+	"sync"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/monitoring"
+	"github.com/google/trillian/storage"
+	"github.com/google/trillian/storage/postgres"
+
+	// Load MySQL driver
+	_ "github.com/lib/pq"
+)
+
+var (
+	pgConnStr         = flag.String("pg_uri", "user=postgres dbname=test port=5432 sslmode=disable", "Connection URI for Postgres database")
+	pgOnce            sync.Once
+	pgStorageInstance *pgProvider
+)
+
+func init() {
+	if err := RegisterStorageProvider("postgres", newPGProvider); err != nil {
+		glog.Fatalf("Failed to register storage provider postgres: %v", err)
+	}
+}
+
+type pgProvider struct {
+	db *sql.DB
+	mf monitoring.MetricFactory
+}
+
+func newPGProvider(mf monitoring.MetricFactory) (StorageProvider, error) {
+	var err error
+
+	pgOnce.Do(func() {
+		var db *sql.DB
+		db, err = postgres.OpenDB(*pgConnStr)
+		if err != nil {
+			return
+		}
+
+		pgStorageInstance = &pgProvider{
+			db: db,
+			mf: mf,
+		}
+	})
+	if err != nil {
+		return nil, err
+	}
+	return pgStorageInstance, nil
+}
+
+func (s *pgProvider) LogStorage() storage.LogStorage {
+	panic("Not Implemented")
+}
+
+func (s *pgProvider) MapStorage() storage.MapStorage {
+	panic("Not Implemented")
+}
+
+func (s *pgProvider) AdminStorage() storage.AdminStorage {
+	return postgres.NewAdminStorage(s.db)
+}
+
+func (s *pgProvider) Close() error {
+	return s.db.Close()
+}
--- a/server/postgres_storage_provider.go
+++ b/server/postgres_storage_provider.go
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	pgConnStr         = flag.String("pg_uri", "user=postgres dbname=test port=5432 sslmode=disable", "Connection URI for Postgres database")
+	pgConnStr         = flag.String("pg_uri", "user=postgres dbname=test port=5432 sslmode=disable", "Connection string for Postgres database")
 	pgOnce            sync.Once
 	pgStorageInstance *pgProvider
 )
--- a/server/postgres_storage_provider.go
+++ b/server/postgres_storage_provider.go
@@ -24,13 +24,14 @@ import (
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/postgres"
 
-	// Load MySQL driver
+	// Load PG driver
 	_ "github.com/lib/pq"
 )
 
 var (
-	pgConnStr         = flag.String("pg_uri", "user=postgres dbname=test port=5432 sslmode=disable", "Connection string for Postgres database")
+	pgConnStr         = flag.String("pg_conn_str", "user=postgres dbname=test port=5432 sslmode=disable", "Connection string for Postgres database")
 	pgOnce            sync.Once
+	pgOnceErr         error
 	pgStorageInstance *pgProvider
 )
 
@@ -46,12 +47,10 @@ type pgProvider struct {
 }
 
 func newPGProvider(mf monitoring.MetricFactory) (StorageProvider, error) {
-	var err error
-
 	pgOnce.Do(func() {
 		var db *sql.DB
-		db, err = postgres.OpenDB(*pgConnStr)
-		if err != nil {
+		db, pgOnceErr = postgres.OpenDB(*pgConnStr)
+		if pgOnceErr != nil {
 			return
 		}
 
@@ -60,8 +59,8 @@ func newPGProvider(mf monitoring.MetricFactory) (StorageProvider, error) {
 			mf: mf,
 		}
 	})
-	if err != nil {
-		return nil, err
+	if pgOnceErr != nil {
+		return nil, pgOnceErr
 	}
 	return pgStorageInstance, nil
 }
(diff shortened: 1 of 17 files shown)


storage/postgres/testdb/testdb.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Create a randomly-named database and then connect using the new name.
+	name := fmt.Sprintf("trl_%v", time.Now().UnixNano())
+
+	stmt := fmt.Sprintf("CREATE DATABASE %v", name)
+	if _, err := db.ExecContext(ctx, stmt); err != nil {
+		return nil, fmt.Errorf("error running statement %q: %v", stmt, err)
+	}
+
+	connStr := fmt.Sprintf("%s database=%s", dataSource, name)
```
AlCutter: this looks like it'll be `"database=database=test sslmode=disable"` is that right?

storage/postgres/testdb/testdb.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	sqlBytes, err := ioutil.ReadFile(trillianSQL)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, stmt := range strings.Split(sanitize(string(sqlBytes)), ";") {
```
AlCutter: I guess it's not killer, but it looks like this would break if anyone ever put a semicolon that wasn't used as a line terminator anywhere in the file? 

`sanitize` attempts to fix this for comments (which was my first thought), so maybe it's good enough, just thought I'd mention it.

storage/postgres/testdb/testdb.go
```diff
@@ -0,0 +1,125 @@
...
+			continue
+		}
+		if _, err := db.ExecContext(ctx, stmt); err != nil {
+			return nil, fmt.Errorf("error running statement %q: %v", stmt, err)
+		}
+	}
+	return db, nil
+}
+
+func sanitize(script string) string {
```
AlCutter: Possibly worth a func comment here?
`sanitize` is perhaps a bit vague.

storage/postgres/testdb/testdb.go
```diff
@@ -0,0 +1,125 @@
...
+			continue // skip empty lines and comments
+		}
+		buf.WriteString(line)
+		buf.WriteString("\n")
+	}
+	return buf.String()
+}
+
+// SkipIfNoPG is a test helper that skips tests that require a local PG.
+func SkipIfNoPG(t *testing.T) {
```
AlCutter: does this need to be exported?

storage/postgres/tree_storage.go
```diff
@@ -0,0 +1,33 @@
...
+
+package postgres
+
+import (
+	"database/sql"
+
+	"github.com/golang/glog"
+)
+
+// OpenDB opens a database connection for all MySQL-based storage implementations.
```
AlCutter: "all PostgreSQL-based"

storage/postgres/tree_storage.go
```diff
@@ -0,0 +1,33 @@
...
+func OpenDB(dbURL string) (*sql.DB, error) {
```
AlCutter: is dbURL a URL still, or should it be something about db connection parameters (or whatever terminology PGSQL uses for this)

storage/postgres/tree_storage.go
```diff
@@ -0,0 +1,33 @@
...
+	db, err := sql.Open("postgres", dbURL)
+	if err != nil {
+		// Don't log uri as it could contain credentials
```
AlCutter: Don't log xxx

storage/storage_helpers.go
```diff
@@ -0,0 +1,162 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"database/sql"
+	"fmt"
+	time "time"
```
AlCutter: no need to rename `time` to `time` :)

storage/storage_helpers.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+// ToMillisSinceEpoch converts a timestamp into milliseconds since epoch
+func ToMillisSinceEpoch(t time.Time) int64 {
+	return t.UnixNano() / 1000000
+}
+
+// FromMillisSinceEpoch converts an UNIX typestamp to a time struct
+func FromMillisSinceEpoch(ts int64) time.Time {
+	secs := int64(ts / 1000)
```
AlCutter: can just be `return time.Unix(0, ts*1000000)`

storage/storage_helpers.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
AlCutter: Maybe this should go under `storage/sql` or something - explicitly helpers for golang SQL  compatible implementations?

storage/mysql/admin_storage.go
```diff
@@ -161,7 +158,7 @@ func (t *adminTX) GetTree(ctx context.Context, treeID int64) (*trillian.Tree, er
 	defer stmt.Close()
 
 	// GetTree is an entry point for most RPCs, let's provide somewhat nicer error messages.
-	tree, err := readTree(stmt.QueryRowContext(ctx, treeID))
+	tree, err := storage.ReadTree(stmt.QueryRowContext(ctx, treeID))
```
AlCutter: I don't know if this is a good idea or not...
OTOH it does promote reuse and reduce boilerplate, but on the other it forces some restrictions on implementations, maybe that's a good trade off to make until we discover a scenario where it doesn't work, let's see what others think.

@pphaneuf @Martin2112 @RJPercival  @daviddrysdale

storage/storage_helpers.go
```diff
@@ -0,0 +1,162 @@
...
+
+// ValidateStorageSettings checks for storage settings and returns an error if they exist
+func ValidateStorageSettings(tree *trillian.Tree) error {
+	if tree.StorageSettings != nil {
+		return fmt.Errorf("storage_settings not supported, but got %v", tree.StorageSettings)
+	}
+	return nil
+}
+
+// Row defines a common interface between sql.Row and sql.Rows(!), so we have to
```
RJPercival: Looks like this comment got a bit mangled. It was originally "There's no common interface between sql.Row and sql.Rows(!), so we have to define one."

storage/storage_helpers.go
```diff
@@ -0,0 +1,162 @@
...
+
+// SetNullStringIfValid assigns src to dest if src is Valid.
+func SetNullStringIfValid(src sql.NullString, dest *string) {
+	if src.Valid {
+		*dest = src.String
+	}
+}
+
+// ValidateStorageSettings checks for storage settings and returns an error if they exist
+func ValidateStorageSettings(tree *trillian.Tree) error {
```
RJPercival: It doesn't seem like this should be a general helper, since storage_settings is supposed to be specific to individual storage implementations.

server/postgres_storage_provider.go
```diff
@@ -0,0 +1,83 @@
...
+		glog.Fatalf("Failed to register storage provider postgres: %v", err)
+	}
+}
+
+type pgProvider struct {
+	db *sql.DB
+	mf monitoring.MetricFactory
+}
+
+func newPGProvider(mf monitoring.MetricFactory) (StorageProvider, error) {
```
RJPercival: You need to apply a fix here similar to PR #1251.

storage/mysql/admin_storage.go
vishal-stripe: Yeah, definitely wasn't sure how much to pull out/leave in. My general understanding was that `db/sql` abstracted _most_ differences away to only being at the query level but I could be wrong about that (especially with types, etc.)

vishalkuo: Yeah, definitely wasn't sure how much to pull out/leave in. My general understanding was that db/sql abstracted most differences away to only being at the query level but I could be wrong about that (especially with types, etc.)

storage/postgres/admin_storage_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func cleanTestDB(db *sql.DB) {
+	for _, table := range allTables {
+		if _, err := db.ExecContext(context.TODO(), fmt.Sprintf("DELETE FROM %s", table)); err != nil {
+			panic(fmt.Sprintf("Failed to delete rows in %s: %v", table, err))
+		}
+	}
+}
+
+func openTestDBOrDie() *sql.DB {
```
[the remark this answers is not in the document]
vishal-stripe: This is only used in `TestMain`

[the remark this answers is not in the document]
vishalkuo: This is only used in `TestMain`

storage/postgres/testdb/testdb.go
```diff
@@ -0,0 +1,131 @@
...
+			continue // skip empty lines and comments
+		}
+		buf.WriteString(line)
+		buf.WriteString("\n")
+	}
+	return buf.String()
+}
+
+// skipIfNoPG is a test helper that skips tests that require a local PG.
+func skipIfNoPG(t *testing.T) {
```
golangcibot: `skipIfNoPG` is unused

storage/sql.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, fmt.Errorf("unknown HashAlgorithm: %v", hashAlgorithm)
+	}
+	if sa, ok := spb.DigitallySigned_SignatureAlgorithm_value[signatureAlgorithm]; ok {
+		tree.SignatureAlgorithm = spb.DigitallySigned_SignatureAlgorithm(sa)
+	} else {
+		return nil, fmt.Errorf("unknown SignatureAlgorithm: %v", signatureAlgorithm)
+	}
+
+	// Let's make sure we didn't mismatch any of the casts above
+	ok := tree.TreeState.String() == treeState
```
AlCutter: Maybe just make this: 

```golang
ok := tree.TreeState.String() == treestate &&
      tree.TreeType.String() == treeType &&
      tree.HashStrategy.String() == hashStrategy &&
      ...
```

?

AlCutter:
Thanks, Vishal!

PR #1309, opened 2018-10-05
Completes admin storage endpoints for postgres

vishalkuo:
Adds missing admin endpoints: ListTrees, ListTreeIDs, UpdateTree, SoftDeleteTree, HardDeleteTree, and UndeleteTree. Most logic is copied from the mysql implementation apart from the sql queries itself. 

Also attempts to add a postgres db to the travis build; will test that to see what effort is necessary to make that happen

Issue: #1298

### DIFF
--- a/scripts/postgres_resetdb.sh
+++ b/scripts/postgres_resetdb.sh
@@ -33,7 +33,7 @@ collect_vars() {
   FLAGS+=(--host "${PG_HOST}")
   FLAGS+=(--port "${PG_PORT}")
 
-    # handle flags
+  # handle flags
   FORCE=false
   while [[ $# -gt 0 ]]; do
     case "$1" in
--- a/scripts/postgres_resetdb.sh
+++ b/scripts/postgres_resetdb.sh
@@ -32,6 +32,17 @@ collect_vars() {
   FLAGS+=(-U "${PG_ROOT_USER}")
   FLAGS+=(--host "${PG_HOST}")
   FLAGS+=(--port "${PG_PORT}")
+
+    # handle flags
+  FORCE=false
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      --force) FORCE=true ;;
+      --help) usage; exit ;;
+      *) FLAGS+=("$1")
+    esac
+    shift 1
+  done
 }
 
 main() {
--- a/storage/postgres/admin_storage.go
+++ b/storage/postgres/admin_storage.go
@@ -52,7 +52,7 @@ const (
 		delete_time_millis
 	FROM trees`
 
-	nonDeletedWhere       = " WHERE (deleted is NULL OR deleted = false)"
+	nonDeletedWhere       = " WHERE deleted = false"
 	selectNonDeletedTrees = selectTrees + nonDeletedWhere
 
 	selectTreeIDs           = "SELECT tree_id FROM trees"
--- a/storage/postgres/admin_storage.go
+++ b/storage/postgres/admin_storage.go
@@ -17,7 +17,6 @@ package postgres
 import (
 	"context"
 	"database/sql"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -52,6 +51,13 @@ const (
 		deleted,
 		delete_time_millis
 	FROM trees`
+
+	nonDeletedWhere       = " WHERE (deleted is NULL OR deleted = false)"
+	selectNonDeletedTrees = selectTrees + nonDeletedWhere
+
+	selectTreeIDs           = "SELECT tree_id FROM trees"
+	selectNonDeletedTreeIDs = selectTreeIDs + nonDeletedWhere
+
 	selectTreeByID = selectTrees + " WHERE tree_id = $1"
 
 	insertSQL = `INSERT INTO trees(
@@ -76,11 +82,18 @@ const (
 		sequencing_enabled,
 		sequence_interval_seconds)
 	VALUES($1, $2, $3, $4)`
-)
 
-var (
-	// errNotImplemented is an error indicating a function's implementation has not been completed
-	errNotImplemented = errors.New("not implemented")
+	updateTreeSQL = `UPDATE trees SET tree_state = $1, tree_type = $2, display_name = $3, 
+		description = $4, update_time_millis = $5, max_root_duration_millis = $6, private_key = $7
+		WHERE tree_id = $8`
+
+	softDeleteSQL = "UPDATE trees SET deleted = $1, delete_time_millis = $2 WHERE tree_id = $3"
+
+	selectDeletedSQL = "SELECT deleted FROM trees WHERE tree_id = $1"
+
+	deleteFromTreeControlSQL = "DELETE FROM tree_control WHERE tree_id = $1"
+
+	deleteFromTreesSQL = "DELETE FROM trees WHERE tree_id = $1"
 )
 
 // NewAdminStorage returns a storage.AdminStorage implementation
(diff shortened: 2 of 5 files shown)


storage/postgres/admin_storage.go
```diff
@@ -76,6 +83,20 @@ const (
 		sequencing_enabled,
 		sequence_interval_seconds)
 	VALUES($1, $2, $3, $4)`
+
+	updatedDeletedSQL = "UPDATE trees SET deleted = $1, delete_time_millis = $2 WHERE tree_id = $3"
```
golangcibot: `updatedDeletedSQL` is unused

scripts/postgres_resetdb.sh
```diff
@@ -32,6 +32,17 @@ collect_vars() {
   FLAGS+=(-U "${PG_ROOT_USER}")
   FLAGS+=(--host "${PG_HOST}")
   FLAGS+=(--port "${PG_PORT}")
+
+    # handle flags
```
RJPercival: nit: indentation is a little off here.

storage/postgres/admin_storage.go
```diff
@@ -52,6 +51,13 @@ const (
 		deleted,
 		delete_time_millis
 	FROM trees`
+
+	nonDeletedWhere       = " WHERE (deleted is NULL OR deleted = false)"
```
RJPercival: Is there a reason for `deleted` being nullable?

storage/postgres/admin_storage.go
```diff
@@ -188,11 +201,63 @@ func (t *adminTX) GetTree(ctx context.Context, treeID int64) (*trillian.Tree, er
...
+	}
+	defer rows.Close()
+	trees := []*trillian.Tree{}
+	for rows.Next() {
+		tree, err := storage.ReadTree(rows)
+		if err != nil {
+			return nil, err
+		}
+		trees = append(trees, tree)
+	}
```
RJPercival: Should check `rows.Err()` here.

storage/postgres/admin_storage.go
```diff
@@ -188,11 +201,63 @@ func (t *adminTX) GetTree(ctx context.Context, treeID int64) (*trillian.Tree, er
...
+		query = selectTrees
+	} else {
+		query = selectNonDeletedTrees
+	}
+
+	stmt, err := t.tx.PrepareContext(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	defer stmt.Close()
```
RJPercival: nit: It'd be nice if the whitespace were consistent between methods. In `ListTreeIDs()`, for example, you have a blank line here.

storage/postgres/admin_storage.go
```diff
@@ -188,11 +201,63 @@ func (t *adminTX) GetTree(ctx context.Context, treeID int64) (*trillian.Tree, er
...
+	defer rows.Close()
+
+	treeIDs := []int64{}
+	var treeID int64
+	for rows.Next() {
+		if err := rows.Scan(&treeID); err != nil {
+			return nil, err
+		}
+		treeIDs = append(treeIDs, treeID)
+	}
```
RJPercival: Should check `rows.Err()` here.

storage/postgres/admin_storage.go
```diff
@@ -278,19 +343,112 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
...
+	if err := validateStorageSettings(tree); err != nil {
+		return nil, err
+	}
+
+	// Use the time truncated-to-millis throughout, as that's what's stored.
+	nowMillis := storage.ToMillisSinceEpoch(time.Now())
+	now := storage.FromMillisSinceEpoch(nowMillis)
+	tree.UpdateTime, err = ptypes.TimestampProto(now)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build update time: %v", err)
```
RJPercival: s/update time/tree.UpdateTime/

Makes it a bit easier to recognise, from the error message, what it was trying to do.

storage/postgres/admin_storage_test.go
```diff
@@ -84,6 +85,119 @@ func TestCreateTreeInvalidStates(t *testing.T) {
...
+		}
+	}
+}
+
+func TestAdminTX_StorageSettingsNotSupported(t *testing.T) {
+	cleanTestDB(DB, t)
+	s := postgres.NewAdminStorage(DB)
+	ctx := context.Background()
+
+	settings, err := ptypes.MarshalAny(&keyspb.PEMKeyFile{})
```
RJPercival: Are you just using `keyspb.PEMKeyFile` here as a convenient empty proto? If so, it'd be a bit clearer to use the [empty.Empty](https://godoc.org/github.com/golang/protobuf/ptypes/empty#Empty) proto.

storage/postgres/admin_storage_test.go
```diff
@@ -100,6 +214,17 @@ func openTestDBOrDie() *sql.DB {
 	}
 	return db
 }
+
+func setNulls(ctx context.Context, db *sql.DB, treeID int64) error {
+	stmt, err := db.PrepareContext(ctx, "UPDATE trees SET display_name = NULL, description = NULL WHERE tree_id = $1")
```
RJPercival: `deleted` and `delete_time_millis` are also nullable.

storage/postgres/admin_storage.go
vishalkuo: The PG schema followed the example of `storage.sql` for MySQL: https://github.com/google/trillian/blob/master/storage/mysql/storage.sql#L23-L24 if you think that this shouldn't be the case let me know.

RJPercival: I'm not sure why it was made nullable for MySQL. It may have been because it was added to the schema later. In this case, I don't see any reason to make it nullable and it is causing more code complexity.

storage/postgres/admin_storage.go
```diff
@@ -188,11 +201,71 @@ func (t *adminTX) GetTree(ctx context.Context, treeID int64) (*trillian.Tree, er
...
+
+	rows, err := stmt.QueryContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	trees := []*trillian.Tree{}
+	for rows.Next() {
+		if err := rows.Err(); err != nil {
```
RJPercival: You need to check `rows.Err()` after the loop (see the documentation for `rows.Next()`), since `rows.Next()` may return false either because there are no more rows, or an error occurred.

storage/postgres/admin_storage.go
```diff
@@ -278,19 +351,112 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
...
+		deleted,
+		deleteTimeMillis,
+		treeID); err != nil {
+		return nil, err
+	}
+	return t.GetTree(ctx, treeID)
+}
+
+func validateDeleted(ctx context.Context, tx *sql.Tx, treeID int64, wantDeleted bool) error {
+	var nullDeleted sql.NullBool
```
RJPercival: The `deleted` column is no longer nullable, so this code can be simplified.

vishalkuo: Sorry for all the confusion, I appreciate the patience

AlCutter:
How are we doing on this one?

vishalkuo:
Apologies for the delay, I believe I've addressed all review comments but happy to adjust again if necessary

AlCutter:
No problem, just checking in!
@RJPercival are you happy to approve?

PR #1311, opened 2018-10-08
Flag Postgres tests as skipped when no Postgres DB available

RJPercival:


Martin2112:
Do we have to do this per-test. I thought we did it differently for MySQL but it's a long time ago so I might have made that up.

RJPercival:
The MySQL tests don't mark the tests as skipped either so, if I'm going to make this change, I should probably do it there too. It's not possible to mark them all as skipped in a single call, because only `testing.T` has a `Skip()` method.

RJPercival:
I'm going to close this PR, on the basis that it introduces the risk of future MySQL/Postgres tests accidentally not being skipped. It's a shame that `testing.M` doesn't have a `Skip()` method.

PR #1312, opened 2018-10-10
Simplify GKE configs

AlCutter:
- Remove node-pools (they add complexity, and we'll use priority/preemption instead)
- Reuse a single `deploy.sh` to push to CI too.

_This is going to break CI, but I'll come back and fix that after the fact_

### DIFF
--- a/examples/deployment/kubernetes/deploy.sh
+++ b/examples/deployment/kubernetes/deploy.sh
@@ -12,39 +12,46 @@ export LOG_URL=TODO
 export MAP_URL=TODO
 export TAG=$(git rev-parse HEAD)
 
-# Get Trillian
-go get github.com/google/trillian/...
-cd $GOPATH/src/github.com/google/trillian
-
-# Build docker images
-docker build -f examples/deployment/docker/log_server/Dockerfile -t gcr.io/$PROJECT_NAME/log_server:$TAG .
-docker build -f examples/deployment/docker/log_signer/Dockerfile -t gcr.io/$PROJECT_NAME/log_signer:$TAG .
-# TODO(al): when cloudspanner supports maps:
-# docker build -f examples/deployment/docker/map_server/Dockerfile -t us.gcr.io/$PROJECT_NAME/map_server:$TAG .
-
 # Connect to gcloud
 gcloud config set project "${PROJECT_NAME}"
-gcloud config set compute/zone "${ZONE}"
+gcloud config set compute/zone "${MASTER_ZONE}"
 
 # Configure Docker to use gcloud credentials with Google Container Registry
 gcloud auth configure-docker
 
-# Push docker images
-docker push "gcr.io/${PROJECT_NAME}/log_server:${TAG}"
-docker push "gcr.io/${PROJECT_NAME}/log_signer:${TAG}"
+# Get Trillian
+go get github.com/google/trillian/...
+cd $GOPATH/src/github.com/google/trillian
+
+echo "Building docker images..."
+docker build --quiet -f examples/deployment/docker/log_server/Dockerfile -t gcr.io/$PROJECT_NAME/log_server:$TAG .
+docker build --quiet -f examples/deployment/docker/log_signer/Dockerfile -t gcr.io/$PROJECT_NAME/log_signer:$TAG .
+# TODO(al): when cloudspanner supports maps:
+# docker build -f examples/deployment/docker/map_server/Dockerfile -t us.gcr.io/$PROJECT_NAME/map_server:$TAG .
+
+echo "Pushing docker images..."
+gcloud docker -- push gcr.io/${PROJECT_NAME}/log_server:${TAG}
+gcloud docker -- push gcr.io/${PROJECT_NAME}/log_signer:${TAG}
+
+echo "Tagging docker images..."
+gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME}/log_server:${TAG} gcr.io/${PROJECT_NAME}/log_server:latest
+gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME}/log_signer:${TAG} gcr.io/${PROJECT_NAME}/log_signer:latest
 
 # TODO(al): when cloudspanner supports maps:
 # gcloud docker -- push "us.gcr.io/${PROJECT_NAME}/map_server:${TAG}"
 
+echo "Updating jobs..."
 # Prepare configmap:
-kubectl delete configmap deploy-config
-envsubst < examples/deployment/kubernetes/trillian-cloudspanner.yaml | kubectl create -f -
+kubectl delete configmap deploy-config || true
+envsubst < ${DIR}/${CONFIGMAP} | kubectl create -f -
 
 # Launch with kubernetes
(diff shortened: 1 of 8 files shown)


examples/deployment/kubernetes/deploy.sh
```diff
@@ -3,50 +3,72 @@
 # - Cluster has already been created & configured using the create.sh script
 # - Go 1.9 is installed
 
+function checkEnv() {
+  if [ -z ${PROJECT_NAME+x} ] ||
+     [ -z ${CLUSTER_NAME+x} ] ||
+     [ -z ${MASTER_ZONE+x} ] ||
+     [ -z ${CONFIG_MAP+x} ]; then
+    echo "You must either pass an argument which is a config file, or set all the required environment variables"
```
RJPercival: nit: Add `>&2` to send this error message to stderr.

examples/deployment/kubernetes/deploy.sh
```diff
@@ -3,50 +3,72 @@
...
+echo "Pushing docker images..."
+gcloud docker -- push gcr.io/${PROJECT_NAME}/log_server:${IMAGE_TAG}
+gcloud docker -- push gcr.io/${PROJECT_NAME}/log_signer:${IMAGE_TAG}
 
-# Push docker images
-docker push "gcr.io/${PROJECT_NAME}/log_server:${TAG}"
-docker push "gcr.io/${PROJECT_NAME}/log_signer:${TAG}"
+echo "Tagging docker images..."
+gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME}/log_server:${IMAGE_TAG} gcr.io/${PROJECT_NAME}/log_server:latest
+gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME}/log_signer:${IMAGE_TAG} gcr.io/${PROJECT_NAME}/log_signer:latest
```
RJPercival: Why do they need to be explicitly tagged as "latest"?

examples/deployment/kubernetes/deploy.sh
```diff
@@ -3,50 +3,72 @@
...
 
 # TODO(al): when cloudspanner supports maps:
 # gcloud docker -- push "us.gcr.io/${PROJECT_NAME}/map_server:${TAG}"
 
+echo "Updating jobs..."
 # Prepare configmap:
-kubectl delete configmap deploy-config
-envsubst < examples/deployment/kubernetes/trillian-cloudspanner.yaml | kubectl create -f -
+kubectl delete configmap deploy-config || true
+envsubst < ${CONFIGMAP} | kubectl create -f -
```
RJPercival: `checkEnv` checks for `${CONFIG_MAP}`, but here you use `${CONFIGMAP}`.

scripts/deploy_gce_ci.sh
```diff
@@ -9,34 +9,8 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-kubectl delete configmap deploy-config
-envsubst < ${DIR}/../examples/deployment/kubernetes/${CONFIGMAP} | kubectl create -f -
-
-envsubst < ${DIR}/../examples/deployment/kubernetes/trillian-log-deployment.yaml | kubectl apply -f -
-envsubst < ${DIR}/../examples/deployment/kubernetes/trillian-log-service.yaml | kubectl apply -f -
-envsubst < ${DIR}/../examples/deployment/kubernetes/trillian-log-signer-deployment.yaml | kubectl apply -f -
-envsubst < ${DIR}/../examples/deployment/kubernetes/trillian-log-signer-service.yaml | kubectl apply -f -
-kubectl set image deployment/trillian-logserver-deployment trillian-logserver=gcr.io/${PROJECT_NAME}/log_server:${TRAVIS_COMMIT}
-kubectl set image deployment/trillian-logsigner-deployment trillian-log-signer=gcr.io/${PROJECT_NAME}/log_signer:${TRAVIS_COMMIT}
+./examples/kubernetes/deploy.sh
```
RJPercival: I think this should be `${DIR}/../examples/deployment/kubernetes/deploy.sh`.

examples/deployment/kubernetes/deploy.sh
```diff
@@ -3,50 +3,72 @@
...
-kubectl get all
-kubectl get services
+envsubst < ${DIR}/trillian-log-deployment.yaml | kubectl apply -f -
+envsubst < ${DIR}/trillian-log-service.yaml | kubectl apply -f -
+envsubst < ${DIR}/trillian-log-signer-deployment.yaml | kubectl apply -f -
+envsubst < ${DIR}/trillian-log-signer-service.yaml | kubectl apply -f -
+kubectl set image deployment/trillian-logserver-deployment trillian-logserver=gcr.io/${PROJECT_NAME}/log_server:${IMAGE_TAG}
+kubectl set image deployment/trillian-logsigner-deployment trillian-log-signer=gcr.io/${PROJECT_NAME}/log_signer:${IMAGE_TAG}
 
 # TODO(al): Create trees
```
RJPercival: nit: Could either delete this TODO or replace it with `./provision_tree.sh` (and change README.md appropriately).

examples/deployment/kubernetes/deploy.sh
AlCutter: done

examples/deployment/kubernetes/deploy.sh
AlCutter: That's currently what the `*-deployment.yaml` files look for, is there a better way?

examples/deployment/kubernetes/deploy.sh
AlCutter: good spot, thanks!

scripts/deploy_gce_ci.sh
AlCutter: yup - done

examples/deployment/kubernetes/deploy.sh
AlCutter: Deleted - I think it makes more sense to leave it as a separate step.

AlCutter:
PTAL, thanks!

examples/deployment/kubernetes/deploy.sh
RJPercival: Further down, you use `kubectl set image` to override the default image (i.e. the image with the "latest" tag) and use `${IMAGE_TAG}` instead. Actually, it might be best to replace "latest" in `*-deployment.yaml` with an env var that `envsubst` can substitute. That could be done as a separate PR though.

examples/deployment/kubernetes/deploy.sh
```diff
@@ -3,54 +3,69 @@
...
+  source $1
+else
+  checkEnv
+fi
+
 
 export LOG_URL=TODO
 export MAP_URL=TODO
-export TAG=$(git rev-parse HEAD)
+export IMAGE_TAG=${IMAGE_TAG:-$(git rev-parse HEAD)}
```
RJPercival: nit: The default `IMAGE_TAG` could be misleading, since the user may have modified but not committed some files which will subsequently get built into the image.

AlCutter: I've made it append a `-dirty` string to the hash if that's the case.

PR #1324, opened 2018-10-18
Bump to prom2sd v0.3.2

AlCutter:


examples/deployment/kubernetes/trillian-log-deployment.yaml
```diff
@@ -66,7 +66,8 @@ spec:
           - --source=logserver:http://localhost:8091/metrics
           - --pod-id=$(POD_NAME)
           - --namespace-id=$(POD_NAMESPACE)
-          - --metrics-resolution=60s
+          - --scrape-interval=5s
```
Martin2112: Fairly idle question - are these related to the prom2sd upgrade or the cost of metrics you mentioned yesterday.

AlCutter: prom2sd upgrade - the flags have changed...

PR #1326, opened 2018-10-22
Kube deploy fixups

AlCutter:
- PROJECT_NAME -> PROJECT_ID
- Make Map deployment optional
- Fix name of map container

### DIFF
--- a/examples/deployment/kubernetes/create.sh
+++ b/examples/deployment/kubernetes/create.sh
@@ -27,7 +27,7 @@ if ! envsubst --help > /dev/null; then
 fi
 
 echo "Creating new Trillian deployment"
-echo "  Project name: ${PROJECT_NAME}"
+echo "  Project name: ${PROJECT_ID}"
 echo "  Cluster name: ${CLUSTER_NAME}"
 echo "  Region:       ${REGION}"
 echo "  Node Locs:    ${NODE_LOCATIONS}"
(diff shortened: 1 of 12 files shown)


examples/deployment/kubernetes/create.sh
```diff
@@ -23,7 +23,7 @@ if ! jq --help > /dev/null; then
 fi
 
 echo "Creating new Trillian deployment"
-echo "  Project name: ${PROJECT_NAME}"
+echo "  Project name: ${PROJECT_ID}"
 echo "  Cluster name: ${CLUSTER_NAME}"
 echo "  Region:       ${REGION}"
 echo "  Node Locs:    ${NODE_LOCATIONS}"
```
Martin2112: Add `${RUN_MAP} to the summary?

examples/deployment/kubernetes/deploy.sh
```diff
@@ -56,15 +60,21 @@ kubectl delete configmap deploy-config || true
...
-kubectl set image deployment/trillian-logsigner-deployment trillian-log-signer=gcr.io/${PROJECT_NAME}/log_signer:${IMAGE_TAG}
-kubectl set image deployment/trillian-mapserver-deployment trillian-mapserver=gcr.io/${PROJECT_NAME}/map_server:${IMAGE_TAG}
+kubeconfigs="trillian-log-deployment.yaml trillian-log-service.yaml trillian-log-signer-deployment.yaml trillian-log-signer-service.yaml"
+if ${RUN_MAP}; then
+  kubeconfigs+=" trillian-map-deployment.yaml trillian-map-service.yaml"
+fi
+for thing in ${kubeconfigs}; do
+  echo ${thing}
+  envsubst < ${DIR}/${thing} | kubectl apply -f -
+don
```
Martin2112: Who is don?

examples/deployment/kubernetes/create.sh
AlCutter: I figured that was more about the deploy than the creation?

examples/deployment/kubernetes/deploy.sh
AlCutter: Don's done.

examples/deployment/kubernetes/create.sh
Martin2112: Yeah maybe. Might be worth adding it somewhere that there's no map being set up but not a big deal.

PR #1327, opened 2018-10-22
Replace `gcloud docker` with `docker`

gdbelvin:
`gcloud docker` will be deprecated soon.  This PR cleans up the following warnings: 

WARNING: `gcloud docker` will not be supported for Docker client versions above 18.03.

As an alternative, use `gcloud auth configure-docker` to configure `docker` to
use `gcloud` as a credential helper, then use `docker` as you would for non-GCR
registries, e.g. `docker pull gcr.io/project-id/my-image`. Add
`--verbosity=error` to silence this warning: `gcloud docker
--verbosity=error -- pull gcr.io/project-id/my-image`.

See: https://cloud.google.com/container-registry/docs/support/deprecation-notices#gcloud-docker

### DIFF
--- a/storage/mysql/kubernetes/image/push.sh
+++ b/storage/mysql/kubernetes/image/push.sh
@@ -42,6 +42,7 @@ while getopts "ht:" opt; do
   esac
 done
 
-gcloud docker -- build -t "$tag" .
-gcloud docker -- push "$tag"
+gcloud auth configure-docker
+docker build -t "$tag" .
+docker push "$tag"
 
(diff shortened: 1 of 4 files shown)


storage/mysql/kubernetes/image/push.sh
```diff
@@ -42,6 +42,7 @@ while getopts "ht:" opt; do
   esac
 done
 
-gcloud docker -- build -t "$tag" .
-gcloud docker -- push "$tag"
+gcloud auth configure-docker
+docker build -t "$tag" .
```
AlCutter: Could you use the `${...}` form for the variables?

gdbelvin: done

PR #1329, opened 2018-10-23
Remove Deprecated fields in SignedLogRoot

gdbelvin:
The fields have been moved to the serialized and signed LogRoot TLS data
structure. Clients SHOULD verify the signature and parse the struct to
retrieve the fields within.

Cleans up several outstanding TODO's


RJPercival:
Is this ready for review, given the "WIP" label?

gdbelvin:
It's not ready for review.
It turns out that the CTFE is currently using these deprecated fields. So
the first thing that I need you do is add sth verification to the CTFE

Martin2112:
Going to close as it's been done in #1462 with additional tests.

Issue #1338, opened 2018-10-31
Refactor compact Merkle tree


pavelkalinnikov:
The current code of `CompactMerkleTree` needs a bit of refreshment. Some things to improve:

- [x] Move it to a separate package and shorten the name to something Go-style like `compact.Tree`. This package can also contain other "compact" things, for example compact ranges.
- [x] The code of the tree tries to achieve multiple things at the same time when it initializes: fetch the nodes / handle errors, verify that the root hash matches, and actually build the compact tree. This can be split into 3 separate actions.
- [x] Initialization code fetches nodes sequentially. Instead, we should be able to make a list of nodes needed in advance, fetch them all simultaneously ([TreeStorage](https://github.com/google/trillian/blob/f9c892bee35a4b538142812396096bb1a7150956/storage/tree_storage.go#L81) allows to), and just pass them in to the compact Tree constructor. This will improve transactions startup time. Another reason to have a simplified constructor (that just takes a list of hashes) is that clients will be able to deserialize compact Merkle trees without writing a complicated `getNodeFunc`.
- [x] The tree changes behavior when the size is 2^k, including the way it gives out hashes for serialization. Doing so does not provide too much benefit in the first place, because tree size is very rarely equal to powers of 2. But, it complicates the code, and makes clients vulnerable to subtle bugs. Instead, we can just let the users of the type decide whether they need to omit 2^k hash or not.

PR #1339, opened 2018-10-31
Move compact Merkle tree to compact package

pavelkalinnikov:
This change introduces 'compact' package that currently includes the compact Merkle tree, but soon will have other compact structures.

It also shortens the name of the type to a higher-context Go-style name.

Addresses #1338 
Better reviewed commit by commit.


merkle/compact/tree.go
```diff
@@ -179,35 +179,35 @@ func (c *CompactMerkleTree) AddLeaf(data []byte, f setNodeFunc) (int64, []byte,
...
 // AddLeafHash adds the specified |leafHash| to the tree.
 // |f| is a callback which will be called multiple times with the full MerkleTree coordinates of nodes whose hash should be updated.
-func (c *CompactMerkleTree) AddLeafHash(leafHash []byte, f setNodeFunc) (int64, error) {
+func (t *Tree) AddLeafHash(leafHash []byte, f setNodeFunc) (int64, error) {
 	defer func() {
-		c.size++
+		t.size++
 		// TODO(al): do this lazily
-		c.recalculateRoot(f)
+		t.recalculateRoot(f)
```
golangcibot: Error return value of `t.recalculateRoot` is not checked

Martin2112:
Hmmm. We're losing Merkle from names of things and that doesn't seem ideal. OTOH we want the package name to be short.

pavelkalinnikov:
`compact.MerkleTree`?

Martin2112:
I think the notion of a "compact" as a thing isn't that meaningful for a package as it's more an implementation property but I'm struggling to think of an improvement. Maybe I'm overthinking it.

merkle/compact/tree.go
pavelkalinnikov: Not a scope of this PR.

Martin2112:
OTOH merkle/compact is not that dissimilar to Go libraries like crypto/whatever. Maybe it's fine.

merkle/compact/tree.go
Martin2112: Yes that's not easily fixed unfortunately. Can you add TODO to fix ignored error.

pavelkalinnikov: Done.

pavelkalinnikov:
After in-person conversations, we seem to agree that `compact.Tree` is fine. The `compact` name is quite specific in the context of Trillian, we don't have other "compact" stuff. And "merkle/compact" package path can help disambiguate in the unlikely case we do. E.g. we can import it as `compactmerkle "github.com/google/trillian/merkle/compact"`, and use it like this: `compactmerkle.Tree`.

PR #1343, opened 2018-11-11
Adds postgres implementation for `tree_storage`

vishalkuo:
This PR adds a partial implementation for the `ReadOnlyTreeTX` in postgres.

This is mostly a copy from mySQL's tree_storage.go file. The only major difference is the `expandPlaceholderSQL` method which needs a little more attention due to formatting requiring numeric placeholders for pg instead of `?` in mySQL. 

I'm not sure how to test this PR without starting work on the `log_storage` implementation and I was worried at that point it'd be really long and difficult to review. If we're fine merging a NOOP PR in so that it's in a manageable size, I can add the rest in subsequent PRs with tests. 

Alternatively, I can try to break this PR down into smaller ones that tests functions solely in tree storage OR create a much larger PR with tests via log storage. 

I have no strong preference to either option, I'd just love to get some feedback on how to proceed.

Issue: #1298

### DIFF
--- a/storage/postgres/admin_storage_test.go
+++ b/storage/postgres/admin_storage_test.go
@@ -34,15 +34,15 @@ const selectTreeControlByID = "SELECT signing_enabled, sequencing_enabled, seque
 
 func TestPgAdminStorage(t *testing.T) {
 	tester := &testonly.AdminStorageTester{NewAdminStorage: func() storage.AdminStorage {
-		cleanTestDB(DB, t)
-		return NewAdminStorage(DB)
+		cleanTestDB(db, t)
+		return NewAdminStorage(db)
 	}}
 	tester.RunAllTests(t)
 }
 
 func TestAdminTX_CreateTree_InitializesStorageStructures(t *testing.T) {
-	cleanTestDB(DB, t)
-	s := NewAdminStorage(DB)
+	cleanTestDB(db, t)
+	s := NewAdminStorage(db)
 	ctx := context.Background()
 
 	tree, err := storage.CreateTree(ctx, s, testonly.LogTree)
@@ -53,7 +53,7 @@ func TestAdminTX_CreateTree_InitializesStorageStructures(t *testing.T) {
 	// Check if TreeControl is correctly written.
 	var signingEnabled, sequencingEnabled bool
 	var sequenceIntervalSeconds int
-	if err := DB.QueryRowContext(ctx, selectTreeControlByID, tree.TreeId).Scan(&signingEnabled, &sequencingEnabled, &sequenceIntervalSeconds); err != nil {
+	if err := db.QueryRowContext(ctx, selectTreeControlByID, tree.TreeId).Scan(&signingEnabled, &sequencingEnabled, &sequenceIntervalSeconds); err != nil {
 		t.Fatalf("Failed to read TreeControl: %v", err)
 	}
 	// We don't mind about specific values, defaults change, but let's check
@@ -64,8 +64,8 @@ func TestAdminTX_CreateTree_InitializesStorageStructures(t *testing.T) {
 }
 
 func TestCreateTreeInvalidStates(t *testing.T) {
-	cleanTestDB(DB, t)
-	s := NewAdminStorage(DB)
+	cleanTestDB(db, t)
+	s := NewAdminStorage(db)
 	ctx := context.Background()
 
 	states := []trillian.TreeState{trillian.TreeState_DRAINING, trillian.TreeState_FROZEN}
@@ -80,8 +80,8 @@ func TestCreateTreeInvalidStates(t *testing.T) {
 }
 
 func TestAdminTX_TreeWithNulls(t *testing.T) {
-	cleanTestDB(DB, t)
-	s := NewAdminStorage(DB)
+	cleanTestDB(db, t)
+	s := NewAdminStorage(db)
 	ctx := context.Background()
 
 	// Setup: create a tree and set all nullable columns to null.
@@ -93,7 +93,7 @@ func TestAdminTX_TreeWithNulls(t *testing.T) {
 	}
 	treeID := tree.TreeId
 
-	if err := setNulls(ctx, DB, treeID); err != nil {
+	if err := setNulls(ctx, db, treeID); err != nil {
 		t.Fatalf("setNulls() = %v, want = nil", err)
 	}
 
@@ -149,8 +149,8 @@ func TestAdminTX_TreeWithNulls(t *testing.T) {
 }
 
 func TestAdminTX_StorageSettingsNotSupported(t *testing.T) {
-	cleanTestDB(DB, t)
-	s := NewAdminStorage(DB)
+	cleanTestDB(db, t)
+	s := NewAdminStorage(db)
 	ctx := context.Background()
 
 	settings, err := ptypes.MarshalAny(&empty.Empty{})
--- a/storage/postgres/admin_storage_test.go
+++ b/storage/postgres/admin_storage_test.go
@@ -12,43 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package postgres_test
+package postgres
 
 import (
 	"context"
 	"database/sql"
-	"flag"
 	"fmt"
-	"os"
 	"testing"
 
-	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/trillian"
 	"github.com/google/trillian/storage"
-	"github.com/google/trillian/storage/postgres"
-	"github.com/google/trillian/storage/postgres/testdb"
 	"github.com/google/trillian/storage/testonly"
 )
 
 var allTables = []string{"unsequenced", "tree_head", "sequenced_leaf_data", "leaf_data", "subtree", "tree_control", "trees"}
-var DB *sql.DB
 
 const selectTreeControlByID = "SELECT signing_enabled, sequencing_enabled, sequence_interval_seconds FROM tree_control WHERE tree_id = $1"
 
 func TestPgAdminStorage(t *testing.T) {
 	tester := &testonly.AdminStorageTester{NewAdminStorage: func() storage.AdminStorage {
 		cleanTestDB(DB, t)
-		return postgres.NewAdminStorage(DB)
+		return NewAdminStorage(DB)
 	}}
 	tester.RunAllTests(t)
 }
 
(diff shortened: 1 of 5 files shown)


storage/postgres/storage_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"database/sql"
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/storage/postgres/testdb"
+)
+
+var DB *sql.DB
```
AlCutter: This _probably_ doesn't need to be exported, but it'd be worth adding a comment here explaining that it's shared by all tests etc.

storage/postgres/storage_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestMain(m *testing.M) {
+	flag.Parse()
+	if !testdb.PGAvailable() {
+		glog.Errorf("PG not available, skipping all PG storage tests")
+		return
+	}
+	DB = testdb.OpenTestDBOrDie()
+	defer DB.Close()
+	ec := m.Run()
+	os.Exit(ec)
```
AlCutter: I think this might defeat the deferred `Close` call above, you could probably work around that with something like:

```golang
  ec := 0
  defer func() { os.Exit(ec) }()
  ...
  defer DB.Close()
  ...
  ec = m.Run()
}
```

storage/postgres/testdb/testdb.go
```diff
@@ -120,3 +120,13 @@ func sanitize(script string) string {
 func getConnStr(name string) string {
 	return fmt.Sprintf("database=%s %s", name, *pgOpts)
 }
+
+// OpenTestDBOrDie attempts to return a connection to a new postgres
+// test database and fails if unable to do so
```
AlCutter: nitty: missing trailing `.`
Actually this is missing throughout, can you pop them on all the comments?

storage/postgres/testdb/testdb.go
```diff
@@ -120,3 +120,13 @@ func sanitize(script string) string {
...
+func OpenTestDBOrDie() *sql.DB {
+	db, err := NewTrillianDB(context.TODO())
```
AlCutter: Any reason not to pass a `context` in here rather than adding a TODO? (Does this have to fit with an existing interface?)

storage/postgres/tree_storage.go
```diff
@@ -15,11 +15,52 @@
...
+	"github.com/google/trillian/storage"
+	"github.com/google/trillian/storage/cache"
+	"github.com/google/trillian/storage/storagepb"
 )
 
+const (
+	placeholderSQL        = "<placeholder>"
+	insertSubtreeMultiSQL = `INSERT INTO subtree(tree_id, subtree_id, nodes, subtree_revision) ` + placeholderSQL
+	selectSubtreeSQL      = `
+ SELECT x.subtree_id, x.max_revision, subtree.nodes
```
AlCutter: Can probably indent this all to help with readability.

storage/postgres/tree_storage.go
```diff
@@ -31,3 +72,317 @@ func OpenDB(connStr string) (*sql.DB, error) {
...
+	}
+	stx := t.tx.StmtContext(ctx, tmpl)
+	defer stx.Close()
+
+	r, err := stx.ExecContext(ctx, args...)
+	if err != nil {
+		glog.Warningf("Failed to set merkle subtrees: %s", err)
+		return err
+	}
+	_, _ = r.RowsAffected()
```
AlCutter: Does this have a side-effect?

storage/postgres/tree_storage.go
```diff
@@ -31,3 +72,317 @@ func OpenDB(connStr string) (*sql.DB, error) {
...
+func (t *treeTX) Commit() error {
+	if t.writeRevision > -1 {
+		if err := t.subtreeCache.Flush(func(st []*storagepb.SubtreeProto) error {
+			return t.storeSubtrees(context.TODO(), st)
+		}); err != nil {
+			glog.Warningf("TX commit flush error: %v", err)
+			return err
+		}
+	}
+	t.closed = true
```
AlCutter: Should the reads/writes to `closed` not be covered by a mutex?

AlCutter: Looks like it's the same in the `mysql` version, @Martin2112 do you have thoughts?
The `cloudspanner` implementation by comparison does hold `mu` in `Close()`

storage/postgres/tree_storage_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+)
+
+type expandTestcase struct {
+	input    *statementSkeleton
+	expected string
+}
+
+func TestInitializes(t *testing.T) {
+	_ = &statementSkeleton{}
```
AlCutter: Sorry, I feel like I'm missing something obvious, but other than the call to `newTreeStorage`, what are the rest of these assignment statements testing?

storage/postgres/tree_storage_test.go
```diff
@@ -0,0 +1,72 @@
...
+			expected: strings.Replace(selectSubtreeSQL, placeholderSQL, "$1,$2", 1),
+		},
+		{
+			input: &statementSkeleton{
+				sql:               insertSubtreeMultiSQL,
+				firstInsertion:    "VALUES(%s, %s, %s, %s)",
+				firstPlaceholders: 4,
+				restInsertion:     "(%s, %s, %s, %s)",
+				restPlaceholders:  4,
+				num:               2,
```
AlCutter: Maybe add another test with a different number of repeated `rest`? Currently both tests have `num=2`, and the `insertSubtree` case will almost certainly get used with multiple subtrees in the real code.

storage/postgres/admin_storage_test.go
```diff
@@ -12,43 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package postgres_test
+package postgres
```
AlCutter: Is it not possible to do black box testing of this? (just curious)

storage/postgres/tree_storage.go
vishalkuo: Not really sure.. It seemed to be in the MySQL one but I can't imagine it does

storage/postgres/tree_storage.go
vishalkuo: Should be fine but it does seem to indent the actual string https://play.golang.org/p/HRgEaPIx4J9

storage/postgres/testdb/testdb.go
vishalkuo: No reason, I was mostly copying things over from the mysql tests. I'll add a context with a 30 second timeout here

storage/postgres/tree_storage_test.go
vishalkuo: This was just to get around gometalinter's megacheck that functions/structs were going unused

storage/postgres/admin_storage_test.go
vishalkuo: Admin storage is actually currently blackbox tested; however I wanted to bring everything under the `postgres` package because 

A) that seems to be the pattern the rest of the codebase follows
B) It allows us to use a unified `TestMain` function.

Happy to change it back though, I don't have strong feelings about this

storage/postgres/tree_storage_test.go
AlCutter: Aha, ok - maybe pop a comment in there for other folks too? :)

storage/postgres/tree_storage.go
AlCutter: Yup, but I don't think it'll cause any problems and we win a bit on tidy source.

storage/postgres/testdb/testdb.go
```diff
@@ -120,3 +120,13 @@ func sanitize(script string) string {
 func getConnStr(name string) string {
 	return fmt.Sprintf("database=%s %s", name, *pgOpts)
 }
+
+// OpenTestDBOrDie attempts to return a connection to a new postgres
+// test database and fails if unable to do so.
+func OpenTestDBOrDie(ctx context.Context) *sql.DB {
```
RJPercival: nit: The convention is to suffix "OrDie" on the end of the function you're calling, i.e. "NewTrillianDBOrDie"

storage/postgres/tree_storage.go
```diff
@@ -15,19 +15,373 @@
...
+	// num is the total repetitions (firstInsertion + restInsertion * num - 1) that
+	// should be inserted.
+	num int
+}
+
+// expandPlaceholderSQL expands an sql statement by adding a specified number of '%s'
+// placeholder slots. At most one placeholder will be expanded.
+func expandPlaceholderSQL(skeleton *statementSkeleton) string {
+	if skeleton.num <= 0 {
+		panic(fmt.Errorf("trying to expand SQL placeholder with <= 0 parameters: %s", skeleton.sql))
```
RJPercival: We avoid using `panic` outside of tests.

storage/postgres/testdb/testdb.go
vishalkuo: For clarification, do you mean `NewTrillianDBOrDie` should just panic if there's an error? If that's the case we don't really need the `OpenTestDB` function. I'm going to go ahead with that assumption but let me know if that's wrong.

RJPercival: I just meant you should rename this function. There isn't already a `NewTrillianDBOrDie` function, is there?

vishalkuo: gotcha, will do

vishalkuo:
Polite bump

AlCutter:
@RJPercival you happy?

storage/postgres/tree_storage.go
```diff
@@ -15,19 +15,378 @@
...
+
+func newTreeStorage(db *sql.DB) *pgTreeStorage {
+	return &pgTreeStorage{
+		db:         db,
+		statements: make(map[string]map[int]*sql.Stmt),
+	}
+}
+
+// statementSkeleton contains the structure of a query to create.
+type statementSkeleton struct {
```
RJPercival: I assume this gets used a lot more in future PRs, thereby justifying its existence? It looks rather complicated for the two uses that appear in this PR. All of this string substitution into SQL and use of variables as printf format strings makes me a little uneasy!

RJPercival: Well if @AlCutter is happy with it, I don't want to block you any longer. I'll try and find the time to have a think about whether there's a better way of going about this and come back to it.

AlCutter: Thanks Rob, I just created #1369 to track this.

PR #1345, opened 2018-11-14
Make an experimental Merkle tree visualiser

AlCutter:
Potentially useful for visually understanding/reasoning about inclusion and consistency proofs.
Needs tests, but I still have to figure out how to even do that.

Also, I know nothing at all about Javascript.

Apparently you can view this using a prefix of `http://htmlpreview.github.io/?` and appending the github URL of the index.html blob. (e.g. http://htmlpreview.github.io/?https://github.com/google/trillian/blob/6e688f1afc43716dafa76eae41771cab5a152041/docs/merkletree/index.html)


Martin2112:
Nice!

AlCutter:
Ah, looks like I've introduced a bug on hover-inclusion proofs on the RHS :/
I'll sort that out.

pavelkalinnikov:
Nice tool, thanks for doing it! Would be nice to add compact ranges to it later. I wish it could output things directly to google docs drawings :)

Some observations after playing with it:

- When the tree size is 1, the node circle doesn't have a border (for bigger sizes it does).
- For small non-power-of-2 tree sizes (e.g. 2, 3, 5, 6) the circles are too big, and intersect each other.
- Personal taste: The topmost circles are too big for small trees. Maybe make an upper limit on the node size or something, to make it more balanced?
- When resizing the screen I sometimes get the upper and lower node circles clipped (e.g. if tree size = 8).

AlCutter:
Fixed up a bunch of stuff, PTAL?

Martin2112:
I just used a patch for preview locally.

The proof bug I saw before seems to have been fixed.

Not sure if it's a chrome issue but I see some clipping in some cases e.g. at tree size 7. Doesn't seem to happen for sizes > 8.

AlCutter:
Ah, this is dependent on the size of the browser window :/
I'll fix it in a follow-up.

PR #1348, opened 2018-11-14
Create user for pg testing

vishalkuo:
As discussed with @RJPercival on slack; this change creates a new postgres user with the reset db script. This user can be used for testing.

### DIFF
--- a/scripts/postgres_resetdb.sh
+++ b/scripts/postgres_resetdb.sh
@@ -12,6 +12,9 @@ Accepts environment variables:
 - POSTGRES_HOST: The hostname of the PG server (default: localhost).
 - POSTGRES_PORT: The port the PG server is listening on (default: 5432).
 - POSTGRES_DB: The name to give to the new Trillian database (default: test).
+- POSTGRES_USER: The name to give to the new Trillian user (default: test).
+- POSTGRES_PASSWORD: The password to use for the new Trillian user
+  (default: zaphod).
 EOF
 }
 
@@ -26,6 +29,8 @@ collect_vars() {
   [ -z ${POSTGRES_HOST+x} ] && POSTGRES_HOST="localhost"
   [ -z ${POSTGRES_PORT+x} ] && POSTGRES_PORT="5432"
   [ -z ${POSTGRES_DB+x} ] && POSTGRES_DB="test"
+  [ -z ${POSTGRES_USER+x} ] && POSTGRES_USER="test"
+  [ -z ${POSTGRES_PASSWORD+x} ] && POSTGRES_PASSWORD="zaphod"
   FLAGS=()
 
   FLAGS+=(-U "${POSTGRES_ROOT_USER}")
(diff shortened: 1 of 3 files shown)


scripts/postgres_resetdb.sh
```diff
@@ -63,6 +68,16 @@ main() {
         die "Error: Failed to create database '${POSTGRES_DB}'."
       psql "${FLAGS[@]}" -d ${POSTGRES_DB} -f ${TRILLIAN_PATH}/storage/postgres/storage.sql || \
         die "Error: Failed to create tables in '${POSTGRES_DB}' database."
+      if ! psql "${FLAGS[@]}" -t -c "SELECT 1 FROM pg_user WHERE usename = '${POSTGRES_USER}'" | grep -q 1; then
+        psql "${FLAGS[@]}" -c "CREATE USER ${POSTGRES_USER} WITH PASSWORD '${POSTGRES_PASSWORD}';" || \
```
RJPercival: You can add `WITH SUPERUSER` to the `CREATE USER` query to avoid having to use `ALTER ROLE` later.

scripts/postgres_resetdb.sh
```diff
@@ -63,6 +68,16 @@ main() {
...
+          die "Error: Failed to create user '${POSTGRES_USER}'."
+      fi
+      psql "${FLAGS[@]}" -d ${POSTGRES_DB} -c "GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO ${POSTGRES_USER};" || \
+        die "Error: Failed to grant '${POSTGRES_USER}' user all privileges on tables in '${POSTGRES_DB}'."
+      psql "${FLAGS[@]}" -d ${POSTGRES_DB} -c "GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO ${POSTGRES_USER};" || \
+        die "Error: Failed to grant '${POSTGRES_USER}' user all privileges on tables in '${POSTGRES_DB}'."
+      psql "${FLAGS[@]}" -d ${POSTGRES_DB} -c "ALTER ROLE ${POSTGRES_USER} WITH SUPERUSER;" || \
```
RJPercival: Why does the user need superuser privileges? Is this just for convenience?

scripts/postgres_resetdb.sh
```diff
@@ -63,6 +68,16 @@ main() {
...
       psql "${FLAGS[@]}" -d ${POSTGRES_DB} -f ${TRILLIAN_PATH}/storage/postgres/storage.sql || \
         die "Error: Failed to create tables in '${POSTGRES_DB}' database."
+      if ! psql "${FLAGS[@]}" -t -c "SELECT 1 FROM pg_user WHERE usename = '${POSTGRES_USER}'" | grep -q 1; then
+        psql "${FLAGS[@]}" -c "CREATE USER ${POSTGRES_USER} WITH PASSWORD '${POSTGRES_PASSWORD}';" || \
+          die "Error: Failed to create user '${POSTGRES_USER}'."
+      fi
+      psql "${FLAGS[@]}" -d ${POSTGRES_DB} -c "GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO ${POSTGRES_USER};" || \
+        die "Error: Failed to grant '${POSTGRES_USER}' user all privileges on tables in '${POSTGRES_DB}'."
+      psql "${FLAGS[@]}" -d ${POSTGRES_DB} -c "GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO ${POSTGRES_USER};" || \
+        die "Error: Failed to grant '${POSTGRES_USER}' user all privileges on tables in '${POSTGRES_DB}'."
```
RJPercival: s/privileges on tables/privileges on sequences/

scripts/postgres_resetdb.sh
RJPercival: This won't work if `$POSTGRES_PASSWORD` contains a `'` character. I'd be fine with just mentioning in the usage message that the password can't contain one, unless you can think of a simple way to avoid this problem (parameterised queries would be nice, but this is Bash).

PR #1350, opened 2018-11-16
Make Docker images consistent

RJPercival:
Consistently use:
- Go 1.10
- $GOFLAGS
- gcr.io/distroless/base

### DIFF
--- a/examples/deployment/docker/db_client/Dockerfile
+++ b/examples/deployment/docker/db_client/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.9
+FROM golang:1.10
 
 RUN apt-get update && \
     apt-get install -y mysql-client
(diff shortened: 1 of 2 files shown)


examples/deployment/docker/db_client/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.9
+FROM golang:1.10
```
gdbelvin: Do we want to just use `golang:1` to keep track with the latest revision of Go?

RJPercival: We could, but then the Docker images might be built with a newer version of Go than Travis uses.

Martin2112: Might be better to keep it explicit for the moment to avoid surprises.

PR #1351, opened 2018-11-16
Use specific tag for Docker images in Kubernetes configs

RJPercival:
The "latest" flag is just whatever the most recently pushed image is, which might not actually be the most up-to-date. It also results in automatic updates of the image used in the Kubernetes cluster, which might be unexpected. It's better to use a specific tag (e.g. a commit hash or release version). This makes it obvious which version of the code is being used in Kubernetes.


AlCutter:
Will you do `ctfe` too?

RJPercival:
Sure, shouldn't be a problem.

PR #1355, opened 2018-11-19
Embed formatting into RetriableError

pavelkalinnikov:
Follow up from [discussion](https://github.com/google/certificate-transparency-go/pull/400/files#r234285599) in #400.

### DIFF
--- a/client/backoff/backoff.go
+++ b/client/backoff/backoff.go
@@ -17,6 +17,7 @@ package backoff
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"time"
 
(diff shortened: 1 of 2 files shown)


gdbelvin:
That looks like a glog flag. 
Did this PR remove all references to glog?

pavelkalinnikov:
Oh, **that's** why, thank you. I wonder how this test is different from the others that it needs this flag.

pavelkalinnikov:
I guess, all the other tests happen to inherit the `glog` flag implicitly, and we enforce it by calling `go test` with `-alsologtostderr` in Travis. This package is quite trivial, and does not inherit this flag, so importing `glog` is a workaround.

PR #1360, opened 2018-11-29
Pull Request Template

gdbelvin:
We need a way to remind submitters to update the release notes. 

A Pull Request Template could be one such way.


AlCutter:
Presumably 3rd party authors of PRs won't be able to update the draft release notes, assuming that's the case should we mention it to not put folks off from raising a PR when they can't check all the boxes?

RJPercival:
It'd probably be best to move to just updating the [CHANGELOG](https://github.com/google/trillian/blob/master/CHANGELOG.md) file directly; that way, third-party contributors could propose changes to it as well. We could always reverse our current process and update the GitHub releases page from that file when we finalise a release.

AlCutter:
SGTM

PR #1361, opened 2018-11-29
Require docker-compose v3.1 or higher

RJPercival:
This is for compatibility with the upcoming CTFE docker-compose manifest,
which uses secrets and so requires docker-compose v3.1. If they don't
have the same version requirement, they won't be compatible.

### Checklist


RJPercival:
I should probably document that docker-compose v3.1 is now required in the CHANGELOG :slightly_smiling_face:

PR #1362, opened 2018-11-29
Update CHANGELOG with our draft release notes

RJPercival:
We will be keeping draft release notes in CHANGELOG.md from now on. This makes it possible for third party contributors to add their own, and for reviewers to easily see if the CHANGELOG has been updated when appropriate.

### Checklist


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # TRILLIAN Changelog
 
+## HEAD
```
RJPercival: Any preference on what to put here? "HEAD", "master", etc? Looking to communicate that it's what you'll get if you pull master and build Trillian.

RJPercival:
Oh this isn't the complete list for this release; this is just what was already in the draft release notes.

CHANGELOG.md
Martin2112: I think HEAD is unambiguous?

Martin2112:
OK.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+Not yet released; provisionally v2.0.0 (may change).
```
RJPercival: Since this mentions an API change, I guess there should be a major version bump. It does say an "internal" API though, so perhaps third-party storage implementations wouldn't be affected, and so it can be just a minor version bump?

Martin2112: I think it would need changes to third party storage but fairly minor ones. I think "internal" here means it doesn't affect end-users.

RJPercival: Ok. I think the idea of semantic versioning is that *any* backwards-incompatible change should result in a major version bump, so I think jumping to v2.0.0 is appropriate.

Martin2112: Not sure about that as it would only affect code outside this repository.

PR #1364, opened 2018-11-29
Re-generate API doc

RJPercival:
It was out-of-date and Travis didn't catch this because its test was broken (https://github.com/google/trillian/pull/1363).


RJPercival:
Yea, I mentioned some of the deficiencies in that auto-generated documentation when I was soliciting opinions on it prior to merging it. People said they'd rather have it than not though. There's an argument that we should stick to the Go rules of comment-writing for our protobufs as well (we've partly done that already), which bars things like lists in comments. That may or may not be a net gain in readability.

PR #1370, opened 2018-12-06
[Map] Add revision parameter to SetLeaves

### LINKED ISSUE #559: Map: Use "ExpectedMapRevision" in SetLeaves to kill race conditions.
A mapper's current operational cycle looks like:
- ReadLeaves
- Apply mutations
- SetLeaves

If two mappers are running at the same time for redundancy, they may trample each other's SetLeaves operations. To eliminate this, propose including an `expected_map_revision` field in the SetLeaves API which will cause the whole operation to fail if there already exists a SignedMapHead for the given revision.

pavelkalinnikov:
This change allows clients to specify the expected write revision when
they call SetLeaves. This enables idempotent retriable writes to maps.

Fixes #559

 

### Checklist

- [x] I have updated [documentation](docs/) accordingly (go generate).


pavelkalinnikov:
PTAL

trillian_map_api.proto
```diff
@@ -93,6 +93,11 @@ message SetMapLeavesRequest {
   // to continue mapping from an external data source.
   reserved 4;
   bytes metadata = 5;
+  // The map revision to associate the leaves with. The request will fail if
+  // this revision already exists, does not match the current write revision, or
```
pavelkalinnikov: Hm, probably suffices to remove all except "does not match the current write revision", as the others follow from it.

PR #1373, opened 2018-12-07
Remove HTTP API

### LINKED ISSUE #1279: HTTP API is untested
Trillian offers a HTTP API that is auto-generated (with the help of some annotations) from our RPC API, e.g. 
https://github.com/google/trillian/blob/670a91ca44c098055e1cf5339708e5e7e026f654/trillian_log_api.proto#L37-L42

However, we do not test it or use it ourselves, so its reliability and usability are unknown. We should either properly support it or remove it.

RJPercival:
We don't test, support or recommend the HTTP API, so we're removing it in favour of the gRPC API.

Fixes #1279. Requires #1372. 

### Checklist


RJPercival:
It turns out that the integration tests use the quota HTTP API. I'm going to create a CLI for the quota gRPC API and make the integration tests use that instead.

gdbelvin:
It would be nice to keep the work we did creating the HTTP apis - just move them to an example Trillian Front end

RJPercival:
I guess I can leave all of the HTTP options in the proto files, and even leave the generated grpc-gateway code, but just refactor the grpc-gateway server out into a personality. I'd have to run that personality in the integration tests in order for them to continue being able to use the quota HTTP API though. It might be simpler to still create a quota CLI.

RJPercival:
Superceded by #1677.

PR #1375, opened 2018-12-12
Client: AddSequencedLeaves

gdbelvin:
Allow others to use the connection to the log server to call methods
that the thick log client doesn't explicitly wrap.

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -358,7 +358,16 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
 }
 
 // AddSequencedLeaves adds any number of pre-sequenced leaves to the log.
-func (c *LogClient) AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) error {
+func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int64][]byte) error {
+	leaves := make([]*trillian.LogLeaf, 0, len(dataByIndex))
+	for index, data := range dataByIndex {
+		leaf, err := c.BuildLeaf(data)
+		if err != nil {
+			return err
+		}
+		leaf.LeafIndex = index
+		leaves = append(leaves, leaf)
+	}
 	_, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
 		LogId:  c.LogID,
 		Leaves: leaves,
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -357,6 +357,15 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
 	return err
 }
 
+// AddSequencedLeaves adds any number of pre-sequenced leaves to the log.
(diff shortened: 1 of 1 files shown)


RJPercival:
Why not instead add methods that wrap those API calls? There are some API methods marked as "DO NOT USE", which it seems less than ideal to expose.

gdbelvin:
Updated to wrap the new API call

client/log_client.go
```diff
@@ -357,6 +357,15 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
 	return err
 }
 
+// AddSequencedLeaves adds any number of pre-sequenced leaves to the log.
+func (c *LogClient) AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) error {
```
RJPercival: For consistency and ease of use, it might be better for this to work like `AddSequencedLeaf()`, e.g.
```go
func (c *LogClient) AddSequencedLeaves(ctx context.Context, data map[int64][]byte) error {
	leaves := make([]*LogLeaf, 0, len(data))
	for i, d := range data {
		leaf, err := c.BuildLeaf(d)
		if err != nil {
			return fmt.Errorf("error building leaf %d: %v", i, err)
		}
		leaf.LeafIndex = i
		leaves = append(leaves, leaf)
	}
	_, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
		LogId:  c.LogID,
		Leaves: leaves,
	})
	return err
}
```

gdbelvin: amazing idea.  Done.

client/log_client.go
```diff
@@ -357,6 +357,24 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
 	return err
 }
 
+// AddSequencedLeaves adds any number of pre-sequenced leaves to the log.
+func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int64][]byte) error {
+	leaves := make([]*trillian.LogLeaf, 0, len(dataByIndex))
+	for index, data := range dataByIndex {
+		leaf, err := c.BuildLeaf(data)
+		if err != nil {
+			return err
```
RJPercival: nit: it might be more helpful to the user if this error message included the leaf index that had a problem

client/log_client.go
```diff
@@ -357,6 +357,24 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
 	return err
 }
 
+// AddSequencedLeaves adds any number of pre-sequenced leaves to the log.
+func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int64][]byte) error {
```
pavelkalinnikov: Does it not allow setting non-data fields of leaves?

gdbelvin: This API does not, just like `AddSequencedLeaf`.  We can revisit the API more generally in another PR if you'd like

pavelkalinnikov: LGTM as long as it's consistent. We can change it when there is a case for it.

PR #1383, opened 2018-12-14
Add TimeSource-driven version of SleepContext

pavelkalinnikov:
### Checklist

### DIFF
--- a/util/clock/time.go
+++ b/util/clock/time.go
@@ -19,24 +19,33 @@ import (
 	"time"
 )
 
-// Sleep sleeps for at least the specified duration. Returns false iff done is
-// closed before the timer fires.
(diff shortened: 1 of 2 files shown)


util/clock/time.go
```diff
@@ -19,24 +19,33 @@ import (
...
 	return nil
 }
+
+// sleepImpl sleeps for at least the specified duration, as measured by the
+// TimeSource. Returns false iff done is closed before the deadline.
+func sleepImpl(done <-chan struct{}, d time.Duration, s TimeSource) bool {
+	timer := s.NewTimer(d)
+	defer timer.Stop()
+	select {
+	case <-timer.Chan():
```
gdbelvin: Would it be better to use time.After() here?

gdbelvin: no, that would leak a timer.

PR #1384, opened 2018-12-14
Contiguous AddSequencedLeaves

gdbelvin:
Ensure that the AddSequencedLeaves API submits leaves in contiguous, ascending order.

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


client/log_client_test.go
```diff
@@ -310,3 +311,27 @@ func TestUpdateRoot(t *testing.T) {
...
+		dataByIndex map[int64][]byte
+		wantErr     bool
+	}{
+		{desc: "empty", dataByIndex: nil},
+		{desc: "non-contiguous", dataByIndex: map[int64][]byte{
+			0: []byte("A"),
+			2: []byte("C"),
+		}, wantErr: true},
+	} {
+
```
pavelkalinnikov: nit: Please remove this empty line.

client/log_client.go
```diff
@@ -358,10 +359,24 @@ func (c *LogClient) AddSequencedLeaf(ctx context.Context, data []byte, index int
...
+		return nil
+	}
 	leaves := make([]*trillian.LogLeaf, 0, len(dataByIndex))
-	for index, data := range dataByIndex {
-		leaf, err := c.BuildLeaf(data)
+	indexes := make([]int64, 0, len(dataByIndex))
+	for index := range dataByIndex {
+		indexes = append(indexes, index)
+	}
+	sort.Slice(indexes, func(a, b int) bool { return indexes[a] < indexes[b] })
```
pavelkalinnikov: A bit more efficient algo would be to find `minIndex` and `maxIndex`, and check that `len(dataByIndex) == maxIndex-minIndex` instead of line 376-378?

pavelkalinnikov: Admittedly a slightly more ugly one because Go doesn't have min/max funcs for `int64`.

gdbelvin: That might be more efficient, but is it easier to follow?

pavelkalinnikov: Nope.

PR #1386, opened 2018-12-14
election2: Add rigorous Election testing

pavelkalinnikov:
This change adds a test-only `Election` implementation and tests for it. It also introduces a `Decorator` that allows injecting errors into `Election` implementations during testing, and controlling how long `Await` method is blocked. The decorator and the bulk of the tests code is re-used by both `testonly.Election` and `etcd.Election` implementations.

Depends on: #1387, #1388

### Checklist

- [ ] I have updated the [CHANGELOG](CHANGELOG.md).
  - Adjust the draft version number according to [semantic versioning](https://semver.org/) rules.
- [ ] I have updated [documentation](docs/) accordingly.

PR #1387, opened 2018-12-17
etcd: Fix first WithMastership call blocking if no Await

pavelkalinnikov:
This change fixes the following bug in `etcd` implementation of `election2`:
1. `Await` is not called.
2. `WithMastership` is called. It is supposed to return a closed context because the instance is not the master.
3. Instead, it blocks forever (assuming there are no other instances calling `Await`) because `Observe` hopes to get some updates which it never gets because nobody was ever the master.

Tests for this case and others will be added in separate PRs like #1386.

### Checklist

### DIFF
--- a/util/election2/etcd/election.go
+++ b/util/election2/etcd/election.go
@@ -50,10 +50,16 @@ func (e *Election) WithMastership(ctx context.Context) (context.Context, error) 
 	// context so that the monitoring goroutine below and the goroutine started
 	// by WithMastership will reliably terminate).
 	cctx, cancel := context.WithCancel(ctx)
-	ch := e.election.Observe(cctx)
 	etcdRev := e.election.Rev() // The revision at which e became the master.
+	if etcdRev == 0 {
+		// Not even tried to become the master. Return a canceled context.
+		cancel()
+		return cctx, nil
+	}
 
-	// Verify mastership before returning context.
+	// Was the master once, so watch for latest mastership updates.
+	ch := e.election.Observe(cctx)
+	// Verify that we are still the master, before returning context.
 	select {
 	case <-ctx.Done():
 		cancel()


Martin2112:
Not sure where we infer from the etcd docs that Observe will not do anything if nobody is master when this starts?

Going to add David to this as well as he's more familiar with etcd.

pavelkalinnikov:
This is an actual bug that I caught with tests in #1386, and this change does fix it. Sorry for not including the tests, I will add them separately as they are quite bulky. The problem is that `Observe` only reports PUT `etcd` values. It does not report DELETE, and does not report absense of values.

Martin2112:
Interesting. I read the etcd API docs and it wasn't obvious that this is what happens. Observe() doesn't list any preconditions and glancing at their code it looks like it bails on errors.

pavelkalinnikov:
[This](https://github.com/etcd-io/etcd/blob/abb57363ba6fce3fea1d427841904f434c863d49/clientv3/concurrency/election.go#L197) and [this](https://github.com/etcd-io/etcd/blob/abb57363ba6fce3fea1d427841904f434c863d49/clientv3/concurrency/election.go#L231) line clarifies not returning DELETE events.
Also, the `Observe` function comment says "observes ordered leader proposals" which is basically values PUT by the leader. No leader -> no PUT values -> no stuff pushed to the channel.

Martin2112:
OK where do these problems manifest? It looks like migrillian always calls Await(). The election interfaces and tracker utils make it harder to see where the logsigner calls it.

pavelkalinnikov:
Correct, the problem does not show itself in reality as `Await` is always called first (in Migrillian and Key Transparency). `logsigner` does not yet use `election2`, but I am in the process of converging the 2 election interfaces that we have.

PR #1388, opened 2018-12-17
election2/etcd: Proclaim resignation

pavelkalinnikov:
The issue that this change addresses is the possibility of leaking mastership contexts if they are not used properly (due to bugs or exceptional cases). `Observe` method of `etcd` election package does not report leadership resignation (i.e. DELETE events in election directory). As a result, the monitoring loop in `WithMastership` doesn't terminate until other instance captures mastership. If there is no other instance, the loop can hang forever, leaving mastership context open.

Caught by tests in #1386.

### Checklist

### DIFF
--- a/util/election2/etcd/election.go
+++ b/util/election2/etcd/election.go
@@ -99,7 +99,10 @@ func (e *Election) WithMastership(ctx context.Context) (context.Context, error) 
 // again using Await. Idempotent, might be useful to retry if fails.
 func (e *Election) Resign(ctx context.Context) error {
 	// Trigger Observe callers to see the update, and cancel mastership contexts.
-	if err := e.election.Proclaim(ctx, resignID); err != nil {
+	err := e.election.Proclaim(ctx, resignID)
+	if err == concurrency.ErrElectionNotLeader {
+		return nil // Resigning if not master is a no-op.
+	} else if err != nil {
 		return err
 	}
 	return e.election.Resign(ctx)
--- a/util/election2/etcd/election.go
+++ b/util/election2/etcd/election.go
@@ -27,6 +27,8 @@ import (
 	"github.com/google/trillian/util/election2"
 )
 
+const resignID = "<resign>"
+
 // Election is an implementation of election2.Election based on etcd.
 type Election struct {
 	resourceID string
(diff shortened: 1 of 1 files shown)


util/election2/etcd/election.go
```diff
@@ -59,8 +61,9 @@ func (e *Election) WithMastership(ctx context.Context) (context.Context, error)
 		cancel()
 		return nil, ctx.Err()
 	case rsp, ok := <-ch:
-		if !ok || rsp.Kvs[0].CreateRevision != etcdRev {
-			// Mastership has been overtaken in the meantime, or not capturead at all.
+		kv := rsp.Kvs[0]
+		if !ok || kv.CreateRevision != etcdRev || string(kv.Value) == resignID {
```
gdbelvin: I might be missing something, but the last two conditions here
`kv.CreateRevision != etcdRev` and `kv.Value == resignID` might be redundant.
If kv.Value changes (eg. to resign), would `etcdRev` get updated?

pavelkalinnikov: `CreateRevision` doesn't change as long as the master keeps mastership. It's the revision of when the `etcd` key was **first** created.

util/election2/etcd/election.go
```diff
@@ -89,6 +98,10 @@ func (e *Election) WithMastership(ctx context.Context) (context.Context, error)
 // Resign releases mastership for this instance. The instance can be elected
 // again using Await. Idempotent, might be useful to retry if fails.
 func (e *Election) Resign(ctx context.Context) error {
+	// Trigger Observe callers to see the update, and cancel mastership contexts.
```
gdbelvin: etcd.Resign returns `nil` when it is not master. The introduction of a proclaim step which will fail when not master changes that API contract. Does the comment above that it "might be useful to retry if it fails" still hold?

util/election2/etcd/election.go
gdbelvin: Got it, thanks.

util/election2/etcd/election.go
pavelkalinnikov: Good point. I think that we should either a) return some sort of `election2.NotMaster` error, or b) stick to the contract and return nil if `Proclaim` returns `concurrency.ErrElectionNotLeader`. I am leaning towards b), WDYT?

gdbelvin: I lean towards B as well

pavelkalinnikov: Done.

Issue #1191, opened 2018-06-05
Support watch-based master election implementations


pavelkalinnikov:
The current implementation of master election polls mastership status using `MasterElection.IsMaster` interface every now and then (`MasterCheckInterval`). The downside of such approach is weak guarantees on what's been happening between `IsMaster` calls, plus a bit of excess `etcd` traffic.

`etcd` has an alternative watch-based solution for [observing](https://github.com/coreos/etcd/blob/438e6750ee4dd3861c83514e405adf2f3e437e4c/clientv3/concurrency/election.go#L158) mastership changes (gRPC stream inside). Perhaps we could refactor the election API to support such mechanisms as well.

pavelkalinnikov:
I can see this as some `GetMasterContext` method which is a bit like `WaitForMastership` but also returning a `context.Context` which is active as long as the instance is the master. In `etcd` implementation it would spin up the watcher internally, and close the returned context as soon as the mastership / connectivity is lost. `MasterElection` implementations that don't have passive watching mechanisms could do the old busy waiting to emulate `GetMasterContext` (something like the code in #1177).

pavelkalinnikov:
@AlCutter @daviddrysdale WDYT?

AlCutter:
Sounds pretty neat on the face of it, I like the `Context` idea.
Any thoughts from @pphaneuf too?

pavelkalinnikov:
A prototype is here: https://github.com/google/certificate-transparency-go/pull/268

pavelkalinnikov:
Followed by https://github.com/google/certificate-transparency-go/pull/279

pavelkalinnikov:
At some point we may need an interface allowing to contact the current master (in a way similar to #1139), for example if we want to submit entries directly through the `logsigner` to improve latency. The right way to do it would probably be having a separate "observer" interface that doesn't participate in master election, only observes it.

pavelkalinnikov:
This has been fixed a while ago when `election2` package replaced the old one.

PR #1393, opened 2018-12-20
Use election2 instead of legacy election interface

pavelkalinnikov:
Part of #1191

### Checklist

### DIFF
--- a/server/trillian_log_signer/main.go
+++ b/server/trillian_log_signer/main.go
@@ -68,10 +68,9 @@ var (
 		"Increase factor for tokens replenished by sequencing-based quotas (1 means a 1:1 relationship between sequenced leaves and replenished tokens)."+
 			"Only effective for --quota_system=etcd.")
 
-	preElectionPause    = flag.Duration("pre_election_pause", 1*time.Second, "Maximum time to wait before starting elections")
-	masterCheckInterval = flag.Duration("master_check_interval", 5*time.Second, "Interval between checking mastership still held")
-	masterHoldInterval  = flag.Duration("master_hold_interval", 60*time.Second, "Minimum interval to hold mastership for")
-	masterHoldJitter    = flag.Duration("master_hold_jitter", 120*time.Second, "Maximal random addition to --master_hold_interval")
+	preElectionPause   = flag.Duration("pre_election_pause", 1*time.Second, "Maximum time to wait before starting elections")
+	masterHoldInterval = flag.Duration("master_hold_interval", 60*time.Second, "Minimum interval to hold mastership for")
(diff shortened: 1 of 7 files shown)


pavelkalinnikov:
PTAL. Commit by commit review might be the best way.

server/trillian_log_signer/main.go
```diff
@@ -66,10 +68,9 @@ var (
 		"Increase factor for tokens replenished by sequencing-based quotas (1 means a 1:1 relationship between sequenced leaves and replenished tokens)."+
 			"Only effective for --quota_system=etcd.")
 
-	preElectionPause    = flag.Duration("pre_election_pause", 1*time.Second, "Maximum time to wait before starting elections")
-	masterCheckInterval = flag.Duration("master_check_interval", 5*time.Second, "Interval between checking mastership still held")
```
pavelkalinnikov: Need to add deletion of this flag to changelog as well.

util/election/runner.go
```diff
@@ -105,80 +99,74 @@ func NewRunner(id string, cfg *RunnerConfig, tracker *MasterTracker, cancel cont
...
-				break  // no longer master
-			}
+		if err := er.beMaster(ctx, pending); err != nil {
+			glog.Errorf("%s: %v", er.id, err)
+			break
 		}
 	}
 }
 
+func (er *Runner) beMaster(ctx context.Context, pending chan<- Resignation) error {
```
gdbelvin: Thanks for splitting this out into a separate function

util/election/runner_test.go
```diff
@@ -59,91 +59,101 @@ func TestConfigResignDelay(t *testing.T) {
...
-			var wg sync.WaitGroup
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
 
-			startTime := time.Now()
-			fakeTimeSource := clock.NewFake(startTime)
+			e := to.NewElection()
+			d := to.NewDecorator(e)
```
gdbelvin: If you don't use e anywhere, you might be able to nest
`d := to.NewDecorator(to.NewElection())`

server/trillian_log_signer/main.go
pavelkalinnikov: Done.

util/election/runner_test.go
pavelkalinnikov: It's used one time in line 134. I could use `d` there, but I would like that line to to ignore any `Resign` errors in the decorator, to simulate mastership loss properly.

pavelkalinnikov:
The containers don't use the flag luckily.

Issue #1397, opened 2019-01-02
[Performance] Parallelize HStar2 


gdbelvin:
`SetLeaves` can take up to 2.5 minutes for 1000 leaves because the `HStar2` algorithm walks the relevant portions of the merkle tree sequentially, incurring many db round trip delays. 

1. Investigate parallelizing `SetLeaves` so that independent subtrees can proceed concurrently.   
2. Investigate fetching all the required nodes at once, since the needed paths can be computed ahead of time.

PR #1398, opened 2019-01-04
Fetch all relevant map nodes in parallel

gdbelvin:
Create a separate go routine per branch in the HStar2 algorithm to set and get map nodes in parallel. 
This creates a large number of parallel database queries, so we also need to set `MaxConns`. 

#1397 

### Checklist

### DIFF
--- a/storage/cache/map_subtree_cache.go
+++ b/storage/cache/map_subtree_cache.go
@@ -18,6 +18,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"math/big"
+	"sync"
 
 	"github.com/golang/glog"
 	"github.com/google/trillian/merkle"
@@ -39,6 +40,7 @@ func NewMapSubtreeCache(mapStrata []int, treeID int64, hasher hashers.MapHasher)
 func populateMapSubtreeNodes(treeID int64, hasher hashers.MapHasher) storage.PopulateSubtreeFunc {
 	return func(st *storagepb.SubtreeProto) error {
 		st.InternalNodes = make(map[string][]byte)
+		var internalNodesMu sync.Mutex
 		leaves := make([]merkle.HStar2LeafHash, 0, len(st.Leaves))
 		for k64, v := range st.Leaves {
 			sfx, err := storage.ParseSuffix(k64)
@@ -72,6 +74,8 @@ func populateMapSubtreeNodes(treeID int64, hasher hashers.MapHasher) storage.Pop
 					}
 					glog.Infof("PopulateMapSubtreeNodes.Set(%x, %d) suffix: %x: %x", index.Bytes(), depth, b, h)
 				}
+				internalNodesMu.Lock()
+				defer internalNodesMu.Unlock()
 				st.InternalNodes[sfxKey] = h
 				return nil
 			})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### Parallel Algorithms
+
+* The HStar2 algorithm for reading and writing to the map now processes each
+  subtree concurrently.
+
+### Default MaxSQLConnections Limit
+The default limit on the maximum number of concurrent mysql connections has been
+changed from unlimited to 1024. See ./scripts/mysqlconnlimit.sh
+
 ### Add Profiling Flags to Binaries
 
 The `trillian_log_server`, `trillian_log_signer` and `trillian_map_server`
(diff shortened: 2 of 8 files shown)


pavelkalinnikov:
When I read this algorithm last time, I thought that it'd be nice to have it implemented in a more direct way: 1. get the list of nodes necessary to read; 2. read the nodes; 3. compute the new nodes; 4. write the new nodes. One of the reasons I thought it would be nice is that it would naturally allow doing steps 2 and 4 more efficiently using batching.

This PR takes a less radical approach, but at a cost of more goroutines / connections created. I suspect that it is also slightly less scalable, particularly because of the number of connections. Do you think it's worth considering the other approach?

gdbelvin:
Your approach would be infinitely better.  This was more of a stop-gap hack.

pavelkalinnikov:
I guess we can submit this, but leave a TODO for a bigger refactoring? Please leave #1397 open, so that option (2) from there is implemented in longer term.

@AlCutter could you take a look as well?

storage/cache/map_subtree_cache.go
```diff
@@ -39,6 +40,7 @@ func NewMapSubtreeCache(mapStrata []int, treeID int64, hasher hashers.MapHasher)
 func populateMapSubtreeNodes(treeID int64, hasher hashers.MapHasher) storage.PopulateSubtreeFunc {
 	return func(st *storagepb.SubtreeProto) error {
 		st.InternalNodes = make(map[string][]byte)
+		var internalNodesMu sync.Mutex
```
AlCutter: I _suspect_ there's probably not that much benefit in making the population of internal subtree nodes concurrent (given there'll be plenty more "real" concurrency at the next level up where disk/DB accesses are happening).

I wonder if it's worth making `NewHStar2` take some sort of concurrency parameter?

gdbelvin: That may well be true, but HStar2 isn't aware of what nodes are in the cache vs. going to result in a db access, and I'm not sure it would be a good idea to give it that education for layering reasons. 

I did think about doing something like worker pools for HStar2 to handle the massive go routine fanout, but worker pools can easily get into a deadlock situation.  In this configuration where threads will block on other threads, we need the more intelligent block/unblocking logic that the native goroutines provide.

AlCutter: > HStar2 isn't aware of what nodes are in the cache vs. going to result in a db access

Exactly - but the caller may have more context - like this site where `hstar2` is being used only to populate the subtree cache structure...

AlCutter:
What was the performance delta you saw when you tested with this change BTW?

On my machine (against a local MySQL DB) I see `./integration/map_integration_test.sh` as taking 25s to complete at `HEAD` of `master`, whereas this PR it takes 31s:

```bash
# git checkout master
$ ./integration/map_integration_test.sh
ok  	github.com/google/trillian/integration/maptest	24.815s


# git fancy-pants magic to pull upstream pull/1398/head
$ ./integration/map_integration_test.sh
ok  	github.com/google/trillian/integration/maptest	31.155s
```

AlCutter:
I had an alternative idea to improve performance which is a bit less invasive - have a look at #1405 and see what you think.

storage/cache/map_subtree_cache.go
pavelkalinnikov: Yes, I think there's hardly a point in populating the cache on **writes** concurrently. But there are also **reads** which need to be sped up: either by batching (#1405 basically achieves the same effect one layer above because it's aware of the cache), or parallelization (this PR). I was starting a PR refactoring the `HStar2` itself, so that it would return a list of nodes to fetch, without relying on the cache.

AlCutter: That refactor is possibly pretty worthwhile - the way the hstar2, cache, and storage interconnect is a little complex as it - pulling it apart to: 
   1. figure out pages we need
   2. load the pages
   3. run hstar2 with *only* access to the loaded pages
   4. flush writes

would be much easier to grok.

There might be some performance improvement from parallelisation, although in the current state there are already 256 goroutines servicing the shards, but with Circus Tent it might be helpful to be able to run sub-shards within each shard server concurrently too.

gdbelvin:
Hmm... Looks like this actually causes a performance regression... 

```
# Master
$ ./integration/map_integration_test.sh
ok  	github.com/google/trillian/integration/maptest	20.681s

# This PR
$ ./integration/map_integration_test.sh
ok  	github.com/google/trillian/integration/maptest	26.836s
```

storage/cache/map_subtree_cache.go
pavelkalinnikov: I agree. The step 1 will probably decompose into two (processed at different levels):
1.1. Figure out **nodes** that HStar2 needs
1.2. Transform it to the list of **pages** to load

PR #1399, opened 2019-01-05
GetMapLeaves in parallel

### LINKED ISSUE #1396: [Performance] Parallelize GetMapLeavesByRevision
`GetMapLeavesByRevision` has batch API semantics, but under-the-hood, the implementation is fetching each leaf sequentially.  For ~1000 leaves, this can take up to 10s. 

1. Explore fetching leaves in parallel. 
2. Explore caching common nodes between the leaves to avoid multiple db lookups of the same value.

gdbelvin:
Initial hacky attempt at #1396 

Use batch get methods for getting map leaves where we have them.
Where we don't have a batch method for fetching inclusion proofs, execute the fetches in parallel.

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -17,6 +17,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -100,6 +101,12 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 	if err != nil {
 		return nil, fmt.Errorf("could not get map %v: %v", mapID, err)
 	}
+	for _, index := range indices {
+		if err := checkIndexSize(index, hasher); err != nil {
+			return nil, err
+		}
+	}
+
 	ctx = trees.NewContext(ctx, tree)
 
 	tx, err := t.registry.MapStorage.SnapshotForTree(ctx, tree)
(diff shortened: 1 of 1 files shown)


server/map_rpc_server.go
```diff
@@ -128,45 +135,50 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
...
-			Inclusion: proof,
-		})
+	// Fetch inclusion proofs in parallel.
+	smtReader := merkle.NewSparseMerkleTreeReader(int64(mapRoot.Revision), hasher, tx)
+	inclusions := make([]*trillian.MapLeafInclusion, len(indices))
+	errs := make(chan error, len(indices))
+	var wg sync.WaitGroup
+	wg.Add(len(indices))
+	for i, index := range indices {
+		go func(i int, index []byte) {
```
pavelkalinnikov: I am concerned about 1 goroutine per index approach w.r.t. scale. Would introducing a batch request instead be more reliable?

server/map_rpc_server.go
```diff
@@ -128,45 +135,50 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
...
+	}
 
-		inclusions = append(inclusions, &trillian.MapLeafInclusion{
-			Leaf:      leaf,
-			Inclusion: proof,
-		})
+	// Fetch inclusion proofs in parallel.
+	smtReader := merkle.NewSparseMerkleTreeReader(int64(mapRoot.Revision), hasher, tx)
+	inclusions := make([]*trillian.MapLeafInclusion, len(indices))
+	errs := make(chan error, len(indices))
```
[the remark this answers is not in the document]
gdbelvin: Do you think the extra code is worth the memory savings?

server/map_rpc_server.go
gdbelvin: A batch index request would be infinitely better, but I'm working with what I've got.

pavelkalinnikov: Could you please leave a TODO to improve this?

server/map_rpc_server.go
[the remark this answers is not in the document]
pavelkalinnikov: Dunno, maybe not.

server/map_rpc_server.go
pavelkalinnikov: Maybe even leave the issue open, to address it in a bit.

PR #1409, opened 2019-01-18
Add experimental mysql option to fetch subtrees one at a time

Martin2112:
Instead of using the potentially expensive join get subtree protos one at time by scanning for the revision. This should be cheaper as we almost always want the most recent version and there is an index by revision. However, we need to test this assumption.

The option is off by default and will be documented if it proves to be useful. At the moment it should not be enabled in production.

### DIFF
--- a/cmd/merkle_path/main.go
+++ b/cmd/merkle_path/main.go
@@ -1,3 +1,17 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package main
 
 import (
--- a/cmd/merkle_path/main.go
+++ b/cmd/merkle_path/main.go
@@ -42,9 +42,7 @@ func main() {
 		glog.Exitf("Failed to build the Merkle path: %v", err)
 	}
 
-	fmt.Printf("Resulting path length: %d\n"+
-		"ls -l"+
-		"", len(path))
+	fmt.Printf("Resulting path length: %d\n", len(path))
 	for _, fetch := range path {
 		nodeID := fetch.NodeID
 		fmt.Printf("%v %v\n", nodeID.CoordString(), fetch.Rehash)
--- a/cmd/merkle_path/main.go
+++ b/cmd/merkle_path/main.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"encoding/hex"
+	"flag"
+	"fmt"
+
+	"github.com/golang/glog"
+	_ "github.com/golang/protobuf/proto"
+	_ "github.com/google/trillian/crypto/keyspb"
+	_ "github.com/google/trillian/crypto/sigpb"
+	"github.com/google/trillian/merkle"
+	_ "github.com/google/trillian/storage/storagepb"
+)
+
+var (
+	size1    = flag.Int64("size1", 0, "First tree size (or only for inclusion proof)")
+	size2    = flag.Int64("size2", 0, "Second tree size (for consistency proof)")
+	index    = flag.Int64("index", 0, "Desired leaf index (for inclusion proof)")
+	treeSize = flag.Int64("tree_size", 0, "Size of the Merkle tree")
+)
+
+func main() {
+	flag.Parse()
+	defer glog.Flush()
+
+	var err error
+	var path []merkle.NodeFetch
+	subtrees := make(map[string]bool)
+
+	if *size2 == 0 {
+		// Do an inclusion proof
+		fmt.Printf("Inclusion proof for index %d at tree size %d in a tree of size %d\n\n", *index, *size1, *treeSize)
+		path, err = merkle.CalcInclusionProofNodeAddresses(*size1, *index, *treeSize, 64)
+	} else {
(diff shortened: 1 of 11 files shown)


testonly/integration/logenv.go
```diff
@@ -87,7 +87,10 @@ func NewLogEnvWithGRPCOptions(ctx context.Context, numSequencers int, serverOpts
 
 	registry := extension.Registry{
 		AdminStorage: mysql.NewAdminStorage(db),
-		LogStorage:   mysql.NewLogStorage(db, nil),
+		// *****************************************************************
+		// TODO(Martin2112): Do something about hardcoded LogStorageOptions.
+		// *****************************************************************
+		LogStorage:   mysql.NewLogStorage(db, nil, mysql.LogStorageOptions{mysql.TreeStorageOptions{FetchSingleSubtrees: true}}),
```
golangcibot: `mysql.LogStorageOptions` composite literal uses unkeyed fields (from `govet`)

testonly/integration/registry.go
```diff
@@ -30,9 +30,12 @@ func NewRegistryForTests(ctx context.Context) (extension.Registry, error) {
 		return extension.Registry{}, err
 	}
 
+	// *****************************************************************
+	// TODO(Martin2112): Do something about hardcoded LogStorageOptions.
+	// *****************************************************************
 	return extension.Registry{
 		AdminStorage: mysql.NewAdminStorage(db),
-		LogStorage:   mysql.NewLogStorage(db, nil),
+		LogStorage:   mysql.NewLogStorage(db, nil, mysql.LogStorageOptions{mysql.TreeStorageOptions{FetchSingleSubtrees: true}}),
```
golangcibot: `mysql.LogStorageOptions` composite literal uses unkeyed fields (from `govet`)

server/mysql_storage_provider.go
```diff
@@ -70,7 +71,11 @@ func newMySQLStorageProvider(mf monitoring.MetricFactory) (StorageProvider, erro
 }
 
 func (s *mysqlProvider) LogStorage() storage.LogStorage {
-	return mysql.NewLogStorage(s.db, s.mf)
+	return mysql.NewLogStorage(s.db, s.mf, mysql.LogStorageOptions{
```
golangcibot: `mysql.LogStorageOptions` composite literal uses unkeyed fields (from `govet`)

cmd/merkle_path/main.go
```diff
@@ -0,0 +1,65 @@
...
+		nodeID := fetch.NodeID
+		fmt.Printf("%v %v\n", nodeID.CoordString(), fetch.Rehash)
+		prefixBytes := nodeID.PrefixLenBits >> 3
+		// Might have to split the path if it isn't the root of a subtree.
+		suffixBits := nodeID.PrefixLenBits - (prefixBytes << 3)
+		subtree := nodeID.Path
+		if suffixBits > 0 {
+			subtree, _ = nodeID.Split(prefixBytes, suffixBits)
+		}
+		subtrees[hex.EncodeToString(subtree)] = true
```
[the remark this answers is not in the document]
Martin2112: You can get multiple reads at the same depth. I was going to sort it by length for display but it's a nice to have.

storage/mysql/tree_storage.go
```diff
@@ -232,42 +257,136 @@ func (t *treeTX) getSubtrees(ctx context.Context, treeRevision int64, nodeIDs []
...
+	defer stmt.Close()
+
+	// Go through each requested nodeID in turn. Then scan for a revision that
+	// meets our requirements.
+	ret := make([]*storagepb.SubtreeProto, 0, len(nodeIDs))
+	for _, nodeID := range nodeIDs {
+		nodeIDBytes := nodeID.Path[:nodeID.PrefixLenBits/8]
+		glog.V(4).Infof("  nodeID: %x", nodeIDBytes)
+
+		rows, err := stmt.QueryContext(ctx, t.treeID, nodeIDBytes, treeRevision)
```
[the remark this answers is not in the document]
Martin2112: Yes, it probably is.

testonly/integration/logenv.go
Martin2112: Isn't the point of an embedded field that it doesn't have a name?

cmd/merkle_path/main.go
```diff
@@ -0,0 +1,65 @@
...
+		glog.Exitf("Failed to build the Merkle path: %v", err)
+	}
+
+	fmt.Printf("Resulting path length: %d\n"+
+		"ls -l"+
+		"", len(path))
+	for _, fetch := range path {
+		nodeID := fetch.NodeID
+		fmt.Printf("%v %v\n", nodeID.CoordString(), fetch.Rehash)
+		prefixBytes := nodeID.PrefixLenBits >> 3
```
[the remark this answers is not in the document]
Martin2112: Not sure what you mean? We don't have any subtree protos, just the IDs of what would be fetched.

testonly/integration/logenv.go
RJPercival: Oh it does have a name, it's just automatically the name of the type. In this case, the field is called "TreeStorageOptions".

Martin2112:
Yeah I'm probably not going to do the cleanups until we know that this option is worth having. I'll see how much time I get.

RJPercival:
I just tried this out and unfortunately found that getting consistency proofs took ~5 times longer (almost 2 seconds, rather than ~400ms). While each query is cheaper than the original (combined) query, their combined cost is greater.

Martin2112:
OK back to the drawing board. Next thing to try is scanning for all of the subtrees in one go.

Martin2112:
Before I look at this again can we check that the subtree cache is working correctly for this case. The sql code assumes the deduping has been done already. If there's a bug in this then it could be doing extra work.

Adding a couple of metrics should do it. Subtrees fetched in the SQL code v. nodes requested at the level up should probably be enough to confirm it.

Martin2112:
Going to close this as it's proven to be of no use in testing.

PR #1418, opened 2019-01-31
Remove warning logging from log/sequencer.go

RJPercival:
Removed in favour of including the same information in error messages. The caller can choose to log the error if they wish. This avoids misleading warning messages in logs when the storage implementation retries transaction errors. While storage can swallow the errors when a retry succeeds, it can't cancel the logging.

### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -136,14 +136,12 @@ func (s Sequencer) buildMerkleTreeFromStorageAtRoot(ctx context.Context, root *t
 	mt, err := compact.NewTreeWithState(s.hasher, int64(root.TreeSize), func(depth int, index int64) ([]byte, error) {
 		nodeID, err := storage.NewNodeIDForTreeCoords(int64(depth), index, maxTreeDepth)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


log/sequencer.go
```diff
@@ -136,14 +136,12 @@ func (s Sequencer) buildMerkleTreeFromStorageAtRoot(ctx context.Context, root *t
 	mt, err := compact.NewTreeWithState(s.hasher, int64(root.TreeSize), func(depth int, index int64) ([]byte, error) {
 		nodeID, err := storage.NewNodeIDForTreeCoords(int64(depth), index, maxTreeDepth)
 		if err != nil {
-			glog.Warningf("%x: Failed to create nodeID: %v", s.signer.KeyHint, err)
-			return nil, err
+			return nil, fmt.Errorf("%x: Failed to create nodeID: %v", s.signer.KeyHint, err)
```
Martin2112: This is fairly serious. We might still want to log it but not necessary if it caller definitely logs.

RJPercival: It ends up getting logged at ERROR level here: https://github.com/google/trillian/blob/2609dcb936ec037623152c28146867fffdc07276/server/log_operation_manager.go#L414

PR #1425, opened 2019-02-08
Support testing against a locally running mysql docker image

gdbelvin:
Allow testing against a local docker image running mysql.  

Some testers prefer to run the docker db image for ease of setup, rather than installing mysql locally. 
This change should be compatible with both a local installation as well as a local docker image.

### Checklist

### DIFF
--- a/storage/testdb/testdb.go
+++ b/storage/testdb/testdb.go
@@ -33,7 +33,7 @@ import (
 
 var (
 	trillianSQL = testonly.RelativeToPackage("../mysql/storage.sql")
-	dataSource  = "root@/"
+	dataSource  = "root@tcp(127.0.0.1)/"
 )
 
 // MySQLAvailable indicates whether a default MySQL database is available.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### Testing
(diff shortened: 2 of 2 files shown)


RJPercival:
Why does this remove objhasher?

gdbelvin:
Rebased on master to remove spurious commits.

storage/testdb/testdb.go
```diff
@@ -33,7 +33,7 @@ import (
 
 var (
 	trillianSQL = testonly.RelativeToPackage("../mysql/storage.sql")
-	dataSource  = "root@/"
+	dataSource  = "root@tcp(127.0.0.1)/"
```
RJPercival: According to https://github.com/go-sql-driver/mysql#dsn-data-source-name, TCP to "localhost" is the default. How is 127.0.0.1 different to the default?

gdbelvin:
By default, it looks for a local unix socket which is not compatible with a
docker image

RJPercival:
> By default, it looks for a local unix socket which is not compatible with a docker image

 What makes you think that? Looking at both [the examples](https://github.com/go-sql-driver/mysql#examples) and [the code](https://github.com/go-sql-driver/mysql/blob/1b9eda25029c206f36637dde8e2acad26df99dab/dsn.go#L82), it appears that TCP is the default.

PR #1429, opened 2019-02-12
Expose Trillian services via load balancer

RJPercival:
Expose trillian-log-service and trillian-log-signer via load balancer.

On Google Cloud, this will use an internal load balancer.

### Checklist


RJPercival:
I've just added a mention of this to CHANGELOG as well.

PR #1430, opened 2019-02-12
Run docker-compose integration test as separate Travis job

RJPercival:
This avoids slowing down the jobs that are required to complete in order to merge a pull request, without entirely removing this test. If a pull request author waits the extra 5 minutes or so required by this test, then they can have confidence that their change doesn't break our Docker images or docker-compose configs.

### Checklist


RJPercival:
While the Docker integration test is 4 minutes slower than the fastest of the Travis jobs, it's about 9 minutes faster than the slowest job, so I might as well not make it optional.

PR #1432, opened 2019-02-13
Check for nil config in client

gdbelvin:
Check for a nil config when building log and map verifiers.


client/log_verifier.go
```diff
@@ -53,17 +54,20 @@ func NewLogVerifier(hasher hashers.LogHasher, pubKey crypto.PublicKey, sigHash c
...
 	}
 
-	logHasher, err := hashers.NewLogHasher(config.GetHashStrategy())
+	logHasher, err := hashers.NewLogHasher(config.HashStrategy)
 	if err != nil {
 		return nil, fmt.Errorf("client: NewLogVerifierFromTree(): NewLogHasher(): %v", err)
 	}
 
-	logPubKey, err := der.UnmarshalPublicKey(config.GetPublicKey().GetDer())
+	logPubKey, err := der.UnmarshalPublicKey(config.PublicKey.GetDer())
```
Martin2112: Looks like that will crash if public key is nil.

gdbelvin: `GetDer` will check if PublicKey is nil and won't crash.

Martin2112: Is that case tested?

gdbelvin: Added a test

gdbelvin:
@Martin2112 PTAL - added a test to verify that a nil value in `PublicKey` will not cause a nil pointer exception.

PR #1439, opened 2019-02-26
Test MySQL error persistence with a garbage DSN

AlCutter:
An empty DSN is no longer an error when opening a MySQL DB, so instead use a garbage DSN in the error persistence test.

From https://github.com/go-sql-driver/mysql:
```
DSN (Data Source Name)
The Data Source Name has a common format, like e.g. PEAR DB uses it, but without type-prefix (optional parts marked by squared brackets):

[username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
A DSN in its fullest form:

username:password@protocol(address)/dbname?param=value
Except for the databasename, all values are optional. So the minimal DSN is:

/dbname
If you do not want to preselect a database, leave dbname empty:

/
This has the same effect as an empty DSN string:


Alternatively, Config.FormatDSN can be used to create a DSN string by filling a struct.
```

### Checklist

### DIFF
--- a/server/mysql_storage_provider_test.go
+++ b/server/mysql_storage_provider_test.go
@@ -23,11 +23,11 @@ import (
 
 func TestMySQLStorageProviderErrorPersistence(t *testing.T) {
 	defer flagsaver.Save().Restore()
-	if err := flag.Set("mysql_uri", ""); err != nil {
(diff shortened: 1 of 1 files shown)


Martin2112:
Travis not happy, looks like an import.

PR #1444, opened 2019-03-01
mapclient: GetLeaves: Verify Indexes Before Sending

gdbelvin:
Verify that no duplicate map indexes are being requested by the client before sending requests to the server.  

If the client requests the same index more than once, the server is happy to oblige by returning the same leaf more than once in return. However, this will immediately fail the client verification that ensures there is only one returned value per index. 

Question: should the server be doing a similar check?
### Checklist

### DIFF
--- a/client/map_client.go
+++ b/client/map_client.go
@@ -61,7 +61,7 @@ func (c *MapClient) GetAndVerifyMapLeaves(ctx context.Context, indexes [][]byte)
 	// Verify that there are no duplicates
 	set := make(map[string]bool)
 	for _, i := range indexes {
-		if ok := set[string(i)]; ok != false {
+		if set[string(i)] {
 			return nil, status.Errorf(codes.InvalidArgument,
 				"map.GetLeaves(): index %x requested more than once", i)
 		}
--- a/client/map_client.go
+++ b/client/map_client.go
@@ -58,6 +58,16 @@ func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRo
 
 // GetAndVerifyMapLeaves verifies and returns the requested map leaves.
 func (c *MapClient) GetAndVerifyMapLeaves(ctx context.Context, indexes [][]byte) ([]*trillian.MapLeaf, error) {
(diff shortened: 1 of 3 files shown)


Martin2112:
Can you give this PR a more descriptive title?

client/map_client.go
```diff
@@ -58,6 +58,16 @@ func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRo
 
 // GetAndVerifyMapLeaves verifies and returns the requested map leaves.
 func (c *MapClient) GetAndVerifyMapLeaves(ctx context.Context, indexes [][]byte) ([]*trillian.MapLeaf, error) {
+	// Verify that there are no duplicates
+	set := make(map[string]bool)
+	for _, i := range indexes {
+		if ok := set[string(i)]; ok != false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `ok` (from `gosimple`)

client/map_client_test.go
```diff
@@ -86,17 +86,32 @@ func TestGetLeavesAtRevision(t *testing.T) {
...
-	if got, want := leaves[0].Index, index; !bytes.Equal(got, want) {
-		t.Errorf("LeafIndex: %v, want %v", got, want)
+
+	for _, tc := range []struct {
+		desc    string
+		indexes [][]byte
+		wantErr bool
+	}{
+		{desc: "1", indexes: [][]byte{index}},
+		{desc: "2", indexes: [][]byte{index, index}, wantErr: true},
```
Martin2112: Should test for the invalid argument code to avoid accidental passing.

gdbelvin:
Done PTAL

client/map_client.go
```diff
@@ -58,6 +58,16 @@ func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRo
 
 // GetAndVerifyMapLeaves verifies and returns the requested map leaves.
```
mhutchinson: Worth documenting that each index should be fetched at most in the method doc.

PR #1446, opened 2019-03-04
More efficient Equivalent method for storage.NodeID

Martin2112:
This code was converting them to binary strings and somewhat obfuscating the underlying comparison by the fact that the Bit method counts from the right.

This version does the comparison without memory allocation on the raw path bytes.

This method has been used in a few places where performance matters a bit more, including comparing results in the map hashing code.

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -337,7 +337,27 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, Suffix) {
 
 // Equivalent return true iff the other represents the same path prefix as this NodeID.
 func (n *NodeID) Equivalent(other NodeID) bool {
-	return n.String() == other.String()
+	// If they're different lengths they cannot represent the same path prefix.
+	if n.PrefixLenBits != other.PrefixLenBits {
+		return false
+	}
+
+	// The first depthBytes must be identical.
+	depthBytes := n.PrefixLenBits / 8
+	for i := 0; i < depthBytes; i++ {
+		if n.Path[i] != other.Path[i] {
+			return false
+		}
+	}
+
+	// There may not be a leftover partial byte to compare.
+	if n.PrefixLenBits%8 == 0 {
+		return true
+	}
+
+	// Check the remaining bits after masking off unwanted bits in the last byte.
+	mask := leftmask[n.PrefixLenBits%8]
+	return (n.Path[depthBytes] & mask) == (other.Path[depthBytes] & mask)
 }
 
 // PopulateSubtreeFunc is a function which knows how to re-populate a subtree
--- a/storage/types.go
+++ b/storage/types.go
@@ -344,10 +344,8 @@ func (n *NodeID) Equivalent(other NodeID) bool {
 
 	// The first depthBytes must be identical.
 	depthBytes := n.PrefixLenBits / 8
-	for i := 0; i < depthBytes; i++ {
-		if n.Path[i] != other.Path[i] {
-			return false
-		}
+	if !bytes.Equal(n.Path[:depthBytes], other.Path[:depthBytes]) {
+		return false
 	}
 
 	// There may not be a leftover partial byte to compare.


Martin2112:
Please look at this closely. I'm fairly sure it's right and it passes tests but log rehashing and part of map hashing will break if I'm wrong.

storage/types.go
```diff
@@ -337,7 +337,27 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, Suffix) {
...
+		}
+	}
+
+	// There may not be a leftover partial byte to compare.
+	if n.PrefixLenBits%8 == 0 {
+		return true
+	}
+
+	// Check the remaining bits after masking off unwanted bits in the last byte.
+	mask := leftmask[n.PrefixLenBits%8]
```
pavelkalinnikov: The algorithm LGTM. The ordering of bits in `leftmask` array worries me though.

pavelkalinnikov: Ah, it looks fine after second look.

Martin2112: I think there's confusion because some of the APIs work with the bits from right to left but the storage is a bitstring from path[0] with MSB first. I've done another PR to clarify the doc a bit.

However it's important we get this right so if something still looks wrong let me know.

Martin2112: On the leftmask specifically. If there's one bit left over it will be in the MSB and leftmask[1] == 0x80.

pavelkalinnikov: Was this doc incorrect?
```golang
// in Path[0] are included, along with the lowest bit of Path[1]
PrefixLenBits int

```

Martin2112: Yes. I believe so. See the next PR.

storage/types.go
```diff
@@ -337,7 +337,27 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, Suffix) {
 
 // Equivalent return true iff the other represents the same path prefix as this NodeID.
 func (n *NodeID) Equivalent(other NodeID) bool {
-	return n.String() == other.String()
+	// If they're different lengths they cannot represent the same path prefix.
+	if n.PrefixLenBits != other.PrefixLenBits {
+		return false
+	}
+
+	// The first depthBytes must be identical.
```
AlCutter: You could potentially use `bytes.Equal` with slices into `Path` here

AlCutter:
Since you're down here in the weeds, I wonder if it might be worth a PR to beef up the corresponding test to include checks for different length prefixes etc.? Just thinking that if you got that in first, and tests all passed then you could be more certain about this (and any other follow up) change(s)?

Martin2112:
Yeah I now think I might have to do that.

Martin2112:
Test update is in #1453.

storage/types.go

Martin2112:
This should pass the additional tests that were just merged. I'm fairly confident having reviewed and documented the nodeID stuff. The use of `leftMask` arises because when the path is not a multiple of 8 the right hand bits are the non significant ones.

Martin2112:
Benchmarks show a "slight" difference here:

```
BenchmarkEquivalentEqual-12       	200000000	         7.21 ns/op
BenchmarkEquivalentNotEqual-12    	1000000000	         2.69 ns/op
BenchmarkStringEqual-12           	 1000000	      1806 ns/op
BenchmarkStringNotEqual-12        	 1000000	      1884 ns/op
```

storage/types_test.go
```diff
@@ -834,3 +834,51 @@ func mustDecode(h string) []byte {
...
+	}
+}
+
+func BenchmarkStringEqual(b *testing.B) {
+	n1, err := NewNodeIDForTreeCoords(1, 1934929, 64)
+	if err != nil {
+		b.Fatal(err)
+	}
+	for i := 0; i < b.N; i++ {
+		_ = n1.String() == n1.String()
```
golangcibot: SA4000: identical expressions on the left and right side of the '==' operator (from `staticcheck`)

Martin2112: Removed the benchmarks again.

PR #1447, opened 2019-03-04
Update and clarify some NodeID docs.

Martin2112:
Clarify that the storage of NodeIDs uses an MSB first representation and
that some APIs use a bit indexing from the rightmost significant bit (of the whole path bytes, not all of these may be significant).

I'm hoping these additional explanations will help people understand the code.

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -42,7 +42,9 @@ type NodeID struct {
 	// this NodeID.
 	//
 	// e.g. if Path contains two bytes, and PrefixLenBits is 9, then the 8 bits
-	// in Path[0] are included, along with the lowest bit of Path[1]
+	// in Path[0] are included, along with the MSB of Path[1]. However, note
+	// that some of the APIs count bits with 0 being the rightmost as this
+	// is a natural ordering to use when working from the leaves upwards.
 	PrefixLenBits int
 }
 
@@ -197,6 +199,7 @@ func NewNodeIDForTreeCoords(depth int64, index int64, maxPathBits int) (NodeID, 
 }
 
 // SetBit sets the ith bit to true if b is non-zero, and false otherwise.
+// Note that the bit index 0 is the right most valid bit in the path.
 func (n *NodeID) SetBit(i int, b uint) {
 	bIndex := (n.PathLenBits() - i - 1) / 8
 	if b == 0 {
--- a/storage/types.go
+++ b/storage/types.go
@@ -33,8 +33,17 @@ type Node struct {
 }
 
 // NodeID uniquely identifies a Node within a versioned MerkleTree.
+// NodeIDs as presented to the storage layer will always have a multiple
+// of 8 bits as they identify the root of a 256 element subtree.
+//
+// Reading paths right to left is the natural order when traversing from
+// leaves towards the root. However, for internal nodes the rightmost bits
+// of the IDs are not aligned on a byte boundary so care must be taken.
+//
+// Note that some of the APIs count bits with 0 being the rightmost
+// in the entire path array when some of these might not be significant.
 type NodeID struct {
-	// path is effectively a BigEndian bit set, with path[0] being the MSB
+	// Path is effectively a BigEndian bit set, with Path[0] being the MSB
 	// (identifying the root child), and successive bits identifying the lower
 	// level children down to the leaf.
 	Path []byte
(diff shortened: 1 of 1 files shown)


Martin2112:
Adding comments causes codecov.io to complain. Getting v. tempted to turn that off.

storage/types.go
```diff
@@ -197,6 +199,7 @@ func NewNodeIDForTreeCoords(depth int64, index int64, maxPathBits int) (NodeID,
 }
 
 // SetBit sets the ith bit to true if b is non-zero, and false otherwise.
+// Note that the bit index 0 is the right most valid bit in the path.
```
pavelkalinnikov: Consider the case when `NodeID` is 9 bits. According to the docstring of `PrefixLenBits`, `Path[1]` will have the MSB used as the 9th bit. When we count bits from the other side, this MSB should be the 0-th bit. But `SetBit(0, 1)` will change the **L**SB of `Path[1]`, which is not in the `NodeID`.

pavelkalinnikov: Are we just lucky that our keys are multiples of 8?

Martin2112: OK going to need to look at it again.

The multiple of 8 assumption is baked in pretty much throughout this code.

pavelkalinnikov: Yeah, I think we should either throw away the code that assumes non-8 keys, or fix and test it.

Martin2112: New plan is to write more tests against the current implementation and then use them against this. It's not worth taking chances.

Martin2112: I think part of the confusion arises because Bit() and SetBit() are working with fixed 64-bit (for log trees) arrays whereas the node ID structure is a variable bit length and not necessarily byte aligned (for internal node references within subtrees). I think adding actual examples might make this clearer.

And I also need to avoid adding to the confusion by using terms like "to the right" without qualification.

pavelkalinnikov: Also apparently node IDs mean [different things](https://github.com/google/trillian/blob/master/docs/storage/storage.md#nodeids) for maps and logs, so we should be twice as careful when reasoning about the bit index.

Martin2112: Maps are "easier" as they're a 256 bit path in the notional sparse Merkle tree so don't do the same sort of bit twiddling we do for logs. But "harder" in that they have non uniform strata.

Martin2112:
OK I beefed up the docs. Can you let me know if this makes things clearer.

Martin2112:
Rebased to see if my change to the code coverage config prevents the flake.

Martin2112:
Yay, codecov didn't flake!

Martin2112:
Ping?

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
 }
 
 // NodeID uniquely identifies a Node within a versioned MerkleTree.
```
pavelkalinnikov: It seems like "versioned" is irrelevant here.

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
...
+// NodeIDs as presented to the storage layer will always have a multiple
```
pavelkalinnikov: This "multiple of 8" comment is slightly misleading as it's a storage's restriction rather than NodeID's. Maybe put this comment after the non-multiple-of-8 example?

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
...
+//
+// Reading paths right to left is the natural order when traversing from
+// leaves towards the root. However, for internal nodes the rightmost bits
+// of the IDs are not aligned on a byte boundary so care must be taken.
+//
+// Note that some of the APIs count bits with 0 being the rightmost
+// in the entire path array when some of these might not be significant.
 type NodeID struct {
-	// path is effectively a BigEndian bit set, with path[0] being the MSB
+	// Path is effectively a BigEndian bit set, with Path[0] being the MSB
```
pavelkalinnikov: "B" in MSB is used both for bits and bytes throughout this type. Expand one of them to avoid ambiguity?

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
...
 	// this NodeID.
 	//
 	// e.g. if Path contains two bytes, and PrefixLenBits is 9, then the 8 bits
-	// in Path[0] are included, along with the lowest bit of Path[1]
+	// in Path[0] are included, along with the MSB of Path[1]. The remaining
+	// 7 bits are not significant.
 	PrefixLenBits int
 }
 
+// Some Additional NodeID examples and Documentation.
```
pavelkalinnikov: Would it make sense to be included in the NodeID docstring?

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
...
+// array of [0x5c] and PrefixLength is 6, which means the lowest 2 bits are not
+// significant. In binary 0x5c is: 0b01011100 and 23 decimal is: 0b000010111
+// so the index has been shifted left by 2 places in the NodeID. Reading
+// from left to right after 6 bits the node has been reached. The remaining
+// two bits therefore don't matter.
+//
+// Larger tree sizes as used by Logs and Maps work in exactly the same
+// way but with the Path taking up multiple bytes. For example if we move
+// to a tree size of up to 65535 there will be two Path bytes. For a leaf
+// in this tree at depth = 0, index = 19 the resulting NodeID has a
```
pavelkalinnikov: To me the "depth" word seems counterintuitive, I imagine it being the distance from the root rather than the distance from leaves. IMO "level" describes the thing better, but that might be my English.

storage/types.go
```diff
@@ -33,19 +33,61 @@ type Node struct {
...
+// Path array of [0x00, 0x13] and PrefixLength 16. To see that it's the
+// left bits that are significant consider depth = 2, index = 16383 in this
+// tree (the rightmost node in a complete tree at this level). The resulting
+// NodeID has Path [0xff, 0xfc] and PrefixLength is 14. When processing this
+// NodeID to move up the tree the lowest 2 bits of Path[1] should be ignored as
+// it only takes 14 bits to reach the root of the tree from level 2 in a 16
+// levels deep tree. Also, testing Bit(0) on this NodeID would not be valid as
```
pavelkalinnikov: I think that this Bit(0) case is really confusing. I don't see a reason why NodeID must have this multiple-of-8 peculiarity, it's just a bitmask. The code that uses NodeID - yes, it might be easier to implement with such a restriction, but the bitmask code would be clearer without it and still compatible with the rest of the code.

Add a TODO to fix it?

storage/types.go
Martin2112: We've used both depth and level in many places. In some cases unhelpfully. I'm not trying to fix it in this PR as it's too big a job.

Martin2112: I'm using 'depth' here as it's the name of the parameter passed to `NewNodeIDWithPrefix`.

storage/types.go
Martin2112: I don't think it's that confusing now but maybe I don't understand what you mean. Bit(0) is always well defined, it may just not be a significant bit in a particular NodeID. 

If Go had a suitable type (c.f. Java BitSet) we might have used that. However the patterns must be kept stable as they're used as keys. They would still have to be converted to bytes in some manner as databases don't support bits as a type.

Otherwise, I suggest you write up what you mean as a PR and we'll discuss further.

storage/types.go
Martin2112: Probably true, removed.

storage/types.go
Martin2112: Moved.

storage/types.go
Martin2112: Fixed so MSB is always referring to bits.

storage/types.go
Martin2112: I think it would be too long. I've added pointers to the doc string.

storage/types.go
pavelkalinnikov: Also, [this](https://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height).

How about `TODO(...): Use "level" instead of "depth" consistently.`, and making this PR the first step towards it (the new comments at least)?

storage/types.go
pavelkalinnikov: I mean that for a bitmask of 14 bits I would expect `Bit(0)` to return the LSB of the 14, rather than something undefined. In the example above, to get the LSB I need to call `Bit(2)`, which is obv. not a great API.

If any of these IDs are put into storage then it's probably hard to not break compatibility. I still have some hope that we use it only as keys of in-memory data structures, as then it would be possible to change bits layout. Not in this PR, you are right.

Martin2112: As I said above they are used in storage as keys. The serialized format can't be changed without major pain.

But I think that the bitwise access functions are actually only used by the sparse Merkle tree code and some tests. In the sparse tree case all the paths are fixed height in multiples of 8 so Bit(0) is always valid.

storage/types.go
Martin2112: It affects more than this file though. I've already spent more time on this and related activities than I intended.

PR #1448, opened 2019-03-04
Small subtree cache optimization.

Martin2112:
Don't calculate some things unless we need to use them. Probably not a huge win but we might as well have it as it's slightly clearer to read.

### Checklist

### DIFF
--- a/storage/cache/log_subtree_cache.go
+++ b/storage/cache/log_subtree_cache.go
@@ -78,6 +78,7 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
 				// nodes. If the subtree was saved with internal nodes then we don't touch the map.
 				if height > 0 && len(st.Leaves) == maxLeaves {
 					subDepth := logStrataDepth - height
+					// TODO(Martin2112): See if we can possibly avoid the expense hiding inside NewNodeIDFromPrefix.
 					nodeID := storage.NewNodeIDFromPrefix(st.Prefix, subDepth, index, logStrataDepth, maxLogDepth)
 					_, sfx := nodeID.Split(len(st.Prefix), int(st.Depth))
 					sfxKey := sfx.String()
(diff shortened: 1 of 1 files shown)


storage/cache/log_subtree_cache.go
```diff
@@ -74,13 +74,13 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
...
 
-				subDepth := logStrataDepth - height
-				nodeID := storage.NewNodeIDFromPrefix(st.Prefix, subDepth, index, logStrataDepth, maxLogDepth)
-				_, sfx := nodeID.Split(len(st.Prefix), int(st.Depth))
-				sfxKey := sfx.String()
 				// Don't put leaves into the internal map and only update if we're rebuilding internal
 				// nodes. If the subtree was saved with internal nodes then we don't touch the map.
 				if height > 0 && len(st.Leaves) == maxLeaves {
+					subDepth := logStrataDepth - height
+					nodeID := storage.NewNodeIDFromPrefix(st.Prefix, subDepth, index, logStrataDepth, maxLogDepth)
```
AlCutter: _Might_ be worth seeing if there's a cheaper approach to figuring out the internal node suffix?

`NewNodeIDFromPrefix` looks quite heavyweight, and we don't care about the prefix here.

Martin2112: Yep, was trying to limit the frightfulness factor for the moment. TODO it?

Martin2112: I think we should micro benchmark this code before more serious optimizations.

AlCutter: Sure :)

Martin2112: I suspect the cost of hashing dominates but it would be good to know.

Martin2112: TODO added.

PR #1449, opened 2019-03-04
Add flag to limit max idle database connections for mysql

davbo:
We've been using Trillian in a serverless (function as a service) context and
using Amazon's Aurora serverless database as a backend. Aurora will scale down
the provisioned databases based upon open connections and CPU usage.

Making this a configurable flag allows us to set the idle pool to 0 and for Aurora
to scale down to 0 running databases. 

See before / after in the CloudWatch metrics here:

screenshot 2019-03-01 at 10 33 42

Go currently defaults to pooling 2 idle connections:
  https://golang.org/pkg/database/sql/#DB.SetMaxIdleConns

We have followed Go's current default of 2 but this may change in future
releases.

### Checklist

Couldn't see anywhere in the docs this could be added. 

Looks like this same feature could be added for PostgreSQL but we're only using MySQL at the moment. If keeping feature parity helps accept the PR we can test it with PostgreSQL too.

### DIFF
--- a/server/mysql_storage_provider.go
+++ b/server/mysql_storage_provider.go
@@ -31,6 +31,7 @@ import (
 var (
 	mySQLURI = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	maxConns = flag.Int("mysql_max_conns", 0, "Maximum connections to the database")
+	maxIdle  = flag.Int("mysql_max_idle_conns", -1, "Maximum idle database connections in the connection pool")
 
 	mysqlOnce            sync.Once
 	mysqlOnceErr         error
@@ -57,6 +58,9 @@ func newMySQLStorageProvider(mf monitoring.MetricFactory) (StorageProvider, erro
 		}
 		if *maxConns > 0 {
 			db.SetMaxOpenConns(*maxConns)
+		}
+		if *maxIdle >= 0 {
+			db.SetMaxIdleConns(*maxIdle)
 		}
 		mySQLstorageInstance = &mysqlProvider{
 			db: db,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### Configurable number of idle connections on MySQL
+
+This version adds a new flag `-mysql_max_idle_conns` to specify the number of
+idle database connections in the pool. Defaults to -1 which uses the Go default.
+Go default is currently set at 2 but could change in future releases.
+
 ### Client Verification
 The map client now verifies that every map leaf is being requested at most once.
 This catches potential errors before they go to the server.


davbo:
We've only just signed the CLA today, maybe that takes some time to update?

server/mysql_storage_provider.go
```diff
@@ -31,6 +31,7 @@ import (
 var (
 	mySQLURI = flag.String("mysql_uri", "test:zaphod@tcp(127.0.0.1:3306)/test", "Connection URI for MySQL database")
 	maxConns = flag.Int("mysql_max_conns", 0, "Maximum connections to the database")
+	maxIdle = flag.Int("mysql_max_idle_conns", 2, "Maximum idle database connections in the connection pool")
```
golangcibot: File is not `goimports`-ed (from `goimports`)

Martin2112: Was this fixed?

Martin2112:
Can you look at the CLA again? If it's correctly signed it should be picked up automatically. Sometimes it's an author issue that doesn't match up.

davbo:
Checked it and the emails used, all seems correct. 

Using a Google Group in our G-suite for a list of contributors, do some permissions need to be set on that in order for the members to be visible to @googlebot?

davbo:
Oh I see - [these docs](https://opensource.google.com/docs/cla/#cla-accepted) make it sound like it can take a few days

> If the contributor is covered by a corporate CLA, make sure that the corporation is listed at http://linkremoved/. If it’s not, it simply may not have been processed yet, which typically takes a few days. If you think it’s taking longer than it should, email emailremoved@ to check the status.

RJPercival:
It appears that we have a signed CLA for dave@projectsbyif.com, but not emily@projectsbyif.com. Also, your GitHub username (davbo) hasn't been associated with your email address, which is preventing the bot from associating it with your CLA.

davbo:
Huh - possibly I signed a Google CLA sometime ago for something else.

The first corporate CLA I submitted was voided so we submitted a new one, this time paying closer attention to the restrictions on job titles. Haven't heard back on that one yet but it should cover myself and Emily as our `.gitconfig`'s and Google group look to be setup correctly. Once I hear back on that I'll post here.

Thanks for checking! :-)

server/mysql_storage_provider.go
```diff
@@ -58,6 +59,7 @@ func newMySQLStorageProvider(mf monitoring.MetricFactory) (StorageProvider, erro
 		if *maxConns > 0 {
 			db.SetMaxOpenConns(*maxConns)
 		}
+		db.SetMaxIdleConns(*maxIdle)
```
Martin2112: The go doc for this implies that the default value (currently 2) could change in future. This makes picking a default for the flag somewhat problematic.

I suggest making the default for the flag -1 and making this test `>= 0`. Then if it's not set and Go change their minds in future users will still get the new default.

davbo: Thanks that's a great point. Updated to reflect your suggestion.

Martin2112: Thanks.

PR #1450, opened 2019-03-05
Better gRPC error message forwarding

gdbelvin:
Turns out there's a better way to forward gRPC errors

Before:
`rpc error: code = Unimplemented desc = map.GetLeaves(): rpc error: code = Unimplemented desc = error message`

After:
`rpc error: code = Unimplemented desc = map.GetLeaves(): error message`

### DIFF
--- a/client/map_client.go
+++ b/client/map_client.go
@@ -47,13 +47,10 @@ func NewMapClientFromTree(client trillian.TrillianMapClient, config *trillian.Tr
 func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRootV1, error) {
 	rootResp, err := c.Conn.GetSignedMapRoot(ctx, &trillian.GetSignedMapRootRequest{MapId: c.MapID})
 	if err != nil {
-		return nil, status.Errorf(status.Code(err), "GetSignedMapRoot(%v): %v", c.MapID, status.Convert(err).Message())
+		s := status.Convert(err)
+		return nil, status.Errorf(s.Code(), "GetSignedMapRoot(%v): %v", c.MapID, s.Message())
 	}
-	mapRoot, err := c.VerifySignedMapRoot(rootResp.GetMapRoot())
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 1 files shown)


Martin2112:
This seems like another case where we should be consistent across repositories.

client/map_client.go
```diff
@@ -47,11 +47,11 @@ func NewMapClientFromTree(client trillian.TrillianMapClient, config *trillian.Tr
 func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRootV1, error) {
 	rootResp, err := c.Conn.GetSignedMapRoot(ctx, &trillian.GetSignedMapRootRequest{MapId: c.MapID})
 	if err != nil {
-		return nil, status.Errorf(status.Code(err), "GetSignedMapRoot(%v): %v", c.MapID, err)
+		return nil, status.Errorf(status.Code(err), "GetSignedMapRoot(%v): %v", c.MapID, status.Convert(err).Message())
```
pavelkalinnikov: Nice one. The `err` is parsed twice though, sadly. Probably okay, if we normally don't expect errors.

gdbelvin: I know... I wasn't able to find a better way.

client/map_client.go
```diff
@@ -47,11 +47,11 @@ func NewMapClientFromTree(client trillian.TrillianMapClient, config *trillian.Tr
...
 	}
 	mapRoot, err := c.VerifySignedMapRoot(rootResp.GetMapRoot())
```
pavelkalinnikov: Maybe just `return c.VerifySignedMapRoot(...)`?

client/map_client.go
gdbelvin: I found a better way:

```go
s := status.Convert(err)
return nil, status.Errorf(s.Code(), "GetSignedMapRoot(%v): %v", c.MapID, s.Message())
```

gdbelvin:
Updated PTAL @pavelkalinnikov

PR #1452, opened 2019-03-08
Optimise Subtree query

RJPercival:
Adding `TreeId` to the `GROUP BY` clause allows MySQL and MariaDB to optimize the grouping, because the grouping is now over the primary key. See the [MySQL documentation on GROUP BY optimization](https://dev.mysql.com/doc/refman/8.0/en/group-by-optimization.html) for more information. 

Note that this is incompatible with descending indices, so the SubtreeRevision must remain in ascending order in the Subtree primary key for this to work.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ The `--master_check_interval` flag is removed from `logsigner`.
 ### Performance
 
 The performance of `SetLeaves` requests on the Map has been slightly improved.
+The performance of `GetConsistencyProof` requests has been improved when using
+MySQL.
 
 ### Logging
 
--- a/storage/postgres/tree_storage.go
+++ b/storage/postgres/tree_storage.go
@@ -41,7 +41,7 @@ const (
 			FROM subtree n
 			WHERE n.subtree_id IN (` + placeholderSQL + `) AND
 			n.tree_id = ? AND n.subtree_revision <= ?
-			GROUP BY n.subtree_id
+			GROUP BY n.tree_id, n.subtree_id
 		) AS x
 		INNER JOIN subtree
 		ON subtree.subtree_id = x.subtree_id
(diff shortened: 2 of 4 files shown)


AlCutter:
You can tick the "I've updated the changelog" box :)

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ The `--master_check_interval` flag is removed from `logsigner`.
 ### Performance
 
 The performance of `SetLeaves` requests on the Map has been slightly improved.
+The performance of `GetConsistencyProof` requests has been improved.
```
AlCutter: Maybe mention that this is for MySQL (and pgsql) specifically?

storage/postgres/tree_storage.go
```diff
@@ -41,7 +41,7 @@ const (
 			FROM subtree n
 			WHERE n.subtree_id IN (` + placeholderSQL + `) AND
 			n.tree_id = ? AND n.subtree_revision <= ?
-			GROUP BY n.subtree_id
+			GROUP BY n.tree_id, n.subtree_id
```
AlCutter: Does pgsql have the same optimisation then, or is this speculative?
If you have this in here you might want to update the PR description to mention pgsql too.

RJPercival: After doing a bit of research into Postgres, I've found that [it doesn't support loose index scans](https://wiki.postgresql.org/wiki/Loose_indexscan), which is what this optimisation takes advantage of, so making this change to the Postgres SQL doesn't achieve anything. It appears that it may be possible to modify this SQL to use a recursive CTE (Common Table Expression), which would achieve the same optimisation, but I'm not sure it's worth going to the trouble right now. Perhaps I should just leave a TODO here to consider it?

AlCutter: Yeah, TODO seems like the best idea for now.

RJPercival: Done.

CHANGELOG.md
RJPercival: Done.

pgporada:
I would like to extend my gratitude for the query optimization in this PR. We had been plagued by context deadline exceeded timeouts from `get-sth-consistency` queries that would run for >=6000s. Since deploying a build with `+18d97ed`, `get-sth-consistency` calls return speedily.

PR #1455, opened 2019-03-12
Add a codecov repository yaml override for Trillian

Martin2112:
Try to ignore code that's related to testing, docs or examples. Possibly we pick up some from the google.org yaml but this does not currently exclude the top level `testonly` directory. If it works this should cut down the noise from code coverage.

### Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,19 @@
+# Customizations to codecov for Trillian repo. This will be merged into
+# the team / default codecov yaml file.
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "cmd/createtree/testonly"
+  - "crypto/keys/testonly"
+  - "docs/**/*"
+  - "examples/**/*"
+  - "integration/**/*"
+  - "monitoring/testonly"
+  - "server/mock_log_operation.go"
+  - "storage/cache/mock_node_storage.go"
+  - "storage/mock_storage.go"
+  - "storage/testonly"
+  - "testonly/**/*"
+  - "util/election2/testonly"
+  - "vendor/**/*"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,8 @@
 # Customizations to codecov for Trillian repo. This will be merged into
 # the team / default codecov yaml file.
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
 
 # Exclude code that's for testing, demos or utilities that aren't really
 # part of production releases.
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,22 @@
+# Customizations to codecov for Trillian repo. This will be merged into
+# the team / default codecov yaml file.
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "cmd/createtree/testonly"
```
RJPercival: Could we not use a pattern like "**/testonly" to match all of the testonly directories?

codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+  - "crypto/keys/testonly"
+  - "docs/**/*"
+  - "examples/**/*"
+  - "integration/**/*"
+  - "monitoring/testonly"
+  - "server/mock_log_operation.go"
```
RJPercival: Could we not use a pattern like "mock_*.go" to match all mocks?

Martin2112: I'm not sure the recursive matching is smart enough. It might be but the docs weren't entirely clear.

codecov.yml
Martin2112: According to the docs the recursive match is somewhat limited. It's possible that some variant can be made to work but there's a caveat on the docs here: https://docs.codecov.io/docs/ignoring-paths

RJPercival: Isn't that caveat just that "*" doesn't match recursively but "**" does?

codecov.yml
Martin2112: Maybe. I can try it and see if it excludes the right things.

RJPercival: We can test it by looking at the coverage report, can't we?

Martin2112:
OK let's see what this does.

Martin2112:
Seems like it might have worked.

codecov.yml
```diff
@@ -0,0 +1,16 @@
...
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "**/mock_*.go"
+  - "**/testonly"
+  - "docs/**/*"
+  - "examples/**/*"
+  - "integration/**/*"
+  - "testonly/**/*"
```
RJPercival: I think this is redundant (I'd expect "**/testonly" to match this).

Martin2112: It will take another hour to find out though.

Martin2112: I'll merge this and test that out seperately.

Martin2112: My test shows that this doesn't match. That might be why testonly wasn't included before if the google.org config (which I don't think I can see) has this in it.

PR #1459, opened 2019-03-13
Validate hashes passed to GetInclusionProofByHash and GetLeavesByHash

RJPercival:
Checks that the hashes are the correct length. This provides more helpful error messages when these endpoints are being used incorrectly, e.g. by sending the base64-encoded hash instead of the raw hash. Previously, GetInclusionProofByHash would return an error saying that no leaf was found, and GetLeavesByHash would simply return no results. Neither of these results indicates to the user that they are providing an invalid argument. Now, an InvalidArgument error will be returned.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ This version adds a new flag `-mysql_max_idle_conns` to specify the number of
 idle database connections in the pool. Defaults to -1 which uses the Go default.
 Go default is currently set at 2 but could change in future releases.
 
-### Server Verification
+### Server verification of leaf hashes
 
 The log server now verifies that leaf hashes are the correct length and returns
 an InvalidArgument error if they are not. Previously, GetLeavesByHash would
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,14 +54,14 @@ This version adds a new flag `-mysql_max_idle_conns` to specify the number of
 idle database connections in the pool. Defaults to -1 which uses the Go default.
 Go default is currently set at 2 but could change in future releases.
 
-### Server verification of leaf hashes
+### Server validation of leaf hashes
 
-The log server now verifies that leaf hashes are the correct length and returns
+The log server now checks that leaf hashes are the correct length and returns
 an InvalidArgument error if they are not. Previously, GetLeavesByHash would
 simply not return any matching leaves for invalid hashes, and
-GetInclusionProofByHash would previously have returned a NotFound error.
+GetInclusionProofByHash would return a NotFound error.
 
-### Client Verification
+### Client validation of map leaf requests
 
 The map client now verifies that every map leaf is being requested at most once.
 This catches potential errors before they go to the server.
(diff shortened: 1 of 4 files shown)


server/log_rpc_server.go
```diff
@@ -549,10 +553,18 @@ func (t *TrillianLogRPCServer) GetLeavesByHash(ctx context.Context, req *trillia
...
 
-	tree, ctx, err := t.getTreeAndContext(ctx, req.LogId, optsLogRead)
+	tree, hasher, err := t.getTreeAndHasher(ctx, req.LogId, optsLogRead)
 	if err != nil {
 		return nil, err
 	}
+	ctx = trees.NewContext(ctx, tree)
+
+	for i, hash := range req.LeafHash {
+		if got, want := len(hash), hasher.Size(); got != want {
```
Martin2112: There's a validateLeafHash() method in that file that currently just checks `len(hash) != 0`. Maybe we could move this check into there and the same checks would apply for both hash related API calls.

RJPercival: I did that first, but it required slightly more pervasive changes because that is currently called before the tree is fetched, which is required for the hash size check. I'll add a commit with this refactor and you can decide whether you prefer it.

Martin2112:
OK. I think it would be better if we make the checks consistent. If it
makes it easier feel free to change the existing validation.

RJPercival:
Think this is worth a CHANGELOG entry? It does result in errors being returned where previously it would have succeeded (but returned no results).

Martin2112:
It's an API change so possibly worth a small note.  

It looks like CTFE just passes any non empty hash through so it might change the status code there from `NotFound` to `BadRequest`? Might need a small tweak to `handlers.go`.

Martin2112:
Other than that I think it looks OK.

RJPercival:
I've rebased this on the latest version of log_rpc_server_test.go and added a CHANGELOG.md entry.

RJPercival:
> It looks like CTFE just passes any non empty hash through so it might change the status code there from `NotFound` to `BadRequest`? Might need a small tweak to `handlers.go`.

That's correct, the CTFE would now return BadRequest if the base64-encoded hash decodes to a string of the wrong length. That seems an improvement over NotFound. I'll add a note about this to the certificate-transparency-go CHANGELOG.

Martin2112:
Can you rebase it and I'll take another look.

RJPercival:
Rebased.

CHANGELOG.md
```diff
@@ -54,14 +54,22 @@ This version adds a new flag `-mysql_max_idle_conns` to specify the number of
 idle database connections in the pool. Defaults to -1 which uses the Go default.
 Go default is currently set at 2 but could change in future releases.
 
+### Server Verification
```
Martin2112: That comment is a bit generic. Can we add Hash or something?

RJPercival: I've changed this heading to "Server verification of leaf hashes".

Martin2112: Maybe 'validate' is better as it doesn't raise the spectre of crypto like verification does?

RJPercival: Should we also change the "client verification" section just below, as that appears to have a similar issue?

Martin2112: Might as well, it was also a very generic title.

RJPercival: Done.

Martin2112:
LGTM.

PR #1460, opened 2019-03-15
Cleanup a bunch of code inspection warnings.

Martin2112:
Mostly mechanical changes from code analysis warnings. Should be no functional changes.

All the commits are independent and can be reviewed separately.

### Checklist

### DIFF
--- a/util/election2/testonly/tests.go
+++ b/util/election2/testonly/tests.go
@@ -64,7 +64,7 @@ func checkDone(ctx context.Context, t *testing.T, wait time.Duration) {
 
 // runElectionAwait tests the Await call with different pre-conditions.
 func runElectionAwait(t *testing.T, f election2.Factory) {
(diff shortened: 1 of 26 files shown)


util/election2/testonly/tests.go
```diff
@@ -64,7 +64,7 @@ func checkDone(ctx context.Context, t *testing.T, wait time.Duration) {
 
 // runElectionAwait tests the Await call with different pre-conditions.
 func runElectionAwait(t *testing.T, f election2.Factory) {
-	awaitErr := errors.New("Await error")
+	awaitErr := errors.New("await error")
```
pavelkalinnikov: Not sure about this one, "Await" here means the name of the exported method. But should be fine since this is a test.

Martin2112: It's only a placeholder error for the test. I think it will do as is.

PR #1462, opened 2019-03-19
Remove deprecated signed log root fields

Martin2112:
### Checklist


server/log_rpc_server_test.go
```diff
@@ -2122,3 +2105,15 @@ func addTreeID(tree *trillian.Tree, treeID int64) *trillian.Tree {
 	newTree.TreeId = treeID
 	return newTree
 }
+
+type protoMatcher struct {
```
golangcibot: U1000: type `protoMatcher` is unused (from `unused`)

Martin2112:
Best done commit by commit. It looks scary but the big change is the test structure.

server/log_rpc_server.go
```diff
@@ -485,12 +485,16 @@ func (t *TrillianLogRPCServer) GetLeavesByIndex(ctx context.Context, req *trilli
 		return nil, err
 	}
 
-	root, err := tx.LatestSignedLogRoot(ctx)
+	slr, err := tx.LatestSignedLogRoot(ctx)
 	if err != nil {
 		return nil, err
 	}
+	var root types.LogRootV1
```
gdbelvin: Perhaps I'm missing something, but I don't see `root` being used.

server/log_rpc_server.go
```diff
@@ -564,18 +568,22 @@ func (t *TrillianLogRPCServer) GetLeavesByHash(ctx context.Context, req *trillia
...
```
gdbelvin: is `root` referenced?

Martin2112: It's just being checked for validity (unpacks ok) before we return it to clients.

CHANGELOG.md
```diff
@@ -9,6 +9,30 @@ Not yet released; provisionally v2.0.0 (may change).
...
+a CTFE that's older than about 11 months it won't work properly. Make sure
+to update CTFE first.
+
+All the fields marked as deprecated in this proto have been removed. All
+the same fields are available via the TLS marshalled log root in the proto.
+Updating affected code is straightforward. For example HTTP handler code that
+needs to use these fields can do something like:
+
+```
+var currentRoot types.LogRootV1
```
gdbelvin: Please direct people to use the [`VerifySignedLogRoot`](https://github.com/google/trillian/blob/e6c1e8cf1ccff4929a9a5624f1a4f77c2f5bb40e/crypto/verifier.go#L33) utility functions.  Unmarshal is unsafe before the signature has been verified.

server/log_rpc_server.go
gdbelvin: Perhaps a local call to `VerifySignedLogRoot` would be better?

gdbelvin: It would also be clearer because the returned argument could be set to underscore `_`

Martin2112: I think that's beyond the scope of this PR, it's already a big change.

gdbelvin: Would a TODO be appropriate?

CHANGELOG.md
Martin2112: Unsafe in what way (assuming you trust the server)?

gdbelvin: Clients should never trust the server ;-) - that's the whole point of being verifiable. 
If you trust the server Unmarshal is fine, but we should probably not encourage that.

server/log_rpc_server.go
Martin2112: Maybe an issue as it's not a trivial change. The tests don't include signatures for one thing so would all need updating.

CHANGELOG.md
Martin2112: Well yes but that's how we started out with CT where it's a single system. I'll try to clarify the doc.

gdbelvin: Thanks

server/log_rpc_server.go
Martin2112: On more thought does it really make sense for the server to check the signature it just pulled from storage? It's a fair point for clients but there isn't really a boundary here. If we add checks we prevent clients from detecting misoperation.

gdbelvin: There's no boundary here.  Checking the signature would be similar to the unmarshing check: perform the checks on the server to make sure everything is ok before returning it to the client.

CHANGELOG.md
Martin2112: Done.

PR #1470, opened 2019-03-22
Log when duplicate indices are fetched in the same request.

mhutchinson:
The same index being fetched multiple times in the same getLeavesByRevision request causes duplicate work to happen. This simple logging change allows inspection of the frequency and magnitude of this
work duplication.


server/map_rpc_server.go
```diff
@@ -145,6 +145,9 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 	for i, l := range leaves {
 		leavesByIndex[string(l.Index)] = &leaves[i]
 	}
+	if len(indices) != len(leavesByIndex) {
+		glog.Infof("%v: request had %v indices, %v of these are unique", mapID, len(indices), len(leavesByIndex))
```
Martin2112: That should be a verbose level log I think.

mhutchinson: Yup, this could get quite spammy. Updated in subsequent commit.

Martin2112:
I believe I've fixed the build errors.

PR #1474, opened 2019-03-25
Update GetLatestSignedLogRoot Documentation

gdbelvin:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ Not yet released; provisionally v2.0.0 (may change).
 
 ### GetLatestSignedLogRoot With Consistency Proof
 
-`GetLatestSignedLogRoot` in LogServer also returns a consistency proof if
-possible, in addition to the latest signed log root. This reduces the number
-RPC calls from logClient from 2 to 1 in `getAndVerifyLatestRoot`.
+`GetLatestSignedLogRoot` in the LogServer will return a consistency proof if
+`first_tree_size` > 0. This reduces the number of RPC calls from logClient from
+2 to 1 in `client.getAndVerifyLatestRoot`.
 
 ### Testing
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### GetLatestSignedLogRoot With Consistency Proof
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### GetLatestSignedLogRoot With Consistency Proof
+
+`GetLatestSignedLogRoot` in LogServer also returns a consistency proof if
+possible, in addition to the latest signed log root. This reduces the number
```
RJPercival:
```suggestion
possible, in addition to the latest signed log root. This reduces the number of
```

RJPercival: Also, should mention that it'll only do this if the new `first_tree_size` field is set in `GetLatestSignedLogRootRequest`.

gdbelvin:
Done

trillian_log_api.proto
```diff
@@ -247,7 +249,7 @@ message GetLatestSignedLogRootRequest {
 message GetLatestSignedLogRootResponse {
   SignedLogRoot signed_log_root = 2;
   // proof is filled if first_tree_size in GetLatestSignedLogRootRequest is
-  // non-zero and the new tree size is not smaller than first_tree_size.
+  // non-zero.
```
Martin2112: Adding a proof here has complicated the semantics of caching roots. Plus there's a cost to fetching proofs that users should be aware of.

Please don't make Log API changes without giving me a chance to review them. Not all applications use the log client.

gdbelvin: I'll be sure to add you to the reviewers list in the future.  
Consider adding a `CODEOWNERS` file to make sure you can review all API changes in the future. [example](https://github.com/google/keytransparency/blob/master/CODEOWNERS)

I discussed this extensively with @daviddrysdale and thought this would be a safe API change because it doesn't change any behavior for existing requests.  Proto3 fields are 0 by default, and this change doesn't incur any additional cost when `first_tree_size` == 0.

Martin2112: It's not an unsafe change directly. Previously these responses were unconditionally cacheable. If applications are doing this and then somehow later set the `first_tree_size` field then this could go wrong.

It might have been clearer to have a new RPC method but as it's a theoretical risk this will probably be ok.

PR #1477, opened 2019-03-25
Add map_server to docker-compose.yml

RJPercival:
Makes it easy for people to experiment with the map server.

### Checklist


RJPercival:
Worth a CHANGELOG entry?

AlCutter:
Yup, why not :)

RJPercival:
It turns out that the CHANGELOG already covers this since #1479.

RJPercival:
I added a test to make sure the map server is working correctly, which is passing.

PR #1480, opened 2019-03-26
Fixes for transaction leaks in RPC server and storage level operations

Martin2112:
QueueLeaves / AddSequenced leaves (Storage Level API) in memory / mysql
storage could leak a transaction in 2 ways - on an ErrTreeNeedsInit or
if the operation they call on the transaction failed. Transactions are
created inside these methods.

The log RPC server could leak via the above for write methods or in a
similar manner for read RPCs by not closing a TX when receiving an
ErrTreeNeedsInit from the initial Snapshot() call.

Add tests. Update some tests to ensure Close() is called for errors by
removing AnyTimes().

The Map RPC server had a similar issue to the Log for read methods.

### Checklist

### DIFF
--- a/server/log_rpc_server.go
+++ b/server/log_rpc_server.go
@@ -817,9 +817,9 @@ func spanFor(ctx context.Context, name string) (context.Context, *trace.Span) {
 
 func (t *TrillianLogRPCServer) snapshotForTree(ctx context.Context, tree *trillian.Tree, method string) (storage.ReadOnlyLogTreeTX, error) {
 	tx, err := t.registry.LogStorage.SnapshotForTree(ctx, tree)
-	if err == storage.ErrTreeNeedsInit {
-		// Special case - We have an error but we still have an open transaction.
-		// To avoid leaking it make sure it's closed and indicate with status code.
+	if err != nil && tx != nil {
+		// Special case to handle ErrTreeNeedsInit, which leaves the TX open.
+		// To avoid leaking it make sure it's closed.
 		defer t.closeAndLog(ctx, tree.TreeId, tx, method)
 	}
 	return tx, err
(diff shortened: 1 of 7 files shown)


server/log_rpc_server.go
```diff
@@ -812,3 +812,13 @@ func (t *TrillianLogRPCServer) recordIndexPercent(leafIndex int64, treeSize uint
 func spanFor(ctx context.Context, name string) (context.Context, *trace.Span) {
 	return trace.StartSpan(ctx, fmt.Sprintf("%s.%s", traceSpanRoot, name))
 }
+
+func (t *TrillianLogRPCServer) snapshotForTree(ctx context.Context, tree *trillian.Tree, method string) (storage.ReadOnlyLogTreeTX, error) {
+	tx, err := t.registry.LogStorage.SnapshotForTree(ctx, tree)
+	if err == storage.ErrTreeNeedsInit {
```
AlCutter: For consistency, would it be better to have this close whenever `err != nil && tx != nil`?

Martin2112: Could do if you think it helps. It's already closed though except for when this error happens. I wish we'd never added the special case but too late now.

Martin2112:
I've been thinking how to test the storage level QueueLeaves and I don't see an easy way. It's possible to embed / wrap the log storage but it will still call it's own beginInternal(), which prevents capturing the TX to ensure it's been closed.

server/log_rpc_server.go
Martin2112: Done.

Martin2112:
I think the map_rpc_server has a similar issue. Will look at that next.

Martin2112:
Map server fixes added. PTAL.

Martin2112:
I don't think it affects CloudSpanner as it doesn't implement the storage level Add / Queue leaves. The other fixes should close the TX if it returns `ErrTreeNeedsInit`.

Martin2112:
Sorry, wrong way round. It does implement the storage level QueueLeaves but it doesn't create the tx inside this method like MySQL does, or didn't seem to.

PR #1496, opened 2019-04-04
storage/mysql/queue.go: Properly defer statement close in removeSequencedLeaves

rolandshoemaker:
Fixes a possible leak of statements in `removeSequencedLeaves` when executing the statement fails. I think I may have introduced this leak when adding the batched queue/dequeue mode, so sorry about that.

(Not sure this is a big enough change to need a mention in the changelog, but happy to add one if you folks feel otherwise!)

### Checklist


Martin2112:
The statement is prepared on the tx, which means it shouldn't outlive it. Probably still best to close things consistently though.

Martin2112:
I think the Travis failure was related to issues with gRPC. Running it again.

Martin2112:
This might need a rebase. The bit that's failing has been removed from the mainline version of .travis.yml.

PR #1502, opened 2019-04-09
Add trace spans to main proof_fetcher.go funcs

Martin2112:
### Checklist


Martin2112:
I think we could add more at the level below the RPC handlers and in some of the code we're looking at for performance improvements.

PR #1507, opened 2019-04-11
Make storage.Suffix immutable and share instances for 8 bit values.

Martin2112:
Profiling shows that tree hashing is spending a lot of time in alloc / gc of small objects. This tries to fix one case by making Suffix an immutable object and then sharing pre-allocated instances for the <=8 bit Suffix values used by logs. The internal APIs have been changed to use *Suffix in a few places.

This should gain the following benefits:

* Log rehashing does not constantly allocate the same Suffix objects over and over.
* String() is computed once so for cached values we don't incur base64 / string allocation repeatedly.
* Use of pointers avoids making copies and thus more overhead.
* User code doesn't manipulate these things so external APIs don't change.

Should make sense to look at it commit by commit.

### Checklist

### DIFF
--- a/storage/suffix.go
+++ b/storage/suffix.go
@@ -61,11 +61,11 @@ func (s Suffix) String() string {
 }
 
 // ParseSuffix converts a suffix string back into a Suffix.
-func ParseSuffix(s string) (Suffix, error) {
+func ParseSuffix(s string) (*Suffix, error) {
 	b, err := base64.StdEncoding.DecodeString(s)
 	if err != nil {
-		return EmptySuffix, err
+		return &EmptySuffix, err
 	}
 
-	return *NewSuffix(byte(b[0]), b[1:]), nil
+	return NewSuffix(byte(b[0]), b[1:]), nil
 }
--- a/storage/suffix.go
+++ b/storage/suffix.go
@@ -23,18 +23,28 @@ import (
 type Suffix struct {
 	// bits is the number of bits in the node ID suffix.
 	// TODO(gdbelvin): make bits an integer.
-	Bits byte
+	bits byte
 	// path is the suffix itself.
-	Path []byte
+	path []byte
+}
+
+// Bits returns the number of significant bits in the Suffix path.
+func (s Suffix) Bits() byte {
+	return s.bits
+}
+
+// Path returns a copy of the Suffix path.
+func (s Suffix) Path() []byte {
+	return append([]byte(nil), s.path...)
 }
 
 // String returns a string that represents Suffix.
 // This is a base64 encoding of the following format:
 // [ 1 byte for depth || path bytes ]
 func (s Suffix) String() string {
-	r := make([]byte, 1, 1+(len(s.Path)))
-	r[0] = s.Bits
-	r = append(r, s.Path...)
+	r := make([]byte, 1, 1+(s.bits/8))
+	r[0] = s.bits
+	r = append(r, s.path...)
 	return base64.StdEncoding.EncodeToString(r)
 }
 
@@ -46,7 +56,7 @@ func ParseSuffix(s string) (Suffix, error) {
 	}
 
 	return Suffix{
-		Bits: byte(b[0]),
-		Path: b[1:],
+		bits: byte(b[0]),
+		path: b[1:],
 	}, nil
 }
--- a/storage/suffix.go
+++ b/storage/suffix.go
@@ -18,11 +18,14 @@ import (
 	"encoding/base64"
 )
 
-// EmptySuffix is a reusable suffix of zero bits.
-var EmptySuffix = Suffix{path: []byte{0}}
-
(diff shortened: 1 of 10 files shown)


storage/types.go
```diff
@@ -373,13 +373,14 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, Suffix) {
 		panic(fmt.Sprintf("storage Split: %x(n.PrefixLenBits: %v - prefixBytes: %v *8) <= 0", n.Path, n.PrefixLenBits, prefixBytes))
 	}
 	suffixBytes := bytesForBits(b)
+	maskIndex := (b - 1) / 8
+	maskLowBits := (b-1)%8 + 1
+	path := a[prefixBytes : prefixBytes+suffixBytes]
```
AlCutter: nit: Maybe rename this to `sfxPath` to make the diff a little clearer?

storage/suffix.go
```diff
@@ -18,6 +18,8 @@ import (
 	"encoding/base64"
 )
 
+var EmptySuffix = Suffix{path: []byte{0}}
```
AlCutter: Should this be using `NewSuffix`?

storage/suffix.go
```diff
@@ -61,11 +61,11 @@ func (s Suffix) String() string {
 }
 
 // ParseSuffix converts a suffix string back into a Suffix.
-func ParseSuffix(s string) (Suffix, error) {
+func ParseSuffix(s string) (*Suffix, error) {
 	b, err := base64.StdEncoding.DecodeString(s)
 	if err != nil {
-		return EmptySuffix, err
+		return &EmptySuffix, err
```
AlCutter: this should probably be `nil, err` now - I suspect it was just returning an empty `Suffix` because it couldn't return `nil`.

storage/suffix.go
```diff
@@ -18,10 +18,16 @@ import (
 	"encoding/base64"
 )
 
+// EmptySuffix is a reusable suffix of zero bits.
 var EmptySuffix = Suffix{path: []byte{0}}
 
+var byteToSuffix = make(map[byte]*Suffix)
```
AlCutter:
```golang
var (
...
)```

File: storage/suffix.go [hunk 5]
```diff
@@ -23,18 +23,28 @@ import (
...
 }
 
 // String returns a string that represents Suffix.
 // This is a base64 encoding of the following format:
 // [ 1 byte for depth || path bytes ]
 func (s Suffix) String() string {
-	r := make([]byte, 1, 1+(len(s.Path)))
-	r[0] = s.Bits
-	r = append(r, s.Path...)
+	r := make([]byte, 1, 1+(s.bits/8))
```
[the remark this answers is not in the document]
Martin2112: Done. Might have been one short before if not a multiple of 8, which I think never happens.

storage/suffix.go
```diff
@@ -23,18 +23,28 @@ import (
...
+}
+
+// Bits returns the number of significant bits in the Suffix path.
+func (s Suffix) Bits() byte {
+	return s.bits
+}
+
+// Path returns a copy of the Suffix path.
+func (s Suffix) Path() []byte {
+	return append([]byte(nil), s.path...)
```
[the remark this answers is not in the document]
Martin2112: I think this preserves nil if s.path is nil or I'd have done a copy. I don't think path is legitimately ever nil though?

storage/types.go
```diff
@@ -349,18 +349,18 @@ func (n *NodeID) Siblings() []NodeID {
 func NewNodeIDFromPrefixSuffix(prefix []byte, suffix Suffix, maxPathBits int) NodeID {
 	path := make([]byte, maxPathBits/8)
 	copy(path, prefix)
-	copy(path[len(prefix):], suffix.Path)
+	copy(path[len(prefix):], suffix.Path())
```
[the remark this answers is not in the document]
Martin2112: Done. Was mechanically replaced.

storage/types.go

storage/suffix.go
Martin2112: Done.

storage/suffix.go
Martin2112: Done. Another mechanical replacement.

storage/suffix.go
Martin2112: Done.

storage/suffix.go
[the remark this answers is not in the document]
AlCutter: Could explicitly check for nil?
I was just thinking about the alloc + realloc (although I do wonder if the compiler doesn't just see the pattern and do an implicit len).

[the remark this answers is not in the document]
Martin2112: I think that now it's all forced to create instances via `NewSuffix` it can't be nil. Done.

Martin2112:
Decrease in `suffix.go` coverage is because of the sanity checks / panics that should be unhittable.

PR #1510, opened 2019-04-11
sequencer: Use compact.Range directly

pavelkalinnikov:
The `compact.Range` type provides a superset of what `compact.Tree` can do.
This change removes dependency from the `compact.Tree` type in Trillian, and
allows deleting it completely.

Addresses #1338

### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -17,6 +17,7 @@
 package log
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strconv"
@@ -134,9 +135,10 @@ func NewSequencer(
 
 // initMerkleTreeFromStorage builds a compact tree that matches the latest data
 // in the database. It ensures that the root hash matches the passed in root.
-func (s Sequencer) initMerkleTreeFromStorage(ctx context.Context, root *types.LogRootV1, tx storage.TreeTX) (*compact.Tree, error) {
+func (s Sequencer) initMerkleTreeFromStorage(ctx context.Context, root *types.LogRootV1, tx storage.TreeTX) (*compact.Range, error) {
+	fact := compact.RangeFactory{Hash: s.hasher.HashChildren}
 	if root.TreeSize == 0 {
-		return compact.NewTree(s.hasher), nil
+		return fact.NewEmptyRange(0), nil
 	}
 
 	ids := compact.TreeNodes(root.TreeSize)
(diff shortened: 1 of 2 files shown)


log/sequencer.go
```diff
@@ -416,11 +456,15 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 		stageStart = s.timeSource.Now()
 
 		// Create the log root ready for signing
-		seqTreeSize.Set(float64(merkleTree.Size()), label)
+		rootHash, err := merkleTree.GetRootHash()
```
pavelkalinnikov: Gotcha: `rootHash` can be nil if the tree is empty. Probably won't happen, but we should better double check it here.

log/sequencer.go
```diff
@@ -167,7 +169,19 @@ func (s Sequencer) initMerkleTreeFromStorage(ctx context.Context, root *types.Lo
...
 	}
 
-	return compact.NewTreeWithState(s.hasher, root.TreeSize, hashes, root.RootHash)
+	rng, err := fact.NewRange(0, root.TreeSize, hashes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create compact.Range: %v", err)
+	}
+	hash, err := rng.GetRootHash(nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to restore the root hash: %v", err)
```
Martin2112: Not sure about "restore" here. Can maybe pass the error through directly.

log/sequencer.go
```diff
@@ -211,25 +225,28 @@ func (s Sequencer) prepareLeaves(leaves []*trillian.LogLeaf, begin uint64, label
 
 // updateCompactTree adds the passed in leaves to the compact tree. Returns a
 // map of all updated tree nodes, and the new root hash.
-func (s Sequencer) updateCompactTree(mt *compact.Tree, leaves []*trillian.LogLeaf, label string) (map[compact.NodeID][]byte, []byte, error) {
+func (s Sequencer) updateCompactTree(mt *compact.Range, leaves []*trillian.LogLeaf, label string) (map[compact.NodeID][]byte, []byte, error) {
```
Martin2112: May already be on the list but the name of this func is now out of date.

log/sequencer.go
pavelkalinnikov: Changed to "compute".

log/sequencer.go

PR #1514, opened 2019-04-11
Improve error handling in flagsaver.

Martin2112:
Exclude saving of "go test" related flags and problematic
log_backtrace_at flag. All other errors related to setting flags are
reported. Add MustRestore() to make tests cleaner.

### Checklist

### DIFF
--- a/util/flagsaver/flagsaver.go
+++ b/util/flagsaver/flagsaver.go
@@ -22,7 +22,12 @@
 //   } // flags are reset to their original values here.
 package flagsaver
 
-import "flag"
+import (
+	"flag"
+	"strings"
+
+	"github.com/golang/glog"
+)
 
 // Stash holds flag values so that they can be restored at the end of a test.
 type Stash struct {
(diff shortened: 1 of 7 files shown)


Martin2112:
Ping.

util/flagsaver/flagsaver.go
```diff
@@ -42,9 +50,23 @@ func Save() *Stash {
...
+		if !strings.HasPrefix(f.Name, "test.") && f.Name != "log_backtrace_at" {
+			s.flags[f.Name] = f.Value.String()
+		}
 	})
 
 	return &s
 }
+
+// MustRestore calls Restore and exits on failure. It can be used in a defer for
+// tests. If Restore fails then otherwise the flags may be in an arbitrary
```
RJPercival:
```suggestion
// tests. If Restore fails then the flags may be in an arbitrary
```

util/flagsaver/flagsaver_test.go
```diff
@@ -94,8 +94,12 @@ func TestRestore(t *testing.T) {
 		}
 
 		func() {
-			defer Save().Restore()
-			flag.Set(test.flag, test.newValue)
+			defer Save().MustRestore()
+			// If the Set() fails the value won't have been updated but some of the
+			// test cases set the same value so it's safer to have this check.
+			if err := flag.Set(test.flag, test.newValue); err != nil {
+				t.Errorf("%v: flag.Set(%q)=%v", test.desc, test.flag, err)
```
RJPercival:
```suggestion
				t.Errorf("%v: flag.Set(%q) = %q, want nil", test.desc, test.flag, err)
```

Martin2112:
We should maybe move this code somewhere that's not part of test coverage?

RJPercival:
Perhaps the testonly directory? I'm not sure there's a use case for it outside of tests.

Martin2112:
Yep I think that's a good idea.

PR #1516, opened 2019-04-11
Format comments, remove unused/buggy methods

pavelkalinnikov:
In this change comments of `compact.Tree` are wrapped and cleaned up. Additionally, two unused methods are removed: `Depth` (which also contained a bug) and `DumpNodes`.

### Checklist

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -26,18 +26,21 @@ import (
 	"github.com/google/trillian/merkle/hashers"
 )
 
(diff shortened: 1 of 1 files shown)


Martin2112:
Quick drive by. It's handy to include what's been removed in the PR desc. It looks like two funcs were removed?

pavelkalinnikov:
@Martin2112 Updated the description.

pavelkalinnikov:
Thanks. Yeah, I will gradually replace it with `compact.Range` which is 100% tested.

PR #1518, opened 2019-04-12
Basic bazel build targets for proto files.

mhutchinson:
This allows clients integrating with Trillian from languages such as Python and Java to easily compile the proto definitions if using Bazel.
It also provides a way to query for all proto files that may be needed to build a target:

bazel query --nohost_deps --noimplicit_deps \
  'kind("source file", deps(:trillian_admin_api_proto))'

This is documented within the BUILD file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.tfstate.backup
 *~
 /.idea
+/bazel-*
 /commit_log
 /coverage.txt
 /createtree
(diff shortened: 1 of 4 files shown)


Martin2112:
Are we planning to have Travis tests for this? If we're not actively using it then we might not notice breakages.

mhutchinson:
@Martin2112 - yup, that sounds worthwhile. Other mixins we can bring into this CL:
1) Documentation
2) Example minimal client using one of the targets

As an example of (2) see https://github.com/mhutchinson/trillian-java-3p. Would it be worth including that in this repo?

Martin2112:
Possibly. We also have an examples repository that could be a home for it. I think basic checks that the build works will do for here?

PR #1521, opened 2019-04-12
Fix unhandled errors in coniks.

Martin2112:
Go docs say that these errors won't happen so make ignoring them
explicit.

### Checklist


Martin2112:
Yes but should only be done it when it's definitely safe.

Martin2112:
I think I won't merge this because binary.Write() could change underneath it.

PR #1522, opened 2019-04-12
compact: Implement Tree using Range

pavelkalinnikov:
This change replaces internals of `compact.Tree` with the `compact.Range`.

The old `Hashes` method/format are now used only in tests, and will be removed
in a follow-up after tests refactoring.

#### Benchmarks
No significant changes.
```
Before:
BenchmarkAppendLeaf-12             	     500	   3214217 ns/op
BenchmarkAppendLeafNoVisitor-12    	     500	   3069281 ns/op

After:
BenchmarkAppendLeaf-12             	     500	   3118462 ns/op
BenchmarkAppendLeafNoVisitor-12    	     500	   3089277 ns/op
```

### Checklist

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -32,19 +32,8 @@ import (
 // TODO(pavelkalinnikov): Remove it, use compact.Range instead.
 type Tree struct {
 	hasher hashers.LogHasher
-	// The list of "dangling" left-hand nodes, where entry [0] is the leaf.
-	// So: nodes[0] is the hash of a subtree of size 1 = 1<<0, if included.
-	//     nodes[1] is the hash of a subtree of size 2 = 1<<1, if included.
-	//     nodes[2] is the hash of a subtree of size 4 = 1<<2, if included.
-	//     ....
-	// Nodes are included if the tree size includes that power of two.
-	// For example, a tree of size 21 is built from subtrees of sizes
-	// 16 + 4 + 1, so nodes[1] == nodes[3] == nil.
-	//
-	// For a tree whose size is a perfect power of two, only the last
-	// entry in nodes will be set (which is also the root hash).
-	nodes [][]byte
-	size  int64
+	nodes  [][]byte
+	size   int64
 }
 
 func isPerfectTree(size int64) bool {
@@ -266,6 +255,24 @@ func (t *Tree) Hashes() [][]byte {
 	if isPerfectTree(t.size) {
 		return nil
 	}
+	return t.getNodes()
+}
+
+// getNodes returns the list of "dangling" left-hand nodes on the right border
+// of the tree. They are indexed from the bottom, so the entry [0] is a leaf.
+//
+// So: nodes[0] is the hash of a subtree of size 1 = 1<<0, if included.
+//     nodes[1] is the hash of a subtree of size 2 = 1<<1, if included.
+//     nodes[2] is the hash of a subtree of size 4 = 1<<2, if included.
+//     ....
+//
+// Nodes are included if the tree size includes that power of two. For example,
+// a tree of size 21 is built from subtrees of sizes 16 + 4 + 1, and thus
+// nodes[1] == nodes[3] == nil.
+//
+// For a tree whose size is a perfect power of two, only the last entry in
+// nodes is set, and it also matches the tree root hash.
+func (t *Tree) getNodes() [][]byte {
 	n := make([][]byte, len(t.nodes))
 	copy(n, t.nodes)
 	return n
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -32,8 +32,7 @@ import (
 // TODO(pavelkalinnikov): Remove it, use compact.Range instead.
 type Tree struct {
 	hasher hashers.LogHasher
-	nodes  [][]byte
-	size   int64
+	rng    *Range
 }
 
 func isPerfectTree(size int64) bool {
@@ -56,13 +55,6 @@ type GetNodesFunc func(ids []NodeID) ([][]byte, error)
 // compact Tree. The expectedRoot is the known-good tree root of the tree at
 // the specified size, and is used to verify the initial state.
 func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesFunc, expectedRoot []byte) (*Tree, error) {
-	sizeBits := bits.Len64(uint64(size))
-	r := Tree{
-		hasher: hasher,
-		nodes:  make([][]byte, sizeBits),
-		size:   size,
-	}
-
 	ids := make([]NodeID, 0, bits.OnesCount64(uint64(size)))
 	// Iterate over perfect subtrees along the right border of the tree. Those
 	// correspond to the bits of the tree size that are set to one.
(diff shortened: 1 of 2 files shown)


merkle/compact/tree.go
```diff
@@ -43,21 +43,21 @@ func (r RootHashMismatchError) Error() string {
...
-	return x != 0 && (x&(x-1) == 0)
 }
 
 // GetNodeFunc is a function prototype which can look up particular nodes
 // within a non-compact Merkle tree. Used by the compact Tree to populate
 // itself with correct state when starting up with a non-empty tree.
 type GetNodeFunc func(depth int, index int64) ([]byte, error)
 
+// nodeID identifies a node of a Merkle tree.
+type nodeID struct {
```
Martin2112: This is going to create confusion with storage.NodeID. It's more like the coordinates of a node so could be renamed.

merkle/compact/tree.go
```diff
@@ -71,55 +71,59 @@ type GetNodeFunc func(depth int, index int64) ([]byte, error)
...
-
-	r := Tree{
-		hasher: hasher,
-		nodes:  make([][]byte, sizeBits),
-		root:   hasher.EmptyRoot(),
-		size:   size,
+	ids := make([]nodeID, 0, bits.OnesCount64(uint64(size)))
+	for sz := uint64(size); sz != 0; sz &= sz - 1 { // Iterate 1-bits of size.
+		level := uint(bits.TrailingZeros64(sz))
+		index := (sz - 1) >> level
```
Martin2112: This is the RHS index? Could be renamed for clarity.

merkle/compact/tree.go
```diff
@@ -71,55 +71,59 @@ type GetNodeFunc func(depth int, index int64) ([]byte, error)
...
-				index := size - 1
-				glog.V(1).Infof("fetching d: %d i: %d, leaving size %d", depth, index, size)
-				h, err := f(depth, index)
-				if err != nil {
-					glog.Warningf("Failed to fetch node depth %d index %d: %s", depth, index, err)
-					return nil, err
-				}
-				r.nodes[depth] = h
+		for i, id := range ids {
+			glog.V(1).Infof("Fetching node level %d index %d", id.level, id.index)
```
Martin2112: Reusing `nodeID` struct in this section might be clearer.

merkle/compact/tree.go
```diff
@@ -71,55 +71,59 @@ type GetNodeFunc func(depth int, index int64) ([]byte, error)
...
 // TODO(pavelkalinnikov): Make GetNodeFunc get all nodes at once.
 func NewTreeWithState(hasher hashers.LogHasher, size int64, f GetNodeFunc, expectedRoot []byte) (*Tree, error) {
-	sizeBits := bits.Len64(uint64(size))
-
-	r := Tree{
-		hasher: hasher,
-		nodes:  make([][]byte, sizeBits),
-		root:   hasher.EmptyRoot(),
-		size:   size,
+	ids := make([]nodeID, 0, bits.OnesCount64(uint64(size)))
```
Martin2112: What happens if this is called with size zero? I think it's ok and hasher.EmptyRoot() falls out below?

merkle/compact/tree.go
```diff
@@ -71,55 +71,59 @@ type GetNodeFunc func(depth int, index int64) ([]byte, error)
...
+		root = hasher.EmptyRoot()
+	}
+	t := Tree{hasher: hasher, rng: rng, root: root}
+
+	if !bytes.Equal(root, expectedRoot) {
+		glog.Warningf("Corrupt state, expected root %s, got %s", hex.EncodeToString(expectedRoot), hex.EncodeToString(root))
+		return nil, RootHashMismatchError{ActualHash: root, ExpectedHash: expectedRoot}
+	}
+
+	glog.V(1).Infof("Resuming at size %d, with root: %s", rng.End(), base64.StdEncoding.EncodeToString(root))
```
Martin2112: This looks like a sequencer specific log message. Maybe something about restoring or loading the tree?

merkle/compact/tree.go
```diff
@@ -129,33 +133,31 @@ func (t *Tree) CurrentRoot() []byte {
...
-				newRoot = t.nodes[bit]
-				first = false
-			} else {
-				newRoot = t.hasher.HashChildren(t.nodes[bit], newRoot)
-				if err := f(bit+1, index, newRoot); err != nil {
-					return err
-				}
-			}
+	hash := hashes[ln-1]
+	for i := ln - 2; i >= 0; i-- {
```
Martin2112: Can you document where the `-2` comes from.

merkle/compact/tree.go
```diff
@@ -129,33 +133,31 @@ func (t *Tree) CurrentRoot() []byte {
...
+		size &= size - 1
+		hash = t.hasher.HashChildren(hashes[i], hash)
+		level := bits.TrailingZeros64(size) + 1
+		index := size >> uint(level)
```
Martin2112: Can you document this section of code a bit more. On first reading it looks like there should be a `-1' here.

merkle/compact/tree.go
```diff
@@ -180,87 +182,53 @@ func (t *Tree) AddLeaf(data []byte, f setNodeFunc) (int64, []byte, error) {
...
-			// Don't re-write the leaf hash node (we've done it above already)
-			if bit > 0 {
-				// Store the leaf hash node
-				if err := f(bit, index, hash); err != nil {
-					return 0, err
-				}
-			}
-			return assignedSeq, nil
+	// Store all the new internal nodes.
+	var lastErr error // The last error occurred in Append.
```
Martin2112: Missing 'that' here possibly.

merkle/compact/tree.go
pphaneuf: Is it? It's not exported, so the scope is quite limited, seems fine to me...

Martin2112: I still think renaming it would be clearer.

pavelkalinnikov:
@Martin2112 This is now ready for another review. Commit-by-commit review should work best.

merkle/compact/tree.go
```diff
@@ -78,12 +70,20 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
 	if got, want := len(hashes), len(ids); got != want {
 		return nil, fmt.Errorf("got %d hashes, needed %d", got, want)
 	}
-	for i, id := range ids {
-		r.nodes[id.Level] = hashes[i]
+	// Note: Right border nodes of compact.Range are ordered top-to-bottom.
```
Martin2112: Might be useful to indicate which is the root, given the inconsistency we've had with terms like height and depth.

merkle/compact/tree.go
```diff
@@ -273,7 +201,12 @@ func (t *Tree) hashes() [][]byte {
...
 // nodes is set, and it also matches the tree root hash.
 func (t *Tree) getNodes() [][]byte {
-	n := make([][]byte, len(t.nodes))
-	copy(n, t.nodes)
+	size := t.rng.End()
+	hashes := t.rng.Hashes()
+	n := make([][]byte, bits.Len64(size))
+	for i := len(hashes) - 1; size != 0; i, size = i-1, size&(size-1) {
+		level := bits.TrailingZeros64(size)
+		n[level] = hashes[i]
```
Martin2112: Might need a copy here as it looks like it returns data the Range holds. Possibly that should be inside the Range code depending on expected usage.

pavelkalinnikov: Probably no need to worry about it, as this function is now used only in tests, and I intend to remove it in a follow-up.

Martin2112: OK. We should make sure the Range API is clear on this. We don't want unnecessary copying but we also don't want people breaking it.

Issue #1188, opened 2018-06-01
Cross-transaction subtree caching


rolandshoemaker:
I'm looking into why we are getting linear growth of sequencer latency without any changes in the number of leaves being sequenced. It looks like the only part of the sequencing process that is getting slower as the tree grows is `initMerkleTreeFromStorage` which makes sense, each subsequent run requires retrieving a larger and larger tree (or more accurately set of sub trees).

Originally I was confused as I thought there was persistent caching of subtrees but after looking at the code closer it looks like the existing caching only happens at the [storage.LogTreeTx level](https://github.com/google/trillian/blob/1ac84cfe7b132dc082290d4f0f7c323d552e842c/storage/mysql/log_storage.go#L235). This means that subsequent runs of the signer don't get the benefit of cached subtrees and have to go and build the cache again each time.

Is there a specific reason for not caching this information at a higher level? It seems like it'd be significantly faster to just retrieve whatever new stuff was generated since the last sequencing operation.

Martin2112:
The cache only contains subtrees that will be affected by the subsequent write. So the time shouldn't be increasing linearly because of the subtree packing and I don't think we saw this in testing. Might need more investigation here.

Reading each time ensures that there's a consistent view of the database state. We did consider longer-term caching but went with the simpler approach that was easier to reason about correctness.

rolandshoemaker:
Ah, I think my major mistake here was thinking that the nodes needed to reconstruct the tree were stable, looking closer at the code I now see that's incorrect. It looks like there would be _some_ overlap between signing operations but the required cache size in order to get a viable number of cache hits seems like not a viable trade off.

I'm now even more confused why we are seeing such linear growth (but also why we are getting up to ~1.5 seconds for tree initialization especially since it looks like <10 nodes need to be retrieved). I wonder if the extra latency is just a artifact of retrieving all of the nodes serially instead of in batches (per `Sequencer.buildMerkleTreeFromStorageAtRoot`), I guess if each read took ~150ms this would make sense.

I'll do some experiments and see if I can come back with some more useful information.

Martin2112:
I might look at instrumenting this code so we can see how much work it's doing and get more detail on latency.

We can look at the SQL query plan as well. It's quite an expensive one IIRC and maybe batching it would help.

Martin2112:
Just as a matter of interest how big is your log tree at the moment? Wondering if you've gone past the range we tested with yet.

[...]

PR #1523, opened 2019-04-12
sequencer: Use batch API to initialize compact.Tree

pavelkalinnikov:
This change refactors sequencer so that it uses batch API of tree storage to fetch all nodes necessary to construct `compact.Tree`. This should reduce overhead of each sequencing run.

Addresses #1188

### Checklist


log/sequencer.go
```diff
@@ -130,26 +130,30 @@ func NewSequencer(
...
-		if nodes == nil || len(nodes) != 1 {
-			return nil, fmt.Errorf("%x: Did not retrieve one node while loading compact Merkle tree, got %#v for ID %v@%v", s.signer.KeyHint, nodes, nodeID.String(), root.Revision)
+		if got, want := len(nodes), len(storIDs); got != want {
+			return nil, fmt.Errorf("%x: failed to get %d nodes at rev %d, got %d", s.signer.KeyHint, want, root.Revision, got)
 		}
 
-		return nodes[0].Hash, nil
+		hashes := make([][]byte, len(nodes))
+		for i, node := range nodes {
+			hashes[i] = node.Hash
```
AlCutter: Given the tightening of the `GetMerkleNodes` contract, I think it's probably worth asserting the assumption holds here - the result of it not holding would be catastrophic...

wdyt?

pavelkalinnikov:
There is a pre-existing lack of coverage for `sequencer.go`. I will add some tests in this PR. Meanwhile, PTAL at the implementation.

log/sequencer.go
pavelkalinnikov: Great idea. I think I will sleep better if I do this :)

CHANGELOG.md
```diff
@@ -179,6 +179,10 @@ The `SubtreeCache.GetNodeHash()` method is no longer exported.
 The memory storage provider has been refactored to make it more consistent with
 the other storage providers.
 
+`NodeReader.GetMerkleNodes` now must return `Node` objects in the same order as
```
Martin2112: This was the original intention but the documentation may have accidentally fallen into the sea ...

merkle/compact/tree.go
```diff
@@ -63,27 +69,24 @@ func isPerfectTree(size int64) bool {
...
-
-// getNodeFn will be called a number of times with the coordinates of internal
-// MerkleTree nodes whose hash values are required to initialize the internal
-// state of the compact Tree.  The expectedRoot is the known-good tree root of
-// the tree at size, and is used to verify the correct initial state of the
-// compact Tree after initialisation.
 //
-// TODO(pavelkalinnikov): Make GetNodeFunc get all nodes at once.
-func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodeFn GetNodeFunc, expectedRoot []byte) (*Tree, error) {
+// getNodesFn will be called a number of times with the coordinates of internal
```
Martin2112: Isn't this now called once for all nodes?

CHANGELOG.md
pavelkalinnikov: Yeah, this paragraph is probably redundant. Are you suggesting to drop it?

Martin2112: I think I'd leave it in as I don't think we've documented it properly before.

merkle/compact/tree.go
pavelkalinnikov: Indeed. Updated the comment.

Martin2112:
FYI I don't think this overlaps with my experiment in #1535 but I'm expecting it to speed up compact range hashing.

pavelkalinnikov:
Added tests in #1543, will rebase this one for better coverage.

PR #1524, opened 2019-04-12
Special-case the perfect subtree re-population

AlCutter:
This is a somewhat experimental/proof-of-concept hack to special-case the re-population of full `SubtreeProto`s.

Seems to result in a 400% speedup for full log tiles:

```
    orig BenchmarkLogPopulation-6               1000           2234488 ns/op
    this commit BenchmarkLogPopulation-6        2000            558159 ns/op
```

Not sure this should be merged as is, but suggests that this might be a useful avenue to pursue.
 
### Checklist


storage/cache/log_subtree_cache_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+func buildFullLogSubtree() *storagepb.SubtreeProto {
```
golangcibot: U1000: func `buildFullLogSubtree` is unused (from `unused`)

storage/cache/log_subtree_cache_test.go
```diff
@@ -0,0 +1,116 @@
...
+		Leaves:            make(map[string][]byte),
+		InternalNodeCount: 256 - 1 - 1,
+	}
+	for i := 0; i < 256; i++ {
+		r.Leaves[storage.Suffix{Bits: 8, Path: []byte{byte(i)}}.String()] = []byte(fmt.Sprintf("Leaf Hash %d", i))
+	}
+	return r
+}
+
+func CreateGoldenSubtree(t testing.TB) {
```
golangcibot: U1000: func `CreateGoldenSubtree` is unused (from `unused`)

storage/cache/log_subtree_cache_test.go
```diff
@@ -0,0 +1,116 @@
...
+	st := buildFullLogSubtree()
+	fmt.Printf("no intermediate:\n%s\n", squashSubtree(t, st))
+	LogPopulateFunc(rfc6962.DefaultHasher)(st)
+	fmt.Printf("populated:\n%s\n", squashSubtree(t, st))
+}
+
+func squashSubtree(t testing.TB, st *storagepb.SubtreeProto) string {
```
golangcibot: U1000: func `squashSubtree` is unused (from `unused`)

Martin2112:
Looks like it would also benefit from the Suffix stuff I'm working on.

storage/cache/log_subtree_cache.go
```diff
@@ -59,42 +102,31 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
...
+				return err
 			}
-			seq, err := cmt.AddLeafHash(h, func(height int, index int64, h []byte) error {
-				if height == logStrataDepth && index == 0 {
-					// no space for the root in the node cache
-					return nil
+		} else {
+			// We need to update the subtree root hash regardless of whether it's fully populated
+			for leafIndex := int64(0); leafIndex < int64(len(st.Leaves)); leafIndex++ {
+				nodeID := storage.NewNodeIDFromPrefix(st.Prefix, logStrataDepth, leafIndex, logStrataDepth, maxLogDepth)
```
Martin2112: We should be able to avoid that call to `NewNodeIDFromPrefix` as well? I think it only uses the Suffix.

storage/cache/log_subtree_cache.go
```diff
@@ -59,42 +102,31 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
...
-		for leafIndex := int64(0); leafIndex < int64(len(st.Leaves)); leafIndex++ {
-			nodeID := storage.NewNodeIDFromPrefix(st.Prefix, logStrataDepth, leafIndex, logStrataDepth, maxLogDepth)
-			_, sfx := nodeID.Split(len(st.Prefix), int(st.Depth))
-			sfxKey := sfx.String()
-			h := st.Leaves[sfxKey]
-			if h == nil {
-				return fmt.Errorf("unexpectedly got nil for subtree leaf suffix %s", sfx)
+		// Special case the perfect subtree case for now:
+		if len(st.Leaves) == maxLeaves {
+			if err := populatePerfectLogSubtree(hasher, st); err != nil {
```
Martin2112: Can that not just return it directly?

Martin2112:
That's quite a gain. For the (PR desc) record can you summarize why it's so large?

storage/cache/log_subtree_cache.go
```diff
@@ -35,6 +35,49 @@ func LogPopulateFunc(hasher hashers.LogHasher) storage.PopulateSubtreeFunc {
 	return populateLogSubtreeNodes(hasher)
 }
 
+func suffixKey(d, i int) string {
```
Martin2112: I can add this to my Suffix rework perhaps? Or do you think the logic belongs better here?

storage/cache/log_subtree_cache.go
```diff
@@ -35,6 +35,49 @@ func LogPopulateFunc(hasher hashers.LogHasher) storage.PopulateSubtreeFunc {
...
+	sfx := &storage.Suffix{Bits: byte(d), Path: []byte{byte(i << uint(8-d))}}
+	maskLowBits := (sfx.Bits-1)%8 + 1
+	sfx.Path[0] &= ((0x01 << maskLowBits) - 1) << uint(8-maskLowBits)
+	return sfx.String()
+}
+
+func populatePerfectLogSubtree(hasher hashers.LogHasher, st *storagepb.SubtreeProto) error {
+	if got, want := len(st.Leaves), 256; got != want {
```
Martin2112: Do we have a suitable constant anywhere? Possibly not.

storage/cache/log_subtree_cache.go
```diff
@@ -35,6 +35,49 @@ func LogPopulateFunc(hasher hashers.LogHasher) storage.PopulateSubtreeFunc {
...
+		return fmt.Errorf("subtree is not perfect, tree has %d leaves, expected %d", got, want)
+	}
+	var l, r []byte
+	var ok bool
+	for d := 7; d >= 0; d-- {
+		for i := 0; i < 1<<uint(d); i++ {
+			leftChild := i * 2
+			childDepth := d + 1
```
Martin2112: When I saw that I thought it was going outside the subtree for d=7. Comment about leaves maybe?

storage/cache/log_subtree_cache.go
```diff
@@ -35,6 +35,49 @@ func LogPopulateFunc(hasher hashers.LogHasher) storage.PopulateSubtreeFunc {
...
+			if d == 7 {
+				l, r = st.Leaves[suffixKey(childDepth, leftChild)], st.Leaves[suffixKey(childDepth, leftChild+1)]
+			} else {
+				l, ok = st.InternalNodes[suffixKey(childDepth, leftChild)]
+				if !ok {
+					return fmt.Errorf("internal error, no intermediate node for %d, %d", d, i)
+				}
+				r, ok = st.InternalNodes[suffixKey(childDepth, leftChild+1)]
+				if !ok {
+					return fmt.Errorf("internal error, no intermediate node for %d, %d", d, i)
```
Martin2112: Think this should be `i+1` in the error.

storage/cache/log_subtree_cache_test.go
```diff
@@ -0,0 +1,116 @@
...
...EUJSCfr2wo1314NUpze3szM9jIngYqnGeYI76BWdnP7rmhQjlWka7DjUtt14NJiXMoA0lERDWxQHRUNyZsZS8GZLbO0UatOYyZ0g0VHKu3GglsJF603MDqpthVUPEWuAusu1lkZa2KpkKOjMToh6PVg2AHpwQqbSBpThAoFmM13BVBnO10D1NkCrBHgYszL0AEX2zAysHoxEWo43LW6JZLsDamAG3uLj4I62PyjNM2m4eTk8m5hg8An1SCO1t6AChR8k3BWcjMx5AJzVLGOlHBt/gaV6Q29SPW1C3Pw7KicDJDTow7nCKYetQciWa2x3eR7sRbAXD0r2UmP0XYuKuFsQ7HC6X0w11z9VCC5ZpzeNDp080WzblCqwcZZDuQG4wxHzkHRD9xJdEVoEva2g+xtGYyvYoRWHXVti6SWei1h7ZRwJRfYV+seNMRkaxFPT8aW2x5mbCEv2k/FmK6tTiU12riccV3VqAMq0mpFyM5Oa0oCDNaWfCCNljiZuJrIcHfZTAUjh6EtoOTAbi+TxLq+SQnZPEqMyGGvhS0oUzZAIjO0qvD9NpkCUusmax8ufouZuZjOduVNNgxU4jGYfEVut9YT3M4WWzwaThzr2ICCbWamoHSKVEWxU/Fp4GIxmWcn90aFyPeMjbhIoGzkc96stmOwFw/ai9zwzjqv+TycTVEqMkRXnUgXmneN4C5WF07rzQjJuuw2mqikk83fis9wg4mPgMV2bD4CVpP0m+KpJvBBMxj5crq2pZKPs8226Ygy7laBggqimh6p9LoHJAJi/czVu9VFMYfjkxax5HZlXcctZbSdgR09G8dI1NXEkCP8qOBFDIGT8qPgoNUO4Kt7tHYu+KK30pbw3Cp0tz8QauV2NlszVzU071k0j49Mmy58LOzWWUUOFy/igiXRuUAPk3UQa7LYfQ18nLWsNbjYm4+A5hxtbCX+r7DBcJsuJskFfxRoAUuKlU6zmv9NjEik6GQaVrLdrFVCBewGwY5/MazbmPhuLhI3FyUSX3CtrmbcclpVu4RyryYzgdRYJDYg1ClCfCj0pkj8WTQZm/mgna3cNWXDwqDI9HY6cKjalQlOVuPobdlVOQmcTdop1NHBKpgj6sF68PY3UsnBcm+mrlwbDopewAw66l7EX8GdbuJs+bMmwl+tvTEmUC5tWUp19tf//g4AAP//V1xHUZcWAAA="
```
Martin2112: Might have put these in files in `testdata` but possibly hassle. Could format them as multi line strings though.

Martin2112:
Sorry, the motivation for previous is that if most of the gain came from removing the `NodeID` creation then this would be a simpler intermediate step, combined with not recreating `Suffix` objects.

Martin2112:
I tested a combination of this and #1507 and it seemed to make it slower.

```
goos: linux
goarch: amd64
pkg: github.com/google/trillian/storage/cache
BenchmarkLogPopulation-12    	    2000	    747554 ns/op
```

Martin2112:
Found the issue. My PR was only handling bits = 8 not the 1-8 range. Now it's 10% faster.

```
goos: linux
goarch: amd64
pkg: github.com/google/trillian/storage/cache
BenchmarkLogPopulation-12    	    3000	    516881 ns/op
```

AlCutter:
> Found the issue. My PR was only handling bits = 8 not the 1-8 range. Now it's 10% faster.
> 
> ```
> goos: linux
> goarch: amd64
> pkg: github.com/google/trillian/storage/cache
> BenchmarkLogPopulation-12    	    3000	    516881 ns/op
> ```

\o/

Martin2112:
It's slightly better over more iterations, presumably because of the work done up front. 

```
BenchmarkLogPopulation-12    	  100000	    503885 ns/op
```

PR #1528, opened 2019-04-15
Don't spam the entire team on each pull request

gdbelvin:
CODEOWERS files automatically assign reviewers.
This is useful when open source contributors open pull requests because they cannot assign a reviewer themselves.

However, this has become spamy during the course of normal operation because the entire team is added as a reviewer to every PR.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,5 +5,4 @@
 # https://git-scm.com/docs/gitignore#_pattern_format
 
 # Default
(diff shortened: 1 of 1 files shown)


mhutchinson:
I endorse this change too :-D

RJPercival:
If we want a round-robin style auto-assigner, we could consider https://probot.github.io/apps/auto-assign/.

PR #1533, opened 2019-04-17
Experimental LogHasher API change for performance.

Martin2112:
Allows hash buffers to be reused at the cost of thread safety and the
need to make copies in some places. Apply this to the compact merkle
tree to demonstrate the gains.

Performance improvement is noticeable

```
BenchmarkAddLeaf-12    	  200000	     15023 ns/op
BenchmarkAddLeaf-12    	  200000	     11392 ns/op
```

Also memory profile showed ~20% reduction in memory allocation for the
benchmark run.

```
Showing nodes accounting for 717.58MB, 96.48% of 743.73MB total
Dropped 36 nodes (cum <= 3.72MB)
Showing top 10 nodes out of 32
      flat  flat%   sum%        cum   cum%
  419.06MB 56.35% 56.35%   419.06MB 56.35%  crypto/internal/boring.(*sha256Hash).sum
     150MB 20.17% 76.51%      150MB 20.17%  crypto/internal/boring.(*sha256Hash).Write.func1
   73.51MB  9.88% 86.40%    73.51MB  9.88%  crypto/internal/boring.NewSHA256
      13MB  1.75% 88.15%       14MB  1.88%  testing.(*common).Helper
      13MB  1.75% 89.89%       13MB  1.75%  fmt.Sprintf
      13MB  1.75% 91.64%   432.06MB 58.09%  crypto/internal/boring.(*sha256Hash).Sum
   12.50MB  1.68% 93.32%   449.05MB 60.38%  github.com/google/trillian/merkle/compact.BenchmarkAddLeaf
      10MB  1.34% 94.67%    24.50MB  3.29%  github.com/google/trillian/merkle/compact.(*tree).verifyRange
       7MB  0.94% 95.61%    50.01MB  6.72%  github.com/google/trillian/merkle/rfc6962.(*Hasher).HashChildren
    6.50MB  0.87% 96.48%   475.05MB 63.87%  github.com/google/trillian/merkle/compact.(*Tree).AddLeafHash
```

### Checklist

### DIFF
--- a/client/log_client_test.go
+++ b/client/log_client_test.go
@@ -308,7 +308,7 @@ func TestAddSequencedLeaves(t *testing.T) {
 		}, wantErr: true},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
-			c := &LogClient{LogVerifier: &LogVerifier{Hasher: rfc6962.DefaultHasher}}
+			c := &LogClient{LogVerifier: &LogVerifier{Hasher: rfc6962.NewSHA256()}}
 			err := c.AddSequencedLeaves(ctx, tc.dataByIndex)
 			if gotErr := err != nil; gotErr != tc.wantErr {
 				t.Errorf("AddSequencedLeaves(): %v, wantErr: %v", err, tc.wantErr)
--- a/client/log_verifier_test.go
+++ b/client/log_verifier_test.go
@@ -54,7 +54,7 @@ func TestVerifyRootErrors(t *testing.T) {
 		{desc: "trustedNil", trusted: nil, newRoot: signedRoot},
 	}
 	for _, test := range tests {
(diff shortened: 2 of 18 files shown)


Martin2112:
OK it doesn't work with mutexes and after removing DefaultHasher so there probably is a bug that's missed by the unit tests.

Martin2112:
Now it's fixed up the full extent of the impact is visible. Quite a few places use `DefaultHasher` which is not shareable with the new API. In `tree_hasher.go` the `NewLogHasher` doesn't actually create new instances so code that uses this doesn't work with the new API.

Having gone through all this it might be better to split the new API off to a new type and make the usage requirements / contract for it explicit. This would reduce the impact of the change quite a bit.

Martin2112:
Closing this one as I think #1535 is definitely cleaner.

PR #1534, opened 2019-04-17
Logic bug

gdbelvin:
Fix logic bug caused by variable shadowing.


Martin2112:
Can you make PR titles more descriptive please?

Martin2112:
It looks like the linter version bump might not be included intentionally?

gdbelvin:
This logic bug is caught by the latest version of the linter, which is why
it's included in this PR

Martin2112:
Which is why documenting the intention of the PR is helpful for reviewers.

PR #1535, opened 2019-04-23
Experimental hasher API change for performance improvement (2nd time)

Martin2112:
Improved version of #1533. This time the API change and caller contract change are explicit and the set of changes is much less intrusive as the old objects remain stateless. Added some docs to existing API.

Apply the new APIs to compact range (via `hash_chainer`) to reduce memory allocation.

The current hasher APIs are constantly allocating slices and copying hashes into them when we often have a chain of hashes like in hash_chainer that can be done in place. The benchmarks with compact merkle tree showed about 25% improvement in time and memory allocation. Some numbers in the description for #1533. I'd expect similar gains for compact range as it uses repeated hashing chains.

### Checklist

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -48,7 +48,7 @@ type NodeID struct {
 //
 // TODO(pavelkalinnikov): Remove it, use compact.Range instead.
 type Tree struct {
-	hasher hashers.LogHasher
+	hasher hashers.InplaceLogHasher
 	root   []byte
 	// The list of "dangling" left-hand nodes, where entry [0] is the leaf.
 	// So: nodes[0] is the hash of a subtree of size 1 = 1<<0, if included.
@@ -84,12 +84,12 @@ type GetNodesFunc func(ids []NodeID) ([][]byte, error)
 // whose hash values are required to initialize the internal state of the
 // compact Tree. The expectedRoot is the known-good tree root of the tree at
 // the specified size, and is used to verify the initial state.
-func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesFunc, expectedRoot []byte) (*Tree, error) {
+func NewTreeWithState(hasher hashers.InplaceLogHasher, size int64, getNodesFn GetNodesFunc, expectedRoot []byte) (*Tree, error) {
 	sizeBits := bits.Len64(uint64(size))
 	r := Tree{
 		hasher: hasher,
 		nodes:  make([][]byte, sizeBits),
-		root:   hasher.EmptyRoot(),
+		root:   hasher.EmptyRootInto(nil),
 		size:   size,
 	}
 
@@ -128,10 +128,10 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
 }
 
 // NewTree creates a new compact Tree with size zero.
-func NewTree(hasher hashers.LogHasher) *Tree {
+func NewTree(hasher hashers.InplaceLogHasher) *Tree {
 	return &Tree{
 		hasher: hasher,
-		root:   hasher.EmptyRoot(),
+		root:   hasher.EmptyRootInto(nil),
 		nodes:  make([][]byte, 0),
 		size:   0,
 	}
@@ -168,7 +168,7 @@ func (t *Tree) recalculateRoot(setNodeFn setNodeFunc) error {
 
 	index := t.size
 
-	var newRoot []byte
+	newRoot := make([]byte, t.hasher.Size())
 	first := true
 	mask := int64(1)
 	numBits := bits.Len64(uint64(t.size))
@@ -176,11 +176,12 @@ func (t *Tree) recalculateRoot(setNodeFn setNodeFunc) error {
 		index >>= 1
 		if t.size&mask != 0 {
 			if first {
-				newRoot = t.nodes[bit]
+				copy(newRoot, t.nodes[bit])
 				first = false
 			} else {
-				newRoot = t.hasher.HashChildren(t.nodes[bit], newRoot)
-				if err := setNodeFn(bit+1, index, newRoot); err != nil {
+				newRoot := t.hasher.HashChildrenInto(t.nodes[bit], newRoot, newRoot)
+				// Must make a copy in case the setNodeFn retains the value.
+				if err := setNodeFn(bit+1, index, append(make([]byte, 0, len(newRoot)), newRoot...)); err != nil {
 					return err
 				}
 			}
@@ -200,7 +201,7 @@ func (t *Tree) recalculateRoot(setNodeFn setNodeFunc) error {
 // Returns the index of the new leaf (equal to t.Size()-1) and the Merkle leaf
 // hash for the new leaf.
 func (t *Tree) AddLeaf(data []byte, setNodeFn setNodeFunc) (int64, []byte, error) {
-	h, err := t.hasher.HashLeaf(data)
+	h, err := t.hasher.HashLeafInto(data, nil)
 	if err != nil {
 		return 0, nil, err
 	}
@@ -237,8 +238,8 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 		return assignedSeq, nil
 	}
 
-	// Initialize our running hash value to the leaf hash.
-	hash := leafHash
+	// Initialize our running hash value to a copy of the leaf hash
+	hash := append(make([]byte, 0, len(leafHash)), leafHash...)
 	bit := 0
 	// Iterate over the bits in our existing tree size.
 	for mask := t.size; mask > 0; mask >>= 1 {
@@ -256,9 +257,9 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 			return assignedSeq, nil
 		}
 		// The bit is set so we have a node at that position in the nodes list so hash it with our running hash:
-		hash = t.hasher.HashChildren(t.nodes[bit], hash)
-		// Store the resulting parent hash.
-		if err := setNodeFn(bit+1, index, hash); err != nil {
+		hash := t.hasher.HashChildrenInto(t.nodes[bit], hash, hash)
+		// Store the resulting parent hash. Make a copy in case the function retains it.
+		if err := setNodeFn(bit+1, index, append(make([]byte, 0, len(hash)), hash...)); err != nil {
 			return 0, err
 		}
 		// Now, clear this position in the nodes list as the hash it formerly contained will be propagated upwards.
@@ -272,6 +273,7 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 		} else if mask&0x02 == 0 {
 			// If the node above us is unused at this tree size, then store our
 			// running hash there, and we're done.
+			// No need to copy the value here as we won't modify it again.
 			t.nodes[bit+1] = hash
 			return assignedSeq, nil
 		}
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -257,7 +257,7 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 			return assignedSeq, nil
 		}
 		// The bit is set so we have a node at that position in the nodes list so hash it with our running hash:
-		hash := t.hasher.HashChildrenInto(t.nodes[bit], hash, hash)
+		hash = t.hasher.HashChildrenInto(t.nodes[bit], hash, hash)
 		// Store the resulting parent hash. Make a copy in case the function retains it.
 		if err := setNodeFn(bit+1, index, append(make([]byte, 0, len(hash)), hash...)); err != nil {
 			return 0, err
--- a/merkle/hash_chainer.go
+++ b/merkle/hash_chainer.go
@@ -24,7 +24,7 @@ import (
 //
 // TODO(pavelkalinnikov): Add a Merkle Trees doc with visual explanations.
 type hashChainer struct {
-	hasher hashers.LogHasher
+	hasher hashers.InplaceLogHasher
 }
 
 // chainInner computes a subtree hash for a node on or below the tree's right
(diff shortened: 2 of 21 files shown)


Martin2112:
The race condition is because the CT integration test shares a `LogVerifier` between goroutines.

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
 	Size() int
 }
 
+// InplaceLogHasher provides the hash functions needed to compute dense merkle
```
pavelkalinnikov: nit: Merkle

Maybe also update other occurrences in this file.

pavelkalinnikov:
Could you outline the main idea of the optimization in the PR description?

Martin2112:
Yeah I didn't pull over some of the stuff from the first attempt in #1533. The hasher APIs are constantly allocating slices and copying hashes into them when we often have a chain of hashes like in `hash_chainer` that can be done in place. The benchmarks with compact merkle tree showed about 25% improvement in time and memory allocation.

Martin2112:
Might keep the old API for use by clients as it's harder to get it wrong. Removing it is harder because of the cross repo dependencies. This was really an experiment to scope it out.

merkle/compact/tree.go
```diff
@@ -163,19 +163,20 @@ func (t *Tree) recalculateRoot(setNodeFn setNodeFunc) error {
...
 		index >>= 1
 		if t.size&mask != 0 {
 			if first {
-				newRoot = t.nodes[bit]
+				copy(newRoot, t.nodes[bit])
 				first = false
 			} else {
-				newRoot = t.hasher.HashChildren(t.nodes[bit], newRoot)
-				if err := setNodeFn(bit+1, index, newRoot); err != nil {
+				newRoot := t.hasher.HashChildrenInto(t.nodes[bit], newRoot, newRoot)
```
pavelkalinnikov: `HashChildrenInto` reads and writes to `newRoot` at the same time. Should it be specified in the interface's contract that such pattern is safe? (also applies below, and to other methods)

merkle/hashers/tree_hasher.go
```diff
@@ -93,7 +137,18 @@ func NewLogHasher(h trillian.HashStrategy) (LogHasher, error) {
 	return nil, fmt.Errorf("LogHasher(%s) is an unknown hasher", h)
 }
 
+// NewInplaceLogHasher returns an InplaceLogHasher.
```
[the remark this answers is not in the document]
Martin2112: No. It does create new instances.

merkle/rfc6962/rfc6962_test.go
```diff
@@ -196,5 +198,14 @@ func TestHashChildrenEquivToOld(t *testing.T) {
 			t.Errorf("%d different hashes: %x vs %x", i, oldHash, newHash)
 		}
 	}
+}
 
+// Tests that the rfc6962 log hashers were registered.
+func TestRegistrations(t *testing.T) {
```
pavelkalinnikov: TODO adding other hashers to this test in follow-ups?

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+// Implementations may hold state and are not expected to be thread
+// safe. They should not be shared between multiple callers. Where performance
+// is not critical the LogHasher interface provides a safer alternative.
+//
+// Note: Implementations are not expected to be thread safe. The caller must
+// ensure that concurrent calls are not made and is responsible for making
+// copies of slices where necessary to avoid data being overwritten.
+type InplaceLogHasher interface {
+	// EmptyRootInto supports returning a special case for the root of an empty tree.
+	// The passed in slice is mutated. The caller is responsible for making copies
```
pavelkalinnikov: nit: I feel like the "copies" sentence could be omitted.

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+// InplaceLogHasher provides the hash functions needed to compute dense merkle
+// trees.
+// Implementations may hold state and are not expected to be thread
+// safe. They should not be shared between multiple callers. Where performance
+// is not critical the LogHasher interface provides a safer alternative.
+//
+// Note: Implementations are not expected to be thread safe. The caller must
+// ensure that concurrent calls are not made and is responsible for making
+// copies of slices where necessary to avoid data being overwritten.
+type InplaceLogHasher interface {
```
pavelkalinnikov: Maybe say somewhere that capacity of `res` slices should be at least `Size()`?

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+	// EmptyRootInto supports returning a special case for the root of an empty tree.
```
pavelkalinnikov: 2 sentences could be shortened to:
`// EmptyRootInto writes a root hash of an empty tree to the passed in slice.`

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+	// The passed in slice is mutated. The caller is responsible for making copies
+	// where necessary. Not thread safe.
+	EmptyRootInto([]byte) []byte
+	// HashLeafInto computes the hash of a leaf into an existing slice,
+	// which is mutated. The leaf and res slices may be the same. Not thread safe.
```
pavelkalinnikov: After a second look:

In `HashChildrenInto`, it feels natural to allow `res == l/r` as they are the same "hash" type, and hence all have `cap(...) >= Size()`.

It feels less natural to allow `res == leaf` in `HashLeafInto` though, as they represent different types of data. Moreover, it's more error-prone, as the caller must ensure `cap(leaf) >= h.Size()`, even if `len(leaf) < h.Size()` -- quite an unnatural thing to do, because whoever produces the `leaf` slice might be not aware of the hasher.

I suggest removing this sentence from `HashLeafInto`, to not encourage such use-case.

merkle/compact/tree.go
Martin2112: Clarified docs.

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+	HashLeafInto(leaf, res []byte) ([]byte, error)
+	// HashChildrenInto computes interior nodes into an existing slice, which is
+	// mutated. The res slice may be the same as one of the l or r slices. Not
```
pavelkalinnikov: ignore: I wonder what would happen if `res` was not equal to either `l` or `r`, but overlapped with one or both. I don't think it makes sense to clarify it in the comment though (no need to make this contract any stronger).

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+	// thread safe.
+	HashChildrenInto(l, r, res []byte) []byte
+	// Size is the number of bytes in the underlying hash function.
+	// TODO(gbelvin): Replace Size() with BitLength().
```
pavelkalinnikov: For this particular hasher, the `Size()` method makes more sense than `BitLength()`, as it is used for sizing `res` slices. Maybe remove this TODO?

merkle/hashers/tree_hasher.go
```diff
@@ -33,6 +35,32 @@ type LogHasher interface {
...
+// is not critical the LogHasher interface provides a safer alternative.
+//
+// Note: Implementations are not expected to be thread safe. The caller must
+// ensure that concurrent calls are not made and is responsible for making
+// copies of slices where necessary to avoid data being overwritten.
+type InplaceLogHasher interface {
+	// EmptyRootInto supports returning a special case for the root of an empty tree.
+	// The passed in slice is mutated. The caller is responsible for making copies
+	// where necessary. Not thread safe.
+	EmptyRootInto([]byte) []byte
```
pavelkalinnikov: Maybe name the output parameter `res`, for consistency with the other methods?

merkle/hashers/tree_hasher.go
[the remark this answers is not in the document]
pavelkalinnikov: It takes one from a hashmap, doesn't it?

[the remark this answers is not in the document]
pavelkalinnikov: Oh wait, I'm wrong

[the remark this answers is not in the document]
Martin2112: Yeah the others happened to be stateless so it didn't matter they were shared. But if somebody does register a `LogHasher` that isn't then it will break.

merkle/rfc6962/rfc6962_test.go
Martin2112: These are the ones registered by the code being tested. Others should be tested in an appropriate place.

merkle/hashers/tree_hasher.go
Martin2112: They don't have to be. Tests added for that case.

merkle/hashers/tree_hasher.go
Martin2112: Many things could be done. I'm not sure that wording is an improvement though.

merkle/hashers/tree_hasher.go
Martin2112: It's not necessary that the capacity is large enough as the value is produced by truncate followed by append. Test added to demonstrate.

I'm going to leave the doc in as it does work though it might not be a great idea to do it.

merkle/hashers/tree_hasher.go
Martin2112: Because the source is written to the hasher first before the result is truncated it should at least produce the right answer, not that this is a sensible thing to do.

merkle/hashers/tree_hasher.go
Martin2112: Out of scope for this PR. Probably applies to MapHasher as well. Might want to have a word with Gary.

merkle/hashers/tree_hasher.go
Martin2112: Done.

merkle/hashers/tree_hasher.go
pavelkalinnikov: Ah, indeed, Go can re-allocate slices on `append` under the hood.

merkle/compact/tree.go
```diff
@@ -237,8 +238,8 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 		return assignedSeq, nil
 	}
 
-	// Initialize our running hash value to the leaf hash.
-	hash := leafHash
+	// Initialize our running hash value to a copy of the leaf hash
+	hash := append(make([]byte, 0, len(leafHash)), leafHash...)
```
pavelkalinnikov: `append([]byte(nil), leafHash...)`

merkle/compact/tree.go
```diff
@@ -256,9 +257,9 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
 			return assignedSeq, nil
 		}
 		// The bit is set so we have a node at that position in the nodes list so hash it with our running hash:
-		hash = t.hasher.HashChildren(t.nodes[bit], hash)
-		// Store the resulting parent hash.
-		if err := setNodeFn(bit+1, index, hash); err != nil {
+		hash := t.hasher.HashChildrenInto(t.nodes[bit], hash, hash)
```
pavelkalinnikov: It seems that physically `:=` doesn't change the algorithm (only luckily, because the `hash` size doesn't change), but logically it makes less sense. This line depends on the outer `hash` left from the previous iteration, but it doesn't update it for the next iteration explicitly. I suggest leaving `=`.

merkle/compact/tree.go
```diff
@@ -256,9 +257,9 @@ func (t *Tree) AddLeafHash(leafHash []byte, setNodeFn setNodeFunc) (int64, error
...
+		// Store the resulting parent hash. Make a copy in case the function retains it.
+		if err := setNodeFn(bit+1, index, append(make([]byte, 0, len(hash)), hash...)); err != nil {
```
pavelkalinnikov: Aside: What's the advantage of using the inline hasher here, given that we are still making copies on every iteration? Is it mostly the shared internal pre-image buffer? I wonder if a regular non-thread-safe `LogHasher` implementation that kept a similar pre-image buffer could do the same job.

pavelkalinnikov: `append([]byte(nil), hash...)`

merkle/hash_chainer.go
```diff
@@ -24,41 +24,47 @@ import (
...
-	return seed
+	return res
 }
 
 // chainInnerRight computes a subtree hash like chainInner, but only takes
 // hashes to the left from the path into consideration, which effectively means
 // the result is a hash of the corresponding earlier version of this subtree.
 func (c hashChainer) chainInnerRight(seed []byte, proof [][]byte, index int64) []byte {
+	// Make a copy to begin with, which will be reused throughout.
+	res := append(make([]byte, 0, len(seed)), seed...)
```
pavelkalinnikov: `append([]byte(nil), seed...)`

merkle/compact/tree.go
Martin2112: While it would work the same I think using an empty slice is slightly clearer than a nil one.

merkle/compact/tree.go
Martin2112: That was unintended. I'll fix it.

merkle/compact/tree.go
Martin2112: IIRC it's the sum of a shared hash buffer + not allocating a new result slice. It's pretty clear in the memory profiling results.

pavelkalinnikov: Yeah, so what I am saying is that `InlineLogHasher` does eliminate allocating a new result slice, but we are still allocating one in this line (to make a copy). So the only advantage that remains is the shared hash buffer, isn't it? But for that we don't strictly need a new interface, the buffer can be hidden behind another `LogHasher` implementation.

merkle/compact/tree.go
pavelkalinnikov: Dunno, leaving up to to. To me the shorted the better. Also, consider [this](https://github.com/go101/go101/wiki/How-to-perfectly-clone-a-slice%3F) article, there is even a third way apparently.

Martin2112:
Going to close this. I think the improvement isn't worth the disruption, given the other performance improvements that have been made.

PR #1539, opened 2019-04-25
Remove unnecessary type

pavelkalinnikov:
### Checklist

### DIFF
--- a/storage/testonly/fake_node_reader.go
+++ b/storage/testonly/fake_node_reader.go
@@ -30,13 +30,6 @@ import (
 // to test code reading from multiple tree revisions. It cannot live in the main testonly
 // package as this creates import cycles.
 
-// NodeMapping is a struct we use because we can't use NodeIDs as map keys. Callers pass this
-// and FakeNodeReader internally manages derived keys.
-type NodeMapping struct {
-	NodeID storage.NodeID
-	Node   storage.Node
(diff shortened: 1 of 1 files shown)


storage/testonly/fake_node_reader.go
```diff
@@ -46,21 +39,19 @@ type FakeNodeReader struct {
...
 	nodeMap := make(map[string]storage.Node)
 
-	for _, mapping := range mappings {
-		_, ok := nodeMap[mapping.NodeID.String()]
-
-		if ok {
-			// Duplicate mapping - the test data is invalid so don't continue
-			glog.Fatalf("NewFakeNodeReader duplicate mapping for: %s in:\n%v", mapping.NodeID.String(), mappings)
+	for _, node := range nodes {
+		id := node.NodeID.String()
```
Martin2112: It's generally not safe to rely on string representations as IDs, which is why that mapping exists. As this is test code it might be acceptable.

pavelkalinnikov: The previous code also used the `String()` representation as a hashmap key, this PR doesn't change that.

Martin2112: Yeah. It wasn't a great idea then either. I suppose we do at least define String() on this to return the binary value.

PR #1540, opened 2019-04-25
Delete LogMetadata.GetUnsequencedCounts()

RJPercival:
It is unused and the documentation warns against using it.

### Checklist


Martin2112:
It was used in tests I believe at one point. Is that no longer the case?

It should still be removed as it probably can't be efficiently implemented on a production database.

RJPercival:
This PR deletes all references to it, so it appears that it isn't used in tests (other than its own unit tests).

PR #1542, opened 2019-04-25
Postgres tests should not fail if a Postgres server is not available

RJPercival:
The MySQL tests pass if a MySQL server is not installed, and the Postgres tests should behave similarly. Ideally, both sets of tests would mark themselves as skipped using `t.Skip()`, but that would require more extensive changes.

### Checklist

### DIFF
--- a/storage/postgres/storage_test.go
+++ b/storage/postgres/storage_test.go
@@ -38,6 +38,5 @@ func TestMain(m *testing.M) {
 	defer cancel()
 	db = testdb.NewTrillianDBOrDie(ctx)
 	defer db.Close()
-	ec := m.Run()
-	os.Exit(ec)
+	os.Exit(m.Run())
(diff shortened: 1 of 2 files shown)


storage/postgres/storage_test.go
```diff
@@ -30,16 +30,14 @@ var db *sql.DB
...
 		glog.Errorf("PG not available, skipping all PG storage tests")
-		ec = 1
 		return
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(time.Second*30))
 	defer cancel()
 	db = testdb.NewTrillianDBOrDie(ctx)
 	defer db.Close()
-	ec = m.Run()
+	ec := m.Run()
```
AlCutter: `os.Exit(m.Run())`?

PR #1543, opened 2019-04-26
sequencer: Add golden test for non-perfect tree update

pavelkalinnikov:
This change improves coverage of `sequencer.go` by exercising `IntegrateBatch`
on a non-perfect tree.

### Checklist

### DIFF
--- a/log/sequencer_test.go
+++ b/log/sequencer_test.go
@@ -307,6 +307,9 @@ func createTestContext(ctrl *gomock.Controller, params testParameters) (testCont
 			ids := []storage.NodeID{node.NodeID}
 			nodes := []storage.Node{node}
 			mockTx.EXPECT().GetMerkleNodes(gomock.Any(), params.writeRevision-1, ids).Return(nodes, params.merkleNodesGetError)
+			if params.merkleNodesGetError != nil {
+				break
+			}
 		}
 	}
 
@@ -466,6 +469,20 @@ func TestIntegrateBatch(t *testing.T) {
 				skipStoreSignedRoot:   true,
 			},
 			errStr: "root",
+		},
+		{
+			desc: "get-merkle-nodes-fails",
+			params: testParameters{
+				logID:               154035,
+				writeRevision:       int64(testRoot21.Revision + 1),
+				dequeueLimit:        1,
+				dequeuedLeaves:      []*trillian.LogLeaf{getLeaf42()},
+				latestSignedRoot:    testSignedRoot21,
+				merkleNodesGet:      &compactTree21,
+				merkleNodesGetError: errors.New("getmerklenodes"),
+				skipStoreSignedRoot: true,
+			},
+			errStr: "getmerklenodes",
 		},
 		{
 			desc: "update-seq-leaves-fails",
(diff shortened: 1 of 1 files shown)


log/sequencer_test.go
```diff
@@ -109,6 +116,65 @@ var (
...
+			Hash:         testonly.MustDecodeBase64("BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB="),
+			NodeRevision: 6,
+		},
+		{
+			NodeID:       storage.NodeID{Path: []uint8{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}, PrefixLenBits: 60},
+			Hash:         testonly.MustDecodeBase64("CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC="),
+			NodeRevision: 6,
+		},
+	}
+	// Nodes that will be storead after updating the tree of size 21.
```
AlCutter: /storead/stored/

log/sequencer_test.go
```diff
@@ -109,6 +116,65 @@ var (
...
+	updatedNodes21 = []storage.Node{
+		{
+			NodeID:       storage.NodeID{Path: []uint8{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}, PrefixLenBits: 59},
+			Hash:         testonly.MustDecodeBase64("1oUtLDlyOWXLHLAvL3NvWaO4D9kr0oQYScylDlgjey4="),
+			NodeRevision: 6,
```
AlCutter: Should these `NodeRevision` be 7?

log/sequencer_test.go
```diff
@@ -109,6 +116,65 @@ var (
...
+			NodeID:       storage.NodeID{Path: []uint8{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x15}, PrefixLenBits: 64},
+			Hash:         testonly.MustDecodeBase64("L5Iyd7aFOVewxiRm29xD+EU+jvEo4RfufBijKdflWMk="),
+			NodeRevision: 6,
+		},
+	}
+	// The new root after updating the tree of size 21.
+	updatedRoot21 = &types.LogRootV1{
+		RootHash:       testonly.MustDecodeBase64("1oUtLDlyOWXLHLAvL3NvWaO4D9kr0oQYScylDlgjey4="),
+		TimestampNanos: uint64(fakeTimeForTest.UnixNano()),
+		Revision:       6,
```
AlCutter: 7 here too?

log/sequencer_test.go
pavelkalinnikov: I think it's rather `compactTree21` should be revision 5, because the transaction gets from 5 to 6.

log/sequencer_test.go
pavelkalinnikov: Nah, `testRoot21` is 5, so this one should be 6?

Martin2112:
Al got in first.

log/sequencer_test.go
pavelkalinnikov: I tried changing these to 7 locally, and got an error. As expected.

pavelkalinnikov:
It works :) PTAL another time.

pavelkalinnikov:
@Martin2112 LGTyou?

Martin2112:
I think so.

pavelkalinnikov:
Could you approve? It would unblock rebasing #1523. (that one passes these tests as well, btw)

Martin2112:
I did approve, your push after that must have turned it off.

PR #1544, opened 2019-04-26
Propagate Log Update Failures in AddSequencedLeaves

therealdrake:
AddSequencedLeaves in the log client previously did not evaluate the response proto. Adding some fail case checking and logging.

### DIFF
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -316,7 +316,7 @@ func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int6
 	})
 	for _, leaf := range resp.GetResults() {
 		if stat := leaf.GetStatus().GetCode(); stat != int32(codes.OK) && stat != int32(codes.AlreadyExists) {
-			return fmt.Errorf("unexpected fail status in AddSequencedLeaves: %+v", leaf)
+			return status.Errorf(s.Code(), "unexpected fail status in AddSequencedLeaves: %+v, err: %v", leaf, s.Message())
 		}
 	}
 	return err
--- a/client/log_client.go
+++ b/client/log_client.go
@@ -315,7 +315,7 @@ func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int6
 		Leaves: leaves,
 	})
 	for _, leaf := range resp.GetResults() {
-		if stat := leaf.GetStatus().GetCode(); stat != int32(codes.OK) && stat != int32(codes.AlreadyExists) {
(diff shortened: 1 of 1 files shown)


client/log_client.go
```diff
@@ -399,10 +399,15 @@ func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int6
 		leaf.LeafIndex = index
 		leaves = append(leaves, leaf)
 	}
-	_, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
+	resp, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
 		LogId:  c.LogID,
 		Leaves: leaves,
 	})
+	for _, leaf := range resp.GetResults() {
+		if stat := leaf.GetStatus().GetCode(); stat != int32(codes.OK) && stat != int32(codes.AlreadyExists) {
```
gdbelvin: It might be easier to use the [gRPC status](https://github.com/grpc/grpc-go/blob/master/status/status.go#L119) helper functions here:

```suggestion
		if s := status.FromProto(leaf.GetStatus()); s.Code() != codes.OK && s.Code() != codes.AlreadyExists {
```

client/log_client.go
```diff
@@ -399,10 +399,15 @@ func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int6
...
+			return fmt.Errorf("unexpected fail status in AddSequencedLeaves: %+v", leaf)
```
gdbelvin: Might be a good idea to also propagate the error itself:
```suggestion
			return status.Errorf(s.Code(), "unexpected fail status in AddSequencedLeaves: %+v, err: %v", leaf, s.Message())
```

AlCutter:
wdyt @Martin2112 ?

client/log_client.go
```diff
@@ -399,10 +399,15 @@ func (c *LogClient) AddSequencedLeaves(ctx context.Context, dataByIndex map[int6
 		leaf.LeafIndex = index
 		leaves = append(leaves, leaf)
 	}
-	_, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
+	resp, err := c.client.AddSequencedLeaves(ctx, &trillian.AddSequencedLeavesRequest{
 		LogId:  c.LogID,
 		Leaves: leaves,
 	})
+	for _, leaf := range resp.GetResults() {
```
Martin2112: It looks like the client doesn't get to find out which leaves already exist. Maybe that doesn't matter.

mhutchinson (approved):

PR #1546, opened 2019-04-29
Add last_sth_timestamp to log signer

gdbelvin:
### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -53,6 +53,7 @@ var (
 	seqStoreRootLatency    monitoring.Histogram
 	seqCounter             monitoring.Counter
 	seqMergeDelay          monitoring.Histogram
+	seqTimestamp           monitoring.Gauge
 
 	// QuotaIncreaseFactor is the multiplier used for the number of tokens added back to
 	// sequencing-based quotas. The resulting PutTokens call is equivalent to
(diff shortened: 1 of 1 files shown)


log/sequencer.go
```diff
@@ -80,6 +81,7 @@ func createMetrics(mf monitoring.MetricFactory) {
 	quota.InitMetrics(mf)
 	seqBatches = mf.NewCounter("sequencer_batches", "Number of sequencer batch operations", logIDLabel)
 	seqTreeSize = mf.NewGauge("sequencer_tree_size", "Size of Merkle tree", logIDLabel)
+	seqTimestamp = mf.NewGauge("sequencer_timestamp", "Time of last STH in ms since epoch", logIDLabel)
```
Martin2112: Naming the new one `sequencer_tree_timestamp` would indicate that they're related.

Doc comment should probably say "last STH produced" or "signed" for clarity. Might also want to document both of those as only valid if master for that tree.

pavelkalinnikov:
Is "STH" purely a CT term? Rename it to Trillian terms if so.

log/sequencer.go
gdbelvin: Added documentation suffix: "last SLR signed". 

All the metrics in this file are only valid for when master so I've omitted that clarification for consistency.

log/sequencer.go
```diff
@@ -417,13 +419,15 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
...
 		// Create the log root ready for signing
-		seqTreeSize.Set(float64(merkleTree.Size()), label)
 		newLogRoot = &types.LogRootV1{
 			RootHash:       merkleTree.CurrentRoot(),
 			TimestampNanos: uint64(s.timeSource.Now().UnixNano()),
 			TreeSize:       uint64(merkleTree.Size()),
 			Revision:       uint64(newVersion),
 		}
+		seqTreeSize.Set(float64(newLogRoot.TreeSize), label)
+		seqTimestamp.Set(float64(time.Duration(newLogRoot.TimestampNanos)*time.Nanosecond/
```
Martin2112: `TimestampNanos` is already in nanos and `Duration` is documented as a nanosecond count. So you can remove the multiplier if you like.

PR #1550, opened 2019-05-01
sequencer: Clean up compact Tree updates

pavelkalinnikov:
This change cleans up compact Tree update code:
- Timestamps / merge delay computation is factored out into a separate function.
- Node ID computation and hashmap updating is deduplicated.
- Use compact node IDs as hashmap keys.

The last one also saves some time/memory because we stop converting NodeID to string.

### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -209,6 +209,12 @@ func (s Sequencer) prepareLeaves(leaves []*trillian.LogLeaf, begin int64, label 
 }
 
 func (s Sequencer) updateCompactTree(mt *compact.Tree, leaves []*trillian.LogLeaf, label string) (map[compact.NodeID][]byte, error) {
+	// TODO(pavelkalinnikov): Move this call to IntegrateBatch when gocyclo is
+	// happy with the function complexity.
+	if err := s.prepareLeaves(leaves, mt.Size(), label); err != nil {
+		return nil, err
+	}
+
 	nodeMap := make(map[compact.NodeID][]byte)
 	store := func(level uint, index uint64, hash []byte) {
 		nodeMap[compact.NodeID{Level: level, Index: index}] = hash
@@ -401,9 +407,6 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 		}
 
 		// Collate node updates.
-		if err := s.prepareLeaves(sequencedLeaves, merkleTree.Size(), label); err != nil {
-			return err
-		}
 		nodeMap, err := s.updateCompactTree(merkleTree, sequencedLeaves, label)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
PT al@
Best reviewed commit by commit.

log/sequencer.go
```diff
@@ -401,6 +409,9 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 		}
 
 		// Collate node updates.
+		if err := s.prepareLeaves(sequencedLeaves, merkleTree.Size(), label); err != nil {
```
pavelkalinnikov: I wonder if putting this before `st.update` line is better (because `timeSource.Now()` will be slightly more ahead, and merge delay more pessimistic/realistic).

log/sequencer.go
```diff
@@ -209,32 +209,29 @@ func (s Sequencer) prepareLeaves(leaves []*trillian.LogLeaf, begin int64, label
...
-			nodeID, err := storage.NewNodeIDForTreeCoords(int64(depth), index, maxTreeDepth)
-			if err != nil {
-				return err
-			}
-			nodeMap[nodeID.String()] = storage.Node{
-				NodeID: nodeID,
-				Hash:   hash,
-			}
-			return nil
+			return store(depth, index, hash)
```
AlCutter: Can you not just pass `store` directly?

log/sequencer.go
```diff
@@ -169,12 +169,15 @@ func (s Sequencer) buildMerkleTreeFromStorageAtRoot(ctx context.Context, root *t
 	return mt, nil
 }
 
-func (s Sequencer) buildNodesFromNodeMap(nodeMap map[string]storage.Node, newVersion int64) ([]storage.Node, error) {
+func (s Sequencer) buildNodesFromNodeMap(nodeMap map[compact.NodeID][]byte, newVersion int64) ([]storage.Node, error) {
```
AlCutter: What's the win from changing string -> n.NodeID - is it mainly removing the cost of stringification?

pavelkalinnikov: - Removing stringification cost (time+memory+allocations)
- Using human readable strings as IDs is a slightly worse practice
- Simpler `updateCompactTree` (no error handling), and `storage.NodeID` is pushed closer to the place where it is needed
- In the long term it would be great to use integer node IDs in logs even at the storage layer (again less overhead and error handling)

log/sequencer.go
pavelkalinnikov: Indeed, will do.

pavelkalinnikov: cancel(): In a follow-up commit store() no longer returns an error (it could, but then it would require handling it in 2 other places).

pavelkalinnikov:
Ping. WDYT?

PR #1551, opened 2019-05-02
Workaround for single-tx-mode map performance issues

AlCutter:
Adds a workaround for performance hit when using the map in single transaction mode.

When using the map in this mode, the various map builder go routines experience massive lock contention because they share a single underlying subtreeCache.  This workaround preloads the entire set of nodes we know the set of builders will want and so reduce the time spent with locks held in the cache.

### Checklist

- [ X] I have updated the [CHANGELOG](CHANGELOG.md).

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -53,6 +53,10 @@ type TrillianMapServerOptions struct {
 	// UseSingleTransaction specifies whether updates to a map should be
 	// attempted within a single transaction.
 	UseSingleTransaction bool
+
+	// UseLargePreload enables the performance workaround applied when
+	// UseSingleTransaction is set.
+	UseLargePreload bool
 }
 
 // TrillianMapServer implements the RPC API defined in the proto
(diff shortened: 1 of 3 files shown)


server/trillian_map_server/main.go
```diff
@@ -66,6 +66,7 @@ var (
 	configFile = flag.String("config", "", "Config file containing flags, file contents can be overridden by command line flags")
 
 	useSingleTransaction = flag.Bool("single_transaction", false, "Experimental: use a single transaction when updating the map")
+	largePreload         = flag.Bool("large_preload_fix", true, "Experimental: work-around locking performance issues when using useSingleTransaction mode")
```
Martin2112: "useSingleTransaction" -> "single_transaction", needs the flag name not var.

server/map_rpc_server.go
```diff
@@ -263,6 +267,23 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 				HashedValue: l.LeafHash,
 			})
 		}
+
+		if t.opts.UseSingleTransaction && t.opts.UseLargePreload {
+			readRev, err := tx.ReadRevision(ctx)
+			if err != nil {
+				return err
+			}
+			nids := make([]storage.NodeID, 0, len(hkv)*256)
```
Martin2112: Can any of these IDs be duplicates?

server/trillian_map_server/main.go
AlCutter: Done

server/map_rpc_server.go
AlCutter: They can. I figured that the storage layer will dedupe, but you're right - it's best to be explicit and do it here [too].

Martin2112: Yeah storage will definitely dedupe but I think the map could get fairly large here so worth doing.

server/map_rpc_server.go
```diff
@@ -263,6 +267,33 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		if t.opts.UseSingleTransaction && t.opts.UseLargePreload {
+			readRev, err := tx.ReadRevision(ctx)
+			if err != nil {
+				return err
+			}
+			nidSet := make(map[string]bool)
+			nids := make([]storage.NodeID, 0, len(hkv)*256)
+			for _, i := range hkv {
+				nid := storage.NewNodeIDFromHash(i.HashedKey)
+				sibs := (&nid).Siblings()
```
Martin2112: Not in scope for this PR but we could look at using *NodeIDs more widely. I suspect we could save on some slice copying.

server/map_rpc_server.go
AlCutter: (done)

PR #1557, opened 2019-05-03
Faster version of NodeID.String() for use by sparse_merkle_tree

Martin2112:
For use in mapkeys etc. Add it to the sparse_merkle_tree code where it's
used in proofs. Profiling showed this using over 3% of CPU time in a
maphammer run. This version is 10x faster.

```
BenchmarkString-12    	 1000000	      2989 ns/op
BenchmarkAsKey-12     	 5000000	       307 ns/op
```

The Log sequencer does a similar thing but I've left that alone for now
so this change only affects the map.

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -17,13 +17,18 @@ package storage
 import (
 	"bytes"
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"math/bits"
+	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/google/trillian/storage/storagepb"
 )
+
+const hexChars = "0123456789abcdef"
 
 // Node represents a single node in a Merkle tree.
 type Node struct {
@@ -267,7 +272,7 @@ func (n *NodeID) Bit(i int) uint {
 // String returns a string representation of the binary value of the NodeID.
 // The left-most bit is the MSB (i.e. nearer the root of the tree). The
 // length of the returned string will always be the same as the prefix length
-// of the node.
+// of the node. For a string ID to use as a map key see AsKey().
 func (n *NodeID) String() string {
 	var r bytes.Buffer
 	limit := n.PathLenBits() - n.PrefixLenBits
@@ -275,6 +280,34 @@ func (n *NodeID) String() string {
 		r.WriteRune(rune('0' + n.Bit(i)))
 	}
 	return r.String()
+}
+
+// AsKey returns a string identifier for this NodeID suitable for
+// short term use e.g. as a Map key. It is more efficient to use this than
+// String() as it's not constrained to return a binary string.
+func (n *NodeID) AsKey() string {
+	var b strings.Builder
+	fullBytes := n.PrefixLenBits / 8
+	bitsLeft := n.PrefixLenBits % 8
+
+	// Note: all Builder write methods are documented never to return errors.
+	// Write the length first.
+	b.WriteString(strconv.Itoa(n.PrefixLenBits))
+	b.WriteRune(':')
+	// We can do all the full bytes in one go.
+	if fullBytes > 0 {
+		buf := make([]byte, hex.EncodedLen(fullBytes))
+		hex.Encode(buf, n.Path[:fullBytes])
+		b.Write(buf)
+	}
+	// If there's bits left over write them out.
+	if bitsLeft > 0 {
+		bits := n.Path[fullBytes] & leftmask[bitsLeft]
+		b.WriteByte(hexChars[bits>>4])
+		b.WriteByte(hexChars[bits&0xf])
+	}
+
+	return b.String()
 }
 
 // CoordString returns a string representation assuming that the NodeID represents a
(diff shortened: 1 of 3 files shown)


pavelkalinnikov:
Re: log sequencer, #1550 makes it no longer use stringification.

storage/types.go
```diff
@@ -277,6 +282,34 @@ func (n *NodeID) String() string {
...
+	var b strings.Builder
+	fullBytes := n.PrefixLenBits / 8
+	bitsLeft := n.PrefixLenBits % 8
+
+	// Note: all Builder write methods are documented never to return errors.
+	// Write the length first.
+	b.WriteString(strconv.Itoa(n.PrefixLenBits))
+	b.WriteRune(':')
+	// We can do all the full bytes in one go.
+	if fullBytes > 0 {
```
pavelkalinnikov:
```golang
fmt.Fprintf(&b, "%d:%x", n.PrefixLenBits, n.Path[:fullBytes])
```
?

storage/types.go
```diff
@@ -277,6 +282,34 @@ func (n *NodeID) String() string {
...
+		buf := make([]byte, hex.EncodedLen(fullBytes))
+		hex.Encode(buf, n.Path[:fullBytes])
+		b.Write(buf)
+	}
+	// If there's bits left over write them out.
+	if bitsLeft > 0 {
+		bits := n.Path[fullBytes] & leftmask[bitsLeft]
+		b.WriteByte(hexChars[bits>>4])
```
pavelkalinnikov: ```fmt.Fprintf(&b, "%x", bits)```?

storage/types.go
```diff
@@ -277,6 +282,34 @@ func (n *NodeID) String() string {
 	return r.String()
 }
 
+// AsKey returns a string identifier for this NodeID suitable for
+// short term use e.g. as a Map key. It is more efficient to use this than
+// String() as it's not constrained to return a binary string.
+func (n *NodeID) AsKey() string {
```
pavelkalinnikov: A string in Go can hold arbitrary bytes. Why not just copy `n.PrefixLenBits + n.Path` as bytes for more efficiency?

storage/types.go
Martin2112: hex.Encode looks more efficient.

storage/types.go
Martin2112: If you mean Sprintf that's more string copying.

storage/types.go
Martin2112: Because this gives a short ID that's human readable.

pavelkalinnikov: Why does it need to be human readable? Its primary usage is being a key for in-memory hashmaps, isn't it? The only places where it needs to be human readable is tests and logs, but there you could print them with "%x" which would result in the same string as currently (also in tests the "golden" values can be hex-decoded).

Martin2112: It seems like a nice property and it's fairly cheap.

storage/types.go
pavelkalinnikov: Could you try the benchmark? Although `Fprintf` has more indirections, it writes the result straight to the buffer, whereas the current code allocates an extra string in line 295. Not a clear win to me.

storage/types.go
pavelkalinnikov: `Fprintf` writes directly to the passed in output buffer.

storage/types.go
pavelkalinnikov: Tried it locally on my machine, `fmt.Fprintf` is 50% slower than `hex.Encode`. Makes sense to me now.

storage/types.go
pavelkalinnikov: Tried this locally, `Fprintf` here makes the thing 25% slower (prob. due to indirection).

storage/types.go
pavelkalinnikov: I don't think it's cheap on the relative scale. Compare:

```golang
func (n *NodeID) AsKey() string {
	var b strings.Builder
	fullBytes := n.PrefixLenBits / 8
	bitsLeft := n.PrefixLenBits % 8

	lb := make([]byte, 2)
	binary.LittleEndian.PutUint16(lb, uint16(n.PrefixLenBits)) // Or uint32, or varint.
	b.Write(lb[:])
	b.Write(n.Path[:fullBytes])
	// If there's bits left over write them out.
	if bitsLeft > 0 {
		bits := n.Path[fullBytes] & leftmask[bitsLeft]
		b.WriteByte(bits)
	}

	return b.String()
}
```

```
Before:	BenchmarkAsKey-12    	 5000000	       286 ns/op
After:	BenchmarkAsKey-12    	20000000	       109 ns/op
```

There is already one human readable format in `String()`. As another option, I think it would make sense to replace `String()` with your optimized version (I think there is no code other than in-memory stuff that relies on this format currently), and refactor `AsKey()` to return bytes packed to a string.

pphaneuf:
My 2p is that `String()` should be the more readable one (it's the one that would likely be invoked in a log message, for example), and `AsKey()` should focus on being fast and a good key, unreadable binary would be totally fine.

storage/types.go
therealdrake: Very very small aside. Could:
b.Write(lb[:])
be written as:
b.Write(lb)?

pavelkalinnikov: Oh, I actually meant it to be `var lb [2]byte`, and then it would imply`lb[:]`.

Martin2112: It's hard to be certain that it's not used for something important. My IDE says it isn't but `String()` can be called implicitly by `fmt` methods, which it might not be able to pick up. I think a `String()` method should only return something readable so we're stuck with it.

Martin2112:
We can revisit it again if this change makes a noticeable difference on real requests.

PR #1558, opened 2019-05-03
Fix the bug which made the map go-faster change go slower.

AlCutter:
### Checklist


Martin2112:
Drat. Should have spotted that.

PR #1559, opened 2019-05-03
Do preload nodeID calculations in parallel

AlCutter:
Do nodeID calculations for set leaves preload concurrently.

For large batches there is a lot of time spent creating NodeID strings here, so do the work concurrently.

### Checklist

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -30,7 +30,6 @@ import (
 	"github.com/google/trillian/types"
 
 	"github.com/golang/glog"
-	"go.opencensus.io/trace"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -319,6 +319,13 @@ func doPreload(ctx context.Context, tx storage.MapTreeTX, treeDepth int, hkv []m
 	if err != nil {
 		return err
 	}
+
+	nids := calcAllSiblingsParallel(ctx, treeDepth, hkv)
+	_, err = tx.GetMerkleNodes(ctx, readRev, nids)
+	return err
+}
+
+func calcAllSiblingsParallel(ctx context.Context, treeDepth int, hkv []merkle.HashKeyValue) []storage.NodeID {
 	nidSet := make(map[string]bool)
 	type nodeAndID struct {
 		id   string
@@ -358,8 +365,7 @@ func doPreload(ctx context.Context, tx storage.MapTreeTX, treeDepth int, hkv []m
 
 	<-done
 
-	_, err = tx.GetMerkleNodes(ctx, readRev, nids)
-	return err
+	return nids
 }
 
 func (t *TrillianMapServer) makeSignedMapRoot(ctx context.Context, tree *trillian.Tree, smrTs time.Time,
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -341,7 +341,7 @@ func calcAllSiblingsParallel(ctx context.Context, treeDepth int, hkv []merkle.Ha
 			nid := storage.NewNodeIDFromHash(k)
 			sibs := (&nid).Siblings()
 			for _, sib := range sibs {
-				sibID := sib.String()
+				sibID := sib.AsKey()
 				sib := sib
 				c <- nodeAndID{sibID, sib}
 			}
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -326,7 +326,6 @@ func doPreload(ctx context.Context, tx storage.MapTreeTX, treeDepth int, hkv []m
 }
 
 func calcAllSiblingsParallel(ctx context.Context, treeDepth int, hkv []merkle.HashKeyValue) []storage.NodeID {
-	nidSet := make(map[string]bool)
 	type nodeAndID struct {
 		id   string
 		node storage.NodeID
@@ -334,6 +333,7 @@ func calcAllSiblingsParallel(ctx context.Context, treeDepth int, hkv []merkle.Ha
 	c := make(chan nodeAndID, 2048)
 	var wg sync.WaitGroup
 
+	// Kick off producers.
 	for _, i := range hkv {
 		wg.Add(1)
 		go func(k []byte) {
@@ -348,22 +348,21 @@ func calcAllSiblingsParallel(ctx context.Context, treeDepth int, hkv []merkle.Ha
 		}(i.HashedKey)
 	}
 
-	done := make(chan bool)
-	nids := make([]storage.NodeID, 0, len(hkv)*treeDepth)
+	// monitor for all the producers being complete to close the channel.
 	go func() {
-		for nai := range c {
-			if _, ok := nidSet[nai.id]; !ok {
-				nidSet[nai.id] = true
-				nids = append(nids, nai.node)
-			}
-		}
-		close(done)
+		wg.Wait()
+		close(c)
 	}()
 
-	wg.Wait()
-	close(c)
-
-	<-done
+	nidSet := make(map[string]bool)
+	nids := make([]storage.NodeID, 0, len(hkv)*treeDepth)
+	// consume the produced IDs until the channel is closed.
+	for nai := range c {
+		if _, ok := nidSet[nai.id]; !ok {
+			nidSet[nai.id] = true
+			nids = append(nids, nai.node)
+		}
+	}
 
 	return nids
 }
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -17,6 +17,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


server/map_rpc_server.go
```diff
@@ -321,6 +305,54 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		id   string
+		node storage.NodeID
+	}
+	c := make(chan nodeAndID, 2048)
+	var wg sync.WaitGroup
+
+	for _, i := range hkv {
+		wg.Add(1)
+		go func() {
+			nid := storage.NewNodeIDFromHash(i.HashedKey)
```
golangcibot: loopclosure: loop variable i captured by func literal (from `govet`)

server/map_rpc_server.go
```diff
@@ -29,13 +30,16 @@ import (
 	"github.com/google/trillian/types"
 
 	"github.com/golang/glog"
+	"go.opencensus.io/trace"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		id   string
+		node storage.NodeID
+	}
+	c := make(chan nodeAndID, 2048)
+	var wg sync.WaitGroup
+
+	for _, i := range hkv {
+		wg.Add(1)
+		go func(k []byte) {
+			nid := storage.NewNodeIDFromHash(k)
```
therealdrake: does go let you make this a pointer on this line? Then you could probably leave out the line below.

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+			sibs := (&nid).Siblings()
+			for _, sib := range sibs {
+				sibID := sib.String()
+				sib := sib
+				c <- nodeAndID{sibID, sib}
+			}
+			wg.Done()
```
therealdrake: I usually like to use the pattern "defer wg.Done()" at the top. This helps protect you from future changes that might return before this line.

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		}(i.HashedKey)
+	}
+
+	done := make(chan bool)
+	nids := make([]storage.NodeID, 0, len(hkv)*256)
+	go func() {
+		for nai := range c {
```
therealdrake: is there a possibility that this loop ends with go routines from above writing to it after?

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	return &trillian.SetMapLeavesResponse{MapRoot: newRoot}, nil
 }
 
+func doPreload(ctx context.Context, tx storage.MapTreeTX, hkv []merkle.HashKeyValue) error {
```
mhutchinson: For the ease of understanding for future readers, I recommend documenting briefly what this method is going to preload, and where things will be preloaded to.

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	var wg sync.WaitGroup
+
+	for _, i := range hkv {
+		wg.Add(1)
+		go func(k []byte) {
+			nid := storage.NewNodeIDFromHash(k)
+			sibs := (&nid).Siblings()
+			for _, sib := range sibs {
+				sibID := sib.String()
+				sib := sib
```
mhutchinson: Is "sib := sib" needed here? This reminds me of a warning some Go Elders once gave me, but I thought this situation only occurred when there were go functions being kicked off inside the range loop?

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+				c <- nodeAndID{sibID, sib}
+			}
+			wg.Done()
+		}(i.HashedKey)
+	}
+
+	done := make(chan bool)
+	nids := make([]storage.NodeID, 0, len(hkv)*256)
```
mhutchinson: Consider extracting this 256 to a constant for SparseMerkleTreeDepth. If we ever wanted to have a Map of different depth then this would make it easier to evaluate where updates are needed. It also avoids the confusion I first had when reading this of assuming the 256 == 2^8 == number of leaves in a single stratum of the tree (I've spent some time today understanding NodeID, which may explain my incorrect first reading of this. YMMV).

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	ctx, span := spanFor(ctx, "doPreload")
+	defer span.End()
+
+	readRev, err := tx.ReadRevision(ctx)
+	if err != nil {
+		return err
+	}
+	nidSet := make(map[string]bool)
+	type nodeAndID struct {
+		id   string
```
gdbelvin: can we avoid a stringification?

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	done := make(chan bool)
+	nids := make([]storage.NodeID, 0, len(hkv)*256)
+	go func() {
+		for nai := range c {
+			if _, ok := nidSet[nai.id]; !ok {
+				nidSet[nai.id] = true
+				nids = append(nids, nai.node)
+			}
+		}
+		done <- true
```
gdbelvin:
```suggestion
		close(done)
```

server/map_rpc_server.go
AlCutter: Sadly not :/

server/map_rpc_server.go
AlCutter: 👍done

server/map_rpc_server.go
AlCutter: No - this loop will not terminate until `c` has been closed (if `c` is still open and there's nothing to read it'll just block), and `c` cannot be closed until all the goroutines above have finished their work and called `wg.Done()` (since that unblocks the `wg.Sync()` below, after which `close(c)` will be called).

(More details [here](https://tour.golang.org/concurrency/4)

server/map_rpc_server.go
```diff
@@ -321,6 +306,57 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	readRev, err := tx.ReadRevision(ctx)
+	if err != nil {
+		return err
+	}
+	nidSet := make(map[string]bool)
+	type nodeAndID struct {
+		id   string
+		node storage.NodeID
+	}
+	c := make(chan nodeAndID, 2048)
```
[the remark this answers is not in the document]
AlCutter: It's an arbitrary number - it'll should work with any channel depth, I just picked one which seemed like it gave a "reasonable" buffer space but wouldn't eat all the memories.

server/map_rpc_server.go
AlCutter: 👍 done

server/map_rpc_server.go
AlCutter: _Possibly_ not, but I figured it was worth having anyway as a warning sign since the loop variable does get sent down a channel; if the `node` in `nodeAndID` were to be changed to take a pointer to a `storage.Node` at some point in the future then this would fail in a nasty way...

server/map_rpc_server.go
AlCutter: Passed in `hasher.BitLen()` for this for now, as this currently defines the height of the tree.

server/map_rpc_server.go
AlCutter: I don't think so - `storage.NodeID` cannot be used as a map key, although Martin's `AsKey` should make it cheaper.

server/map_rpc_server.go
AlCutter: 👍 done

AlCutter:
PTAL

server/map_rpc_server.go
```diff
@@ -321,6 +306,68 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	var wg sync.WaitGroup
+
+	for _, i := range hkv {
+		wg.Add(1)
+		go func(k []byte) {
+			defer wg.Done()
+			nid := storage.NewNodeIDFromHash(k)
+			sibs := (&nid).Siblings()
+			for _, sib := range sibs {
+				sibID := sib.String()
```
gdbelvin:
```suggestion
				sibID := sib.AsKey()
```

AlCutter: I don't think that change has been submitted yet... :)

gdbelvin: @Martin2112 is #1557 ready yet?

server/map_rpc_server.go
```diff
@@ -321,6 +306,68 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+				sib := sib
+				c <- nodeAndID{sibID, sib}
+			}
+		}(i.HashedKey)
+	}
+
+	done := make(chan bool)
+	nids := make([]storage.NodeID, 0, len(hkv)*treeDepth)
+	go func() {
```
gdbelvin: Is this goroutine needed?  Might be simpler to just run this in-line

server/map_rpc_server.go
Martin2112: I think so. Trying to make sure it doesn't break anything.

Martin2112: Just merged.

server/map_rpc_server.go
gdbelvin: I see how it's needed now.  Just wondering if this function would be clearer if the producer function was being run in parallel: 

```go
    chan c
    go func() {
      ...
      wg.Wait()
      close(c) 
   }
```

server/map_rpc_server.go
AlCutter: Rebased

server/map_rpc_server.go
AlCutter: yeah, we need another go-routine one way or the other - you might be right that it's easier to understand with the consumer inline, so I've done that.

PR #1562, opened 2019-05-07
Add a bunch of tracing spans in sparse Merkle writer

AlCutter:
This should give a bit more insight into what's going on inside RPCs to the map server.

### Checklist

### DIFF
--- a/merkle/sparse_merkle_tree.go
+++ b/merkle/sparse_merkle_tree.go
@@ -25,6 +25,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/google/trillian/merkle/hashers"
 	"github.com/google/trillian/storage"
+	"go.opencensus.io/trace"
 )
 
 // For more information about how Sparse Merkle Trees work see the Revocation Transparency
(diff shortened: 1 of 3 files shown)


merkle/sparse_merkle_tree.go
```diff
@@ -442,12 +478,16 @@ func (s SparseMerkleTreeReader) BatchInclusionProof(ctx context.Context, rev int
 		}
 		r[string(index)] = ri
 	}
+	span.End()
```
Martin2112: Might be Github diffs confusing things but does that match the span that's had the End() already deferred?

AlCutter: should be the one at L456, no?
(I can rename them if you prefer?

Martin2112: No I see it now. I think renaming just adds noise.

Martin2112:
Looks like some of the tests might be too strict on the context they accept.

AlCutter:
> Looks like some of the tests might be too strict on the context they accept.

Should be fixed now...

PR #1564, opened 2019-05-08
Clarified documentation for important NodeID methods.

mhutchinson:
These methods have names which have different behaviour than a newcomer to Trillian may expect. Having a little more documentation may save someone from misunderstanding these methods, or from needing to work it out by reading the code.

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -375,8 +375,8 @@ func (n *NodeID) Neighbor() *NodeID {
 // (including this one) on the path up to the root. The array is of length
 // PrefixLenBits and is ordered such that the nodes closest to the leaves are
 // earlier in the array.
-// These nodes are the ones that would be required for a Merkle proof for this
-// node.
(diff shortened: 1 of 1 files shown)


storage/types.go
```diff
@@ -362,13 +362,21 @@ func (n *NodeID) MaskLeft(depth int) *NodeID {
...
 	n.FlipRightBit(height)
 	return n
 }
 
 // Siblings returns the siblings of the given node.
+// In terms of a tree traversal, this returns the Neighbour() of every node
+// (including this one) on the path up to the root. The array is of length
+// PrefixLenBits and is ordered such that the nodes closest to the leaves are
+// earlier in the array.
+// These nodes are the ones that would be required for a Merkle proof for this
```
pavelkalinnikov: Clarify the kind of proof:
`... required for a Merkle tree inclusion proof ...` ?

PR #1565, opened 2019-05-08
compact: Do not return error from visitor

pavelkalinnikov:
This change makes the visitor function that is passed in to AddLeafHash
never return an error. This simplifies code in many contexts.

### Checklist


storage/testonly/fake_node_reader.go
```diff
@@ -154,7 +147,12 @@ func NewMultiFakeNodeReaderFromLeaves(batches []LeafBatch) *MultiFakeNodeReader
 		// Unroll the update map to []storage.Node to retain the most recent node update within
 		// the batch for each ID. Use that to create a new FakeNodeReader.
 		nodes := make([]storage.Node, 0, len(nodeMap))
-		for _, node := range nodeMap {
+		for id, hash := range nodeMap {
+			nID, err := storage.NewNodeIDForTreeCoords(int64(id.Level), int64(id.Index), 64)
+			if err != nil {
+				panic(fmt.Errorf("failed to create NodeID for %+v: %v", id, err))
```
Martin2112: Not really in favour of adding panics but it's test code and it already panics so it will be ok.

PR #1567, opened 2019-05-08
Enforce serialised access to MySQL transaction

AlCutter:
The mysql driver is deeply unhappy if you make multiple concurrent queries on the same tx, this PR attempts to enforce that this doesn't happen.

### Checklist

- [X ] I have updated the [CHANGELOG](CHANGELOG.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,9 @@ if err != nil {
 }
 ```
 
-### Configurable number of connections for MySQL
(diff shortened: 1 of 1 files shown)


Martin2112:
I agree it's not explicit. Sharing TXs will result in pauses where they have to wait for the lock. This will give different behaviour on MySQL to cloudspanner etc. I haven't seen where you're going with this yet though.

PR #1568, opened 2019-05-08
TXRunner Interface

gdbelvin:
Simplify the API semantics for the transaction runner by making it an
interface rather than an anonymous function.

This makes my head hurt significantly less when I look at the code.


server/map_rpc_server.go
```diff
@@ -306,6 +291,35 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+func (r *txRunner) RunTX(ctx context.Context, f func(context.Context, storage.MapTreeTX) error) error {
+	if r.opts.UseSingleTransaction {
+		glog.V(1).Infof("Using enclosing tx for subtree operation %d", atomic.LoadUint64(&r.txRolledUp))
+		atomic.AddUint64(&r.txRolledUp, 1)
+		return f(ctx, r.tx)
+	}
+	return r.mapStorage.ReadWriteTransaction(ctx, r.tree, f)
+}
+
+func (r *txRunner) Close() {
```
AlCutter: You can get rid of this function and just ding the txRolledUp counter - it's not used anywhere.
(It's got potential for confusion currently since Close() doesn't actually close the transaction.)

gdbelvin:
@AlCutter - this is working now. PTAL

merkle/sparse_merkle_tree.go
```diff
@@ -40,9 +40,11 @@ type SparseMerkleTreeReader struct {
 	treeRevision int64
 }
 
-// runTXFunc is the interface for a function which produces something which can
+// TXRunner is the interface for a function which produces something which can
```
AlCutter: Comment needs updating since it's not a func anymore.

merkle/sparse_merkle_tree_test.go
```diff
@@ -66,19 +66,21 @@ func getSparseMerkleTreeReaderWithMockTX(ctrl *gomock.Controller, rev int64) (*S
...
-	defer tx.Close()
+type producerTXRunner struct {
+	tx storage.MapTreeTX
+}
 
-	return func(ctx context.Context, f func(context.Context, storage.MapTreeTX) error) error {
-		return f(ctx, tx)
-	}
+func (r *producerTXRunner) RunTX(ctx context.Context, f func(context.Context, storage.MapTreeTX) error) error {
+	r.tx.Close()
```
AlCutter: This closes the transaction before running `f` on it?

server/map_rpc_server.go
```diff
@@ -366,6 +349,34 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
 
+func (t *TrillianMapServer) newTXRunner(tree *trillian.Tree, tx storage.MapTreeTX) merkle.TXRunner {
+	if t.opts.UseSingleTransaction {
+		return &singleTXRunner{tx: tx}
+	}
+	return &multiTXRunner{tree: tree, mapStorage: t.registry.MapStorage}
+}
+
+// singleTXRunner satisfies the merkle.TXRunner interface.
+type singleTXRunner struct {
```
AlCutter: Maybe just add a one-line comment on this and `multiTXRunner` describing what they do and why?

merkle/sparse_merkle_tree_test.go
gdbelvin: should be a defer.

gdbelvin:
@AlCutter Updated with better comments

gdbelvin:
PTAL

server/map_rpc_server.go
```diff
@@ -414,7 +414,7 @@ func (t *TrillianMapServer) newTXRunner(tree *trillian.Tree, tx storage.MapTreeT
 	return &multiTXRunner{tree: tree, mapStorage: t.registry.MapStorage}
 }
 
-// singleTXRunner executes f using a single transaction.
+// singleTXRunner executes all callse to Run with the same underlying transaction.
```
AlCutter: /callse/calls/

PR #1572, opened 2019-05-09
compact: Move NodeID and clean up comments

pavelkalinnikov:
### Checklist

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -26,12 +26,6 @@ import (
 	"github.com/google/trillian/merkle/hashers"
 )
 
-// NodeID identifies a node of a log type Merkle tree.
-type NodeID struct {
-	Level uint
-	Index uint64
-}
-
 // Tree is a compact Merkle tree representation. It uses O(log(size)) nodes to
 // represent the current on-disk tree.
 //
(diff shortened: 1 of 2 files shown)


merkle/compact/tree.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/google/trillian/merkle/hashers"
 )
 
-// NodeID identifies a node of a Merkle tree.
+// NodeID identifies a node of a log type Merkle tree.
```
mhutchinson: What properties of a "log type" Merkle tree does this require to be useful? Is it primarily the density (i.e. not sparse - it doesn't have missing elements)?

merkle/compact/tree.go
```diff
@@ -26,7 +26,7 @@ import (
...
 type NodeID struct {
 	Level uint
```
mhutchinson: While you're commenting, I would find it useful as an outsider to this code to have a 1-liner explaining where you are counting from with these values. Is Level 0 the root or leaf? Is Index counting from the left or right?

merkle/compact/tree.go
```diff
@@ -83,7 +83,9 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
 	}
 
 	ids := make([]NodeID, 0, bits.OnesCount64(uint64(size)))
-	for sz := uint64(size); sz != 0; sz &= sz - 1 { // Iterate 1-bits of size.
+	// Iterate over perfect subtrees along the right border of the tree. Those
+	// correspond to 1-bits of the tree size.
```
mhutchinson: "1-bits of the tree size" isn't an expression that parsed easily to me. From a sample size of 2, 50% of participants found it easy to parse.

merkle/compact/tree.go
pavelkalinnikov: After some thinking, this `NodeID` type on its own only restricts the max size of the tree (it doesn't have to be log, or dense). I omitted this because it follows from the fact that it's a pair of integers.

merkle/compact/tree.go
pavelkalinnikov: Moved this type to `range.go` and reused an existing comment explaining the tree geometry.

merkle/compact/tree.go
pavelkalinnikov: Done. Does it look clearer?

PR #1575, opened 2019-05-09
CONIKS, hstar2 and SMT Writer performance improvements

Martin2112:
Bunch of performance improvements to CONIKS hasher, hstar2 and Sparse Merkle Tree Writer. Only affects Map trees so should be fairly safe.

Benchmarking shows ~20% speed increase and ~3% RAM allocation reduction for the combined changes. Some of the speedup comes from switching to pointers for `HStar2LeafHash`.

### Checklist


merkle/coniks/coniks.go
```diff
@@ -106,25 +113,22 @@ func (m *hasher) BitLen() int {
...
 var leftmask = [8]byte{0xFF, 0x80, 0xC0, 0xE0, 0xF0, 0xF8, 0xFC, 0xFE}
 
-// maskIndex returns index with only the left depth bits set.
-// index must be of size m.Size() and 0 <= depth <= m.BitLen().
-// e.g.
-func (m *hasher) maskIndex(index []byte, depth int) []byte {
+// maskIndex writes the masked value directly to a Buffer (which never
+// returns an error on writes). This avoids the need to allocate space for and
+// copy a value that will then be discarded immediately.
+func (m *hasher) maskIndex(b *bytes.Buffer, index []byte, depth int) {
```
AlCutter: rename to `writeMaskedIndex`?

merkle/coniks/coniks.go
```diff
@@ -124,11 +125,19 @@ func (m *hasher) maskIndex(b *bytes.Buffer, index []byte, depth int) {
 		panic(fmt.Sprintf("depth: %d, want <= %d && > 0", got, want))
 	}
 
+	prevLen := b.Len()
 	if depth > 0 {
 		// Copy the first depthBytes.
 		depthBytes := (depth + 7) >> 3
-		b.Write(index[:depthBytes])
+		if depth > 1 {
```
AlCutter: Should we add a test for this case?

Martin2112: I think it's covered by existing tests but I'll go back and make sure.

merkle/coniks/coniks.go
```diff
@@ -60,11 +62,13 @@ func (m *hasher) EmptyRoot() []byte {
 func (m *hasher) HashEmpty(treeID int64, index []byte, height int) []byte {
 	depth := m.BitLen() - height
 
+	buf := new(bytes.Buffer)
```
gdbelvin: Do you think it would be worth preallocating the buffer to avoid any runtime resizing?

Martin2112: Probably only if it resizes more than once, which is a possibility. I can test that with a benchmark.

merkle/coniks/coniks.go
Martin2112: This was incorrect, should have been `depthBytes`. However for depthBytes to be zero depth must be zero so it worked anyway by luck.

merkle/coniks/coniks.go
Martin2112: Done.

merkle/sparse_merkle_tree.go
```diff
@@ -257,7 +254,7 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context, queueSize int) {
 		}
 		span.End()
 
-		hsCtx, span := spanFor(ctx, "buildSubtree.runTX.htar2")
+		hsCtx, span := spanFor(ctx, "buildSubtree.runTX.hstar2")
```
AlCutter: Aha, all this extra `tar` must've been slowing it down! :)

merkle/coniks/coniks.go
Martin2112: Additional tests added as a crosscheck against bit twiddling on a big.Int.

merkle/coniks/coniks.go
Martin2112: Seems about 5% faster. Might be worth it though the time in this function is probably way less than in the hashing.

PR #1577, opened 2019-05-10
Add setter/getter for span func

AlCutter:
Turns out this is helpful after all...

### Checklist

### DIFF
--- a/monitoring/trace.go
+++ b/monitoring/trace.go
@@ -18,9 +18,23 @@ import (
 	"context"
 )
 
+// startSpanFunc is the signature of a function which can start tracing spans.
+type startSpanFunc func(context.Context, string) (context.Context, func())
+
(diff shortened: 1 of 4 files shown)


monitoring/trace.go
```diff
@@ -18,9 +18,23 @@ import (
 	"context"
 )
 
+// startSpanFunc is the signature of a function which can start tracing spans.
+type startSpanFunc func(context.Context, string) (context.Context, func())
+
+var startSpan startSpanFunc = func(ctx context.Context, name string) (context.Context, func()) { return ctx, func() {} }
```
phad: You might move the remark about no-op below up here to indicate why 'startSpan' does so little.
Maybe rename it as noopStartSpan.

AlCutter: I think having the comment where it is on the exported func is good so that the behaviour is discoverable via godoc, but naming the default impl as a self-describing this is a really good suggestion - thanks!

Done.

PR #1580, opened 2019-05-11
compact: Remove leaf index from method signatures

pavelkalinnikov:
This change removes leaf index from `AddLeaf` and `AddLeafHash` returned
values, as the same information can be obtained using the `Size` method.

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -169,7 +169,7 @@ func (t *Tree) recalculateRoot(visit VisitFn) error {
 }
 
 // AddLeaf calculates the Merkle leaf hash of the given leaf data and appends
-// it to the tree. Returns Merkle hash of the new leaf.
+// it to the tree. Returns the Merkle hash of the new leaf.
 //
 // visit is a callback which will be called multiple times with the coordinates
 // of the Merkle tree nodes whose hash should be updated.
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -168,33 +168,27 @@ func (t *Tree) recalculateRoot(visit VisitFn) error {
 	return nil
 }
 
-// AddLeaf calculates the Merkle leaf hash of the given leaf data and appends it
-// to the tree.
+// AddLeaf calculates the Merkle leaf hash of the given leaf data and appends
+// it to the tree. Returns Merkle hash of the new leaf.
 //
 // visit is a callback which will be called multiple times with the coordinates
 // of the Merkle tree nodes whose hash should be updated.
-//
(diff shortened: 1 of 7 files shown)


merkle/compact/tree.go
```diff
@@ -168,33 +168,27 @@ func (t *Tree) recalculateRoot(visit VisitFn) error {
 	return nil
 }
 
-// AddLeaf calculates the Merkle leaf hash of the given leaf data and appends it
-// to the tree.
+// AddLeaf calculates the Merkle leaf hash of the given leaf data and appends
+// it to the tree. Returns Merkle hash of the new leaf.
```
taknira: Put a 'the' before the 'Merkle hash' to make the English flow a little better (like the rest of the comment)?

merkle/compact/tree_test.go
```diff
@@ -170,7 +170,8 @@ func TestCompactVsFullTree(t *testing.T) {
 			t.Errorf("AddLeaf(): %v", err)
 		}
 
-		cSeq, cHash, err := cmt.AddLeaf(newLeaf, func(id NodeID, hash []byte) {
+		cSeq := cmt.Size()
```
taknira: Hmm it is interesting (and maybe a little counter-intuitive) that you call this *before* actually adding the leaf... I get why it is this way, but to make it super clear to users it might be worth putting a comment somewhere (on the AddLeaf/AddLeafHash methods maybe?) saying that to get the index of the leaf being added you have to either call .Size() *before* adding it, or do index = cmt.Size()-1.  I feel like people (which people?  Who knows) using those functions could easily end up with off-by-1 errors by relying on .Size() and not using it quite correctly.

merkle/compact/tree_test.go
```diff
@@ -196,12 +197,12 @@ func TestCompactVsFullTree(t *testing.T) {
...
 		newLeaf := []byte(fmt.Sprintf("Leaf %d", i))
-		seq, _, err := cmt.AddLeaf(newLeaf, func(NodeID, []byte) {})
+		_, err := cmt.AddLeaf(newLeaf, func(NodeID, []byte) {})
 		if err != nil {
 			t.Fatalf("AddLeaf(%d)=_,_,%v, want _,_,nil", i, err)
 		}
-		if seq != i {
-			t.Fatalf("AddLeaf(%d)=%d, want %d", i, seq, i)
+		if got, want := cmt.Size(), i+1; got != want {
+			t.Fatalf("new size=%d, want %d", got, want)
```
taknira: "new tree size"?

merkle/compact/tree_test.go
pavelkalinnikov: Put `Size-1` after the `AddLeaf`.

PR #1581, opened 2019-05-12
compact.Tree: Handle root hash recalculation error

pavelkalinnikov:


merkle/compact/tree.go
```diff
@@ -188,11 +188,12 @@ func (t *Tree) AddLeaf(data []byte, visit VisitFn) ([]byte, error) {
 //
 // visit is a callback which will be called multiple times with the coordinates
 // of the Merkle tree nodes whose hash should be updated.
-func (t *Tree) AddLeafHash(leafHash []byte, visit VisitFn) error {
+func (t *Tree) AddLeafHash(leafHash []byte, visit VisitFn) (res error) {
 	defer func() {
 		t.size++
```
RJPercival: Do you want to run any of the code in this deferred func if `res != nil`? I see that the only code path that leads to that is a "should never happen" but, if it did happen, you presumably wouldn't want to increment the tree size, right?

pavelkalinnikov: If `AddLeafHash` returns an error, the type becomes unusable anyway.

RJPercival: Is that documented somewhere? Might be helpful to mention that in the comment above the method.

RJPercival: It seems like it'd be cleaner to do this at the start of the defer:
```go
if res != nil {
  return
}
```
It makes it clear that the defer is doing some finalisation that is only necessary if the method succeeded.

pavelkalinnikov: Done.

merkle/compact/tree.go
```diff
@@ -173,6 +173,8 @@ func (t *Tree) recalculateRoot(visit VisitFn) error {
 //
 // visit is a callback which will be called multiple times with the coordinates
 // of the Merkle tree nodes whose hash should be updated.
+//
+// If returns and error then the Tree is no longer usable.
```
RJPercival: typo: "returns and error"

pavelkalinnikov: Fixed.

PR #1582, opened 2019-05-12
compact.Tree: Compute root hash on demand

pavelkalinnikov:
This change removes root hash field from `compact.Tree` so it is no longer
necessary to maintain it after each `AddLeafHash` operation. A side effect
of this is that `CurrentRoot` now can return an error. We make sure that
the error is handled in the client code.

This commit prepares us to remove some unnecessary computation when many
`AddLeafHash` calls are done in a row. As a result, ephemeral nodes along the
right border of the tree and the root hash will be computed only once.

### DIFF
--- a/merkle/compact/tree.go
+++ b/merkle/compact/tree.go
@@ -32,7 +32,6 @@ import (
 // TODO(pavelkalinnikov): Remove it, use compact.Range instead.
 type Tree struct {
 	hasher hashers.LogHasher
-	root   []byte
 	// The list of "dangling" left-hand nodes, where entry [0] is the leaf.
 	// So: nodes[0] is the hash of a subtree of size 1 = 1<<0, if included.
 	//     nodes[1] is the hash of a subtree of size 2 = 1<<1, if included.
@@ -43,7 +42,7 @@ type Tree struct {
 	// 16 + 4 + 1, so nodes[1] == nodes[3] == nil.
 	//
 	// For a tree whose size is a perfect power of two, only the last
-	// entry in nodes will be set (and it will match root).
+	// entry in nodes will be set (which is also the root hash).
 	nodes [][]byte
 	size  int64
 }
@@ -72,7 +71,6 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
 	r := Tree{
 		hasher: hasher,
 		nodes:  make([][]byte, sizeBits),
-		root:   hasher.EmptyRoot(),
 		size:   size,
 	}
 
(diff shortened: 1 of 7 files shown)


merkle/compact/tree.go
```diff
@@ -94,35 +92,39 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
 	for i, id := range ids {
 		r.nodes[id.Level] = hashes[i]
 	}
-	r.recalculateRoot(func(NodeID, []byte) {})
 
-	if !bytes.Equal(r.root, expectedRoot) {
-		glog.Warningf("Corrupt state, expected root %s, got %s", hex.EncodeToString(expectedRoot[:]), hex.EncodeToString(r.root[:]))
-		return nil, fmt.Errorf("root hash mismatch: got %v, expected %v", r.root, expectedRoot)
+	// TODO(pavelkalinnikov): This check should be done externally.
```
AlCutter: OTOH having it here forces it to be done...
In the interests of making it hard to hurt yourself with the tools is it work keeping it here?

merkle/compact/tree.go
```diff
@@ -94,35 +92,39 @@ func NewTreeWithState(hasher hashers.LogHasher, size int64, getNodesFn GetNodesF
...
 		nodes:  make([][]byte, 0),
 		size:   0,
 	}
 }
 
 // CurrentRoot returns the current root hash.
-func (t *Tree) CurrentRoot() []byte {
-	return t.root
+func (t *Tree) CurrentRoot() ([]byte, error) {
+	return t.calculateRoot(func(NodeID, []byte) {})
```
AlCutter: This makes this function more expensive, are there any existing callers who assume that it isn't?

merkle/compact/tree.go
```diff
@@ -191,8 +192,8 @@ func (t *Tree) AddLeaf(data []byte, visit VisitFn) ([]byte, error) {
 func (t *Tree) AddLeafHash(leafHash []byte, visit VisitFn) error {
 	defer func() {
 		t.size++
-		// TODO(pavelkalinnikov): Handle recalculateRoot errors.
-		t.recalculateRoot(visit)
+		// TODO(pavelkalinnikov): Handle calculateRoot errors.
```
AlCutter: Since this is being recalculated in `CurrentRoot` do you still need this deferred func here? 
I guess yes because it's how intermediate nodes get written - but then is it necessary to recalculate it in `CurrentRoot`?

Is there a bigger picture that this PR is working toward?

pavelkalinnikov: Correct, I keep it here only because all the code currently relies on the fact that `AddLeafHash` also reports **ephemeral** nodes via `visit` function. 

The bigger picture (see also PR description):
- `AddLeafHash` (I will probably rename it because of changed semantics) reports only non-ephemeral nodes, and thus calling `calculateRoot` is no longer necessary (key result: better performance).
- `calculateRoot` is exported and replaces `CurrentRoot`. It allows clients to compute root hash and get ephemeral nodes at the same time, as most of them need both operations anyway: e.g. the sequencer needs both after completing a series of `AddLeafHash` calls.
- Result: the API of `compact.Tree` is so close to `compact.Range` that replacing one with the other is a short PR.

merkle/compact/tree.go
pavelkalinnikov: I didn't find any dramatic ones. Usage patterns that I found:

1. Call `CurrentRoot` after every `AddLeafHash`. Only in tests.
2. Call it after a series of `AddLeafHash`, like in sequencer or tree cache.
3. Call it once, e.g. during serialization/deserialization in the ingestor.

We can ignore case 1 because it's tests, and case 3 is not affected. In case 2 there will be one extra `calculateRoot` call in the whole sequence (so N+1 calls instead of N, for N leaves added). But, see the "bigger picture" comment below, in a follow-up PR N+1 will be be decreased to 1 (`calculateRoot` will be called "lazily" after all inserts are completed).

merkle/compact/tree.go
pavelkalinnikov: Yeah, it feels safer. On the third hand, maybe there will be use-cases when the expected root hash is unknown, or the check is not necessary for some other reason. I usually advocate for bare minimal interfaces, which is a big pro on its own, and in this case the interface can also be more flexible.

merkle/compact/tree.go
AlCutter: :+1:

PR #1584, opened 2019-05-13
Limit MySQL connections during tests

RJPercival:
The hammer test can exceed MySQL's default limit on connections (150), resulting in it failing. To make this test reliably pass without requiring all developers to modify their MySQL server configuration, limit the number of connections that any test can make to 100 (leaving some spare in case other processes are also accessing the MySQL server).

### Checklist


RJPercival:
It looks like the map hammer times out if it is restricted to 100 database connections unfortunately.

PR #1585, opened 2019-05-13
sequencer: Factor out quota replenishment

pavelkalinnikov:
This change puts quota replenishment into a separate method, which
slightly decreases the complexity of the `IntegrateBatch` method.


pavelkalinnikov:
PTAL. This is aimed at decreasing `gocyclo` complexity to unblock `IntegrateBatch` changes like in #1582.

PR #1590, opened 2019-05-13
log_subtree_cache: Factor out visitor callback

pavelkalinnikov:
This change prepares the visitor callback in log subtree cache to be
used multiple times when `AddLeafHash` of `compact.Tree` is replaced by a
method no longer reporting ephemeral nodes. Besides, it reduces
indentation/nesting in the code.

### DIFF
--- a/storage/cache/log_subtree_cache.go
+++ b/storage/cache/log_subtree_cache.go
@@ -58,6 +58,23 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
 		if st.InternalNodes == nil || len(st.Leaves) == maxLeaves {
 			st.InternalNodes = make(map[string][]byte)
 		}
+		store := func(id compact.NodeID, hash []byte) {
+			if id.Level == logStrataDepth && id.Index == 0 {
+				// no space for the root in the node cache
+				return
+			}
+
+			// Don't put leaves into the internal map and only update if we're rebuilding internal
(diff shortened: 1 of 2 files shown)


storage/cache/log_subtree_cache.go
```diff
@@ -58,6 +58,23 @@ func populateLogSubtreeNodes(hasher hashers.LogHasher) storage.PopulateSubtreeFu
 		if st.InternalNodes == nil || len(st.Leaves) == maxLeaves {
 			st.InternalNodes = make(map[string][]byte)
 		}
+		store := func(id compact.NodeID, hash []byte) {
```
Martin2112: Is there a clearer name than 'store'? Maybe not.

pavelkalinnikov:
The PR description already mentions why it is needed: this function will be used more than once soon.

pavelkalinnikov:
If just reading in isolation, it's maybe slightly better stylistically (populating the map is defined next to the map's definition). But the key reason is still: it's a preparation for a following PR - there will be 2 calls that will need to use the same `store` function. I could include it to the next PR, but just wanted to do it separately to not overload it.

storage/cache/log_subtree_cache.go
pavelkalinnikov: Can't think of a better one :) Could be a longer one, but it seems pretty accurate if considered in context of its signature: "store a node with the specified ID and hash into a map".

Issue #1297, opened 2018-09-18
Need to automate database schema updates


RJPercival:
When we have changed the database schema in the past, we have expected users to either delete and recreate their database (e.g. with scripts/resetdb.sh), or apply schema changes manually (after reverse-engineering them from commits). Users have complained that this (amongst other things) makes it painful to update Trillian. Instead, we should provide database schema migration scripts, e.g. a directory of SQL files containing files such as "increase_display_name_length.sql":
```sql
ALTER TABLE Trees ALTER COLUMN DisplayName VARCHAR (100)
```

We would no longer change the schema itself, but instead add SQL migration scripts to apply the changes we'd like. This does, admittedly, make seeing what the final schema looks like a bit harder - I'm open to suggestions on how to improve this.

RJPercival:
Skeema works for MySQL schema updates, but not for Cloud Spanner (or Postgres, but that's not a supported storage backend at the moment). I've experimented with using [github.com/golang-migrate/migrate](https://github.com/golang-migrate/migrate) instead (https://github.com/RJPercival/trillian/commit/b1a4594d07b5b3a68f7c10efe8d9355c4aaed2fa), which does support Cloud Spanner, but it relies on us writing migration scripts and shipping them in the binaries, and managing a version number for each database schema (so it's more work than Skeema to maintain).

RJPercival:
I investigated doing migrations using Terraform, but this isn't supported for either [Cloud Spanner](https://github.com/terraform-providers/terraform-provider-google/issues/2082) or [MySQL](https://github.com/hashicorp/terraform/issues/10740) (using the most widely used providers, at least).

mhutchinson:
We aren't making schema changes recently. What we'll do to somewhat address this is add the schemas to a restricted code-owner review, and ensure that no breaking schema changes are made without some kind of documentation or process on upgrading.

mhutchinson:
Fixed by #2746

PR #1594, opened 2019-05-13
Use Skeema to manage MySQL schema changes

RJPercival:
Use [Skeema](https://www.skeema.io/) to manage our MySQL schema changes. It will enforce a canonical style on our schema (via `skeema lint`) and ensure that no backwards-incompatible changes are made. It simplifies application of schema changes by simply requiring the user to execute `skeema add-environment` (to configure it with the address of their database) and `skeema push`. In a future change, I intend to add a sidecar to our Kubernetes configs to allow it to automatically update the database schema.

I had to manually replace `CREATE [UNIQUE] INDEX` statements with `[UNIQUE] KEY` options within the `CREATE TABLE` statements, because Skeema doesn't support the former.

I've had to move comments out of the `CREATE TABLE` statements and place them before the statement instead. This is because `Skeema` rewrites those statements when linting and would remove the comments. I experimented with recording those comments as column documentation using the `COMMENT` option, which is compatible with Skeema, but it would have the undesirable side effect of causing Skeema to drop and then recreate indexes to which we add comments, which would be rather expensive on large tables. Therefore, I haven't done this (see cb80c279c72f469d1450e19b5337963314c3740b if you're interested).

Skeema deletes the `IF NOT EXISTS` part of the `CREATE TABLE` statements because it doesn't support them. This should be fine if we switch to using Skeema, because it handles this functionality (creating or altering tables as necessary).

Contributes to #1297. 

### Checklist

### DIFF
--- a/scripts/lintdb.sh
+++ b/scripts/lintdb.sh
@@ -4,7 +4,7 @@
 #
 # This tool lints and diffs the local copy of an SQL schema file against the
 # version stored in the specified branch ("master" by default). If linting
-# detects a problem or the diff identifies a backwards-incompatble change,
+# detects a problem or the diff identifies a backwards-incompatible change,
 # the tool will exit with a non-zero exit code.
 
 die() {
--- a/scripts/lintdb.sh
+++ b/scripts/lintdb.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+# Usage: lintdb.sh [branch] [SQL file]
+#
+# This tool lints and diffs the local copy of an SQL schema file against the
+# version stored in the specified branch ("master" by default). If linting
+# detects a problem or the diff identifies a backwards-incompatble change,
+# the tool will exit with a non-zero exit code.
+
+die() {
+  echo $@ > /dev/stderr
+  exit 1
+}
+
+schema_stashed=false
+
+# Saves the given directory by stashing it if it contains uncommitted changes.
+# Will automatically restore those changes on exit.
+save_schema() {
+  local schema_dir=${1:?Usage: save_schema <schema dir>}
+  local output=$(git diff-index --exit-code HEAD -- ${schema_dir} 2>&1)
+  case $? in
+    0)
+      # No uncommitted changes to database schema
+      return 0
+      ;;
+    1)
+      # Uncommitted changes to database schema - stash them
+      git stash push --quiet -- ${schema_dir} ||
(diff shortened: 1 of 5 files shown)


storage/mysql/schema/.skeema
```diff
@@ -0,0 +1,9 @@
+default-character-set=utf8mb4
```
Martin2112: It might be irritating for example if people have to add a prod section here or edit the defaults to match their setup. Would they have their own config file in that case?

scripts/lintdb.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+
+# Usage: lintdb.sh [branch] [SQL file]
+#
+# This tool lints and diffs the local copy of an SQL schema file against the
+# version stored in the specified branch ("master" by default). If linting
+# detects a problem or the diff identifies a backwards-incompatble change,
```
Martin2112: typo.

storage/mysql/schema/.skeema
RJPercival: This should work as-is for dev purposes but yes, they'd have to add a section if they want to use it to push updates to their prod database (either by editing this file or using `skeema add-environment`). I'm thinking that I'd automate this for Kubernetes deployments.

RJPercival: The alternative would be using a library that executes database migrations from within the server/signer processes, so it already knows the database connection details (e.g. https://github.com/golang-migrate/migrate).

Martin2112: If you had e.g. a different server port do you have to keep this file modified locally? Not sure if there is a way to manage this.

Martin2112: Mixing dev / prod configs in the same file could also be a recipe for disaster, it was just an example.

RJPercival: Unfortunately yes - Skeema doesn't let you override the host or post on the command-line. I don't think mixing dev and prod configs in one file would be too problematic, since you have to specify the config name whenever you invoke a command, so it's fairly explicit about which config you're using.

RJPercival: Oh actually, you can have multiple .skeema files and they'll get combined, but they have to be in subdirectories. So it'd be possible to add subdirectories for each environment and .gitignore them.

PR #1596, opened 2019-05-14
Delete Dockerfile.db

RJPercival:
This file has been superceded by examples/deployment/docker/db_server.

### Checklist


gdbelvin:
I think this file might be relied on by the KT docker images. Could you update the KT references as well?

RJPercival:
I can't see any references to it in Key Transparency: https://github.com/google/keytransparency/search?q=Dockerfile.db&unscoped_q=Dockerfile.db

It appears that KT has already migrated to using the db_server image instead.

gdbelvin:
Thanks for checking

PR #1598, opened 2019-05-14
sequencer: Cache compact.Tree between sequencing runs

pavelkalinnikov:
This change adds `compact.Tree` caching to sequencer. This should significantly
reduce cost of sequencing, because most of the time sequencer does many runs
in a row.

### Checklist

### DIFF
--- a/log/cache.go
+++ b/log/cache.go
@@ -1,0 +1,60 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"bytes"
+
+	"github.com/google/trillian/merkle/compact"
+	"github.com/google/trillian/types"
+)
+
+// TreeCache caches per-tree information between sequencing runs.
+//
+// TODO(pavelkalinnikov): Support some cache clearing mechanism.
+type TreeCache struct {
+	cr map[int64]*compact.Range
+}
+
+// NewTreeCache returns a new tree cache.
+func NewTreeCache() TreeCache {
+	return TreeCache{cr: make(map[int64]*compact.Range)}
+}
+
+// CompactRange returns the cached compact range for the specified tree ID. If
+// the passed in root doesn't match the cached tree (for example, the root has
+// moved forward because another master took over sequencing in the meantime),
+// then nil is returned. If the information is not found, nil is returned.
+func (tc TreeCache) CompactRange(treeID int64, root *types.LogRootV1) *compact.Range {
+	cr, ok := tc.cr[treeID]
+	if !ok {
+		return nil
+	}
+	if cr.Begin() != 0 || cr.End() != root.TreeSize {
+		delete(tc.cr, treeID)
+		return nil
+	}
+	hash, err := cr.GetRootHash(nil)
+	if err != nil || !bytes.Equal(hash, root.RootHash) {
+		delete(tc.cr, treeID)
+		return nil
+	}
+	return cr
+}
+
+// Update puts the passed in compact tree into the cache.
+func (tc TreeCache) Update(treeID int64, tree *compact.Range) {
+	tc.cr[treeID] = tree
+}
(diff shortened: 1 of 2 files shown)


pavelkalinnikov:
@Martin2112 @AlCutter What do you think about this idea?

pavelkalinnikov:
Caching is a consequence of `IntegrateBatch` method being stateless. An alternative approach is having per-tree `Sequencer` objects which eliminates need for caching (the "cached" things could be just stored in the `Sequencer` itself). I think the latter approach is better, because it eliminates the need in syncing lifetimes of cached objects and mastership.

Martin2112:
It may be time for some history from the dim and distant past of those long-forgotten days in 2016 ... 

Not retaining state was done for several reasons. 1.) Not holding something that could get corrupted in RAM 2.) Reduced need to think about various complicated distributed failure modes. 3.) Limited time and resources to implement made us pretty cautious.

So fast forward to the glorious days of 2019. The kind of questions that should be asked are whether this opens up any possible ways to go wrong that don't currently exist. I think the versioning protects against most stale database / accidental multiple master cases but it's worth thinking it through with worst case assumptions.

If we're going to do this I think it should check that what comes out of the cache appears to be the same as what went in. Also what happens if time on the machine appears to have gone backwards? I think there is a check somewhere for writing an older than current root but not sure it covers this. It might be possible make it either write an incorrect timestamp or get stuck failing to update the root until it resigns and a new signer takes over.

pavelkalinnikov:
> 1.) Not holding something that could get corrupted in RAM
> ...
> If we're going to do this I think it should check that what comes out of the cache appears to be the same as what went in.

This is a good point. In this PR I check tree size and root hash of the `compact.Tree` to match the latest signed log root read from the storage. Assuming the hash is cryptographic, a match is a strong proof that nothing has been corrupted.

> Also what happens if time on the machine appears to have gone backwards? I think there is a check somewhere for writing an older than current root but not sure it covers this. It might be possible make it either write an incorrect timestamp or get stuck failing to update the root until it resigns and a new signer takes over.

This is an orthogonal question, as the cache only stores hashes. This PR doesn't change behavior w.r.t. timestamps.

Martin2112:
Yes it currently checks what it uses so could be ok but otherwise it would be safer to hang on to the serialized format and verify each time it's used. That may be overkill.

The timestamp thing is an example as there's a timestamp in `LogRootV1`. I don't think the code checks that so if multiple roots are written at the same tree size the cached root gets out of date? Again I don't think that timestamp gets used for anything but if it did then it the door would be open for something along the lines of the bug that caused us an ~8 hour log outage a few years ago.

It's not a bad idea. We just need to make sure we've thought through how it might break. If it can't that's great but we have to do at least do it.

log/cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/google/trillian/merkle/compact"
+	"github.com/google/trillian/types"
+)
+
+// TreeCache caches per-tree information between sequencing runs.
+//
+// TODO(pavelkalinnikov): Support some cache clearing mechanism.
+type TreeCache struct {
+	ct map[int64]*compact.Tree
```
AlCutter: You may want to use `sync.Map` here, or otherwise protect with `sync.Mutex`?

log/sequencer.go
```diff
@@ -401,7 +405,7 @@ func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limi
 		}
 
 		// Collate node updates.
-		nodeMap, err := s.updateCompactTree(merkleTree, sequencedLeaves, label)
+		nodeMap, err := s.updateCompactTree(merkleTree, tree.TreeId, sequencedLeaves, label)
```
AlCutter: _Ideally_ you probably only want this to happen once the `tx` has successfully committed?

It's probably fine because on a rollback the replayed root hash won't match the cached one so you'll end up having to init the cache from storage, but ...

log/cache.go
pavelkalinnikov: Fantastic point.

log/sequencer.go
pavelkalinnikov: That was precisely my thinking. But it's probably still not a bad idea to make it safer. OTOH, it's not trivial: note that the returned `compact.Tree` is mutable, and we can't rollback these updates without some sort of "clone" function.

pavelkalinnikov: This triggers another interesting observation: this code can be retried, as it's within a `RunInTransaction` function. If it is retried, the 1st retry will fail (and clear the cache) because of `compact.Tree` mutability and root hash mismatch, but the 2nd retry has a chance of succeeding again.

pavelkalinnikov: Oh wait, ignore the last comment. The 1st retry will not fail, it will only be forced to reload the `compact.Tree` from storage.

pavelkalinnikov:
Closing in favor of a larger refactoring: #1640

PR #1601, opened 2019-05-15
Switch some low-level logging in hashing related code to 'if' guards

Martin2112:
Didn't expect much gain as the arguments involved don't require complex evaluation. However the CONIKS hasher turned out faster by ~10% using this.

Results were repeatable across multiple runs.

Before:

BenchmarkHStar2Root-12    	     200	  47001010 ns/op
BenchmarkHStar2Root-12    	     200	  46416230 ns/op
BenchmarkHStar2Root-12    	     200	  47124380 ns/op

After:

BenchmarkHStar2Root-12    	     200	  41636550 ns/op
BenchmarkHStar2Root-12    	     200	  41859620 ns/op
BenchmarkHStar2Root-12    	     200	  42533060 ns/op

### Checklist


Martin2112:
Sorry I messed that up and broke it. Should be fixed now.

PR #1602, opened 2019-05-15
Just one copy of bytes from big.Int in NewNodeIDFromBigInt

AlCutter:
`big.Int.Bytes()` makes a new copy of the underlying data, so we shouldn't call it 3 times in the same function!

Might be worth investigating whether calling `Bits()` instead, and doing the conversion ourselves into `path` is worth the saving of one more copy...

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -173,15 +173,17 @@ func NewNodeIDFromBigInt(depth int, index *big.Int, totalDepth int) NodeID {
 		panic(fmt.Sprintf("storage NewNodeFromBitInt(): totalDepth mod 8: %v, want %v", got, want))
 	}
 
+	// TODO(al): We _could_ use Bits() and avoid the extra copy/alloc.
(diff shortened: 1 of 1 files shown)


Martin2112:
Yeah I wish there were better tools for analyzing memory usage.

Martin2112:
Seems like a nice win. Over 15% on the hstar2 root benchmark.

```
BenchmarkHStar2Root-12    	     200	  46194138 ns/op (before)
BenchmarkHStar2Root-12    	     200	  39229419 ns/op (after)
```

gdbelvin:
Post commit comment -- is this a kind of thing that the compiler should optimize away? -- or no because the compiler doesn't know if there are side effects?

Martin2112:
Probably hard to do at compile time for general cases as it needs knowledge of internals of the methods called. Could be detectable by tools that know how the Go library APIs work.

PR #1613, opened 2019-05-16
Couple of changes to make NodeIDs more frugal.

AlCutter:
This change causes less memory to be allocated when manipulating NodeIDs:
 -  Introduces `NodeID.Suffix()` to be used rather than `NodeID.Split()`when only the suffix is required, saves ~30% memory compared to `Split`,
 - `storage.NewSuffix()` now allocates ~90% less memory, and is 15% faster

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -402,29 +402,39 @@ func NewNodeIDFromPrefixSuffix(prefix []byte, suffix *Suffix, maxPathBits int) N
 	}
 }
 
+// Suffix returns a Node's suffix starting at prefixBytes.
+// This is the same Suffix that Split() would return, just without the overhead
+// of also creating the prefix.
+func (n *NodeID) Suffix(prefixBytes, suffixBits int) *Suffix {
+	if n.PrefixLenBits == 0 {
+		return EmptySuffix
+	}
+
+	b := n.PrefixLenBits - prefixBytes*8
+	if b > suffixBits {
+		panic(fmt.Sprintf("Suffix(): %x(n.PrefixLenBits: %v - prefixBytes: %v *8) > %v", n.Path, n.PrefixLenBits, prefixBytes, suffixBits))
+	}
+	if b <= 0 {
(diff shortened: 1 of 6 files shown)


storage/types.go
```diff
@@ -420,11 +418,12 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, *Suffix) {
...
 	maskIndex := (b - 1) / 8
 	maskLowBits := (b-1)%8 + 1
-	sfxPath := a[prefixBytes : prefixBytes+suffixBytes]
+	sfxPath := make([]byte, suffixBytes)
+	copy(sfxPath, n.Path[prefixBytes:prefixBytes+suffixBytes])
 	sfxPath[maskIndex] &= ((0x01 << uint(maskLowBits)) - 1) << uint(8-maskLowBits)
 	sfx := NewSuffix(byte(b), sfxPath)
 
-	return a[:prefixBytes], sfx
+	return n.Path[:prefixBytes], sfx
```
Martin2112: Slightly worried about that one. Some of the NodeID methods are mutators. If they share path slices this could break.

AlCutter: Fair point - I've reverted this, and factored out the suffix creation bit into its own method and switched over to that in the places where we discard the prefix anyway.

phad: Defensive copying LGTM.

storage/types_test.go
```diff
@@ -876,3 +876,19 @@ func BenchmarkFlipRightBit(b *testing.B) {
...
+	for i := 0; i < b.N; i++ {
+		_, _ = n.Split(10, 176)
+	}
+}
+
+func BenchmarkSuffix(b *testing.B) {
+	n := NewNodeIDFromHash(h2b("0000000000000000000000000000000000000000000000000000000000000001"))
+	n.PrefixLenBits = 256
+	for i := 0; i < b.N; i++ {
+		_ = n.Suffix(10, 176)
```
phad: For my info, what's special about benchmarking with 10, 176?

AlCutter: Nothing, it's just an arbitrary node id

PR #1615, opened 2019-05-17
Parallelize VerifyMapLeavesResponse

gdbelvin:
Each inclusion proof can be verified independently which should result in an ~12x speedup

### Checklist

### DIFF
--- a/client/map_verifier.go
+++ b/client/map_verifier.go
@@ -18,6 +18,8 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
+	"runtime"
+	"sync"
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keys/der"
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Is this function used anywhere?

client/map_verifier.go
```diff
@@ -108,11 +110,47 @@ func (m *MapVerifier) VerifyMapLeavesResponse(indexes [][]byte, revision int64,
...
+				return
+			}
+		}
+	}()
+	errors := make(chan error)
+	go func() {
+		defer close(errors)
+		var wg sync.WaitGroup
+		defer wg.Wait()
+		for w := 0; w < runtime.NumCPU(); w++ {
```
pavelkalinnikov: What are the chances that `len(resp.MapLeafInclusion) < runtime.NumCPU()`? Maybe limit the number of goroutines with the number of items?

gdbelvin:
I noticed that this was taking a semi large amount of time when verifying ~1000 inclusion proofs in the Key Transparency Sequencer.  

1) This might be an indication that we need to come up with a more efficient inclusion proof
  a) Removing the many empty branch hashes 
  b) Removing the duplicate hashes near the root of the tree. 
2) This is an optimization that only makes the most sense when we're dealing with a large number of leaves.

PR #1618, opened 2019-05-20
compact: Simplify getting hashes in NewTreeWithState

pavelkalinnikov:
This change factors out getting the list of node IDs needed to initialize `compact.Tree`
into a separate exported function.

This eliminates `GetNodesFunc` callback which makes client code simpler, and also
allows a more direct initialization of `compact.Range`.

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -133,40 +133,35 @@ func NewSequencer(
 }
 
 func (s Sequencer) buildMerkleTreeFromStorageAtRoot(ctx context.Context, root *types.LogRootV1, tx storage.TreeTX) (*compact.Tree, error) {
-	mt, err := compact.NewTreeWithState(s.hasher, int64(root.TreeSize), func(ids []compact.NodeID) ([][]byte, error) {
-		storIDs := make([]storage.NodeID, len(ids))
-		for i, id := range ids {
-			nodeID, err := storage.NewNodeIDForTreeCoords(int64(id.Level), int64(id.Index), maxTreeDepth)
-			if err != nil {
-				return nil, fmt.Errorf("failed to create nodeID: %v", err)
-			}
-			storIDs[i] = nodeID
-		}
-
-		nodes, err := tx.GetMerkleNodes(ctx, int64(root.Revision), storIDs)
+	ids := compact.TreeNodes(root.TreeSize)
+	storIDs := make([]storage.NodeID, len(ids))
+	for i, id := range ids {
+		nodeID, err := storage.NewNodeIDForTreeCoords(int64(id.Level), int64(id.Index), maxTreeDepth)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get Merkle nodes: %v", err)
+			return nil, fmt.Errorf("failed to create nodeID: %v", err)
 		}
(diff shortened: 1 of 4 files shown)


log/sequencer.go
```diff
@@ -241,8 +236,12 @@ func (s Sequencer) initMerkleTreeFromStorage(ctx context.Context, currentRoot *t
 		return compact.NewTree(s.hasher), nil
 	}
 
-	// Initialize the compact tree state to match the latest root in the database
-	return s.buildMerkleTreeFromStorageAtRoot(ctx, currentRoot, tx)
+	// Initialize the compact tree state to match the latest root in the database.
+	mt, err := s.buildMerkleTreeFromStorageAtRoot(ctx, currentRoot, tx)
+	if err != nil {
```
Martin2112: This is going to spam the log if something goes wrong. Unfortunately the glog api doesn't have rate limiting. The failed signing run metric might be enough.

Maybe leave it in for now and TODO it for review later.

merkle/compact/tree.go
```diff
@@ -210,3 +186,22 @@ func (t *Tree) getNodes() [][]byte {
...
+	// Iterate over perfect subtrees along the right border of the tree. Those
+	// correspond to the bits of the tree size that are set to one.
+	for sz := size; sz != 0; sz &= sz - 1 {
+		level := uint(bits.TrailingZeros64(sz))
+		index := (sz - 1) >> level
+		ids = append(ids, NewNodeID(level, index))
+	}
+	// Note: Right border nodes of compact.Range are ordered from root to leaves.
+	for i, j := 0, len(ids)-1; i < j; i, j = i+1, j-1 {
+		ids[i], ids[j] = ids[j], ids[i]
```
Martin2112: Just a thought but we could add the other ordering to Range if reversal is going to crop up a lot.

pavelkalinnikov: Good thought, we'll see if/when another use-case appears. That wouldn't be a trivial change though. `Range` implementation has a very specific ordering: `hashes` slice contains 2 **stacks** of hashes - the front one is ordered bottom-to-top, the back one is ordered top-to-bottom. Stack ordering comes handy when 2 ranges are merged, so that the slice is mutated with minimal copies/reallocations.

log/sequencer.go
pavelkalinnikov: This PR doesn't change errors or `glog` behaviour, so out of scope. The amount of spam is (as before) one line per failed signing run (written somewhere up the stack),  which I think is reasonable.

Martin2112: Is it the same? It looked to be adding extra logging. 

To be clear I was just asking for a TODO, not to change glog or anything else.

pavelkalinnikov: No extra logging: Ctrl+F "glog." finds nothing. Changes in `sequencer.go` are mostly about indentation: before the code was in a callback, now it's not.

Martin2112: It's not clear from the diffs. It's just moved around from one func to another? OK then.

pavelkalinnikov: Yes. Github diff is not great here.

PR #1619, opened 2019-05-20
Remove FlipRightBit and fold into Neighbor func.

Martin2112:
Reduces the number of mutators for NodeID by one. Continuing to try to rationalize this API.

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -377,13 +377,6 @@ func (n *NodeID) Copy() *NodeID {
 	}
 }
 
-// FlipRightBit flips the ith bit from LSB
-func (n *NodeID) FlipRightBit(i int) *NodeID {
-	bIndex := (n.PathLenBits() - i - 1) / 8
-	n.Path[bIndex] ^= 1 << uint(i%8)
-	return n
-}
-
 // leftmask contains bitmasks indexed such that the left x bits are set. It is
 // indexed by byte position from 0-7 0 is special cased to 0xFF since 8 mod 8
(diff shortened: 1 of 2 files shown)


gdbelvin:
Does this make the API harder to test?

storage/types.go
```diff
@@ -412,9 +405,13 @@ func (n *NodeID) MaskLeft(depth int) *NodeID {
 // Neighbor returns the same node with the bit at PrefixLenBits flipped.
 // In terms of a tree traversal, this is the parent node's other child node
 // in the binary tree (often termed sibling node).
+// TODO(Martin2112): This is only used by Siblings() in conjunction with
+// MaskLeft() and could be cleaned up further.
 func (n *NodeID) Neighbor() *NodeID {
 	height := n.PathLenBits() - n.PrefixLenBits
-	return n.FlipRightBit(height)
+	bIndex := (n.PathLenBits() - height - 1) / 8
```
gdbelvin: Maybe add a comment that this is an inlined FlipRightBit function?

gdbelvin: For context, I found all of these manipulation functions very hard to understand and so I broke them down into the smallest functions I could grok and unit test.

Martin2112: These are internal APIs that most users won't see. That said the documentation wasn't as clear as it could be.

The goal is to remove all functions that mutate node ids in place, which prevents their representations from being shared and blocks storage optimizations.

PR #1622, opened 2019-05-21
Clean up compact.Tree tests

pavelkalinnikov:


pavelkalinnikov:
You are right, I will do so in the future. Thanks for the review.

PR #1624, opened 2019-05-22
compact.Tree: Factor out root verification method

pavelkalinnikov:
This change makes root verification in `compact.Tree` optional,
by factoring it out from the constructor.


pavelkalinnikov:
PTAL.

pavelkalinnikov:
Obsolete. I'm going forward replacing `compact.Tree` with `compact.Range`: #1510.

PR #1626, opened 2019-05-22
Avoid an unnecessary Split() in preload()

Martin2112:
The prefix can be used directly from the NodeID path slice, it's only used to create a hash table key. The code already effectively assumes that the prefix is a multiple of 8 bits. For one thing the subtree strata store prefix lengths in terms of bytes not bits.

Not sure how good the tests are for this code. I'll investigate.

### Checklist


AlCutter:
I started on a related change just before I went away (#1636) which I think does this too (as well as trims work in a few other places)

Martin2112:
I'll close this then, the other one is neater.

PR #1627, opened 2019-05-22
Fix bazel build

pavelkalinnikov:


WORKSPACE
```diff
@@ -24,14 +24,22 @@ load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
...
 
 http_archive(
-    name = "googleapi",
-    url = "https://github.com/googleapis/googleapis/archive/master.zip",
+    name = "com_google_googleapis",
     strip_prefix = "googleapis-master",
+    url = "https://github.com/googleapis/googleapis/archive/master.zip",
+)
+
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
```
RJPercival: Instead of renaming "googleapi" to "com_google_googleapis", could we just have used:
```
load("@googleapi//:repository_rules.bzl", "switched_rules_by_language")

switched_rules_by_language(
    name = "googleapi_imports",
    go = True,
    java = True,
    python = True,
)
```

pavelkalinnikov: I think the answer is no. This is the error line from the failing builds:

`ERROR: Analysis of target '//:trillian_log_api_proto' failed; build aborted: error loading package '@googleapi//google/rpc': Extension file not found. Unable to load package for '@com_google_googleapis_imports//:imports.bzl': The repository could not be resolved`

It looks like the `google/rpc` thing requires the `com_google_googleapis_imports` symbol. This is also the reason why I renamed `googleapi` to `com_google_googleapis` - to make it "canonical".

PR #1628, opened 2019-05-23
Use uint64 throughout compact.Tree

pavelkalinnikov:
This change replaces signed `int64` -> `uint64` in `compact.Tree`, to make
it consistent with `compact.Range`. It also makes all the code that utilizes the
type use `uint64` wherever possible, and adds safety checks where
conversion is strictly necessary (e.g. `LeafIndex` field population).

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -167,7 +167,7 @@ func (s Sequencer) initMerkleTreeFromStorage(ctx context.Context, root *types.Lo
 		hashes[i] = node.Hash
 	}
 
-	return compact.NewTreeWithState(s.hasher, int64(root.TreeSize), hashes, root.RootHash)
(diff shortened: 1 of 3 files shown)


gdbelvin:
hmm -- In my experience `uint64` causes more pain than it is worth because many standard libraries and generated protos accept `int64`, not `uint64`.  Is there a strong reason to use `uint64`? If consistency is the motivation for this PR, have you considered switching the other bits  of code to `int64`?

pavelkalinnikov:
Much of the arithmetic in `compact.Range` uses `bits` package with `uint64` methods - that was one of the reasons why I chose to use it throughout these files consistently. It's okay to use `int64` in other places (like protos), but it's not this package's concern - the client needs to make sure the size doesn't overflow (which probably will never happen anyway).

gdbelvin:
Please update with a description of what packages will be uint64 based and what packages will be int64 based.

PR #1633, opened 2019-05-24
Remove all usages of compact.Tree

pavelkalinnikov:
This change removes the remaining dependencies from `compact.Tree`,
by using `compact.Range` instead.

### DIFF
--- a/integration/log.go
+++ b/integration/log.go
@@ -514,15 +514,16 @@ func makeGetLeavesByIndexRequest(logID int64, startLeaf, numLeaves int64) *trill
 }
 
 func buildMemoryMerkleTree(leafMap map[int64]*trillian.LogLeaf, params TestParameters) (*merkle.InMemoryMerkleTree, error) {
-	// Build the same tree with two different Merkle implementations as an additional check. We don't
-	// just rely on the compact tree as the server uses the same code so bugs could be masked
+	// Build the same tree with two different Merkle tree implementations as an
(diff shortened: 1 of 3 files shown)


integration/log.go
```diff
@@ -516,22 +516,28 @@ func makeGetLeavesByIndexRequest(logID int64, startLeaf, numLeaves int64) *trill
 func buildMemoryMerkleTree(leafMap map[int64]*trillian.LogLeaf, params TestParameters) (*merkle.InMemoryMerkleTree, error) {
 	// Build the same tree with two different Merkle implementations as an additional check. We don't
 	// just rely on the compact tree as the server uses the same code so bugs could be masked
```
Martin2112: comment can be updated

integration/log.go
```diff
@@ -516,22 +516,28 @@ func makeGetLeavesByIndexRequest(logID int64, startLeaf, numLeaves int64) *trill
...
 	}
 
 	// If the two reference results disagree there's no point in continuing the checks. This is a
 	// "can't happen" situation.
-	root, err := compactTree.CurrentRoot()
+	root, err := cr.GetRootHash(nil)
 	if err != nil {
-		return nil, fmt.Errorf("failed to compute compact tree root: %v", err)
+		return nil, fmt.Errorf("failed to compute compact range root: %v", err)
+	} else if cr.End() == 0 {
```
Martin2112: For Go style this probably shouldn't be an `else if`. It is annoying.

PR #1635, opened 2019-05-24
Update codecov config. Remove regexes for top level folders.

Martin2112:
Listing the path should be enough.

### Checklist


Martin2112:
Has this broken code coverage?

Martin2112:
The "no coverage uploaded" message was maybe because the diff didn't affect any code.

PR #1636, opened 2019-05-24
Add methods for individually getting at prefix/suffix of NodeID

AlCutter:
This PR adds finer-grained accessors for getting at Prefix and Suffix parts of a given NodeID, rather than forcing the caller to pay the overhead cost of calling `Split` if only one of the parts is required.

It also provides a `PrefixAsKey` method as a means to avoid a double allocation.

### Checklist

### DIFF
--- a/storage/types.go
+++ b/storage/types.go
@@ -469,7 +469,7 @@ func (n *NodeID) Suffix(prefixBytes, suffixBits int) *Suffix {
 	return NewSuffix(byte(b), sfxPath)
 }
 
-// Prefix returns a NodeID's prefix.
+// Prefix returns a copy of NodeID's prefix.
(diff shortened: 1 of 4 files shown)


storage/types.go
```diff
@@ -469,6 +469,29 @@ func (n *NodeID) Suffix(prefixBytes, suffixBits int) *Suffix {
 	return NewSuffix(byte(b), sfxPath)
 }
 
+// Prefix returns a NodeID's prefix.
+// This is the same value that would be returned from Split, but without the
+// overhead of calulating the suffix too.
```
Martin2112: Prob worth documenting that it returns a copy. And possibly in Split().

AlCutter: 👍

PR #1642, opened 2019-05-28
Fold the MaskLeft operation into Neighbour.

Martin2112:
This means Neighbour no longer mutates the input node. There is no
additional copying as MaskLeft was always previous applied.

### Checklist


Martin2112:
Some sort of `const` would be useful. But it would probably require other changes as the slice API is not very `const` friendly.

PR #1646, opened 2019-05-29
log: Delete unnecessary mutex

pavelkalinnikov:
There is only one sequential pattern of accessing "held IDs" in `OperationManager`,
so the mutex is not needed.

### DIFF
--- a/log/operation_manager.go
+++ b/log/operation_manager.go
@@ -108,7 +108,6 @@ type OperationManager struct {
 	pendingResignations chan election.Resignation
 	runnerWG            sync.WaitGroup
 	tracker             *election.MasterTracker
-	heldMutex           sync.Mutex
 	lastHeld            []int64
 	// Cache of logID => name; assumed not to change during runtime
 	logNamesMutex sync.Mutex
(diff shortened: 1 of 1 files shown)


log/operation_manager.go
```diff
@@ -254,8 +253,6 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
 // updateHeldIDs updates the process status with the number/list of logs that
 // the instance holds mastership for.
 func (o *OperationManager) updateHeldIDs(ctx context.Context, logIDs, activeIDs []int64) {
-	o.heldMutex.Lock()
-	defer o.heldMutex.Unlock()
 	heldInfo := o.heldInfo(ctx, logIDs)
 	msg := fmt.Sprintf("Acting as master for %d / %d active logs: %s", len(logIDs), len(activeIDs), heldInfo)
 	if !reflect.DeepEqual(logIDs, o.lastHeld) {
```
Martin2112: I'm now suspicious of this as if the ID ordering changes then it will compare false and there is no ordering on the query used to get the IDs.  I think it's probably updating the process state more often than necessary.

Possibly this should come directly from the tracker by notifications and remove the `lastHeld` stuff.

pavelkalinnikov: Yeah, good point. Leaving it for another PR.

PR #1647, opened 2019-05-30
GetLeavesByRevisionNoProofs API

gdbelvin:
This allows clients inside the trust boundary to avoid duplicate work

### Checklist


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+
+### OmitInclusionProof Option
+Allow map clients to forgo fetching inclusion proofs.
```
Martin2112: How often will this be set? Is it better to invert it?

Martin2112: Or you could have different methods rather than parameters that drastically effect performance.

gdbelvin: By constructing the option in this way, existing methods that don't set this option will keep their existing behavior.  That said, double negatives are admittedly an anti pattern.

gdbelvin: We expect this flag to not be set the vast majority of the time.

gdbelvin: I'll make this a new API endpoint

integration/maptest/map.go
```diff
@@ -573,6 +575,37 @@ func RunInclusion(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdm
...
+	}
+	batchSize := 10
+	numBatches := 2
+
+	leafMap := writeBatch(ctx, t, tmap, tree, batchSize, numBatches)
+	indexes := make([][]byte, 0, len(leafMap))
+	leaves := make([]*trillian.MapLeaf, 0, len(leafMap))
+	for i, l := range leafMap {
+		indexes = append(indexes, []byte(i))
+		leaves = append(leaves, l)
```
golangcibot: SA4010: this result of append is never used, except maybe in other appends (from `staticcheck`)

server/map_rpc_server.go
```diff
@@ -156,6 +156,34 @@ func (t *TrillianMapServer) GetLeavesByRevision(ctx context.Context, req *trilli
 	return t.getLeavesByRevision(ctx, req.MapId, req.Index, req.Revision)
 }
 
+// GetLeavesByRevisionNoProof implements the GetLeavesByRevision RPC method.
```
Martin2112: Can we add tests for the new RPC?

docs/api.md
```diff
@@ -1052,6 +1068,7 @@ defined in the Verifiable Data Structures paper.
 | GetLeafByRevision | [GetMapLeafByRevisionRequest](#trillian.GetMapLeafByRevisionRequest) | [GetMapLeafResponse](#trillian.GetMapLeafResponse) |  |
 | GetLeaves | [GetMapLeavesRequest](#trillian.GetMapLeavesRequest) | [GetMapLeavesResponse](#trillian.GetMapLeavesResponse) |  |
 | GetLeavesByRevision | [GetMapLeavesByRevisionRequest](#trillian.GetMapLeavesByRevisionRequest) | [GetMapLeavesResponse](#trillian.GetMapLeavesResponse) |  |
+| GetLeavesByRevisionNoProof | [GetMapLeavesByRevisionRequest](#trillian.GetMapLeavesByRevisionRequest) | [MapLeaves](#trillian.MapLeaves) | GetLeavesByRevisionNoProof returns the requested map leaves without inclusion proofs. This API is designed for internal use where verificaion is not needed. |
```
Martin2112: Typo.

server/map_rpc_server.go
gdbelvin: Tests added

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+
+### GetLeavesByRevisionNoProof API
+Allow map clients to forgo fetching inclusion proofs.
+This dramatically speeds things up for clients that don't need verifiability.
```
Martin2112: Not sure if it's worth giving an example here.

server/map_rpc_server.go
Martin2112: There don't seem to be RPC server unit tests though.

CHANGELOG.md
gdbelvin: I don't think it is?

Martin2112: I just meant a sentence on why you might want to do this but it's not really important.

gdbelvin:
@Martin2112 Does this look ok?

PR #1649, opened 2019-05-31
Use []*trillian.MapLeaf consistently

gdbelvin:
Repeated proto fields are slices of pointers.  By using slices of
pointers throught, we can avoid multiple conversions and memcopies.

### DIFF
--- a/log/mock_operation.go
+++ b/log/mock_operation.go
@@ -34,6 +34,7 @@ func (m *MockOperation) EXPECT() *MockOperationMockRecorder {
 
 // ExecutePass mocks base method
 func (m *MockOperation) ExecutePass(arg0 context.Context, arg1 int64, arg2 *OperationInfo) (int, error) {
+	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "ExecutePass", arg0, arg1, arg2)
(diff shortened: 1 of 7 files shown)


pavelkalinnikov:
I am not too sure about such kinds of optimizations. See [this](https://segment.com/blog/allocation-efficiency-in-high-performance-go-services/) article explaining memory quirks in Go (here, in particular, you are introducing "Storing pointers or values containing pointers in a slice"). Surely you avoid memcopies of `MapLeaf` (which are quite cheap cause they don't deep-copy), but I don't know whether you are introducing more heap operations (which are presumably higher cost).

Martin2112:
Pointers seemed better in the optimizations I did according to benchmarks. Can we benchmark this perhaps?

gdbelvin:
That's a very helpful article. 

The main objective of this PR is not performance (though I do think it will help), but rather consistency.
It is awkward to convert `[]trillian.MapLeaf` to `[]*trillian.MapLeaf` in several locations.
Furthermore, in some / most cases, the storage layer at the bottom of the stack produces `[]*trillian.MapLeaf` natively, requiring us to convert it to `[]trillian.MapLeaf` and then back to `[]*trillian.MapLeaf` again.

Issue #1605, opened 2019-05-16
Log Signer hanging when select statement gets stuck


paulmattei:
Hi trillian team, we're encountering some oddities with a particular log. We see the log_signer processing speedily right until a select statement gets stuck and the log_signer halts. The select eventually times out at `300s` per our connection config. Once the query times out, the following goroutine information is dumped to the log, a rollback occurs, and the log_signer begins processing again.It's unclear why the log_signer cruises along right until it hits a wall.

```
I0515 22:11:15.082392       1 log_operation_manager.go:454] Group run completed in 0.02 seconds: 3 succeeded, 0 failed, 3 items processed
I0515 22:11:15.110065       1 sequencer.go:472] 1045204392378480099: sequenced 1 leaves, size 210937358, tree-revision 24322870
I0515 22:11:15.110078       1 log_operation_manager.go:435] 1045204392378480099: processed 1 items in 0.02 seconds (43.67 qps)
I0515 22:11:15.110091       1 log_operation_manager.go:454] Group run completed in 0.02 seconds: 3 succeeded, 0 failed, 1 items processed
I0515 22:11:15.156972       1 sequencer.go:472] 1045204392378480099: sequenced 1 leaves, size 210937359, tree-revision 24322871
I0515 22:11:15.156985       1 log_operation_manager.go:435] 1045204392378480099: processed 1 items in 0.02 seconds (45.00 qps)
I0515 22:11:15.156994       1 log_operation_manager.go:454] Group run completed in 0.02 seconds: 3 succeeded, 0 failed, 1 items processed
W0515 22:12:15.177265       1 tree_storage.go:224] Failed to get merkle subtrees: context deadline exceeded
W0515 22:12:15.177352       1 tree_storage.go:429] TX rollback error: sql: transaction has already been committed or rolled back, stack:
goroutine 2329070 [running]:
runtime/debug.Stack(0xc000222380, 0x10a5e20, 0xc000142950)
	/usr/local/go/src/runtime/debug/stack.go:24 +0xa7
github.com/google/trillian/storage/mysql.(*treeTX).rollbackInternal(0xc000176280, 0xf94dd0, 0xc000af2368)
	/go/src/github.com/google/trillian/storage/mysql/tree_storage.go:429 +0x5c
github.com/google/trillian/storage/mysql.(*treeTX).Close(0xc000176280, 0x0, 0x0)
	/go/src/github.com/google/trillian/storage/mysql/tree_storage.go:440 +0x82
[...]

PR #1650, opened 2019-06-03
 Ensure that TreeID is part of the MySQL join condition.

Martin2112:
Experiments show that this reduces query cost. May help with #1605. Hard to be sure as so far the situation isn't reproducible in the same way.

### Checklist

### DIFF
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -39,9 +39,9 @@ const (
 		 VALUES(?,?,?,?,?,?)`
 
 	selectSubtreeSQL = `
- SELECT x.SubtreeId, x.MaxRevision, Subtree.Nodes
+ SELECT x.TreeId, x.SubtreeId, x.MaxRevision, Subtree.Nodes
  FROM (
- 	SELECT n.SubtreeId, max(n.SubtreeRevision) AS MaxRevision
+ 	SELECT n.TreeId, n.SubtreeId, max(n.SubtreeRevision) AS MaxRevision
 	FROM Subtree n
 	WHERE n.SubtreeId IN (` + placeholderSQL + `) AND
 	 n.TreeId = ? AND n.SubtreeRevision <= ?
@@ -50,6 +50,7 @@ const (
  INNER JOIN Subtree 
  ON Subtree.SubtreeId = x.SubtreeId 
  AND Subtree.SubtreeRevision = x.MaxRevision 
+ AND Subtree.TreeId = x.TreeId
  AND Subtree.TreeId = ?`
 	placeholderSQL = "<placeholder>"
 )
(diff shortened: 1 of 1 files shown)


storage/mysql/tree_storage.go
```diff
@@ -39,9 +39,9 @@ const (
 		 VALUES(?,?,?,?,?,?)`
 
 	selectSubtreeSQL = `
- SELECT x.SubtreeId, x.MaxRevision, Subtree.Nodes
+ SELECT x.TreeId, x.SubtreeId, x.MaxRevision, Subtree.Nodes
```
RJPercival: Do you need to select `x.TreeId` or is just including it in the join sufficient?

Martin2112: I don't think I can reference it in the join if it's not part of the X result set. If you feel like experimenting then feel free to have a go at it.

Martin2112:
I think it should see more testing before it gets merged.

storage/mysql/tree_storage.go
RJPercival: I've just tested it and you indeed don't need `x.TreeId` here. You only need it in the inner SELECT and on the join.

storage/mysql/tree_storage.go
```diff
@@ -50,6 +50,7 @@ const (
  INNER JOIN Subtree 
  ON Subtree.SubtreeId = x.SubtreeId 
  AND Subtree.SubtreeRevision = x.MaxRevision 
+ AND Subtree.TreeID = x.TreeId
```
RJPercival:
```suggestion
 AND Subtree.TreeId = x.TreeId
```

storage/mysql/tree_storage.go
```diff
@@ -39,9 +39,9 @@ const (
 		 VALUES(?,?,?,?,?,?)`
 
 	selectSubtreeSQL = `
- SELECT x.SubtreeId, x.MaxRevision, Subtree.Nodes
+ SELECT x.TreeId, x.SubtreeId, x.MaxRevision, Subtree.Nodes
  FROM (
- 	SELECT n.SubtreeId, max(n.SubtreeRevision) AS MaxRevision
+ 	SELECT n.TreeID, n.SubtreeId, max(n.SubtreeRevision) AS MaxRevision
```
RJPercival:
```suggestion
 	SELECT n.TreeId, n.SubtreeId, max(n.SubtreeRevision) AS MaxRevision
```

Martin2112:
Did you check that the query cost was the same with that change?

storage/mysql/tree_storage.go
Martin2112: Done.

RJPercival:
Yes, it's the same cost.

RJPercival:
Was there any further testing you wanted to do before merging this?

Martin2112:
I think it can be merged. Was waiting to see if there were more comments.

PR #1653, opened 2019-06-04
Introduce experimental package for Skylog

pavelkalinnikov:


skylog/README.md
```diff
@@ -0,0 +1,6 @@
+# Skylog
+
+Skylog stands for *scalable log*. This is the new experimental design for
+Trillian logs that supports horizontal scaling. It is not for production use.
```
Martin2112: (yet)? Also might be worth at least a sentence on what horizontal scaling means in this context.

pavelkalinnikov: Done.

PR #1654, opened 2019-06-04
skylog: Introduce tree storage interfaces

pavelkalinnikov:
This change introduces storage API for Merkle trees in Skylog.

It differs from the current production tree storage API in that it doesn't
have revisions because all nodes of the tree are immutable and written
only when a corresponding perfect subtree is fully populated.

### DIFF
--- a/skylog/storage/types.go
+++ b/skylog/storage/types.go
@@ -1,3 +1,17 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


skylog/storage/types.go
```diff
@@ -0,0 +1,9 @@
+package storage
+
+import "github.com/google/trillian/merkle/compact"
+
+// Node represents a Merkle tree node.
+type Node struct {
```
Martin2112: Might want to think about making `Node` immutable, though I don't think the `compact.NodeID` is so might not help that much.

pavelkalinnikov: Could you elaborate what you mean by immutable? Like unexporting the fields and adding read-only getters?

Martin2112: Cannot be modified after construction. Making IDs opaque outside the defining package is another option, but not as easy in Go where it's all or nothing visibility. 

In `storage.NodeID` we ended up with a performance penalty copying byte slices because there were mutator methods directly on the IDs. That doesn't apply here though so the decision is really about whether it has benefits for correctness. It's also possibly too early days to say so maybe carry on like this for the moment until more of it takes shape.

PR #1655, opened 2019-06-04
New API service for writing to the Trillian Map

mhutchinson:
Separating the Verifiable Map along the concerns of reading vs writing leads to a cleaner API.

This change simply adds a new API service for writing to the map. Future changes will remove the methods from the read API when the service has been implemented and clients have been migrated.

### Checklist


trillian_map_api.proto
```diff
@@ -137,6 +137,25 @@ message SetMapLeavesResponse {
...
+message WriteMapLeavesRequest {
+  int64 map_id = 1;
+  // The leaves being set must have unique Index values within the request.
+  repeated MapLeaf leaves = 2;
+  bytes metadata = 3;
+  // The map revision to associate the leaves with. The request will fail if
+  // this revision already exists, does not match the current write revision, or
+  // is negative. If revision = 0 then the leaves will be written to the current
+  // write revision.
+  int64 revision = 4;
```
AlCutter: Would `expect_revision` be a touch clearer to reflect the check-and-set functionality this supports?

trillian_map_api.proto
```diff
@@ -137,6 +137,25 @@ message SetMapLeavesResponse {
   SignedMapRoot map_root = 2;
 }
 
+message WriteMapLeavesRequest {
+  int64 map_id = 1;
+  // The leaves being set must have unique Index values within the request.
+  repeated MapLeaf leaves = 2;
+  bytes metadata = 3;
```
AlCutter: Worth a comment on `metadata`? Presumably this is popped into the SMR?

trillian_map_api.proto
```diff
@@ -137,6 +137,25 @@ message SetMapLeavesResponse {
...
+  // The map revision to associate the leaves with. The request will fail if
+  // this revision already exists, does not match the current write revision, or
+  // is negative. If revision = 0 then the leaves will be written to the current
+  // write revision.
+  int64 revision = 4;
+}
+
+message WriteMapLeavesResponse {
+  // The map revision that the leaves will be published as.
```
AlCutter: /as/at/?

trillian_map_api.proto
```diff
@@ -200,3 +219,16 @@ service TrillianMap {
     };
   }
 }
+
+// TrillianWriteMap defines a service to allow writes against a Verifiable Map
+// that will be readable via the TrillianMap service. The write API does not
+// expose any Merkle Tree properties. This allows key/value writes to be
+// decoupled from the Merkle Tree synthesis and publishing.
+service TrillianWriteMap {
```
AlCutter: `TrillianMapWrite` (or `TrillianMapUpdate`) feel a little more natural to me, but I don't feel strongly about it :)

trillian_map_api.proto
mhutchinson: Good call.

trillian_map_api.proto
Martin2112: Along similar lines I was going to suggest something like `TrillianMapWriter`?

trillian_map_api.proto
mhutchinson: Done.

trillian_map_api.proto
mhutchinson: SGTM.

mhutchinson:
Thanks. Changes included.

PR #1657, opened 2019-06-04
skylog: Add Cloud Spanner tree storage implementation

pavelkalinnikov:
This change introduces Skylog tree storage implementation in Cloud Spanner.
Data in the nodes table is sharded in a way that allows massive parallelism if
nodes arrive "mostly" sequentially.

### DIFF
--- a/skylog/storage/cloudspanner/schema.sdl
+++ b/skylog/storage/cloudspanner/schema.sdl
@@ -1,9 +1,9 @@
 -- TreeNodes stores Merkle tree nodes. The nodes are sharded according to
--- ShardId, which is computed based on the tree shape and write pattern, in
+-- ShardID, which is computed based on the tree shape and write pattern, in
 -- order to maximize writing parallelism.
 CREATE TABLE TreeNodes (
-  TreeId       INT64 NOT NULL,
-  ShardId      INT32 NOT NULL,
-  NodeId       UINT64 NOT NULL,
+  TreeID       INT64 NOT NULL,
+  ShardID      INT32 NOT NULL,
+  NodeID       UINT64 NOT NULL,
   SubtreeHash  BYTES(32) NOT NULL,
-) PRIMARY KEY (TreeId, ShardId, NodeId);
+) PRIMARY KEY (TreeID, ShardID, NodeID);
--- a/skylog/storage/cloudspanner/schema.sdl
+++ b/skylog/storage/cloudspanner/schema.sdl
@@ -1,0 +1,9 @@
+-- TreeNodes stores Merkle tree nodes. The nodes are sharded according to
+-- ShardId, which is computed based on the tree shape and write pattern, in
+-- order to maximize writing parallelism.
+CREATE TABLE TreeNodes (
+  TreeId       INT64 NOT NULL,
+  ShardId      INT32 NOT NULL,
+  NodeId       UINT64 NOT NULL,
+  SubtreeHash  BYTES(32) NOT NULL,
+) PRIMARY KEY (TreeId, ShardId, NodeId);
--- a/skylog/storage/cloudspanner/schema.sdl
+++ b/skylog/storage/cloudspanner/schema.sdl
@@ -2,8 +2,8 @@
 -- ShardID, which is computed based on the tree shape and write pattern, in
 -- order to maximize writing parallelism.
 CREATE TABLE TreeNodes (
-  TreeID       INT64 NOT NULL,
-  ShardID      INT32 NOT NULL,
-  NodeID       UINT64 NOT NULL,
-  SubtreeHash  BYTES(32) NOT NULL,
+  TreeID   INT64 NOT NULL,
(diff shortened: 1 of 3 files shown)


skylog/storage/cloudspanner/tree.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Read fetches Merkle tree hashes of the passed in nodes from the storage.
+// TODO(pavelkalinnikov): Add nodes cache.
+func (t *TreeStorage) Read(ctx context.Context, ids []compact.NodeID) ([][]byte, error) {
+	var keys []spanner.KeySet
+	for _, id := range ids {
+		keys = append(keys, spanner.Key{t.id, t.opts.shardID(id), packNodeID(id)})
+	}
+	keySet := spanner.KeySets(keys...)
+	hashes := make([][]byte, len(ids))
```
pavelkalinnikov: `make([][]byte, 0, len(ids))`

skylog/storage/cloudspanner/schema.sdl
```diff
@@ -0,0 +1,9 @@
+-- TreeNodes stores Merkle tree nodes. The nodes are sharded according to
+-- ShardId, which is computed based on the tree shape and write pattern, in
+-- order to maximize writing parallelism.
+CREATE TABLE TreeNodes (
+  TreeId       INT64 NOT NULL,
+  ShardId      INT32 NOT NULL,
+  NodeId       UINT64 NOT NULL,
+  SubtreeHash  BYTES(32) NOT NULL,
```
Martin2112: I think this is a hash of the nodes as stored and is not related to tree hashing? If so then that might be a useful comment to add.

skylog/storage/cloudspanner/schema.sdl
```diff
@@ -0,0 +1,9 @@
+-- TreeNodes stores Merkle tree nodes. The nodes are sharded according to
+-- ShardId, which is computed based on the tree shape and write pattern, in
+-- order to maximize writing parallelism.
+CREATE TABLE TreeNodes (
+  TreeId       INT64 NOT NULL,
+  ShardId      INT32 NOT NULL,
```
Martin2112: Looks like there could be a key hotspot here if I read the docs attached to the code correctly. Not sure about the plans for splits though.

skylog/storage/cloudspanner/schema.sdl
```diff
@@ -0,0 +1,9 @@
...
+  NodeId       UINT64 NOT NULL,
```
Martin2112: Possibly another key hotspot, though probably not as bad given it's a packed encoding.

skylog/storage/cloudspanner/tree.go
```diff
@@ -0,0 +1,99 @@
...
+// TreeStorage allows reading and writing from a tree storage.
+type TreeStorage struct {
+	c    *spanner.Client
+	id   int64
+	opts TreeOpts
+}
+
+// NewTreeStorage returns a new TreeStorage for the specified tree and options.
+func NewTreeStorage(c *spanner.Client, treeID int64, opts TreeOpts) *TreeStorage {
+	return &TreeStorage{c: c, id: treeID, opts: opts}
```
Martin2112: It looks to me that the opts used for a tree must not change over time? If that's correct then they should maybe be stored in the database.

skylog/storage/cloudspanner/schema.sdl
pavelkalinnikov: The idea is that if the `TreeOpts` is good then the load on a single shard will be `O(2^ShardLevels)` nodes/s, so we have control on this hotspot.

skylog/storage/cloudspanner/schema.sdl
Martin2112: OK but be aware that the order fields appear in the primary key defines the physical data layout and IIRC you can only split on a key prefix. When there are multiple trees how do we want the split to work? TreeId, ShardId seems plausible but if we wanted to split by ShardId then they'd probably need to be swapped round.

Just for my interest for reads do you know roughly how many shards a proof path is expected to cross?

pavelkalinnikov: I'm fairly sure that splits can be done by (TreeID, ShardID) pairs, but I will double check once I get to test it.

pavelkalinnikov: A proof path will probably cross about 2 shards - one for levels below `ShardLevels`, and one above.

Martin2112: Yep I think if you use this layout you can split by TreeID, ShardID but not by ShardID on it's own.

Overall, I think it's OK.

skylog/storage/cloudspanner/schema.sdl
Martin2112: SGTM.

PR #1658, opened 2019-06-04
Modify licenses tool to use "golang.org/x/tools/go/packages"

RJPercival:
Using this package, instead of "go/build", enables the tool to work with packages that are using Go Modules. It also transparently supports $GOFLAGS and handles walking the package dependency tree, which makes it possible to delete some code.

Contributes to #1400.

### Checklist

### DIFF
--- a/scripts/licenses/licenses/library.go
+++ b/scripts/licenses/licenses/library.go
@@ -78,6 +78,9 @@ func Libraries(ctx context.Context, importPaths ...string) ([]*Library, error) {
 			// No license requirements for the Go standard library.
 			return false
 		}
+		if len(p.OtherFiles) > 0 {
+			glog.Warningf("%q contains non-Go code that can't be inspected for further dependencies:\n%s", p.PkgPath, strings.Join(p.OtherFiles, "\n"))
+		}
 		var pkgDir string
 		switch {
 		case len(p.GoFiles) > 0:
--- a/scripts/licenses/licenses/library.go
+++ b/scripts/licenses/licenses/library.go
@@ -15,52 +15,99 @@
 package licenses
 
 import (
+	"context"
 	"fmt"
 	"go/build"
+	"path/filepath"
 	"sort"
-	"sync"
+	"strings"
 
 	"github.com/golang/glog"
-)
-
-var (
-	pkgCache sync.Map
+	"golang.org/x/tools/go/packages"
 )
 
 // Library is a collection of packages covered by the same license file.
 type Library struct {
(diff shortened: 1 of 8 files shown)


scripts/licenses/licenses/library.go
```diff
@@ -15,78 +15,133 @@
...
+		}
 		if isStdLib(p) {
 			// No license requirements for the Go standard library.
-			continue
+			return false
+		}
+		var pkgDir string
+		switch {
+		case len(p.GoFiles) > 0:
+			pkgDir = filepath.Dir(p.GoFiles[0])
```
gdbelvin: do you want to concat all these options instead of terminating at the first case?

gdbelvin:
Would you mind testing in a directory outside of GOPATH as well? 
I'm noticing that the original version of this tool fails when doing a fresh checkout with export GO111MODULE=on

RJPercival:
I've tested it outside of $GOPATH and found that the `csv` subcommand is unable to determine URLs for the license files, because Go modules appear to have the `.git` directories deleted (which are necessary in order to determine where the repository is hosted). However, the `save` subcommand works, which is all we need for Travis to succeed.

gdbelvin:
gosimple errors

gdbelvin:
From the cgo docs:  https://golang.org/cmd/cgo/

When the Go tool sees that one or more Go files use the special import "C", it will look for other non-Go files in the directory and compile them as part of the Go package. Any .c, .s, or .S files will be compiled with the C compiler. Any .cc, .cpp, or .cxx files will be compiled with the C++ compiler. Any .f, .F, .for or .f90 files will be compiled with the fortran compiler. Any .h, .hh, .hpp, or .hxx files will not be compiled separately, but, if these header files are changed, the package (including its non-Go source files) will be recompiled. **Note that changes to files in other directories do not cause the package to be recompiled, so all non-Go source code for the package should be stored in the package directory, not in subdirectories.** The default C and C++ compilers may be changed by the CC and CXX environment variables, respectively; those environment variables may include command line options.

RJPercival:
While that documentation doesn't preclude the C/C++/Fortran/etc. code from pulling in other dependencies, it seems like downgrading the existence of such code from an error to a warning isn't unreasonable. Perhaps I should add a user guide for this tool though so that people understand that this warning warrants investigation.

RJPercival:
@gdbelvin: Could you take a look at the README and, if you're happy with it, I'll merge this PR.

gdbelvin:
This looks good!

PR #1664, opened 2019-06-05
Remove toy map client

gdbelvin:
A real map client is located in `client/map_client.go`

### Checklist


AlCutter:
isn't `client/map_client.go` a library, whereas this is a binary?

gdbelvin:
That is true, but do we want people to use this?
For making rpc calls there is the grpc_cli command line client which may be
more canonical.

AlCutter:
Having a simple toy to play with the map is helpful, I think - perhaps a better idea would be to re-implement this binary using `client/`?

Martin2112:
There is also an example of map usage at: https://github.com/google/trillian-examples/blob/master/etherslurp/mapper/mapper.go

I think removing this early example will be helpful.

PR #1665, opened 2019-06-05
Use errgroup in VerifyMapLeavesResponse

gdbelvin:
This greatly simplifies the logic needed to capture the first error and cancel
outstanding work.

### Checklist

### DIFF
--- a/client/map_verifier.go
+++ b/client/map_verifier.go
@@ -18,8 +18,6 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
-	"runtime"
-	"sync"
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keys/der"
@@ -27,6 +25,7 @@ import (
 	"github.com/google/trillian/merkle/hashers"
 	"github.com/google/trillian/trees"
 	"github.com/google/trillian/types"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
This is awesome :)

client/map_verifier.go
```diff
@@ -111,42 +110,15 @@ func (m *MapVerifier) VerifyMapLeavesResponse(indexes [][]byte, revision int64,
...
-		}
-	}()
-	for err := range errors {
-		if err != nil {
-			return nil, status.Errorf(status.Code(err), "map: VerifyMapLeafInclusion(): %v", err)
-		}
+	var g errgroup.Group
+	for _, p := range resp.MapLeafInclusion {
+		p := p
+		g.Go(func() error {
```
pavelkalinnikov: This code always starts all the verifications, whereas the previous one could be canceled if one failed? Is this critical? I guess, we should care mostly about the success case.

gdbelvin: Thanks for catching that. The documentation said it would cancel things, but it only cancels when called from `WithContext` I'll look into this more tomorrow.

gdbelvin: Thinking about this more... 
Explicit cancels are useful when one is in the middle of a pipeline and one can use
```go
select {
data <- inputpipe:
    process
<- ctx.Done():
    exit pipeline
}
```

Because this instantiation does not have a pipeline, I don't think checking ctx.Done() is useful since it will only create an unreliable race condition. 

Your observation that this is mostly about the success case is also correct.

PR #1667, opened 2019-06-07
NodeID.SetRight

gdbelvin:
The NodeID.SetRight function supports computing the left and right edges of any
subtree at a given level.

### Checklist


storage/types_test.go
```diff
@@ -26,6 +26,55 @@ import (
...
+		{val: 0xFF, index: h2b("0000000000000000000000000000000000000000"), depth: 159, want: h2b("0000000000000000000000000000000000000001")},
+		{val: 0xFF, index: h2b("0000000000000000000000000000000000000000"), depth: 160, want: h2b("0000000000000000000000000000000000000000")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 1, want: h2b("7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 17, want: h2b("00017FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 159, want: h2b("000102030405060708090A0B0C0D0E0F10111213")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 160, want: h2b("000102030405060708090A0B0C0D0E0F10111212")},
+	} {
+		nID := NewNodeIDFromHash(tc.index)
+		if got, want := nID.SetRight(tc.depth, tc.val).Path, tc.want; !bytes.Equal(got, want) {
+			t.Errorf("SetRight(%x, %v, 0x%X): %x, want %x", tc.index, tc.depth, tc.val, got, want)
```
pavelkalinnikov: It seems like it's important to verify the bit length of the result, not only the byte length + byte contents.

storage/types.go
```diff
@@ -370,6 +370,32 @@ func (n *NodeID) MaskLeft(depth int) *NodeID {
...
+		// Copy the first depthBytes-1.
+		copy(r, n.Path[:depthBytes-1])
+		// Set the first leftmask bits of the last byte.
+		left := n.Path[depthBytes-1] & leftmask[depth%8]
+		right := val &^ leftmask[depth%8]
+		r[depthBytes-1] = left | right
+	}
+	b := n.PrefixLenBits
+	if depth < n.PrefixLenBits {
+		b = depth
```
pavelkalinnikov: Why do you change the bit length here? Isn't this method supposed to only flip bits?

pavelkalinnikov:
I wonder if we need to unconditionally make a copy in this method. Maybe it should change the bits in-place, and then if the client needs a copy they can call `Copy()+Method()`. I feel like the same argument can be applied to many of other `NodeID` methods.

storage/types.go
gdbelvin: Good point. I'll remove it.

gdbelvin:
@Martin2112 has been working hard to make sure that the `NodeID` modifiers do not modify the receiver.  I believe making the receiver immutable opens up some efficiency gains IIRC. 

In any case, the way I expect to use this function will make a copy anyways.

storage/types_test.go
```diff
@@ -26,6 +26,59 @@ import (
 
 var h2b = testonly.MustHexDecode
 
+func TestSetRight(t *testing.T) {
```
pavelkalinnikov: TestSetLowerBits

storage/types_test.go
```diff
@@ -26,6 +26,59 @@ import (
...
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 17, want: h2b("00017FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 159, want: h2b("000102030405060708090A0B0C0D0E0F10111213")},
+		{val: 0xFF, index: h2b("000102030405060708090A0B0C0D0E0F10111212"), depth: 160, want: h2b("000102030405060708090A0B0C0D0E0F10111212")},
+	} {
+		nID := NewNodeIDFromHash(tc.index)
+		gotNode := nID.SetLowerBits(tc.depth, tc.val)
+		if got, want := gotNode.Path, tc.want; !bytes.Equal(got, want) {
+			t.Errorf("SetLowerBits(%x, %v, 0x%X): %x, want %x", tc.index, tc.depth, tc.val, got, want)
+		}
+		if got, want := gotNode.PrefixLenBits, 160; got != want {
```
pavelkalinnikov: All the tests have bit length multiple of 8. Should we test the case when it's not?

gdbelvin: Since we are no longer manipulating the bit length, I don't think that's needed.  
`NewNodeIDFromHash` will always produce a bit length that is a multiple of 8.
Furthermore, `PrefixLenBits` is not used in the the computation.

PR #1668, opened 2019-06-07
Use go list to find certificate-transparency-go path

gdbelvin:
This enables other repos that have go modules enabled to use
the scripts in the Trillian repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,9 +114,9 @@ script:
   - |
     if [[ "${INTEG_TESTS}" == "true" ]]; then
       ./integration/integration_test.sh
-      pushd .
-      cd `go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go`
+      pushd `go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go`
       chmod +x ./trillian/integration/integration_test.sh
+      ./trillian/integration/integration_test.sh
       popd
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -114,9 +114,11 @@ script:
   - |
     if [[ "${INTEG_TESTS}" == "true" ]]; then
       ./integration/integration_test.sh
-      cd "$HOME/gopath/src/github.com/google/certificate-transparency-go"
+      pushd .
+      cd `go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go`
```
RJPercival: You can combine the above two lines:
```shell
pushd `go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go`
```

.travis.yml
```diff
@@ -114,9 +114,11 @@ script:
...
+      chmod +x ./trillian/integration/integration_test.sh
```
RJPercival: Why is this chmod necessary?

RJPercival: Oh is this that annoying Go Modules behaviour where they overwrite all permissions on the files with 0400 (https://github.com/golang/go/issues/31481)?

gdbelvin: yes, indeed it is.  I'm doing my best to see if this is going to cause other issues, but I haven't seen any so far other than this.

PR #1671, opened 2019-06-10
skylog: Add Merkle tree build worker

pavelkalinnikov:
This change introduces `BuildWorker` which is responsible for building lower
levels of the Merkle tree in accordance with leaf hashes that it gets.

### DIFF
--- a/skylog/core/builder.go
+++ b/skylog/core/builder.go
@@ -1,0 +1,68 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package core contains code for a scalable Merkle tree construction.
+package core
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/trillian/merkle/compact"
+	"github.com/google/trillian/skylog/storage"
+)
+
+// BuildJob is a slice of leaf data hashes spanning a range of a Merkle tree.
+type BuildJob struct {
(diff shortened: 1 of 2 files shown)


skylog/core/builder.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"fmt"
+
+	"github.com/google/trillian/merkle/compact"
+	"github.com/google/trillian/skylog/storage"
+)
+
+// BuildJob is a slice of leaf data hashes spanning a range of a Merkle tree.
+type BuildJob struct {
+	Begin  uint64   // The beginning of the range.
```
AlCutter: How about `RangeStart`? (Begin sounds more like a function name)

skylog/core/builder_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/google/trillian/merkle/compact"
+	"github.com/google/trillian/merkle/rfc6962"
+	"github.com/google/trillian/merkle/testonly"
+	"github.com/google/trillian/skylog/storage"
+)
+
+// TODO(pavelkalinnikov): Use GoMock?
```
AlCutter: Although the golang folks [used to?] hate on mocks, it does seem a little like you're having to reimplement some of that functionality here.

skylog/core/builder_test.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("got nodes: %+v\nwant: %+v", got, want)
+	}
+	return nil
+}
+
+func TestBuildWorker(t *testing.T) {
+	hashes := testonly.NodeHashes()
+	n := len(hashes[0])
+	if got, want := n, 8; got != want {
+		t.Fatalf("got %d leaves, want %d", got, want)
```
AlCutter: nit: maybe change the logged message here - "prerequisite: testinly.NodeHashes() returned %d leaves, but this test wants 8" - just to make it clear that it's not the _test per se_ which has failed.

skylog/core/builder.go
pavelkalinnikov: Done.

skylog/core/builder_test.go
pavelkalinnikov: Done.

skylog/core/builder_test.go
pavelkalinnikov: Yep, made this a more definite TODO.

PR #1673, opened 2019-06-11
Scripts may not be executable, invoke with bash

gdbelvin:
When referenced from a go-module enabled project, the scripts referenced by `functions.sh` may not be executable. 

Solution: invoke them directly with bash.


gdbelvin:
Related issue: https://github.com/golang/go/issues/31481

PR #1676, opened 2019-06-12
Implementation of Map Write API that delegates to legacy methods.

mhutchinson:
The write API would preferably only be enabled if the single transaction mode is enabled, however forcing this would require splitting the integration tests to allow testing the write API in an optional way. This overhead seems more likely to introduce errors than temporarily supporting the multi-transaction writes in the Write API.

Future work in this area will move the implementation of the write methods into the write server, which will allow synthesis & publishing of the Merkle Tree to be performed asynchronously from the key/value write operations.

### DIFF
--- a/integration/map.go
+++ b/integration/map.go
@@ -49,6 +49,7 @@ var (
 // MapInfo describes an in-progress integration test.
 type MapInfo struct {
 	cl       trillian.TrillianMapClient
+	wcl      trillian.TrillianMapWriteClient
 	id       int64
 	tree     *trillian.Tree
 	verifier *client.MapVerifier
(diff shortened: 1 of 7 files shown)


server/map_write_rpc_server.go
```diff
@@ -0,0 +1,79 @@
...
+		MapId: req.MapId,
+		Leaves: req.Leaves,
+		Metadata: req.Metadata,
+		Revision: req.ExpectRevision}
+
+	resp, err := t.mapServer.SetLeaves(ctx, &setLeavesReq)
+	if err != nil {
+		return nil, err
+	}
+	root, err := tcrypto.VerifySignedMapRoot(pub, hash, resp.MapRoot)
```
gdbelvin: You might be able to use some of the convenience methods in `client/map_verifier.go`

integration/map.go
```diff
@@ -49,20 +49,22 @@ var (
 // MapInfo describes an in-progress integration test.
```
gdbelvin: There are also map tests in `integration/maptest/map.go`

mhutchinson:
This PR has gone a bit bonkers. I think my rebasing/squashing strategy has upset the universe. I'll see if I can fix it, but I may clone the changes into a new branch instead.

server/map_write_rpc_server.go
gdbelvin: Hmm that's interesting. Let's add this as an issue somewhere for tracking

mhutchinson: Done.

PR #1677, opened 2019-06-12
Don't register HTTP/JSON handlers

gdbelvin:
Remove the HTTP/JSON web gateway. 

No clients were using the HTTP gateway, the gateway was untested, and introduces additional code complexity and dependencies. 

### Checklist


RJPercival:
This has the same problem that I ran into with PR #1373 - the integration tests use the HTTP API to control quota. We need to make a tool that talks to the quota API via gRPC before we can get rid of the HTTP API. I started on this, but haven't finished it because it hasn't been a priority. Is it higher priority now?

gdbelvin:
Indeed it does.  I think the solution is to migrate to grpc_cli, but the
challenge there is that grpc_cli has a bit of a complicated install
process.

The reason for the (small) bump in priority is that we're spending some
more time in the inner_main.go trampolines and there was a desire to see if
we could clean them up a bit.

gdbelvin:
Supersedes #1373  and #1996

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 Not yet released; provisionally v1.4.0 (may change).
 
+### HTTP APIs
+
+The HTTP/JSON APIs have been deprecated in favor of a pure gRPC intereface.
```
RJPercival: I don't think we can call them "deprecated" when we've deleted them. I think this should just say "removed".

cmd/trillian_map_server/main.go
```diff
@@ -40,7 +40,6 @@ import (
 	"github.com/google/trillian/server"
 	"github.com/google/trillian/storage"
 	etcdutil "github.com/google/trillian/util/etcd"
-	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"google.golang.org/grpc"
 
 	// Register key ProtoHandlers
```
RJPercival: The httpEndpoint flag says "Endpoint for HTTP metrics and **REST requests**" - should that second part be removed now?

cmd/trillian_log_server/main.go
```diff
@@ -43,7 +43,6 @@ import (
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/util/clock"
 	etcdutil "github.com/google/trillian/util/etcd"
-	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"google.golang.org/grpc"
 
 	// Register key ProtoHandlers
```
RJPercival: The httpEndpoint flag says "Endpoint for HTTP metrics and **REST requests**" - should that second part be removed now?

cmd/internal/serverutil/main.go
```diff
@@ -69,8 +68,6 @@ type Main struct {
 	StatsPrefix string
```
RJPercival: References to "REST" in the comment for HTTPEndpoint above should also possibly be removed.

cmd/trillian_log_server/main.go
gdbelvin: yep

gdbelvin:
Voting on removal: https://gtrillian.slack.com/archives/C5E5P19A6/p1582026823006300

PR #1680, opened 2019-06-12
skylog: Introduce sequence storage

pavelkalinnikov:
This change introduces Skylog sequence storage and its implementation in
Cloud Spanner. Data in the sequence table is sharded in a way that allows
massive parallelism if entires arrive "mostly" sequentially.


skylog/storage/cloudspanner/sequence.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, nil
+	}
+	// TODO(pavelkalinnikov): Restrict the range length.
+	// TODO(pavelkalinnikov): Don't return tail entries if all empty.
+	ret := make([]storage.Entry, end-begin)
+
+	// We only read entries from the shard that includes the begin-th entry.
+	// TODO(pavelkalinnikov): Unit-test this logic.
+	offset := begin / s.opts.BatchSize
+	if next := (offset + 1) * s.opts.BatchSize; end < next {
```
pavelkalinnikov: Oups. Should be:
```
if next < end {
  end = next
}
```

The TODO above is right :)

skylog/storage/cloudspanner/sequence.go
```diff
@@ -0,0 +1,98 @@
...
+		end = next
+	}
+	shardID := int64(offset % s.opts.Shards)
```
AlCutter: This seems potentially scary/brittle - if I decide I want to change the number of shards I'm running (because performance/scaling), and I start updating my jobs then my read results are liable to be incorrect? (if this is possible I guess it'll result in a SEGV/panic at L74?]

skylog/storage/cloudspanner/schema.sdl
```diff
@@ -1,3 +1,13 @@
+-- Entries stores sequenced entries of the log, which also correspond to leaves
+-- of the Merkle Tree.
+CREATE TABLE Entries (
```
pavelkalinnikov: Hm, I'm wondering if interleaving this into `TreeNodes` is a good idea (mapping between `EntryIndex` and `NodeID` is trivial).

skylog/storage/api.go
```diff
@@ -21,6 +21,20 @@ import (
...
+// SequenceReader allows reading from a sequence storage.
+type SequenceReader interface {
+	// Read fetches the specified [begin, end) range of entries, and returns them
+	// in order. May return a prefix of the requested range (potentially empty).
+	Read(ctx context.Context, begin, end uint64) ([]Entry, error)
+}
+
+// SequenceWriter allows writing to a sequence storage.
+type SequenceWriter interface {
+	// Write puts all the passed in entries to the sequence starting at the
```
AlCutter: /puts/stores/?

/passed in/passed-in/

skylog/storage/cloudspanner/sequence.go
```diff
@@ -0,0 +1,98 @@
...
+		ret[uint64(index)-begin] = storage.Entry{Data: data, Extra: extra}
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return ret, nil
+}
+
+// Write puts all the passed in entries to the sequence starting at the
```
AlCutter: /puts/stores/

skylog/storage/cloudspanner/schema.sdl
pavelkalinnikov: The upside of doing that is having `NodeHash` close to `Data`.

skylog/storage/cloudspanner/sequence.go
AlCutter: this looks like it'll make this method read more entries than requested?

skylog/storage/cloudspanner/sequence.go
```diff
@@ -0,0 +1,98 @@
...
+// specified begin index.
+func (s *SequenceStorage) Write(ctx context.Context, begin uint64, entries []storage.Entry) error {
+	ms := make([]*spanner.Mutation, 0, len(entries))
+	for i, entry := range entries {
+		index := int64(begin) + int64(i)
+		shardID := index / int64(s.opts.BatchSize) % int64(s.opts.Shards)
+		// TODO(pavelkalinnikov): Consider doing just Insert when it is clear what
```
AlCutter: Is there a valid use case where the caller would want to overwrite a previously sequenced entry?

skylog/storage/cloudspanner/sequence.go
pavelkalinnikov: In your example you will end up not finding entries, but it's never going to return incorrect ones because the actual leaf index is part of the key.

skylog/storage/cloudspanner/sequence.go
pavelkalinnikov: Not that I'm aware of. However, `Write`s can be retried. We don't want to fail the whole write if a single entry is already present. Ideally, we need a "mutation pool" here that filters out duplicates.

skylog/storage/cloudspanner/sequence.go
pavelkalinnikov: Yep, the condition should be reversed.

skylog/storage/cloudspanner/schema.sdl
pavelkalinnikov: OTOH, `Entries` is tailored for horizontal reads, whereas `TreeNodes` is (mostly) for vertical inclusion-proof-like reads. I don't want to mix the two in one table, so keeping them separate for now.

skylog/storage/cloudspanner/sequence.go
AlCutter: But if it returns `[L, L, L, L, nil, nil, nil, nil, nil, L, L, L, L]` is that ok?

pavelkalinnikov: That's a bit wacky. OTOH, I'm not thinking about dynamic resharding yet, cause it would be complicated anyway. In your example the right solution would be to continue serving data using the old sharding scheme, until the data is copied to new shards, then switch to the new scheme, and then delete the old data. Alternatively, we can do a live migration to a new log tree.

There is another case when `[L, L, L, L, nil, nil, nil, nil, nil, L, L, L, L]` can be returned - while the tree is under construction. But I guess we can disallow it for now, and return just the `[L, L, L, L]` prefix.

pavelkalinnikov:
@AlCutter PTAL. I made some changes in `Read` implementation - now it never returns data with gaps.

skylog/storage/cloudspanner/sequence.go
AlCutter: Worth popping a `TODO` or comment about sharding in there, just as a warning to future you/anyone else who comes visiting?

pavelkalinnikov: Added comment to `SequenceOpts`. PTAL.

PR #1682, opened 2019-06-13
Add a MapWriteStress test

AlCutter:
Adds a simple write stress test for the map.

### Checklist

### DIFF
--- a/integration/maptest/map.go
+++ b/integration/maptest/map.go
@@ -654,6 +654,16 @@ func RunInclusionBatch(ctx context.Context, t *testing.T, tadmin trillian.Trilli
 	}
 }
 
+// RunWriteBatchStress performs checks on Trillian Map's SetLeaves call.
+func RunWriteBatchStress(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient, batchSize int, numBatches int) {
+	tree, err := newTreeWithHasher(ctx, tadmin, tmap, trillian.HashStrategy_TEST_MAP_HASHER)
+	if err != nil {
+		t.Fatalf("%v: newTreeWithHasher(): %v", trillian.HashStrategy_TEST_MAP_HASHER, err)
+	}
+
+	_ = writeBatch(ctx, t, tmap, tree, batchSize, numBatches)
+}
+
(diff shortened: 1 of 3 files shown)


integration/maptest/map.go
```diff
@@ -646,6 +646,16 @@ func RunInclusionBatch(ctx context.Context, t *testing.T, tadmin trillian.Trilli
 	}
 }
 
+// RunWriteBatchStress performs checks on Trillian Map's SetLeaves call.
+func RunWriteBatchStress(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient, batchSize int, numBatches int) {
```
mhutchinson: You need to add this to AllTests at the top of this file or it won't run. But nobody would fall into that bear trap.

AlCutter: Just call me "Stumpy"

gdbelvin:
Please rebase on master to get the build unstuck

integration/maptest/map.go
```diff
@@ -55,6 +55,7 @@ var AllTests = TestTable{
 	{"LeafHistory", RunLeafHistory},
 	{"Inclusion", RunInclusion},
 	{"InclusionBatch", RunInclusionBatch},
+	{"WriteStress", RunWriteStress},
```
gdbelvin: This table is what the default RunIntegration test iterates through. 

It looks like you've added a new entry point in `map_test.go` We should probably pick just one of them.  Otherwise, this test will be run twice.

AlCutter: That's by design, and it won't be run twice since the new entry point will skip by default.
I added a new target  so that I can run stress tests *only* and be able to tune the level.

gdbelvin: Not sure if I completely follow. 
if we pass `--stress` and run `go test ./integration` the following tests will be run:

```
TestIntegration/etc...
TestIntegration/InclusionBatch
TestIntegration/WriteStress
TestMapWriteStress
```

If you'd like to only run one test does 
`go test ./integration -run TestIntegration/WriteStress` accomplish the job?

AlCutter: That's probably good enough.

AlCutter:
ping

PR #1683, opened 2019-06-13
Check for duplicates server side, not client side.

gdbelvin:
### Checklist

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -160,6 +160,9 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req 
 	if req.Revision < 0 {
 		return nil, fmt.Errorf("map revision %d must be >= 0", req.Revision)
 	}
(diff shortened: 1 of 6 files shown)


server/map_rpc_server.go
```diff
@@ -657,3 +661,16 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
 	return tx, err
 }
+
+// hasDuplicates returns an error if there are duplicates in indexes.
+func hasDuplicates(indexes [][]byte) error {
+	set := make(map[string]bool)
+	for _, i := range indexes {
+		if set[string(i)] {
+			return status.Errorf(codes.InvalidArgument,
+				"map.GetLeaves(): index %x requested more than once", i)
```
mhutchinson: nit: this error message assumes who will call it. Not really a fan. I think it would be better to say "duplicate index detected: %x". If the caller wants to wrap that in an error string that identifies the exact method being called then fair enough.

PR #1684, opened 2019-06-13
Don't return LeafHash in RunGetLeafByRevisionNoProof

gdbelvin:
Since the Callers of `RunGetLeafByRevisionNoProof` don't rely on `LeafHash`, don't return it. 

This saves on bandwidth and is functionally cleaner. 

This PR also fixes the tests of `RunGetLeafByRevisionNoProof`

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -180,6 +180,11 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req 
 		return nil, err
 	}
 
+	// Remove LeafHash because SetLeaves does not supply it.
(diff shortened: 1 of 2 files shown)


server/map_rpc_server.go
```diff
@@ -180,6 +180,11 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req
 		return nil, err
 	}
 
+	// Remove LeafHash because SetLeaves does not supply it.
+	for _, l := range leaves {
```
mhutchinson: Please update the CL description to explicitly call out this change in functional behaviour (is this also worth updating changelog for?).

PR #1685, opened 2019-06-13
Run CT Go tests with Go modules

gdbelvin:
Now that certificate-transprency-go is Go Module enabled, Trillian needs to use it with Go Modules enabled.

### Checklist


gdbelvin:
Error: `cmd/createtree/pkcs11.go:53:42: config.PublicKeyPath undefined`

Because the CT tests call back into the Trillian repo, it's not possible to enable go modules for certificate-transparency-go and not also for Trillian.

I think the way forward is to turn on Go modules for Trillian #1400

gdbelvin:
Reopened in #1686

PR #1687, opened 2019-06-14
skylog: Add build worker based on Cloud Functions

pavelkalinnikov:
This change introduces a prototype of a Cloud Function that processes
Merkle tree building jobs. This is a building block for scalable Merkle tree
construction on GCP.


skylog/storage/cloudfunctions/worker.go
```diff
@@ -0,0 +1,98 @@
...
+// BuildSubtree consumes builder job message.
+func BuildSubtree(ctx context.Context, msg BuildMessage) error {
+	var job BuildJob
+	if err := json.Unmarshal(msg.Data, &job); err != nil {
+		return err
+	}
+
+	log.Printf("Accepted job: %+v", job)
+	if job.End < job.Begin {
+		return errors.New("invalid job: begin > end")
```
AlCutter: >= ?

skylog/storage/cloudfunctions/worker.go
```diff
@@ -0,0 +1,98 @@
...
+type BuildMessage struct {
+	// Data contains a JSON-encoded BuildJob.
+	// TODO(pavelkalinnikov): Consider protobuf instead.
+	Data []byte `json:"data"`
+}
+
+// BuildJob describes a Merke tree building job.
+type BuildJob struct {
+	TreeID int64  `json:"tree_id"`
+	Begin  uint64 `json:"begin"`
```
AlCutter: Is `End` inclusive or exclusive?
Would `Start`, `Num` be clearer?

pavelkalinnikov: Clarified this in the type comment. I prefer `[begin, end)` when possible, which I believe to be a commonly recognized convention (I might be biased towards people who heard about C++). Compact range is also designed this way.

skylog/storage/cloudfunctions/worker.go
pavelkalinnikov: Done.

PR #1688, opened 2019-06-14
Export GO111MODULE Variable in Docker Builds

gdbelvin:
Allow the docker images to be built with modules off (by default) or on if GO111MODULE=on is passed as an argument.

### Checklist

### DIFF
--- a/examples/deployment/docker/log_server/Dockerfile
+++ b/examples/deployment/docker/log_server/Dockerfile
@@ -4,6 +4,8 @@ ADD . /go/src/github.com/google/trillian
 WORKDIR /go/src/github.com/google/trillian
 
 ARG GOFLAGS=""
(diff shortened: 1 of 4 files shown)


examples/deployment/docker/log_server/Dockerfile
```diff
@@ -4,6 +4,8 @@ ADD . /go/src/github.com/google/trillian
 WORKDIR /go/src/github.com/google/trillian
 
 ARG GOFLAGS=""
```
RJPercival: Separately, at should add an ENV directive for $GOFLAGS too, since it'll only apply to the `go get` right now (ARG drops out of scope between layers).

gdbelvin: That sounds like a separate PR

PR #1691, opened 2019-06-17
Add GOFLAGS environment variable to Dockerfiles

gdbelvin:
The $GOFLAGS variable was being passed as an argument, but not being propagated to an environment variable inside the Docker container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,6 +144,9 @@ inclusion proof queries.
 
 ### Deployments
 
+The Trillian Docker images now accept GOFLAGS and GO111MODULE arguments
+and set them as environment variables inside the Docker container.
+
 The [db\_server Docker image](examples/deployment/docker/db_server/Dockerfile)
 is now based on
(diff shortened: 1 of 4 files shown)


RJPercival:
Might be worth a CHANGELOG entry, since it could surprise someone that $GOFLAGS is now propagated from the environment when running the image.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 Not yet released; provisionally v2.0.0 (may change).
 
+### New Docker ARGS: GOFLAGS and GO111MODULE 
+The Trillian Docker images now accept GOFLAGS and GO111MODULE arguments
+and set them as environment variables inside the Docker container.
```
RJPercival: It might be best to put this in the "Deployments" section, since that already has lots of information about changes to Docker-related stuff.

RJPercival: It might be best to put this in the "Deployments" section, since that already has lots of information about changes to Docker-related stuff.

PR #1693, opened 2019-06-18
rm -rf vendor

gdbelvin:
Go modules does not use the vendor folder. 
Removing it now to reduce confusion


Martin2112:
Let's just see whether the skylog stuff etc. builds in case we have to go back.

gdbelvin:
As a side effect, golangci.com can now process Trillian again:
https://github.com/google/trillian/commit/e4067aefa14764b0651d9293022ed43d7e292cf5

gdbelvin:
https://golangci.com/r/github.com/google/trillian

PR #1697, opened 2019-06-18
skylog: Add GCP proto package

pavelkalinnikov:
This change introduces a protobuf package which will be used by both
senders and receivers (e.g. the tree builder Cloud Function).


skylog/storage/gcp/gcppb/gen.go
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package gcppb contains proto messages for GCP-based Skylog storage.
+package gcppb
+
+//go:generate protoc -I=. --go_out=$GOPATH/src gcp.proto
```
AlCutter: Do this need new `go list` stuff because of gomodules?

pavelkalinnikov: Not sure how to use it here. Could you elaborate?

At this point I do it the same way as in other proto files in this repo. Perhaps there should be a separate PR fixing all protos.

pavelkalinnikov: @gdbelvin WDYT?

pavelkalinnikov: Btw, https://github.com/golang/protobuf/issues/692

gdbelvin: This is consistent with the other proto generation methods, so this LGTM. 
However, the issue you've linked to does suggest a better way.  Could you experiment with using `paths=source_relative` ?  That would allow us to truly remove any references to $GOPATH.

PR #1701, opened 2019-06-19
skylog: Add binary for testing BuildJob publishing

pavelkalinnikov:


skylog/storage/gcp/pubsub/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+func runSender(ctx context.Context, cli *pubsub.Client) error {
+	top := cli.Topic(*psTopic)
+	defer top.Stop()
+
+	var results []*pubsub.PublishResult
+
+	lim := rate.NewLimiter(rate.Limit(*maxRate), *maxJobSize)
+	for index, next := *beginIndex, int64(0); index < *endIndex; index = next {
+		if next = index + int64(*maxJobSize); next > *endIndex {
```
AlCutter: I'd break this out into two lines for clarity - there's a "leaking" side-effect in the if (`next = ...` as opposed to `next := ...`) which is unusual.

PR #1702, opened 2019-06-19
Use a function to unmarshal MapLeaf

gdbelvin:
This allows other functions (next PR) to use it

### DIFF
--- a/storage/mysql/map_storage.go
+++ b/storage/mysql/map_storage.go
@@ -240,7 +240,7 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
 	ret := make([]*trillian.MapLeaf, 0, len(indexes))
 	for rows.Next() {
 		var mapKeyHash, flatData []byte
-		if err = rows.Scan(&mapKeyHash, &flatData); err != nil {
+		if err := rows.Scan(&mapKeyHash, &flatData); err != nil {
 			return nil, err
 		}
 		mapLeaf, err := unmarshalMapLeaf(flatData, mapKeyHash)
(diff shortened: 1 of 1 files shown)


storage/mysql/map_storage.go
```diff
@@ -213,6 +199,21 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
 	if len(indexes) == 0 {
 		return []*trillian.MapLeaf{}, nil
 	}
+	selectMapLeafSQL := `
```
AlCutter: This should probably still be a const unless you had some reason to change it to a var?
(`const selectMapLeafSQL = ...`)

storage/mysql/map_storage.go
```diff
@@ -237,32 +238,33 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
...
 	ret := make([]*trillian.MapLeaf, 0, len(indexes))
-	nr := 0
-	er := 0
 	for rows.Next() {
-		var mapKeyHash []byte
-		var mapRevision int64
-		var flatData []byte
-		err = rows.Scan(&mapKeyHash, &mapRevision, &flatData)
+		var mapKeyHash, flatData []byte
+		err = rows.Scan(&mapKeyHash, &flatData)
```
AlCutter: ` if err := rows.Scan(...); err != nil {`

storage/mysql/map_storage.go
```diff
@@ -240,8 +240,7 @@ func (m *mapTreeTX) Get(ctx context.Context, revision int64, indexes [][]byte) (
 	ret := make([]*trillian.MapLeaf, 0, len(indexes))
 	for rows.Next() {
 		var mapKeyHash, flatData []byte
-		err = rows.Scan(&mapKeyHash, &flatData)
-		if err != nil {
+		if err = rows.Scan(&mapKeyHash, &flatData); err != nil {
```
AlCutter: Probably should be `:=` as elsewhere in this loop.

PR #1703, opened 2019-06-19
skylog: Add tree sharding description to job messages

pavelkalinnikov:
This change adds tree sharding info to the BuildJob proto message, in
order for the build workers to avoid reading it from storage on each
invocation.


pavelkalinnikov:
The worker code that uses the new field will be added in a follow-up PR, cause it's in a different module.

skylog/storage/gcp/pubsub/main.go
```diff
@@ -28,13 +28,15 @@ import (
...
 var (
-	projectID  = flag.String("project", "skylog-test", "The GCP project ID")
-	psTopic    = flag.String("topic", "build-jobs", "The Pub/Sub topic for build jobs")
-	treeID     = flag.Int64("tree_id", 1, "The ID of the tree under construction")
-	beginIndex = flag.Int64("begin", 0, "The beginning of the tree building range")
-	endIndex   = flag.Int64("end", 0, "The ending of the tree building range")
-	maxJobSize = flag.Int("job_size", 256, "The maximal number of entries in a build job")
-	maxRate    = flag.Float64("rate", 100000, "The average rate of adding entries per second")
+	projectID   = flag.String("project", "skylog-test", "The GCP project ID")
+	psTopic     = flag.String("topic", "build-jobs", "The Pub/Sub topic for build jobs")
```
AlCutter: How about `buildJobsTopic`/`build_jobs_topic` for the var and flag name?

skylog/storage/gcp/pubsub/main.go
```diff
@@ -28,13 +28,15 @@ import (
...
+	treeID      = flag.Int64("tree_id", 1, "The ID of the tree under construction")
+	beginIndex  = flag.Int64("begin", 0, "The beginning of the tree building range")
+	endIndex    = flag.Int64("end", 0, "The ending of the tree building range")
```
AlCutter: Consider specifying inclusive/exclusive in the var name/description to reduce potential of off-by-ones in the future?

skylog/storage/gcp/pubsub/main.go
```diff
@@ -43,13 +45,15 @@ func runSender(ctx context.Context, cli *pubsub.Client) error {
...
+	sharding := pb.TreeSharding{Levels: uint32(*shardLevels), Shards: uint32(*treeShards)}
+
 	lim := rate.NewLimiter(rate.Limit(*maxRate), *maxJobSize)
 	for index, next := *beginIndex, int64(0); index < *endIndex; index = next {
 		next = index + int64(*maxJobSize)
 		if next > *endIndex {
 			next = *endIndex
 		}
-		job := pb.BuildJob{TreeId: *treeID, Begin: uint64(index), End: uint64(next)}
+		job := pb.BuildJob{TreeId: *treeID, Begin: uint64(index), End: uint64(next), TreeSharding: &sharding}
```
AlCutter: Would it be better to make this a ptr from the outset?

skylog/storage/gcp/pubsub/main.go
pavelkalinnikov: Ack. Let me address this comment and the one below in a separate PR, as it's orthogonal.

skylog/storage/gcp/pubsub/main.go
pavelkalinnikov: Will do in another PR.

skylog/storage/gcp/pubsub/main.go
pavelkalinnikov: Done.

skylog/storage/gcp/pubsub/main.go
pavelkalinnikov: #1710

PR #1704, opened 2019-06-19
Use *trillian.MapLeaf in mapTX.Set

gdbelvin:
The offical way to use protos is to pass by pointer, not by value.


Martin2112:
Towards #1738.

PR #1706, opened 2019-06-20
Fix a couple of reflection based comparisons on protos in tests

Martin2112:
This is not safe as protos have additional internal fields.

### Checklist

### DIFF
--- a/server/log_rpc_server_test.go
+++ b/server/log_rpc_server_test.go
@@ -19,7 +19,6 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -409,7 +408,7 @@ func TestGetLeavesByRange(t *testing.T) {
 		if test.wantErr != "" {
 			t.Errorf("GetLeavesByRange(%d, %+d)=_,nil; want nil, err containing %q", req.StartIndex, req.Count, test.wantErr)
 		}
-		if got := rsp.Leaves; !reflect.DeepEqual(got, test.want) {
+		if got := rsp.Leaves; !leavesEqual(got, test.want) {
 			t.Errorf("GetLeavesByRange(%d, %+d)=%+v; want %+v", req.StartIndex, req.Count, got, test.want)
 		}
 	}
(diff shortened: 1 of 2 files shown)


server/log_rpc_server_test.go
```diff
@@ -409,7 +408,7 @@ func TestGetLeavesByRange(t *testing.T) {
 		if test.wantErr != "" {
 			t.Errorf("GetLeavesByRange(%d, %+d)=_,nil; want nil, err containing %q", req.StartIndex, req.Count, test.wantErr)
 		}
-		if got := rsp.Leaves; !reflect.DeepEqual(got, test.want) {
+		if got := rsp.Leaves; !leavesEqual(got, test.want) {
```
gdbelvin:
```suggestion
		if got := rsp.Leaves; !gocmp.Equal(got, test.want, cmp.Comparer(proto.Equal)) {
```

 consider using [`go-cmp`](https://godoc.org/github.com/google/go-cmp/cmp)?

Martin2112: I was put off by their doc that says it "may" use `Equal` methods. As protos have exported internal fields if it doesn't use Equal it's still wrong.

gdbelvin: My understanding was that it would use the comparer when possible, and `reflect.Equal` otherwise:

https://godoc.org/github.com/google/go-cmp/cmp#Comparer

> Comparer returns an Option that determines whether two values are equal to each other.
> 
> The comparer f must be a function "func(T, T) bool" and is implicitly filtered to input values assignable to T. If T is an interface, it is possible that f is called with two values of different concrete types that both implement T.

Martin2112: It could just be poor wording in the go-cmp docs. We need to prevent `reflect.Equal` being used on protos as it contains exported fields that are not really part of the public state. If go-cmp always uses `Equal` then it's probably the right choice.

gdbelvin: `cmp.Comparer(proto.Equal)` would prevent `reflect.Equal` from being used on protos. 

I've had troubles before where reflect.Equal complained about the `XXX_` fields not being the same and I resolved it with `cmp.Comparer(proto.Equal)`

Martin2112: Done.

Martin2112:
Towards #1738.

PR #1715, opened 2019-06-21
Pass `trillian.SignedMapRoot` by reference

gdbelvin:
Modify the storage layer to accept and return `*trillian.SignedMapRoot`
Passing by reference is the only supported method of passing protobufs. 

- StoreSignedMapRoot
- LatestSignedMapRoot
- GetSignedMapRoot


Martin2112:
Yeah I was thinking we were OK because roots aren't protos but that's at
the CT level for LogRootV1.

Martin

PR #1716, opened 2019-06-21
Raise lint timeout from 5 to 8 minutes.

mhutchinson:
Out of 3 trials it takes 6:15 +/- 1s on my machine. Giving 8 minutes because other machines likely have fewer and/or slower cores.

Longer term, it would be great if golangci-lint could be made faster. This feels like too long for linting a codebase of this size.


mhutchinson:
Thanks @Martin2112

PR #1721, opened 2019-06-21
Cloud Build should use latest images as cache when building new images

RJPercival:
This will accelerate builds when some layers haven't changed or the image has already been built. It is recommended by [the Cloud Build docs](https://cloud.google.com/cloud-build/docs/speeding-up-builds#using_a_cached_docker_image).

### Checklist


RJPercival:
I'm not sure whether this will be sabotaged by the fact that the first thing our Docker images do is add the entire Git repo, which always changes between commits, invalidating all subsequent layers.

PR #1724, opened 2019-06-24
MapStorage Integration Tests

gdbelvin:
A testharness skeleton that can become a single unified test suite for the storage layer.

### Checklist

### DIFF
--- a/integration/storagetest/maptestrunner.go
+++ b/integration/storagetest/maptestrunner.go
@@ -1,0 +1,75 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storagetest
+
+import (
+	"context"
+	"path/filepath"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/google/trillian/storage"
+)
+
+type MapStorageFactory func(ctx context.Context, t *testing.T) (storage.MapStorage, storage.AdminStorage)
+type MapStorageTest func(ctx context.Context, t *testing.T, s storage.MapStorage, as storage.AdminStorage)
+
+// RunMapStorageTests runs all the map storage tests against the provided map storage implementation.
+func RunMapStorageTests(t *testing.T, storageFactory MapStorageFactory) {
+	ctx := context.Background()
+	for _, f := range mapTestFunctions(t, &MapTests{}) {
+		ms, as := storageFactory(ctx, t)
+		t.Run(functionName(f), func(t *testing.T) { f(ctx, t, ms, as) })
+	}
+}
+
+func functionName(x interface{}) string {
+	pc := reflect.ValueOf(x).Pointer()
+	nameFull := runtime.FuncForPC(pc).Name() // main.foo
+	nameEnd := filepath.Ext(nameFull)        // .foo
+	name := strings.TrimPrefix(nameEnd, ".") // foo
+	return strings.TrimPrefix(name, "Test")
+}
+
+func mapTestFunctions(t *testing.T, x interface{}) []MapStorageTest {
+	prefix := "Test"
+	xt := reflect.TypeOf(x)
+	xv := reflect.ValueOf(x)
+
+	tests := []MapStorageTest{}
+	for i := 0; i < xt.NumMethod(); i++ {
+		methodName := xt.Method(i).Name
+		if !strings.HasPrefix(methodName, prefix) {
+			continue
+		}
+		tests = append(tests, getTestFunc(t, xv, methodName))
+	}
+	return tests
+}
+
+func getTestFunc(t *testing.T, xv reflect.Value, name string) MapStorageTest {
+	m := xv.MethodByName(name)
+	if !m.IsValid() {
+		t.Fatalf("storagetest: function %v is not valid", name)
+	}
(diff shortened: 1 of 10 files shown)


Martin2112:
Not quite sure where this will end up but Go doesn't really encourage testing frameworks. It's maybe something we should talk about.

mhutchinson:
It seems reasonable to me to have a One True Storage Layer Test - test the storage layer API as a blackbox and then check all the implementations satisfy it. Perhaps testharness is a bit misleading. I think this is definitely something we should talk about.

gdbelvin:
We'll queue it up for rubber duck :-)

Martin2112:
OK, thanks.

Martin2112:
One option might be for us to stop pretending that these are unit tests. Anything that relies on an environment being set up could be moved to `integration` because they're not actually being run a lot of the time unless the right things are set up.

Need to think through how coverage etc. works in that model on Travis. Could maybe have a wrapper test that runs mysql integration or a build flag that pulls this in.

Disclaimer: These are just ideas, may not be easy or even feasible. The value of these ideas may go up or down.

gdbelvin:
Agreed that these are integration tests and belong in the integration test
folder

gdbelvin:
#1740

storage/testharness/maptest.go
```diff
@@ -0,0 +1,112 @@
...
+		TestCheckDatabaseAccessible,
+		TestMapSnapshot,
+	} {
+		s, as := storageFactory(ctx, t)
+		t.Run(functionName(f), func(t *testing.T) { f(ctx, t, s, as) })
+	}
+}
+
+func functionName(i interface{}) string {
+	pc := reflect.ValueOf(i).Pointer()
```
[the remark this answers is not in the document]
gdbelvin: I could manually maintain the list of names is done in `maptest`, but this seemed much cleaner.
https://github.com/google/trillian/blob/master/integration/maptest/map.go#L52

gdbelvin:
Addressed all comments to my knowledge. PTAL

phad:
Summary of an in-person discussion with @gdbelvin and @Martin2112 

In general I think the approach here is OK.  Reasons:
 * We don't get SetUp / TearDown support in the Go testing.T framework, and must work around that
 * The tests being done here require a protocol for creation of a specific database, cleaning between tests running the test, and teardown; codifing that protocol in one place is better than alternatives which could lead to copy/pasted code
 * Writing the tests as a library does have the effect of removing them from the code being tested, but in this situation we are concerned with integration tests and not unit tests, where you normally expect to see foo.go and foo_test.go side-by-side
 * Using reflection seems like a necessary step to avoid repetitive code such as {"TestMyThing", TestMyThing}.  I think over-use of reflection in a program leads to metaprogramming which can obscure what actually takes place if you're not very familiar with the setting.  If we can keep the use of reflection constrained to a small area of the code then that would help.
 * A bit more reflection to introspect the list of TestMapDoesFoo functions in TestMapStorage is probably justifiable since it's in the same file as the reflection that casts the test funcs to string func names; and because it will help avoid the developer error of adding a new TestMapDoesFoo while not extending the table

My suggested changes mainly try to avoid unforced errors for developers who work with this code:
 * We should avoid the misconception that this is a regular Go unit test suite, I suggest not naming the test runner file maptest.go (too similar to map_test.go) but something more like map_tester.go.
 * Avoid the misconception that a new TestMapDoesFoo func will be automatically run just because it starts 'Test' and takes a 't *testing.T', by either:
   - naming it differently; e.g. InnerTestMapDoesFoo  ;  DoTestMapDoesFoo; etc.
   - making them member functions of a new struct so that reflection can be used to build the test table in TestMapStorage  -  i.e. func TestMapStorage(...) becomes e.g. func (mt *MapTester) TestMapStorage(...)

Another suggestion is to have the storageFactory also return a string labeling the storage type, and using that in a top-level t.Run(storage_label, ...) so that the eventual test names are like:

MySql/TestMapSnapshot/activeMapSnapshot

Also, consider giving func(context.Context, *testing.T, storage.MapStorage, storage.AdminStorage)
a type name.

integration/storagetest/maptestrunner.go
```diff
@@ -0,0 +1,75 @@
...
+// RunMapStorageTests runs all the map storage tests against the provided map storage implementation.
+func RunMapStorageTests(t *testing.T, storageFactory MapStorageFactory) {
+	ctx := context.Background()
+	for _, f := range mapTestFunctions(t, &MapTests{}) {
+		ms, as := storageFactory(ctx, t)
+		t.Run(functionName(f), func(t *testing.T) { f(ctx, t, ms, as) })
+	}
+}
+
+func functionName(x interface{}) string {
```
phad: You could write a short doc comment here with an example to show what is returned.

integration/storagetest/maptestrunner.go
```diff
@@ -0,0 +1,75 @@
...
+	pc := reflect.ValueOf(x).Pointer()
+	nameFull := runtime.FuncForPC(pc).Name() // main.foo
+	nameEnd := filepath.Ext(nameFull)        // .foo
+	name := strings.TrimPrefix(nameEnd, ".") // foo
+	return strings.TrimPrefix(name, "Test")
+}
+
+func mapTestFunctions(t *testing.T, x interface{}) []MapStorageTest {
+	prefix := "Test"
```
phad: const

integration/storagetest/maptestrunner.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return tests
+}
+
+func getTestFunc(t *testing.T, xv reflect.Value, name string) MapStorageTest {
+	m := xv.MethodByName(name)
+	if !m.IsValid() {
+		t.Fatalf("storagetest: function %v is not valid", name)
+	}
+	f, ok := m.Interface().(func(ctx context.Context, t *testing.T, s storage.MapStorage, as storage.AdminStorage))
```
phad: Can't this be: `f, ok := m.Interface().(MapStorageTest)`  ?

gdbelvin: For some reason, that doesn't satisfy the type cast.  I suspect there is something special with Go types happening that I don't understand.

phad: OK, sounds like a type and another typedef'd to it are separate things as far as reflection is concerned.  That would be consistent with other strict behaviours we're used to seeing when trying to substitute types.

PR #1727, opened 2019-06-25
Pass context to storeSubtrees()

gdbelvin:
Commit() calls subtreeCache.Flush which calls storeSubtrees which
requires a context:

```
-                       return t.storeSubtrees(context.TODO(), st)
+                       return t.storeSubtrees(ctx, st)
```

### DIFF
--- a/storage/memory/log_storage.go
+++ b/storage/memory/log_storage.go
@@ -128,7 +128,7 @@ func (m *memoryLogStorage) Snapshot(ctx context.Context) (storage.ReadOnlyLogTX,
 	return &readOnlyLogTX{m.TreeStorage}, nil
 }
 
-func (t *readOnlyLogTX) Commit(_ context.Context) error {
+func (t *readOnlyLogTX) Commit(context.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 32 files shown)


storage/memory/log_storage.go
```diff
@@ -128,7 +128,7 @@ func (m *memoryLogStorage) Snapshot(ctx context.Context) (storage.ReadOnlyLogTX,
 	return &readOnlyLogTX{m.TreeStorage}, nil
 }
 
-func (t *readOnlyLogTX) Commit() error {
+func (t *readOnlyLogTX) Commit(_ context.Context) error {
```
Martin2112: Don't think you need the '_'.

storage/mysql/log_storage.go
```diff
@@ -187,7 +187,7 @@ func (m *mySQLLogStorage) Snapshot(ctx context.Context) (storage.ReadOnlyLogTX,
 	return &readOnlyLogTX{m, &sync.Mutex{}, tx}, nil
 }
 
-func (t *readOnlyLogTX) Commit() error {
+func (t *readOnlyLogTX) Commit(_ context.Context) error {
```
Martin2112: Same here.

storage/postgres/log_storage.go
```diff
@@ -216,7 +216,7 @@ func (m *postgresLogStorage) Snapshot(ctx context.Context) (storage.ReadOnlyLogT
 	return &readOnlyLogTX{m, tx}, nil
 }
 
-func (t *readOnlyLogTX) Commit() error {
+func (t *readOnlyLogTX) Commit(_ context.Context) error {
```
Martin2112: And here.

PR #1728, opened 2019-06-25
Storage test helpers accept storage interfaces

gdbelvin:
By passing storage.{Log,Map,Admin}Storage to test helpers, we can abstract away the specifics of the storage implementation from its tests.

### DIFF
--- a/storage/mysql/storage_test.go
+++ b/storage/mysql/storage_test.go
@@ -244,7 +244,7 @@ func cleanTestDB(db *sql.DB) {
 	}
 }
 
-func createFakeSignedLogRoot(ctx context.Context, t *testing.T, l storage.LogStorage, tree *trillian.Tree, treeSize uint64) {
+func mustSignAndStoreLogRoot(ctx context.Context, t *testing.T, l storage.LogStorage, tree *trillian.Tree, treeSize uint64) {
 	t.Helper()
 	signer := tcrypto.NewSigner(0, testonly.NewSignerWithFixedSig(nil, []byte("notnil")), crypto.SHA256)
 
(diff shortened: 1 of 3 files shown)


storage/mysql/storage_test.go
```diff
@@ -240,11 +244,9 @@ func cleanTestDB(db *sql.DB) {
 	}
 }
 
-func createFakeSignedLogRoot(db *sql.DB, tree *trillian.Tree, treeSize uint64) {
+func createFakeSignedLogRoot(ctx context.Context, t *testing.T, l storage.LogStorage, tree *trillian.Tree, treeSize uint64) {
```
Martin2112: Looks like this was missing a `t.Helper()`

storage/mysql/storage_test.go
```diff
@@ -256,36 +258,20 @@ func createFakeSignedLogRoot(db *sql.DB, tree *trillian.Tree, treeSize uint64) {
...
-		panic(fmt.Sprintf("ReadWriteTransaction() = %v", err))
+		t.Fatalf("ReadWriteTransaction() = %v", err)
 	}
 }
 
 // createTree creates the specified tree using AdminStorage.
-func createTree(db *sql.DB, tree *trillian.Tree) (*trillian.Tree, error) {
-	ctx := context.Background()
-	s := NewAdminStorage(db)
+func createTree(ctx context.Context, t *testing.T, s storage.AdminStorage, tree *trillian.Tree) *trillian.Tree {
```
Martin2112: can you rename this to something like `mustCreateTree` or `createTreeOrDie` to make it obvious it can exit.

storage/mysql/storage_test.go
Martin2112: Pre-existing but would be better if named `signRootOrDie` or whatever matches rename of fn below.

AlCutter:
Might be worth raising a tracking issue about the larger goal of factoring out the MySQL (etc.) tests into a common storage impl test framework?

PR #1729, opened 2019-06-25
Remove unused Snapshot function

gdbelvin:


gdbelvin:
This interface just had a bunch of TODOs about deleting. 
It's not part of any storage interface and can't be called anyway.
Since I'm working on improving testing, it made sense to remove unused code.

We can put it back in with proper tests when we need it.

PR #1730, opened 2019-06-25
Take advantage of Go Modules in Docker images

RJPercival:
Force `$GO111MODULE=on`, since we don't support non-Modules builds, and fetch Trillian's dependencies before adding it to the image, as this improves cacheability of layers and therefore speeds up image builds.

### Checklist


RJPercival:
I wonder if we'd be better off having only a single Docker image that contains all 3 binaries?

PR #1732, opened 2019-06-25
Added interceptor for Map WriteLeaves added in #1655.

mhutchinson:
Without this change, quota enforcement for WriteLeaves will not be applied even in environments where quota enforcement is enabled.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,7 +5,8 @@ run:
 linters-settings:
   gocyclo:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
-    min-complexity: 25
+    # TODO(mhutchinson): lower this again after reworking interceptor
+    min-complexity: 26
   depguard:
     list-type: blacklist
     packages:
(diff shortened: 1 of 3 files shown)


gdbelvin:
server/interceptor/interceptor.go:332:1: cyclomatic complexity 26 of func `newRPCInfoForRequest` is high (> 25) (gocyclo)

mhutchinson:
As an update on this, I've gone down a rabbit hole investigating whether using google.protobuf.FieldOptions will allow us to include the relevant information as metadata on the request/response types instead of having this global interceptor knowing the mechanics of all other surfaces in the system.

In the interests of getting this fix in, I've taken the pragmatic step of increasing the threshold by 1 for now.

gdbelvin:
Sounds good.
Not to further your rabbit hole adventure, but this might be relevant:
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master//ratelimit/https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master///ratelimit/ratelimit.go#L14:2

gdbelvin:
and this: https://godoc.org/github.com/golang/protobuf/descriptor

PR #1735, opened 2019-06-27
Fix storage API Get / Set root passing SignedLogRoot proto by value

Martin2112:
And associated fixups and test updates.

# Checklist


gdbelvin:
Maybe we can link all these CLs together with an issue?

Martin2112:
Sounds organized. I think there were a few PRs last week as well.

Martin2112:
Towards #1738.

PR #1745, opened 2019-07-01
Increase Cloud Build timeout from 15 mins to 20 mins

RJPercival:
Since we enabled Go Modules, builds are taking significantly longer and
often timing out.

### Checklist

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,4 +1,4 @@
-timeout: 900s
+timeout: 1200s
 substitutions:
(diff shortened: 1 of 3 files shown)


RJPercival:
> Is it worth writing the timeouts in `minutes`?

You'd think that'd be possible, given the "s" suffix, wouldn't you? However, it turns out [the only supported suffix is "s"](https://cloud.google.com/cloud-build/docs/build-config#timeout_2), i.e. you can't write the timeout in minutes :disappointed:

PR #1761, opened 2019-07-09
merkle: Helper to get node IDs of a compact range

pavelkalinnikov:
This change adds the `RangeNodes` function that returns IDs of nodes comprising a compact range.
It supersedes `RangeNodesForPrefix` function, which is now removed.


pavelkalinnikov:
@googlebot rescan

PR #1782, opened 2019-07-23
Use letsencrypt/pkcs11key/v3

gdbelvin:
Resolves letsencrypt/pkcs11key#22


cmd/createtree/pkcs11.go
```diff
@@ -26,7 +26,8 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/trillian/cmd/createtree/keys"
 	"github.com/google/trillian/crypto/keyspb"
-	"github.com/letsencrypt/pkcs11key"
+
+	pkcs11key "github.com/letsencrypt/pkcs11key/v3"
```
RJPercival: This import alias shouldn't be necessary - it should automatically be imported as "pkcs11key", assuming the version suffix is being treated correctly. However, I'm not entirely sure that letsencrypt/pkcs11key#22 did the right thing by moving all of the code into a "v3" directory. I was under the impression that the version suffix isn't supposed to be a literal filepath, but rather just an identifier of which major version to use?

```suggestion
	"github.com/letsencrypt/pkcs11key/v3"
```

RJPercival: [Reading into this further](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher), it seems that moving all of the code into a subdirectory is a supported option. Does it break the ability of Go to provide the correct package name on import though?

gdbelvin: I think the recommended pattern is to have `/vN` at the root of the module path.
This module however only has one package, so ending in `v3` is a bit unfortunate.  Perhaps it should be `github.com/letsencrypt/pkcs11key/v3/pkcs11key` ?

RJPercival: I guess so, but it's not a big deal.

PR #1784, opened 2019-08-02
Ensure MapHammer deadline errors are propagated

AlCutter:
Previously, if the MapHammer simply timed out on an RPC, no error would be returned.
This PR attempts to correct that.

### Checklist

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -166,8 +166,7 @@ func (c MapConfig) String() string {
 }
 
 // HitMap performs load/stress operations according to given config.
-func HitMap(cfg MapConfig) error {
-	ctx := context.Background()
+func HitMap(ctx context.Context, cfg MapConfig) error {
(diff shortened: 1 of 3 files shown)


testonly/hammer/hammer.go
```diff
@@ -481,6 +480,10 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
 		}
 
 		if time.Now().After(deadline) {
+			if firstErr == nil {
+				// If there was no other error, we've probably hit the deadline - make sure we bubble that up.
+				firstErr = ctx.Err()
```
Martin2112: It's a bit annoying that the context error doesn't have a `DEADLINE_EXCEEDED`  status. Not sure it's worth returning one explicitly though as it won't be handled differently.

AlCutter: Yeah, tempted to leave it for now.

PR #1791, opened 2019-08-21
Update go.mod with go 1.13

gdbelvin:
Verify that everything works with go 1.13. 

Blocked on https://github.com/golangci/golangci-lint/pull/643

Currently golangci-lint has dependencies with invalid pseudo-versions.
This breaks things in go 1.13 because go 1.13 has stricter dependency checks and goproxy is enabled by default.


RJPercival:
Since #1801, we should no longer be blocked on golangci/golangci-lint#643.

PR #1793, opened 2019-08-30
Removed overly spammy trace spans from buildSubtree.

mhutchinson:
buildSubtree is called potentially 256 + 1 times each time a batch of leaves is written. This could cause ~1k spans to be written, which is more than this tracing library anticipated.

Also added some documentation to make immediately clear the expected ebb and flow of the buildSubtree method.

### DIFF
--- a/merkle/sparse_merkle_tree.go
+++ b/merkle/sparse_merkle_tree.go
@@ -239,12 +239,15 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context, queueSize int) {
 					NodeRevision: s.treeRevision,
 				})
 		}
+		ctx, postQueueCloseEnd := spanFor(ctx, "buildSubtree.runTX.postQueueClose")
+		defer postQueueCloseEnd()
 
 		// Prewarm the cache:
 		if _, err := tx.GetMerkleNodes(ctx, s.treeRevision, sibs); err != nil {
 			return fmt.Errorf("failed to preload node hash cache: %s", err)
 		}
 
+		hsCtx, hstar2SpanEnd := spanFor(ctx, "buildSubtree.runTX.hstar2")
 		// calculate new root, and intermediate nodes:
 		hs2 := NewHStar2(s.treeID, s.hasher)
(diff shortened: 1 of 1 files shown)


merkle/sparse_merkle_tree.go
```diff
@@ -242,16 +239,12 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context, queueSize int) {
...
 
-		preloadCtx, preloadSpanEnd := spanFor(ctx, "buildSubtree.runTX.preload")
 		// Prewarm the cache:
-		if _, err := tx.GetMerkleNodes(preloadCtx, s.treeRevision, sibs); err != nil {
+		if _, err := tx.GetMerkleNodes(ctx, s.treeRevision, sibs); err != nil {
 			return fmt.Errorf("failed to preload node hash cache: %s", err)
 		}
-		preloadSpanEnd()
 
-		hsCtx, hstar2SpanEnd := spanFor(ctx, "buildSubtree.runTX.hstar2")
```
pavelkalinnikov: FYI: This bit contributes the most to the running time of `buildSubtree`. We wouldn't know that without this trace.
Although, I agree with the motivation of this PR, and that we should delete these spans, or at least disable them by default.

pavelkalinnikov: There is one idea on why hstar2 is slow: there is much of lock contention on the tree cache. Can we land this after we fix that and see the impact?

mhutchinson: Take another look. I've left 2 spans per buildSubtree call:
- an outer span to measure everything after we close the queue
- an inner span to measure the subset of that time we spend doing hstar2

I think this meets your use-case, and reduces the number of spans by 60%, and makes the information more useful (the time spent doing stuff before the queue is closed isn't particularly worrying). Sounds like a win.

pavelkalinnikov: LGTM. You can also put a TODO to remove the remaining spans.

PR #1795, opened 2019-09-03
HStar2: Add and use prefetching

pavelkalinnikov:
This change adds a new `Prefetch` method to `HStar2` algorithm, which lists all nodes that the algorithm will read. This method is now used instead of the half-baked cache prefetch optimization that read all siblings of nodes on the paths to all the written leaves.

### DIFF
--- a/merkle/hstar2.go
+++ b/merkle/hstar2.go
@@ -59,7 +59,11 @@ func NewHStar2(treeID int64, hasher hashers.MapHasher) HStar2 {
 // which contains the given set of non-null leaves.
 func (s *HStar2) HStar2Root(depth int, values []*HStar2LeafHash) ([]byte, error) {
 	sort.Sort(ByIndex{values})
-	return s.hStar2b(0, depth, values, smtZero, nil, nil)
+	combine := func(depth int, offset *big.Int, lhs, rhs []byte) ([]byte, error) {
+		h := s.hasher.HashChildren(lhs, rhs)
+		return h, nil
+	}
+	return s.hStar2b(0, depth, values, smtZero, nil, combine)
 }
 
 // SparseGetNodeFunc should return any pre-existing node hash for the node address.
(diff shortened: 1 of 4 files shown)


merkle/hstar2_test.go
```diff
@@ -99,7 +99,8 @@ func TestHStar2SimpleDataSetKAT(t *testing.T) {
 }
 
 // TestHStar2GetSet ensures that we get the same roots as above when we
-// incrementally calculate roots.
+// incrementally calculate roots. It also makes sure that Prefetch visits the
+// same set of node that HStar2Nodes fetches.
```
pavelkalinnikov: nodes

merkle/sparse_merkle_tree.go
```diff
@@ -243,22 +237,34 @@ func (s *subtreeWriter) buildSubtree(ctx context.Context, queueSize int) {
...
-		var err error
+		// Calculate new root hash, and intermediate nodes.
 		root, err = hs2.HStar2Nodes(s.prefix, s.subtreeDepth, leaves,
 			func(depth int, index *big.Int) ([]byte, error) {
 				nodeID := storage.NewNodeIDFromBigInt(depth, index, s.hasher.BitLen())
 				if glog.V(4) {
 					glog.Infof("buildSubtree.get(%x, %d) nid: %x, %v",
 						index.Bytes(), depth, nodeID.Path, nodeID.PrefixLenBits)
 				}
+				// TODO(pavelkalinnikov): Reuse the result of prefetch GetMerkleNodes.
```
pavelkalinnikov: The corresponding PR is ready, blocked on this one.

merkle/hstar2.go
```diff
@@ -83,19 +87,51 @@ func (s *HStar2) HStar2Nodes(prefix []byte, subtreeDepth int, values []*HStar2Le
...
+	combine := func(depth int, offset *big.Int, lhs, rhs []byte) ([]byte, error) {
+		return nil, nil
+	}
+	_, err := s.run(prefix, subtreeDepth, values, visit, combine)
+	return err
+}
+
+// combineFunc returns a node hash based on two child hashes. It may also do
+// side effects, e.g. put the resulting node to storage.
+type combineFunc func(depth int, offset *big.Int, lhs, rhs []byte) ([]byte, error)
```
mhutchinson: It took me a little while to work out why this function was needed over the logically cleaner "set" function. My calculated answer is that in the prefetch case you don't want to actually be doing any lookups for node values in-line, and thus you don't have any hashes to combine. By passing in the combine function you can avoid the nil hash error. Consider making this explicit in this comment, e.g. "this method may perform no work if none is needed, such as the prefetch case."

merkle/hstar2.go
```diff
@@ -83,19 +87,51 @@ func (s *HStar2) HStar2Nodes(prefix []byte, subtreeDepth int, values []*HStar2Le
...
+			return nil, err
+		}
+		return h, nil
+	}
+	return s.run(prefix, subtreeDepth, values, get, combine)
+}
+
+// Prefetch does a dry run of HStar2 algorithm, and reports all Merkle tree
+// nodes that it needs through the passed-in visit function. Note that the
+// return value of the visit function is ignored, unless it is an error.
```
mhutchinson: Consider whether a different function signature for the visit argument is better than SparseGetNodeFunc. Having a return argument that isn't used is a little messy, and I can't see too many use-cases where clients would want to pass the same SparseGetNodeFunc into this as the other places it is used: if a prefetch is useful then the visit will likely always be a Node ID accumulator.

pavelkalinnikov: Done.

merkle/hstar2.go
pavelkalinnikov: Done.

merkle/hstar2.go
```diff
@@ -59,13 +59,16 @@ func NewHStar2(treeID int64, hasher hashers.MapHasher) HStar2 {
...
-	combine := func(depth int, offset *big.Int, lhs, rhs []byte) ([]byte, error) {
+	combine := func(depth int, index *big.Int, lhs, rhs []byte) ([]byte, error) {
 		h := s.hasher.HashChildren(lhs, rhs)
 		return h, nil
 	}
 	return s.hStar2b(0, depth, values, smtZero, nil, combine)
 }
 
+// PrefetchNodeFunc reports coordinates of a Merkle tree node to prefetch.
+type PrefetchNodeFunc func(depth int, index *big.Int)
```
mhutchinson: Optional: You could call this PrefetchNodeVisitor if you want to give a hat tip to the design pattern it mirrors:
https://en.wikipedia.org/wiki/Visitor_pattern

pavelkalinnikov: Done.

PR #1796, opened 2019-09-04
Sparse Merkle tree: Prefetch nodes to a local cache

pavelkalinnikov:
This change speeds up sparse Merkle tree construction, by using a local per-HStar2 node cache instead of the global per-transaction cache, which significantly reduces lock contention.

Benchmarking on `integration/maptest/map.go/WriteStress` test with 1024 leaves shows a 2x speedup of `WriteLeaves` (before: 3.57s; after: 1.88s).

### DIFF
--- a/merkle/sparse_merkle_tree_test.go
+++ b/merkle/sparse_merkle_tree_test.go
@@ -443,12 +443,14 @@ func testSparseTreeFetches(ctx context.Context, t *testing.T, vec sparseTestVect
 		readMutex.Lock()
 		defer readMutex.Unlock()
 
-		state, ok := reads[ids[0].String()]
-		reads[ids[0].String()] = "met"
-
-		// Allow re-reads of the same nodes too (state=="met") - this is due to
-		// cache warming:
-		return ok && (state == "unmet" || state == "met")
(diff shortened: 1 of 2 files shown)


merkle/sparse_merkle_tree_test.go
```diff
@@ -443,12 +443,14 @@ func testSparseTreeFetches(ctx context.Context, t *testing.T, vec sparseTestVect
...
-
-		// Allow re-reads of the same nodes too (state=="met") - this is due to
-		// cache warming:
-		return ok && (state == "unmet" || state == "met")
+		for _, id := range ids {
+			strID := id.String()
+			if reads[strID] != "unmet" {
+				return false
+			}
+			reads[strID] = "met"
```
gdbelvin: will this line ever be reached?

pavelkalinnikov: Yes, it must be, otherwise the test will fail.

pavelkalinnikov: The "unmet" line above will never happen, for the same reason - the test will fail. But it was there for ease of debugging in case it does fail.

PR #1798, opened 2019-09-04
subtree_cache: Lock reads and writes separately

pavelkalinnikov:
This PR removes a mutex lock in `SubtreeCache` from the place where database reads are. Now only the in-memory reads and writes are under lock. This allows double-reads from DB if multiple parallel reads of the same node hash occur, but, having looked at the call sites, it never happens. The cache is used on per-transaction basis, and within one transaction all reads are either done in one batch, or in batches isolated from each other.

This speeds up Map integration tests by 5-10%.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -149,30 +149,34 @@ func (s *SubtreeCache) stratumInfoForPrefixLength(l int) stratumInfo {
 // passed in node IDs, uses getSubtrees to retrieve them, and finally populates
 // the cache structures with the data.
 func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc) error {
-	s.mutex.Lock()
-	defer s.mutex.Unlock()
+	// Figure out the set of subtrees we need.
+	want := make(map[string]storage.NodeID)
+	func() {
+		s.mutex.RLock()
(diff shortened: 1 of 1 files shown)


storage/cache/subtree_cache.go
```diff
@@ -149,30 +149,34 @@ func (s *SubtreeCache) stratumInfoForPrefixLength(l int) stratumInfo {
...
+	func() {
+		s.mutex.RLock()
+		defer s.mutex.RUnlock()
+
+		for _, id := range ids {
+			sInfo := s.stratumInfoForNodeID(id)
+			pxKey := id.PrefixAsKey(sInfo.prefixBytes)
+			// TODO(al): Fix for non-uniform strata.
+			id.PrefixLenBits = sInfo.prefixBytes * depthQuantum
+			if _, ok := s.subtrees[pxKey]; !ok {
```
mhutchinson: I assume the lines above this that are effectively statically mapping a node ID to a strata key are super fast. If that isn't the case, the creation of the set of all required strata could happen outside the critical block, and then the subtraction of the subtrees already in the cache could be the only part inside the mutex.

pavelkalinnikov: It's not *super* fast, as it involves one string copy, but it doesn't look horrible.

PR #1802, opened 2019-09-05
Release v1.3.2

RJPercival:
Patch release that just adds the fix in PR #1801 on top of v1.3.1.

### Checklist


RJPercival:
Sadly this didn't have the desired effect because `replace` directives are ignored when building other modules. Therefore, things that depends on us have to copy those `replace` directives into their own go.mod files (requiring them to create one if they don't have one already).

RJPercival:
It may be better to remove golangci-lint from tools.go until they've fixed their go.mod file, otherwise we'll continue propagating this problem to anyone that depends on us.

mhutchinson:
See https://github.com/golangci/golangci-lint/pull/643 - it's blocked at the moment.

PR #1805, opened 2019-09-06
Refactored logical operations into methods from within SetLeaves.

mhutchinson:
This herds the code in a direction where updateTree() can be called asynchronously, thus allowing for greater parallelism in the write path. This is a pure refactoring; no functional changes are intended here.

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -358,6 +358,7 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 		if err != nil {
 			return err
 		}
+		glog.V(2).Infof("%v: Writing at revision %v", tree.TreeId, writeRev)
 
 		hkv, err := t.writeLeaves(ctx, tree, hasher, tx, req.Leaves, writeRev)
 		if err != nil {
@@ -365,10 +366,7 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 		}
 
 		newRoot, err = t.updateTree(ctx, tree, hasher, tx, hkv, req.Metadata, writeRev)
-		if err != nil {
-			return err
-		}
-		return nil
+		return err
 	})
 	if err != nil {
 		return nil, err
@@ -387,7 +385,6 @@ func (t *TrillianMapServer) writeRevision(ctx context.Context, tree *trillian.Tr
 	if rev != 0 && writeRev != rev {
 		return 0, status.Errorf(codes.FailedPrecondition, "can't write to revision %v", rev)
 	}
-	glog.V(2).Infof("%v: Writing at revision %v", tree.TreeId, writeRev)
 	return writeRev, nil
 }
 
@@ -395,28 +392,29 @@ func (t *TrillianMapServer) writeRevision(ctx context.Context, tree *trillian.Tr
 func (t *TrillianMapServer) writeLeaves(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, leaves []*trillian.MapLeaf, rev int64) ([]merkle.HashKeyValue, error) {
 	hkv := make([]merkle.HashKeyValue, 0, len(leaves))
 	for _, l := range leaves {
+		// TODO(mhutchinson): make this check and perform hashing and hkv building outside of the transaction scope
 		if err := checkIndexSize(l.Index, hasher); err != nil {
 			return nil, err
 		}
 		l.LeafHash = hasher.HashLeaf(tree.TreeId, l.Index, l.LeafValue)
-
-		if err := tx.Set(ctx, l.Index, l); err != nil {
-			return nil, err
-		}
 		hkv = append(hkv, merkle.HashKeyValue{
 			HashedKey:   l.Index,
 			HashedValue: l.LeafHash,
 		})
+		if err := tx.Set(ctx, l.Index, l); err != nil {
+			return nil, err
+		}
 	}
 	return hkv, nil
 }
 
-// updateTree calculates the latest merkle values within the tree for the given leaf updates at
-// the specified revision. The new signed map root is committed to storage and returned.
+// updateTree updates the sparse Merkle tree at the specified revision based on the passed-in
+// leaf changes, and writes it to the storage. Returns the new signed map root, which is also
+// submitted to storage.
 func (t *TrillianMapServer) updateTree(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, hkv []merkle.HashKeyValue, metadata []byte, rev int64) (*trillian.SignedMapRoot, error) {
 	// Work around a performance issue when using the map in
 	// single-transaction mode by preloading all the nodes we know the
-	// sparse merkle writer is going to need.
+	// sparse Merkle writer is going to need.
 	if t.opts.UseSingleTransaction && t.opts.UseLargePreload {
 		if err := doPreload(ctx, tx, hasher.BitLen(), hkv); err != nil {
 			return nil, err
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -359,7 +359,7 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 			return err
 		}
 
-		hkv , err := t.writeLeaves(ctx, tree, hasher, tx, req.Leaves, writeRev)
+		hkv, err := t.writeLeaves(ctx, tree, hasher, tx, req.Leaves, writeRev)
 		if err != nil {
 			return err
 		}
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 
 	var newRoot *trillian.SignedMapRoot
 	err = t.registry.MapStorage.ReadWriteTransaction(ctx, tree, func(ctx context.Context, tx storage.MapTreeTX) error {
-		writeRev, err := tx.WriteRevision(ctx)
-		if err != nil {
-			return err
-		}
-		if rev := req.Revision; rev != 0 && writeRev != rev {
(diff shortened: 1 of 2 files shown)


server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
 	err = t.registry.MapStorage.ReadWriteTransaction(ctx, tree, func(ctx context.Context, tx storage.MapTreeTX) error {
-		writeRev, err := tx.WriteRevision(ctx)
+		writeRev, err := t.writeRevision(ctx, tree, tx, req.Revision)
 		if err != nil {
 			return err
 		}
-		if rev := req.Revision; rev != 0 && writeRev != rev {
-			return status.Errorf(codes.FailedPrecondition, "can't write to revision %v", rev)
+
+		hkv , err := t.writeLeaves(ctx, tree, hasher, tx, req.Leaves, writeRev)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		hkv, err := t.writeLeaves(ctx, tree, hasher, tx, req.Leaves, writeRev)
```

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
-			}
-			hkv = append(hkv, merkle.HashKeyValue{
-				HashedKey:   l.Index,
-				HashedValue: l.LeafHash,
-			})
+// writeLeaves updates the leaf values, but does not calculate nor update the Merkle tree.
+func (t *TrillianMapServer) writeLeaves(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, leaves []*trillian.MapLeaf, rev int64) ([]merkle.HashKeyValue, error) {
+	hkv := make([]merkle.HashKeyValue, 0, len(leaves))
+	for _, l := range leaves {
+		if err := checkIndexSize(l.Index, hasher); err != nil {
```
gdbelvin: I wonder if this should be part of input validation much earlier?

mhutchinson: Yup. I am considering that for a follow-up CL which would change the behaviour slightly. It would be good to do this work outside of the transaction.

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+	return &trillian.SetMapLeavesResponse{MapRoot: newRoot}, nil
+}
 
-		hkv := make([]merkle.HashKeyValue, 0, len(req.Leaves))
-		for _, l := range req.Leaves {
-			if err := checkIndexSize(l.Index, hasher); err != nil {
-				return err
-			}
-			l.LeafHash = hasher.HashLeaf(mapID, l.Index, l.LeafValue)
+// writeRevision ensures the desired write revision is available and claim a lock on this revision
```
pavelkalinnikov: The lock claiming behavior is DB-specific and doesn't necessarily take place. In some there is optimistic / multi-versioned concurrency control, where the fact of revision not progressing is checked when the TX is committed.

I suggest focusing on the logical meaning of this "locking", e.g.:

```golang
// writeRevision returns the map revision that the passed-in transaction writes
// to. It verifies that this revision corresponds to the expected value, which
// helps sorting out conflicting transactions, should they occur.
```

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+// committed.
+func (t *TrillianMapServer) writeRevision(ctx context.Context, tree *trillian.Tree, tx storage.MapTreeTX, rev int64) (int64, error) {
+	writeRev, err := tx.WriteRevision(ctx)
+	if err != nil {
+		return 0, err
+	}
+	if rev != 0 && writeRev != rev {
+		return 0, status.Errorf(codes.FailedPrecondition, "can't write to revision %v", rev)
+	}
+	glog.V(2).Infof("%v: Writing at revision %v", tree.TreeId, writeRev)
```
pavelkalinnikov: nit: I'd put the logging up the stack, and leave `writeRevision` a "pure" function.

server/map_rpc_server.go
pavelkalinnikov: +1, but probably out of scope of this PR. TODO it?

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+		l.LeafHash = hasher.HashLeaf(tree.TreeId, l.Index, l.LeafValue)
 
-		// Work around a performance issue when using the map in
-		// single-transaction mode by preloading all the nodes we know the
-		// sparse merkle writer is going to need.
-		if t.opts.UseSingleTransaction && t.opts.UseLargePreload {
-			if err := doPreload(ctx, tx, hasher.BitLen(), hkv); err != nil {
-				return err
-			}
+		if err := tx.Set(ctx, l.Index, l); err != nil {
```
pavelkalinnikov: out-of-scope nit to the API design: The `l.Index` parameter is redundant.

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+			return nil, err
 		}
+		hkv = append(hkv, merkle.HashKeyValue{
```
pavelkalinnikov: for future PR: This can also be bundled with the initial validation.

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
 		}
-		glog.V(2).Infof("%v: Writing at revision %v", mapID, writeRev)
-		txRunner := t.newTXRunner(tree, tx)
-		smtWriter, err := merkle.NewSparseMerkleTreeWriter(ctx, req.MapId, writeRev, hasher, txRunner)
+
+		newRoot, err = t.updateTree(ctx, tree, hasher, tx, hkv, req.Metadata, writeRev)
 		if err != nil {
 			return err
 		}
+		return nil
```
pavelkalinnikov: Squash it with the previous 3 lines to just `return err`.

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+			HashedKey:   l.Index,
+			HashedValue: l.LeafHash,
+		})
+	}
+	return hkv, nil
+}
 
-		if err = smtWriter.SetLeaves(ctx, hkv); err != nil {
-			return err
+// updateTree calculates the latest merkle values within the tree for the given leaf updates at
```
pavelkalinnikov: nit1: **M**erkle
nit2: Not only calculates, but stores it

Suggestion:
```golang
// updateTree updates the sparse Merkle tree at the specified revision based on
// the passed-in leaf changes, and writes it to the storage. Returns the new
// signed map root, which is also submitted to storage.
```

server/map_rpc_server.go
```diff
@@ -354,66 +354,98 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
...
+// the specified revision. The new signed map root is committed to storage and returned.
+func (t *TrillianMapServer) updateTree(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, hkv []merkle.HashKeyValue, metadata []byte, rev int64) (*trillian.SignedMapRoot, error) {
+	// Work around a performance issue when using the map in
+	// single-transaction mode by preloading all the nodes we know the
+	// sparse merkle writer is going to need.
```
pavelkalinnikov: While you are here: **M**erkle

server/map_rpc_server.go
mhutchinson: Hah, good call. This looked quite n00b now I read it again :-D

server/map_rpc_server.go
mhutchinson: I feel like it should be a requirement of our data layer that it won't allow multiple updates to be stored under the same revision number. Which data layers allow transactions with conflicting revision IDs to be applied?

server/map_rpc_server.go
mhutchinson: Added a TODO to map_storage.go

server/map_rpc_server.go
mhutchinson: Added this to the TODO above.

server/map_rpc_server.go
mhutchinson: Nothing like a refactoring to find out all the stuff that has bothered people quietly for a long time :-D

server/map_rpc_server.go
pavelkalinnikov: Sure, I'm just saying that the comment probably shouldn't have details on how the conflict resolution works, because it depends on the database used within the storage layer. MySQL has locks, but something like CockroachDB is optimistic (I think).

pavelkalinnikov: From CockroachDB [docs](https://www.cockroachlabs.com/blog/serializable-lockless-distributed-isolation-cockroachdb/):

> **Lockless**: Operations execute without taking locks on resources. Correctness is enforced by aborting transactions which would violate either serializability or strict scheduling.

pavelkalinnikov: > I feel like it should be a requirement of our data layer that it won't allow multiple updates to be stored under the same revision number.

I think it is true now, multiple updates aren't possible.

Unrelated to this PR: In a massively parallel world though, you can imagine multiple parallel shards (split by leaf hash) writing leaves to the same revision, which is then "finalized" by a "master".

mhutchinson: Leaving as-is for now. The docs are saying what & why the code is at this revision. Currently this is correct and is what prevents us from creating conflicting revision states in the DB. As we add lock-free or massively parallel clever stuff, this will need to be revisited.

PR #1811, opened 2019-09-09
merkle: Drop dependency on storage.NodeID for logs

pavelkalinnikov:
This change pushes `storage.NodeID` usage for logs up the stack. As a
result, the log-specific subset of `merkle` package no longer depends on
`storage`, and uses its own lightweight `compact.NodeID` type.

### DIFF
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -30,7 +30,7 @@ import (
 	"strings"
 
 	"github.com/google/trillian/merkle"
-	"github.com/google/trillian/storage"
+	"github.com/google/trillian/merkle/compact"
 )
 
 const (
@@ -77,9 +77,6 @@ const (
 
 	// Maximum number of ranges to allow.
 	maxRanges = 3
-
-	// maxLen is a suitably large maximum nodeID length for storage.NodeID.
-	maxLen = 64
 )
 
 var (
@@ -247,16 +244,14 @@ func renderTree(prefix string, treeSize, index int64) {
 }
 
 // nodeKey returns a stable node identifier for the passed in node coordinate.
-func nodeKey(height, index int64) string {
-	return fmt.Sprintf("%d.%d", height, index)
+func nodeKey(level, index int64) string {
+	return fmt.Sprintf("%d.%d", level, index)
 }
 
-// toNodeKey converts a storage.NodeID to the corresponding stable node
+// toNodeKey converts a compact.NodeID to the corresponding stable node
 // identifier used by this tool.
-func toNodeKey(n storage.NodeID) string {
-	d := int64(maxLen - n.PrefixLenBits)
-	i := n.BigInt().Int64() >> uint(d)
-	return nodeKey(d, i)
+func toNodeKey(n compact.NodeID) string {
+	return nodeKey(int64(n.Level), int64(n.Index))
 }
 
 // decompose splits out a range into component subtrees.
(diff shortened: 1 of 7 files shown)


merkle/merkle_path.go
```diff
@@ -32,13 +32,13 @@ const vvLevel = 4
...
 type NodeFetch struct {
-	NodeID storage.NodeID
+	ID     compact.NodeID
 	Rehash bool
 }
 
 // Equivalent return true iff the other represents the same rehash state and NodeID as the other.
 func (n NodeFetch) Equivalent(other NodeFetch) bool {
-	return n.Rehash == other.Rehash && n.NodeID.Equivalent(other.NodeID)
+	return n.Rehash == other.Rehash && n.ID == other.ID
```
Martin2112: `Equivalent` in `storage.NodeID` is a deep compare on the path prefix. There's no similar concept in `compact.NodeID` so I think this version of `Equivalent` should be renamed to avoid confusion.

Alternatively, as it's only used in tests so it maybe can be deleted and the structs compared directly in the test.

docs/merkletree/treetex/main.go
```diff
@@ -247,16 +244,14 @@ func renderTree(prefix string, treeSize, index int64) {
...
 }
 
-// toNodeKey converts a storage.NodeID to the corresponding stable node
+// toNodeKey converts a compact.NodeID to the corresponding stable node
 // identifier used by this tool.
-func toNodeKey(n storage.NodeID) string {
-	d := int64(maxLen - n.PrefixLenBits)
-	i := n.BigInt().Int64() >> uint(d)
-	return nodeKey(d, i)
+func toNodeKey(n compact.NodeID) string {
```
Martin2112: Can it not just use the `NodeID` directly?

merkle/merkle_path.go
```diff
@@ -32,13 +32,13 @@ const vvLevel = 4
 // NodeFetch bundles a nodeID with additional information on how to use the node to construct the
 // correct proof.
 type NodeFetch struct {
-	NodeID storage.NodeID
+	ID     compact.NodeID
```
Martin2112: We're losing some error checking here because the storage code has checks on creating IDs e.g that they're inside the tree depth whereas the compact code accepts any values. I suppose almost the exact same checks are done when the fetches are converted so perhaps this isn't important.

docs/merkletree/treetex/main.go
pavelkalinnikov: Great idea. Done.

pavelkalinnikov:
@AlCutter Please review the `treetex` file.

server/proof_fetcher.go
```diff
@@ -24,6 +24,9 @@ import (
 	"github.com/google/trillian/storage"
 )
 
+// proofMaxBitLen is the max depth of a tree. Used for storage.NodeID creation.
+const proofMaxBitLen = 64
```
AlCutter: Worth being explicit in the const name and comment that this is for logs only?

AlCutter: Might be worth renaming these files to `log_proof_...` etc. and maybe prefixing the methods with `log` too?

Fine for a follow-up if you agree in principle.

pavelkalinnikov: Yep, let's do in in another PR.

AlCutter:
LGTM for treetex bits

merkle/merkle_path.go
Martin2112: Well yes but you can create compact node IDs with level > 64. It's often better to prevent the creation of invalid objects rather than detect them every time they're used. This is easier in languages that handle errors better than Go does.

The second part is what makes this probably OK as I don't think it creates any new opportunities for things to go wrong.

pavelkalinnikov:
@Martin2112 Any more concerns?

Martin2112:
No, was waiting for Al to review his bit.

PR #1813, opened 2019-09-09
Be less permissive of multiple trees in the scope of a context.

mhutchinson:
No operations should span multiple trees so we shouldn't be tolerant of this case. This change makes it clearer what one can expect of the tree being set in the context.

### DIFF
--- a/trees/trees.go
+++ b/trees/trees.go
@@ -159,12 +159,16 @@ func GetTree(ctx context.Context, s storage.AdminStorage, treeID int64, opts Get
 	ctx, spanEnd := spanFor(ctx, "GetTree")
 	defer spanEnd()
 	tree, ok := FromContext(ctx)
-	if !ok || tree.TreeId != treeID {
+	if !ok {
 		var err error
 		tree, err = storage.GetTree(ctx, s, treeID)
 		if err != nil {
 			return nil, err
 		}
+	} else if tree.TreeId != treeID {
+		// No operations should span multiple trees. If a tree is already in the context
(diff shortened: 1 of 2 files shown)


trees/trees.go
```diff
@@ -159,12 +159,16 @@ func GetTree(ctx context.Context, s storage.AdminStorage, treeID int64, opts Get
...
-	if !ok || tree.TreeId != treeID {
+	if !ok {
 		var err error
 		tree, err = storage.GetTree(ctx, s, treeID)
 		if err != nil {
 			return nil, err
 		}
+	} else if tree.TreeId != treeID {
+		// No operations should span multiple trees. If a tree is already in the context
+		// it better had be the one that we want.
```
AlCutter: "had better be"?

trees/trees.go
```diff
@@ -159,12 +159,16 @@ func GetTree(ctx context.Context, s storage.AdminStorage, treeID int64, opts Get
...
 	defer spanEnd()
 	tree, ok := FromContext(ctx)
-	if !ok || tree.TreeId != treeID {
+	if !ok {
 		var err error
 		tree, err = storage.GetTree(ctx, s, treeID)
 		if err != nil {
 			return nil, err
 		}
+	} else if tree.TreeId != treeID {
```
AlCutter: Is it worth removing the `else` and checking the ID regardless of whether it came from storage or the context?

mhutchinson: Done. Amusingly this meant another test case needed to be updated.

PR #1815, opened 2019-09-10
Updated golangci-lint to latest version, which supports Go v.13

mhutchinson:
When the same fix is applied to certificate-transparency-go this may allow us to remove the require statements at the bottom of the go.mod file.


gdbelvin:
Would you mind including the commands that you ran to produce these changes?

go.mod
```diff
@@ -94,20 +88,4 @@ require (
...
-
-replace github.com/golangci/gosec v0.0.0-20180901114220-66fb7fc33547 => github.com/golangci/gosec v0.0.0-20190211064107-66fb7fc33547
-
-replace github.com/golangci/ineffassign v0.0.0-20180808204949-42439a7714cc => github.com/golangci/ineffassign v0.0.0-20190609212857-42439a7714cc
-
-replace github.com/golangci/lint-1 v0.0.0-20180610141402-ee948d087217 => github.com/golangci/lint-1 v0.0.0-20190420132249-ee948d087217
-
-replace mvdan.cc/unparam v0.0.0-20190124213536-fbb59629db34 => mvdan.cc/unparam v0.0.0-20190209190245-fbb59629db34
-
 replace google.golang.org/genproto v0.0.0-20170818100345-ee236bd376b0 => google.golang.org/genproto v0.0.0-20170818010345-ee236bd376b0
```
RJPercival: Do you know which dependency has this invalid pseudo-version in their go.mod file?

mhutchinson: github.com/fullstorydev/grpcurl@v1.3.1 requires
        github.com/jhump/protoreflect@v1.4.4 requires
        google.golang.org/genproto@v0.0.0-20170818100345-ee236bd376b0: invalid pseudo-version: does not match version-control timestamp (2017-08-18T01:03:45Z)

mhutchinson:
> Would you mind including the commands that you ran to produce these changes?

It was non-trivial - I had to edit the file. I can try to reproduce it from a fresh client with pure commands if that exercise has value. I was getting surprising errors though; I think go is at least mildly upset we have a dependency cycle between trillian and certificate-transparency-go.

go.mod
RJPercival: It looks like that was fixed about a month ago: https://github.com/jhump/protoreflect/pull/258

mhutchinson: I'll upgrade the uber library which has a stale dep. That seems to fix it.

PR #1817, opened 2019-09-10
subtree_cache: Replace map+sync.RWMutex with sync.Map

pavelkalinnikov:
This change gets rid of `map+sync.RWMutex` combination in favor of `sync.Map` which suits our use of `SubtreeCache`: either multiple parallel workers reading/writing to non-intersecting subsets, or many reads and a few writes.


storage/cache/subtree_cache.go
```diff
@@ -211,28 +211,20 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
...
 	for _, id := range want {
 		prefixLen := id.PrefixLenBits / depthQuantum
 		px := id.Path[:prefixLen]
 		pxKey := string(px)
-		_, exists := s.subtrees[pxKey]
-		if exists {
-			return fmt.Errorf("preload tried to clobber existing subtree for: %v", id)
-		}
-		s.subtrees[pxKey] = s.newEmptySubtree(id, px)
+		s.subtrees.Store(pxKey, s.newEmptySubtree(id, px))
```
AlCutter: Should/could this use `LoadOrStore()` to keep the error-checking behaviour looking for overwrites (if it's now permissible for overwrites, we could check that the stored subtree is identical to the one we were about to store)?

storage/cache/subtree_cache.go
```diff
@@ -288,25 +280,32 @@ func (s *SubtreeCache) GetNodes(ids []storage.NodeID, getSubtrees GetSubtreesFun
 	return ret, nil
 }
 
+func (s *SubtreeCache) unmapSubtree(prefixKey string) *storagepb.SubtreeProto {
```
AlCutter: `getCachedSubtree`?

storage/cache/subtree_cache.go
```diff
@@ -288,25 +280,32 @@ func (s *SubtreeCache) GetNodes(ids []storage.NodeID, getSubtrees GetSubtreesFun
...
+		return nil
+	}
+	res, ok := raw.(*storagepb.SubtreeProto)
+	if !ok {
+		return nil
+	}
+	return res
+}
+
+func (s *SubtreeCache) unmapDirty(prefixKey string) bool {
```
AlCutter: `prefixIsDirty`?

storage/cache/subtree_cache.go
```diff
@@ -151,23 +152,18 @@ func (s *SubtreeCache) stratumInfoForPrefixLength(l int) stratumInfo {
...
+		if _, ok := s.subtrees.Load(pxKey); !ok {
+			want[pxKey] = id
 		}
-	}()
-	// Note: At this point the lock is released, so multiple parallel preload
-	// invocations can happen to getSubtrees with the same node IDs. It's okay
-	// because we collapse results further below.
+	}
+	// Note: At this point multiple parallel preload invocations can happen to
+	// getSubtrees with overlapping sets of IDs. It's okay because we collapse
```
gdbelvin: It looks like multiple overlapping IDs will result in the same tree being fetched more than once?
It also looks like multiple overlapping IDs might invalidate the sync.map assumption that a particular key gets set once and then read many times.

pavelkalinnikov: > It looks like multiple overlapping IDs will result in the same tree being fetched more than once?

Yes. This was also true before this PR. To completely avoid double-read of an ID we'd have to "lock" it while the database read is ongoing, which is quite a complex thing to implement without sacrificing performance (requires some finer grained locking than a "global" cache lock that we had previously). Instead, we can assume that this situation is rare. In fact, it doesn't seem to happen at all, if you consider how we use the cache: it's either read/written once (e.g. on different kinds of proof requests), or goroutines that write/read to/from it do not overlap w.r.t. subtrees (e.g. see the sparse Merkle tree building code).

> It also looks like multiple overlapping IDs might invalidate the sync.map assumption that a particular key gets set once and then read many times.

True. But I assume that this situation is rare or doesn't happen at all (see above). Obv., even if it does, `sync.Map` still works.

storage/cache/subtree_cache.go

storage/cache/subtree_cache.go

storage/cache/subtree_cache.go
pavelkalinnikov: Yup, good idea. Done.
We should probably do the same for non-empty subtrees `Store` call above, but in a separate PR.

pavelkalinnikov: Actually, done it all in this PR.

PR #1818, opened 2019-09-11
Simplify Travis setup with jobs

gdbelvin:
The Travis Jobs API allows us to supply different setup stages for each job.
This should 
- Reduce the massive conflagration of flags that we use.
- Make errors easier to diagnose 
- Make travis faster since each job will only need to install the dependencies that it needs

New Travis wall clock time: `Ran for 11 min 54 sec`, down from 13 min previousy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ jobs:
         # TODO(RJPercival): Make docker-compose integration test work when PKCS#11
         # support is enabled. This requires running softhsm in a Docker container.
         # See https://github.com/rolandshoemaker/docker-hsm for an example.
-        ./integration/docker_compose_integration_test.sh
+      - ./integration/docker_compose_integration_test.sh
     - name: "CT Integration"
       script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,24 +22,93 @@ env:
...
 matrix:
   fast_finish: true
 
+jobs:
+  include:
+    - name: "build"
+      before_install: true
+      install: true
+      before_script: true
+      script: go build ./...
```
mhutchinson: after_success should only be run on the coverage jobs.

.travis.yml
```diff
@@ -22,24 +22,93 @@ env:
   - GO111MODULE=on
   - GOPROXY=https://proxy.golang.org
   matrix:
```
mhutchinson: Add TODO that all of this matrix should be pushed into jobs.

.travis.yml
gdbelvin: I'll move coverage to it's own job in a separate PR

PR #1824, opened 2019-09-12
subtree_cache: Make GetNodes read sync.Map only once

pavelkalinnikov:
Previously: `GetNodes` fills up the cache map with all the missing subtrees; then it tries to read from this map as if it didn't know what it had just written to it.

Now: `GetNodes` returns the results using all the available information about the subtrees that previously the `preload` function was hiding. As a result, it reads from the cached map only once.


gdbelvin:
Does this still need review?

pavelkalinnikov:
Closing this, as the maps no longer use `SubtreeCache`, and I don't want to risk breaking logs. The longer term plan is to remove this code from logs as well.

PR #1825, opened 2019-09-13
NodeID: Remove unnecessary copying

pavelkalinnikov:


storage/node.go
```diff
@@ -460,9 +460,6 @@ func (n *NodeID) Split(prefixBytes, suffixBits int) ([]byte, *Suffix) {
 	if n.PrefixLenBits == 0 {
 		return []byte{}, EmptySuffix
 	}
-	a := make([]byte, prefixBytes)
```
Martin2112: I think this was missed in previous refactorings. Previously not all the methods returned copies so prefix was copied here.

pavelkalinnikov:
I'll be happy to remove the only mutator :)

And then it will mean that we need to copy bytes in `NodeID` calls. We can say that `NodeID` is immutable, and then all prefix-returning methods can just reference the underlying bytes slice. If any use-case needs a copy, we can provide `NodeID.Copy` method.

Martin2112:
Sounds good. `Copy()` already exists. Outside of tests I only see 8 write references to `Path`. Apart from `SetLowerBits` they appear to be `NewXXX()` or modifying copies. So I think it is the last obstacle.

PR #1827, opened 2019-09-13
NodeID: Remove the empty ID constructor

pavelkalinnikov:
Since `NodeID` is now immutable, there is no point to have a constructor which pre-allocates a zeroed byte slice (presumably the idea was that it can later be filled in).

### DIFF
--- a/storage/node.go
+++ b/storage/node.go
@@ -37,7 +37,9 @@ type Node struct {
 	NodeRevision int64
 }
 
-// NodeID uniquely identifies a Node within a MerkleTree.
+// NodeID is an immutable identifier of a Merkle tree Node. A default
+// constructed NodeID is an empty ID representing the root of a (sub-)tree.
+//
 // Reading paths right to left is the natural order when traversing from
 // leaves towards the root. However, for internal nodes the rightmost bits
 // of the IDs are not aligned on a byte boundary so care must be taken.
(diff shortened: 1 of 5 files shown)


storage/node.go
```diff
@@ -37,7 +37,9 @@ type Node struct {
 	NodeRevision int64
 }
 
-// NodeID uniquely identifies a Node within a MerkleTree.
+// NodeID is an immutable identifier of a Merkle tree Node. A default
```
Martin2112: Technically, it's not immutable (yet) as the struct fields are exported and can be modified.

pavelkalinnikov: Made this statement a bit weaker.

storage/node.go
```diff
@@ -233,15 +224,15 @@ func NewNodeIDForTreeCoords(depth int64, index int64, maxPathBits int) (NodeID,
...
+	r := NodeID{Path: make([]byte, maxPathBits/8)} // Note: maxPathBits % 8 == 0.
 	for i := len(r.Path) - 1; uidx > 0 && i >= 0; i-- {
 		r.Path[i] = byte(uidx & 0xff)
 		uidx >>= 8
 	}
 	// In the storage model nodes closer to the leaves have longer nodeIDs, so
 	// we "reverse" depth here:
 	r.PrefixLenBits = int(maxPathBits - int(depth))
-	return *r, nil
+	return r, nil
```
Martin2112: For consistency all the New stuff should return pointers. Can be done later.

pavelkalinnikov: Will do separately. Should they all return `NodeID` by value though? It's a lightweight type, and it's also meant to be immutable.

Martin2112: Not sure how strong the Go convention is. The Effective Go examples return pointers.

PR #1828, opened 2019-09-13
Pass NodeID by value everywhere

pavelkalinnikov:
This change makes `NodeID` always passed by value. It seems that all call sites used it by value anyway. This change also encourages Go compiler to not allocate `NodeID`s in heap.

### DIFF
--- a/merkle/sparse_merkle_tree_test.go
+++ b/merkle/sparse_merkle_tree_test.go
@@ -123,7 +123,7 @@ func getRandomNonRootNode(t *testing.T, rev int64) storage.Node {
 	// Make sure it's not a root node.
 	nodeID.PrefixLenBits = int(1 + randomBytes(t, 1)[0]%254)
 	return storage.Node{
-		NodeID:       *nodeID,
+		NodeID:       nodeID,
 		Hash:         randomBytes(t, 32),
 		NodeRevision: rev,
 	}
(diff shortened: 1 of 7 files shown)


Martin2112:
Can you do a quick benchmark of one of the usages? Pointers seemed to be faster, presumably because of not needing to copy the path bytes.

pavelkalinnikov:
Path bytes are not copied, because slices in Go just reference bytes stored elsewhere.

pavelkalinnikov:
```
$ go test -bench BenchmarkSplit
Before: BenchmarkSplit-12    	 2643762	       457 ns/op
After: BenchmarkSplit-12    	 2649568	       438 ns/op
```

Martin2112:
That's odd because pointers definitely benchmarked faster before.

pavelkalinnikov:
Niceness: Caught one test that was supposed to rely on `NodeID.String` format, but since the argument didn't match the receiver type it just used the default Golang stringer.

pavelkalinnikov:
Pointers being faster is counter-intuitive. Maybe it was in some other type's benchmarks, e.g. `Suffix`? Consider that literally all call sites of `NewNodeID*` constructors converted the returned value back to non-pointer type. Doing it without the intermediate pointer step should definitely be at least as fast.

Martin2112:
Seem to remember it being in rehashing but there were a lot of changes made by me and Al around that time so it's hard to be certain.

PR #1830, opened 2019-09-13
NodeID: Do not copy bytes in Prefix method

pavelkalinnikov:
This change makes `NodeID.Prefix` method return a new `NodeID` which represents a prefix of the passed-in ID. This allows both using the returned `Path` slice as an in-memory map key, and reusing it for querying subtrees. As a result, the `PrefixAsKey` method is redundant, and is being removed.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -155,15 +155,14 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
 	want := make(map[string]storage.NodeID)
 	for _, id := range ids {
 		sInfo := s.stratumInfoForNodeID(id)
-		pxKey := id.PrefixAsKey(sInfo.prefixBytes)
+		pxID := id.Prefix(sInfo.prefixBytes)
+		pxKey := string(pxID.Path)
 		if _, ok := want[pxKey]; ok {
 			// No need to check s.subtrees map twice.
 			continue
 		}
 		if _, ok := s.subtrees.Load(pxKey); !ok {
-			// TODO(al): Fix for non-uniform strata.
-			id.PrefixLenBits = sInfo.prefixBytes * depthQuantum
-			want[pxKey] = id
+			want[pxKey] = pxID
 		}
 	}
 	// Note: At this point multiple parallel preload invocations can happen to
@@ -224,9 +223,7 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
 	// may be be running this code and touch the same keys, although this doesn't happen
 	// normally.
 	for _, id := range want {
-		prefixLen := id.PrefixLenBits / depthQuantum
-		px := id.Path[:prefixLen]
-		if err := s.cacheSubtree(s.newEmptySubtree(id, px)); err != nil {
+		if err := s.cacheSubtree(s.newEmptySubtree(id)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


pavelkalinnikov:
Might be easier to review commit by commit.

storage/cache/subtree_cache.go
```diff
@@ -155,15 +155,14 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
...
 		sInfo := s.stratumInfoForNodeID(id)
-		pxKey := id.PrefixAsKey(sInfo.prefixBytes)
+		pxID := id.Prefix(sInfo.prefixBytes)
+		pxKey := string(pxID.Path)
 		if _, ok := want[pxKey]; ok {
 			// No need to check s.subtrees map twice.
 			continue
 		}
 		if _, ok := s.subtrees.Load(pxKey); !ok {
-			// TODO(al): Fix for non-uniform strata.
```
Martin2112: Before I review this: If "non-uniform" here means "not exactly 8 bits" then this could possibly break the map. If it means "not a multiple of 8 bits" then that has never worked and this is ok.

storage/node.go
```diff
@@ -398,34 +398,22 @@ func (n NodeID) Suffix(prefixBytes, suffixBits int) *Suffix {
...
-// overhead of calculating the suffix too.
-func (n NodeID) Prefix(prefixBytes int) []byte {
+// Prefix returns a NodeID's prefix of the specified byte length, as another
+// NodeID. Panics if the byte length is out of bound for this ID.
+func (n NodeID) Prefix(bytes int) NodeID {
+	// TODO(pavelkalinnikov): This check doesn't eliminate panics, so we should
+	// either delete it, or cover all other panics.
 	if n.PrefixLenBits == 0 {
-		return []byte{}
+		return NodeID{}
```
pavelkalinnikov: Map tests are broken because of this line. If I change this to `NodeID{Path: []byte{}}` so that it corresponds to the previous code, tests pass. It's a puzzling puzzle why.

storage/cache/subtree_cache.go
pavelkalinnikov: I think it meant non-multiples-of-8.

Martin2112: I thought we might have lost the quantum bit but it's reflected in the `sInfo` prefix.

I don't think "non uniform" was the right term in that comment but as it's gone it's no longer an issue.

storage/node.go
Martin2112: OK I'm probably not going to review this fully today so no rush. I've been starting at Merkle trees all week and they're all starting to look the same.

pavelkalinnikov: After a couple of hours debugging I concluded that it's a bug in MySQL tree storage implementation [here](https://github.com/google/trillian/blob/17aea688cfaaf8d6d64f69982a091903569fdcc8/storage/mysql/tree_storage.go#L212). To read a subtree, it converts `NodeID.Path` directly to the `SubtreeId` field. So, for an empty slice it can either put `nil` or `[]byte{}` to the request. MySQL probably treats `nil` as a SQL `NULL`, while an empty slice corresponds to a blob of length 0.

pavelkalinnikov: Good news: That bug is probably what caused the existence of `if n.PrefixLenBits == 0` clause here and in a couple of other places in `NodeID`. Now we understand that it should be removed :)

Martin2112: Good find. We might have to leave that there unless we know that other storage implementations won't have the same issue. If things would break in an obvious way if they did then maybe we can remove it?

pavelkalinnikov:
Obsoleting this because `NodeID` will be replaced by `NodeID2`.

PR #1831, opened 2019-09-16
storage/mysql: Add tree storage tests

pavelkalinnikov:
The new tests exercise scenarios of reading nodes missing from storage.

### DIFF
--- a/storage/mysql/storage_test.go
+++ b/storage/mysql/storage_test.go
@@ -39,44 +39,60 @@ import (
 )
 
 func TestNodeRoundTrip(t *testing.T) {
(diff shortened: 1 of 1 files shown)


Martin2112:
The codecov report implies no change of coverage (if it's correct, which isn't necessarily the case). Was that expected?

pavelkalinnikov:
It's probably expected. MySQL tree storage passes all reads/writes through the cache, where all the logic of handling missing subtrees/nodes resides. This PR probably doesn't change an already good coverage of `SubtreeCache`. However, the new tests examine the higher level contract, so I think it's good to have them.

PR #1832, opened 2019-09-16
storage/mysql: Ensure node prefix is never nil

pavelkalinnikov:
This change makes MySQL tree storage more robust by strongly ensuring that node IDs are converted to non-nil byte slices before converting to a SQL query. Previously it depended on `NodeID` implementation to always keep a non-nil slice, which is not documented anywhere.

### DIFF
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -437,7 +437,7 @@ func (t *treeTX) IsOpen() bool {
 func subtreeKey(id storage.NodeID) ([]byte, error) {
 	// TODO(pavelkalinnikov): Extend this check to verify strata boundaries.
 	if id.PrefixLenBits%8 != 0 {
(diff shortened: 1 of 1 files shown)


Martin2112:
It looks like the original only returned `nil` if the input node id was `nil`. Is this fixing a symptom or a cause?

pavelkalinnikov:
See https://github.com/google/trillian/pull/1830#discussion_r324230582. If we remove the seemingly redundant `if n.PrefixLenBits == 0` clauses throughout the `NodeID`, we'll see map tests failing (because the tree storage tries to load the root subtree from `NULL` rather than an empty prefix). This PR fixes the problem (or rather strengthens the insurance against it) at the place where it makes more sense - the storage implementation. I don't think that ensuring `NodeID.Path != nil` should be part of `NodeID` contract.

PR #1837, opened 2019-09-16
Factor out treeLayout type from SubtreeCache

pavelkalinnikov:
This change moves the subtree strata calculus into a separate `treeLayout` type. This is the first step towards having cleaner abstraction levels around `SubtreeCache`. As an extra small optimization, this type can have 2 global instances (for log and for map), and be reused everywhere instead of being instantiated whenever a new transaction is started.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -54,7 +54,7 @@ const maxSupportedTreeDepth = 256
 //  1. Parallel readers/writers working on non-intersecting subsets of subtrees/nodes.
 //  2. Subtrees/nodes are rarely written, and mostly read.
 type SubtreeCache struct {
-	layout *TreeLayout
+	layout *treeLayout
 
 	// subtrees contains the Subtree data read from storage, and is updated by
 	// calls to SetNodeHash.
@@ -79,7 +79,7 @@ func NewSubtreeCache(strataDepths []int, populateSubtree storage.PopulateSubtree
 	// TODO(al): pass this in
 	maxTreeDepth := maxSupportedTreeDepth
 	glog.V(1).Infof("Creating new subtree cache maxDepth=%d strataDepths=%v", maxTreeDepth, strataDepths)
-	layout := NewTreeLayout(strataDepths)
+	layout := newTreeLayout(strataDepths)
 
 	// TODO(al): This needs to be passed in, particularly for Map use cases where
 	// we need to know it matches the number of bits in the chosen hash function.
@@ -106,7 +106,7 @@ func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc
 	// Figure out the set of subtrees we need.
 	want := make(map[string]storage.NodeID)
 	for _, id := range ids {
-		pxID := s.layout.GetSubtreeRoot(id)
+		pxID := s.layout.getSubtreeRoot(id)
 		pxKey := string(pxID.Path)
 		if _, ok := want[pxKey]; ok {
 			// No need to check s.subtrees map twice.
@@ -265,7 +265,7 @@ func (s *SubtreeCache) getNodeHash(id storage.NodeID, getSubtree GetSubtreeFunc)
 		glog.Infof("cache: getNodeHash(path=%x, prefixLen=%d) {", id.Path, id.PrefixLenBits)
 	}
 
-	prefixID, sx := s.layout.Split(id)
+	prefixID, sx := s.layout.split(id)
 	prefixKey := string(prefixID.Path)
 	c := s.getCachedSubtree(prefixKey)
 	if c == nil {
@@ -326,7 +326,7 @@ func (s *SubtreeCache) SetNodeHash(id storage.NodeID, h []byte, getSubtree GetSu
 		glog.Infof("cache: SetNodeHash(%x, %d)=%x", id.Path, id.PrefixLenBits, h)
 	}
 
-	prefixID, sx := s.layout.Split(id)
+	prefixID, sx := s.layout.split(id)
 	prefixKey := string(prefixID.Path)
 	c := s.getCachedSubtree(prefixKey)
 	if c == nil {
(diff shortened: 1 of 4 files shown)


storage/cache/subtree_cache.go
```diff
@@ -479,17 +426,17 @@ func (s *SubtreeCache) Flush(ctx context.Context, setSubtrees SetSubtreesFunc) e
 	return err
 }
 
-func (s *SubtreeCache) newEmptySubtree(id storage.NodeID, px []byte) *storagepb.SubtreeProto {
-	sInfo := s.stratumInfoForPrefixLength(id.PrefixLenBits)
+func (s *SubtreeCache) newEmptySubtree(id storage.NodeID) *storagepb.SubtreeProto {
```
mhutchinson: Some kind of assertion that the node id passed in is the root of a subtree would be worthwhile, if cheap. If this has an unbearable cost, then at least document that this is expected to be a subtree root.

storage/cache/tree_layout.go
```diff
@@ -0,0 +1,115 @@
...
+	return &TreeLayout{sIndex: sIndex, height: height}
+}
+
+// GetSubtreeRoot returns the root node ID for the stratum that the passed-in
+// node belongs to.
+//
+// Note that nodes located at multiples of depthQuantum normally belong to
+// subtrees rooted above them. However, the topmost node (with an empty NodeID)
+// is the root for its own subtree since there is nothing above it.
+func (t *TreeLayout) GetSubtreeRoot(id storage.NodeID) storage.NodeID {
```
mhutchinson: We discussed in person, but a reminder (for this, or a future CL - your call) that returning a different type here which allows for type-safe enforcement of the specialness of SubtreeID would be tasty.

You could make GetSubtreeHeight below only accept such types too.

pavelkalinnikov: Great idea. Added a TODO, and will create a follow-up PR.

storage/cache/subtree_cache.go

pavelkalinnikov:
@Martin2112 A second pair of eyes on this would be great.

storage/cache/subtree_cache.go
```diff
@@ -128,38 +92,28 @@ func NewSubtreeCache(strataDepths []int, populateSubtree storage.PopulateSubtree
...
 // preload calculates the set of subtrees required to know the hashes of the
 // passed in node IDs, uses getSubtrees to retrieve them, and finally populates
 // the cache structures with the data.
 func (s *SubtreeCache) preload(ids []storage.NodeID, getSubtrees GetSubtreesFunc) error {
 	// Figure out the set of subtrees we need.
 	want := make(map[string]storage.NodeID)
 	for _, id := range ids {
-		sInfo := s.stratumInfoForNodeID(id)
-		pxKey := id.PrefixAsKey(sInfo.prefixBytes)
+		pxID := s.layout.GetSubtreeRoot(id)
```
Martin2112: If we got the key directly here then it would remove another access to the internals of a NodeID.

storage/cache/subtree_cache.go
```diff
@@ -313,29 +265,26 @@ func (s *SubtreeCache) getNodeHash(id storage.NodeID, getSubtree GetSubtreeFunc)
 		glog.Infof("cache: getNodeHash(path=%x, prefixLen=%d) {", id.Path, id.PrefixLenBits)
 	}
 
-	sInfo := s.stratumInfoForNodeID(id)
-	prefixKey := id.PrefixAsKey(sInfo.prefixBytes)
+	prefixID, sx := s.layout.Split(id)
+	prefixKey := string(prefixID.Path)
```
Martin2112: Again I think it would be an improvement for the layout to return the key + 1 instance below.

storage/cache/subtree_cache.go
Martin2112: Not sure that check is what was meant by @mhutchinson? At least in the map case it could possibly be stronger because of the non-uniform strata depths.  It may not be possible to apply it here though as it's not map specific code.

storage/cache/tree_layout.go
```diff
@@ -0,0 +1,117 @@
...
+	height int
+}
+
+// NewTreeLayout creates a tree layout based on the passed-in strata heights.
+func NewTreeLayout(heights []int) *TreeLayout {
+	// Compute the total tree height.
+	height := 0
+	for _, h := range heights {
+		height += h
+	}
```
Martin2112: Should maybe verify the total height is sensible. If it ends up negative for example it will be used to create a slice before the > 0 checks below are applied.

storage/cache/tree_layout.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/google/trillian/storage"
+)
+
+const (
+	// depthQuantum defines the smallest supported subtree height, which all
+	// subtree heights must also be a multiple of.
+	//
+	// WARNING: The code below breaks if this value is not a multiple of 8,
+	// because it uses storage.NodeID byte representation directly.
+	depthQuantum = 8
```
Martin2112: It's an old comment but I don't think changing that value is safe at all. It might work for maps but I suspect it will cause problems for logs if you set it to e.g 16.

storage/cache/tree_layout.go
Martin2112: This maybe covers the same ground as my previous comments. If we can avoid users of Layout having to look inside the path then that would be better I think.

storage/cache/subtree_cache.go
pavelkalinnikov: Yeah, this is the most I could do here. The TODO that I added below is actually how it should be done - `SubtreeID` will be compliant with strata sizes because it will be generated by `TreeLayout`.

storage/cache/subtree_cache.go
pavelkalinnikov: I started doing this and realised that there is still a couple of places below where we need `NodeID` (for example, the `getSubtrees` function signature). So if we return a prefix string here, we sill convert it back to `NodeID` assuming it's aligned to bytes, and hence there is still a dependency. I will look into breaking this dependency separately.

storage/cache/subtree_cache.go
pavelkalinnikov: See above.

storage/cache/tree_layout.go
pavelkalinnikov: Yeah, I don't think it's a good idea to ever change it. The comment only focuses on `TreeLayout` though, which in isolation doesn't break if this constant changes to another multiple of 8. I updated the comment.

pavelkalinnikov:
Addressed the comments. I will also unexport the new type for now, and export it if/when it's necessary.

pavelkalinnikov:
Ah, missed some comments, sorry. Taking a look.

storage/cache/tree_layout.go
pavelkalinnikov: I moved the checks up a bit to address this issue.

PR #1838, opened 2019-09-16
Refactor: extract validator for requested indices

mhutchinson:
This refactor makes code more readable, reusable, and tested.

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -160,18 +160,17 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req 
 	if req.Revision < 0 {
 		return nil, fmt.Errorf("map revision %d must be >= 0", req.Revision)
 	}
-	if err := hasDuplicates(req.Index); err != nil {
-		return nil, err
-	}
 	tree, hasher, err := t.getTreeAndHasher(ctx, req.MapId, optsMapRead)
 	if err != nil {
 		return nil, fmt.Errorf("could not get map %v: %v", req.MapId, err)
 	}
+	indexValidator := newRequestIndexValidator(hasher.Size())
 	for _, index := range req.Index {
-		if err := checkIndexSize(index, hasher); err != nil {
+		if err := indexValidator.validate(index); err != nil {
 			return nil, err
 		}
 	}
+
 	tx, err := t.snapshotForTree(ctx, tree, "GetLeavesByRevisionNoProof")
 	if err != nil {
 		return nil, fmt.Errorf("could not create database snapshot: %v", err)
@@ -192,15 +191,13 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req 
 }
 
 func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64, indices [][]byte, revision int64) (*trillian.GetMapLeavesResponse, error) {
-	if err := hasDuplicates(indices); err != nil {
-		return nil, err
-	}
 	tree, hasher, err := t.getTreeAndHasher(ctx, mapID, optsMapRead)
 	if err != nil {
 		return nil, fmt.Errorf("could not get map %v: %v", mapID, err)
 	}
+	indexValidator := newRequestIndexValidator(hasher.Size())
 	for _, index := range indices {
-		if err := checkIndexSize(index, hasher); err != nil {
+		if err := indexValidator.validate(index); err != nil {
 			return nil, err
 		}
 	}
@@ -315,31 +312,8 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 	}, nil
 }
 
-func checkIndexSize(index []byte, hasher hashers.MapHasher) error {
-	// The parameter is named 'index' (here and in the RPC API) because it's the ordinal number
-	// of the leaf, but that number is obtained by hashing the key value that corresponds to the
-	// leaf.  Leaf "indices" are therefore sparsely scattered in the range [0, 2^hashsize) and
-	// are represented as a []byte, and every leaf must have an index that is the same size.
-	//
-	// We currently police this by requiring that the hash size for the index space be the same
-	// as the hash size for the tree itself, although that's not strictly required (e.g. could
-	// have SHA-256 for generating leaf indices, but SHA-512 for building the root hash).
-	if len(index) != hasher.Size() {
-		return status.Errorf(codes.InvalidArgument, "index len(%x) is not %d", index, hasher.Size())
-	}
-	return nil
-}
-
 // SetLeaves implements the SetLeaves RPC method.
 func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapLeavesRequest) (*trillian.SetMapLeavesResponse, error) {
-	indexes := make([][]byte, 0, len(req.Leaves))
-	for _, l := range req.Leaves {
-		indexes = append(indexes, l.Index)
-	}
-	if err := hasDuplicates(indexes); err != nil {
-		return nil, err
-	}
-
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
 
@@ -349,6 +323,12 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	tree, hasher, err := t.getTreeAndHasher(ctx, mapID, optsMapWrite)
 	if err != nil {
 		return nil, err
+	}
+	indexValidator := newRequestIndexValidator(hasher.Size())
+	for _, l := range req.Leaves {
+		if err := indexValidator.validate(l.Index); err != nil {
+			return nil, err
+		}
 	}
 	ctx = trees.NewContext(ctx, tree)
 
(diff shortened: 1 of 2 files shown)


server/map_rpc_server.go
```diff
@@ -697,14 +674,36 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
+func newRequestIndexValidator(indexSize int) *requestIndexValidator {
+	return &requestIndexValidator{
+		indexSize:   indexSize,
+		seenIndices: make(map[string]bool),
+	}
+}
+
+func (c *requestIndexValidator) validate(index []byte) error {
+	// The parameter is named 'index' (here and in the RPC API) because it's the ordinal number
+	// of the leaf, but that number is obtained by hashing the key value that corresponds to the
```
pavelkalinnikov: Probably out of this PR's scope, but OTOH maybe a follow-up PR could start a little naming revolution:

The notion of "hashes of keys" feels confusing, and mixes abstractions a bit. We could just say that the map tree has a fixed-length keyspace, and then it's up to the client/personality to satisfy this limitation (e.g. by hashing the "real" keys, zero-padding them, or doing magic like 248bytes(hash)+8bytes(local_index)). The advantage: this allows throwing away "index" (well, maybe not if it's explicitly used as "leaf index") and "hash of a key" terms, in favor of just "key".

server/map_rpc_server.go
```diff
@@ -394,10 +374,7 @@ func (t *TrillianMapServer) getWriteRevision(ctx context.Context, tree *trillian
 func (t *TrillianMapServer) writeLeaves(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, leaves []*trillian.MapLeaf, rev int64) ([]merkle.HashKeyValue, error) {
 	hkv := make([]merkle.HashKeyValue, 0, len(leaves))
 	for _, l := range leaves {
-		// TODO(mhutchinson): make this check and perform hashing and hkv building outside of the transaction scope
-		if err := checkIndexSize(l.Index, hasher); err != nil {
-			return nil, err
-		}
+		// TODO(mhutchinson): perform hashing and hkv building outside of the transaction scope
```
pavelkalinnikov: supernit: **P**erform and period in the end.

server/map_rpc_server.go
```diff
@@ -350,6 +324,12 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	if err != nil {
 		return nil, err
 	}
+	indexValidator := newRequestIndexValidator(hasher.Size())
+	for _, l := range req.Leaves {
```
pavelkalinnikov: Consider adding this loop's index to error messages instead of the byte array: `for i, l := range ...` + `fmt.Errorf("key #%d: ...", i, ...)`
Advantage: If an RPC client sends a bonkersly long leaf index, we won't send it back to them.

server/map_rpc_server.go
```diff
@@ -697,14 +674,36 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
-// hasDuplicates returns an error if there are duplicates in indexes.
-func hasDuplicates(indexes [][]byte) error {
-	set := make(map[string]bool)
-	for _, i := range indexes {
-		if set[string(i)] {
-			return status.Errorf(codes.InvalidArgument, "index %x requested more than once", i)
-		}
-		set[string(i)] = true
+// requestIndexValidator validates the indices in a single map request.
+type requestIndexValidator struct {
```
pavelkalinnikov: Takes a while to guess how the nouns are related in this name. How about `leafIndexValidator`?

server/map_rpc_server.go
```diff
@@ -160,18 +160,17 @@ func (t *TrillianMapServer) GetLeavesByRevisionNoProof(ctx context.Context, req
...
 		return nil, fmt.Errorf("map revision %d must be >= 0", req.Revision)
 	}
-	if err := hasDuplicates(req.Index); err != nil {
-		return nil, err
-	}
 	tree, hasher, err := t.getTreeAndHasher(ctx, req.MapId, optsMapRead)
 	if err != nil {
 		return nil, fmt.Errorf("could not get map %v: %v", req.MapId, err)
 	}
+	indexValidator := newRequestIndexValidator(hasher.Size())
```
pavelkalinnikov: This object is stateful, and its entire state (lots of keys potentially) will stick around in memory while the transaction is ongoing. How about moving it to a function, for better GC-ability?
Also, when I see a function name like "validate", I think of something stateless.

Based on these 2 arguments, how about just having a simple `validateLeafIndices` function instead of a dedicated type?

pavelkalinnikov: Oh I see that the 2 things loop over different types of arrays. That's a bit sad then...

mhutchinson: Turned into a single method that sucks in the iteration and thus avoids being stateful.

server/map_rpc_server.go
mhutchinson: Done.

server/map_rpc_server.go
mhutchinson: No longer exists.

server/map_rpc_server.go
```diff
@@ -697,14 +674,36 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
-func hasDuplicates(indexes [][]byte) error {
-	set := make(map[string]bool)
-	for _, i := range indexes {
-		if set[string(i)] {
-			return status.Errorf(codes.InvalidArgument, "index %x requested more than once", i)
-		}
-		set[string(i)] = true
+// requestIndexValidator validates the indices in a single map request.
+type requestIndexValidator struct {
+	// The index size in bytes
```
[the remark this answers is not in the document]
mhutchinson: Gone.

server/map_rpc_server.go
mhutchinson: A future PR can introduce a consistent happy naming scheme here. It's beyond the scope of this small refactor. This comment has been moved down from above.

server/map_rpc_server.go
```diff
@@ -697,14 +674,36 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
+func (c *requestIndexValidator) validate(index []byte) error {
+	// The parameter is named 'index' (here and in the RPC API) because it's the ordinal number
+	// of the leaf, but that number is obtained by hashing the key value that corresponds to the
+	// leaf.  Leaf "indices" are therefore sparsely scattered in the range [0, 2^hashsize) and
+	// are represented as a []byte, and every leaf must have an index that is the same size.
+	//
+	// We currently police this by requiring that the hash size for the index space be the same
+	// as the hash size for the tree itself, although that's not strictly required (e.g. could
+	// have SHA-256 for generating leaf indices, but SHA-512 for building the root hash).
+	if len(index) != c.indexSize {
```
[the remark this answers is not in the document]
mhutchinson: I think you might actually want "want" to be in the if assignment here? But no worries, I've put both for absolute clarity.

server/map_rpc_server.go
```diff
@@ -697,14 +674,36 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
...
+		return status.Errorf(codes.InvalidArgument, "index len(%x) is not %d", index, c.indexSize)
+	}
+	s := string(index)
```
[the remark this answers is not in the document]
mhutchinson: That's really weird and I agree with rsc that it's weird to see extracting a common variable being slower than duplicating the function call. But whatever, it is what it is and I've inlined this []byte->string transform.

server/map_rpc_server_test.go
```diff
@@ -352,3 +351,46 @@ func fakeAdminStorageForMap(ctrl *gomock.Controller, times int, treeID int64) st
 
 	return adminStorage
 }
+
+func TestRequestIndexValidator(t *testing.T) {
+	tests := []struct {
+		desc    string
+		indices [][]byte
+		wantErr bool
+	}{
```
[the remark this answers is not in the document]
mhutchinson: Done.

server/map_rpc_server_test.go
```diff
@@ -352,3 +351,46 @@ func fakeAdminStorageForMap(ctrl *gomock.Controller, times int, treeID int64) st
...
+		{
+			desc:    "Multiple indices of correct length with duplicates",
+			indices: [][]byte{{'a'}, {'a'}},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			c := newRequestIndexValidator(1)
+			var gotErr bool
```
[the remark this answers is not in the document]
mhutchinson: Yup, done.

server/map_rpc_server_test.go
```diff
@@ -352,3 +351,55 @@ func fakeAdminStorageForMap(ctrl *gomock.Controller, times int, treeID int64) st
...
+			indices:   [][]byte{{'a'}, {'a'}},
+			wantErr:   true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			err := validateIndices(tt.indexSize, len(tt.indices), func(i int) []byte { return tt.indices[i] })
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("requestIndexValidator.validate() error = %v, wantErr %v", err, tt.wantErr)
```
pavelkalinnikov: validateIndices

server/map_rpc_server.go
```diff
@@ -697,14 +666,29 @@ func (t *TrillianMapServer) snapshotForTree(ctx context.Context, tree *trillian.
 	return tx, err
 }
 
-// hasDuplicates returns an error if there are duplicates in indexes.
-func hasDuplicates(indexes [][]byte) error {
-	set := make(map[string]bool)
-	for _, i := range indexes {
-		if set[string(i)] {
-			return status.Errorf(codes.InvalidArgument, "index %x requested more than once", i)
+// validateIndices confirms that all indexes have the given size and there are no duplicates.
```
pavelkalinnikov: all indices?

PR #1840, opened 2019-09-17
Introduce subtreeID

pavelkalinnikov:
The new `subtreeID` type is a "type-safe" alias to `NodeID` which is used for identifying tree strata.

### DIFF
--- a/storage/cache/tree_layout.go
+++ b/storage/cache/tree_layout.go
@@ -29,6 +29,27 @@ const (
 	depthQuantum = 8
 )
 
+// subtreeID holds an ID of a subtree, which is aligned with the tree layout.
+//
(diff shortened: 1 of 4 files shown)


storage/cache/tree_layout.go
```diff
@@ -29,6 +29,27 @@ const (
 	depthQuantum = 8
 )
 
+// subtreeID holds an ID of a subtree, which is aligned with the tree layout.
+//
+// It assumes that strata heights are multiples of 8, and so the byte
+// representation of the subtree ID matches storage.NodeID.
+type subtreeID struct {
```
mhutchinson: As discussed, consider renaming all of this "subtree" stuff to "tile" in a future PR. It makes it a lot easier to understand.

PR #1841, opened 2019-09-17
Perform the leaf hashing outside of the write transaction scope

mhutchinson:


server/map_rpc_server.go
```diff
@@ -363,20 +374,13 @@ func (t *TrillianMapServer) getWriteRevision(ctx context.Context, tree *trillian
 }
 
 // writeLeaves updates the leaf values, but does not calculate nor update the Merkle tree.
-func (t *TrillianMapServer) writeLeaves(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, leaves []*trillian.MapLeaf, rev int64) ([]merkle.HashKeyValue, error) {
-	hkv := make([]merkle.HashKeyValue, 0, len(leaves))
+func (t *TrillianMapServer) writeLeaves(ctx context.Context, tx storage.MapTreeTX, leaves []*trillian.MapLeaf) error {
```
pavelkalinnikov: Is this used more that once? Consider inlining it.

mhutchinson: I also considered this but am leaving it. One of the next steps will be changing WriteLeaves in map_write_rpc_server.go to also call this method instead of delegating via SetLeaves.

PR #1842, opened 2019-09-17
Deprecated methods that have better replacements on Map Read API

mhutchinson:
This is a fairly hard deprecation where we will migrate all known clients over and delete these methods. Having SetLeaves in its current form will block us from moving to async tree synthesis after WriteLeaves.


trillian_map_api.proto
```diff
@@ -193,20 +193,22 @@ service TrillianMap {
   rpc GetLeafByRevision(GetMapLeafByRevisionRequest) returns (GetMapLeafResponse) {}
   rpc GetLeaves(GetMapLeavesRequest) returns (GetMapLeavesResponse) {}
   rpc GetLeavesByRevision(GetMapLeavesByRevisionRequest) returns (GetMapLeavesResponse) {}
-  // GetLeavesByRevisionNoProof returns the requested map leaves without inclusion proofs.
-  // This API is designed for internal use where verification is not needed.
-  rpc GetLeavesByRevisionNoProof(GetMapLeavesByRevisionRequest) returns (MapLeaves) {}
+  // Deprecated: this should only be used be writers, which should migrate
```
pavelkalinnikov: typo: used by?

passim

pavelkalinnikov:
Tests are failing because the deprecated method is still used in hammer and some tests.

mhutchinson:
This is blocked on #1872 and 1875 to remove the usages of the deprecated methods. If we can't get those merged soon, then the other option is to disable the lint check for these usages.

mhutchinson:
I had to disable one of the lint checks for a single line - I don't want to change the API surface area here, so disabling this check is the only pragmatic way to do that while still maintaining functionality.

PR #1844, opened 2019-09-18
Better error message if inclusion proof fails

mhutchinson:
If this fails it's really useful to know if it was the first index of the first revision, or whether the hammer got quite far before failing.


mhutchinson:
This makes it clearer to see what is going on with #1845

Issue #1845, opened 2019-09-18
Trillian batch operations with > 1k entries fail on MariaDB >= 10.3


mhutchinson:
go test ./testonly/hammer/...
--- FAIL: TestInProcessMapHammer (9.77s)
    hammer_test.go:129: hammer failure: failed to verify inclusion proof for Index=922cab9b7de0da6c6b65792d3030303030303234000000000000000000000000: calculated root: 3af78a4fc25a48d822fd41eed4d56b3afff53cf6ac4f603427120b9127eece9c, want: 8192c68b3ec424a648ef33ae4d3b970b7ff870f12aa6fca03214652d25363e61
FAIL
FAIL    github.com/google/trillian/testonly/hammer      10.203s
?       github.com/google/trillian/testonly/hammer/maphammer    [no test files]
?       github.com/google/trillian/testonly/hammer/mapreplay    [no test files]
FAIL

Whatever else happens, we need to be explicit about which versions of mysql we support.

It's also worth seeing whether this version of MySQL can be supported without breaking backwards compatibility.

gdbelvin:
Occasionally passing for me 

```
go test ./testonly/hammer/
ok  	github.com/google/trillian/testonly/hammer	9.462s

mysql -V
mysql  Ver 8.0.17 for osx10.14 on x86_64 (Homebrew)
```

go env
```
 go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/gbelvin/Library/Caches/go-build"
GOENV="/Users/gbelvin/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/gbelvin/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.13/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.13/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/gbelvin/go/src/github.com/google/trillian/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
[...]

PR #1847, opened 2019-09-18
Refactor: use the verifying client in the hammer

mhutchinson:
The hammer was duplicating logic in the verifying client. It makes sense to reuse this logic.

This addresses some of the issues brought up investigating #1845

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -19,13 +19,13 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/golang/protobuf/proto"
 	"github.com/google/trillian"
 	"github.com/google/trillian/client"
 	"github.com/google/trillian/monitoring"
@@ -257,8 +257,8 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 
 // hammerState tracks the operations that have been performed during a test run.
 type hammerState struct {
-	cfg      *MapConfig
-	verifier *client.MapVerifier
+	cfg *MapConfig
+	vc  *client.MapClient
 
 	start time.Time
 
@@ -273,7 +273,7 @@ type hammerState struct {
 
 	// SMRs are arranged from later to earlier (so [0] is the most recent), and the
 	// discovery of new SMRs will push older ones off the end.
-	smr [smrCount]*trillian.SignedMapRoot
+	smr [smrCount]*types.MapRootV1
 
 	// Counters for generating unique keys/values.
 	keyIdx   int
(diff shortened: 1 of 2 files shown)


testonly/hammer/hammer.go
```diff
@@ -568,62 +567,23 @@ func (s *hammerState) doGetLeaves(ctx context.Context, prng *rand.Rand, latest b
...
-		return fmt.Errorf("failed to verify root: %v", err)
-	}
-	for i, inc := range rsp.MapLeafInclusion {
-		if err := s.verifier.VerifyMapLeafInclusionHash(root.RootHash, inc); err != nil {
-			return fmt.Errorf("failed to verify inclusion proof for Index=%x (%d/%d) in revision %d of tree %d: %v", inc.Leaf.Index, i, len(rsp.MapLeafInclusion), root.Revision, s.cfg.MapID, err)
+			return fmt.Errorf("failed to GetAndVerifyMapLeavesByRevision: %v", err)
 		}
 	}
-
-	if err := contents.CheckContents(rsp.MapLeafInclusion, s.cfg.ExtraSize); err != nil {
```
gdbelvin: Ad a TODO to check whether the contents match the expected values?

testonly/hammer/hammer.go
```diff
@@ -780,50 +732,38 @@ func (s *hammerState) setLeavesInvalid(ctx context.Context, prng *rand.Rand) err
...
-	root, err := s.verifier.VerifySignedMapRoot(rsp.MapRoot)
+	root, err := s.vc.VerifySignedMapRoot(rsp.MapRoot)
 	if err != nil {
 		return err
 	}
 	glog.V(2).Infof("%d: got SMR(time=%q, rev=%d)", s.cfg.MapID, time.Unix(0, int64(root.TimestampNanos)), root.Revision)
 
-	if !proto.Equal(rsp.MapRoot, smr) {
-		return fmt.Errorf("get-smr-rev(@%d)=%+v, want %+v", rev, rsp.MapRoot, smr)
+	if !bytes.Equal(root.RootHash, smrRoot.RootHash) {
```
gdbelvin: If you'd like to check a struct that *contains* a proto, the right way to do this is to use the `cmp` library.

(From memory)
```
cmp.Equal(A,B, cmpopt.Equals(proto.Equal))
```

testonly/hammer/hammer.go
mhutchinson: Added this code back in. Got a bit aggressive with the pruning.

gdbelvin: Awaiting a github push...

testonly/hammer/hammer.go
mhutchinson: No protos involved here. I'm using DeepEqual now.

testonly/hammer/hammer.go
mhutchinson: Done. Normally I feel like github starts a review session for my comments, but this time it pushed directly without the batch. Anyway, race condition has resolved, and this is now addressed.

mhutchinson:
I also took the opportunity to rebase on master so I could verify locally that these changes were working, which require #1849 to work.

Issue #1797, opened 2019-09-04
SubtreeCache is slow


pavelkalinnikov:
[`SubtreeCache`](https://github.com/google/trillian/blob/5eb28bddb27d845397a309c0d1d9e088e447ea2f/storage/cache/subtree_cache.go) is currently at the core of tree transactions. All node reads and writes go through this cache. It was meant to be thread-safe, which is great, but internally it guards all accesses with a `sync.RWMutex`. This creates bottlenecks in various places like Map's `SetLeaves` call.

We need to think about properties that we want from `SubtreeCache`, and also measure how it's used now (hit rate, etc). Question: Do we feel okay if multiple reads of the same thing fall through the cash if made within a very short period of time? This is a weaker guarantee, but it allows more efficient cache implementations for typical-case scenarios.

Hypothesis: The hit rate is basically ~0. The purpose of the cache is to expand the storage layer's subtree strata.

pavelkalinnikov:
Cache slowness makes it necessary to do workarounds like #1796

pavelkalinnikov:
**Statement**: `SubtreeCache` does a lot more stuff than necessary, and kills performance of logs and maps. It should be simplified to be a "local" thread-unsafe cache.

Let's understand the cache usage patterns by looking at our `GetMerkleNodes` & `SetMerkleNodes` storage calls. `SetMerkleLeaves` is used only in 2 types of transactions:
1. `Sequencer.IntegrateBatch` in logs.
2. `subtreeWriter.buildSubtree` in maps (for each shard of the sparse Merkle tree).

[...]

PR #1848, opened 2019-09-18
Introduce the TileSet primitive

pavelkalinnikov:
The new `TileSet` primitive is a "local" cache that will allow the following use-cases:

- Figure out the "read set" of a tree storage transaction.
- Store the tiles read from the storage, and query them to perform by-`NodeID` reads/writes.

Addresses #1797.

### DIFF
--- a/storage/tree/tile_set.go
+++ b/storage/tree/tile_set.go
@@ -1,0 +1,62 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tree
+
+import (
+	"errors"
+
+	"github.com/google/trillian/storage/storagepb"
+)
+
+// TileSet contains an append-only set of tree tiles.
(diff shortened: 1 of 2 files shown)


storage/cache/tile_set_test.go
```diff
@@ -0,0 +1,148 @@
...
+		})
+	}
+}
+
+func TestTileSetAddErrors(t *testing.T) {
+	// Note: This returns new instances each time, for better tests isolation.
+	all := func() []*storagepb.SubtreeProto {
+		return []*storagepb.SubtreeProto{
+			nil,
+			&storagepb.SubtreeProto{Prefix: nil},
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

storage/cache/tile_set.go
```diff
@@ -0,0 +1,63 @@
...
+
+// newTileSet returns a new empty tileSet.
+func newTileSet() *tileSet {
+	return &tileSet{tiles: make(map[string]*storagepb.SubtreeProto)}
+}
+
+// Touch marks the passed-in tile ID as touched. Returns true iff it was not
+// touched before. Note that a tile can be touched but still not present, and
+// so the corresponding Get call can return nil.
+func (ts *tileSet) Touch(id subtreeID) bool {
```
mhutchinson: I like the other methods on this class, but I think the Touch method has some curious semantics that I think are brittle. Recommend moving the Touch method to something like a batchTileQueryBuilder (how Java does that sound?!), which would be something like:

batchTileQueryBuilder{
  getTile(subtreeID) { // add to set, no return type}
  fetchBatch(Transaction) tileset { ... }
}

pavelkalinnikov:
Closing this for reasons:

- `NodeID` is now deprecated.
- Using `storagepb.SubtreeProto` directly seems like a bad practice. Need some intermediate "tile" type.
- After prototyping map refactoring end-to-end the `TileSet` looks very different [#1932].

PR #1852, opened 2019-09-19
Move primitives from storage to tree package

pavelkalinnikov:
Move all primitives that can be used without pairing up with `storage` API, to a separate package. The nice benefit that we get is that some packages no longer depend on `storage`.

### DIFF
--- a/storage/aliases.go
+++ b/storage/aliases.go
@@ -14,10 +14,11 @@
 
 package storage
 
+import tree "github.com/google/trillian/storage/tree"
+
 // TODO(pavelkalinnikov): These aliases were created to not break the code that
 // depended on these type. We should delete this file eventually.
 
-/*
 type NodeID = tree.NodeID
 type Node = tree.Node
 type Suffix = tree.Suffix
@@ -34,4 +35,3 @@ var (
 	EmptySuffix = tree.EmptySuffix
 	ParseSuffix = tree.ParseSuffix
 )
-*/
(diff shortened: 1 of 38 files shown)


pavelkalinnikov:
I recommend reviewing this commit by commit.

storage/aliases.go
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import "github.com/google/trillian/storage/tree"
+
+// TODO(pavelkalinnikov): These aliases were created to not break the code that
+// depended on these type. We should delete this file eventually.
```
pavelkalinnikov: typo: types

pavelkalinnikov:
In the first commit (where I introduced the type and function aliases) I made sure that the whole repo build+tests are working, so any code that depends on it should work.

In the second commit I commented out the aliases, and fixed all references in the repo to point at the new package. Now we don't reference the aliases from anywhere in this repo.

In the third commit I uncommented the aliases again, so that any dependent code (including Google logs) doesn't break.

pavelkalinnikov:
> Only comment is whether these tree nodes should be in storage/tree - I'm not sure what the storage level really provides here. That said, I think separating these out is still a massive step in the right direction and rehoming that package will be a lot easier now.

I agree, it's slightly more generic than `storage` and maybe should be in `/tree`, or `/merkle/tree`. Not moving it too far for now until I add move stuff to this package and see where it logically belongs. I think that module-wise we are free to move it around until we release a new version.

PR #1854, opened 2019-09-20
Move treeLayout to storage/tree package

pavelkalinnikov:
This change moves `treeLayout` from `storage/cache` package to `storage/tree`, and renames it nicely as `tree.Layout`.


pavelkalinnikov:
Good idea, I'll do that at some point when it's more or less final.

The current diagram probably looks like this:
dependencies

Issue #1855, opened 2019-09-20
Proposed NodeID structure


pavelkalinnikov:
Proposal:

```golang
type NodeID struct {
	path string
	last byte
	bits byte
}
```

The whole node path consists of `path` bytes concatenated with 0 to 8 bits of the `last` byte depending on the `bits` field.

Advantages:
- This type can be used as a Go `map` key directly, with zero copy of byte slices (`string` is not copied). A canonical representation must be chosen carefully: e.g. `bits` should always be > 0, unless this is a "zero" ID; all "unused" bits of `last` should be zeroed.
- When used as a `map` key, it doesn't have to be byte-aligned - all non-multiple-of-8 lengths are supported automatically.
- Zero-copy "sibling" operation can be implemented efficiently, by changing the `last` byte.
- Arbitrary non-multipe-of-8 prefixes can be extracted with zero copy.
- It takes less space than the current `[]byte + int` implementation.
- All the fields are hidden, and there is no real way for the client to shoot themselves in the foot when using this type.

Disadvantage (not really):
- The `path` field can't be used directly, even if the bit-length is a multiple of 8, because of the canonicalization requirement that the `last` byte is always used. However, note that the storage layer does not request the node directly, it always needs a prefix of `path` to load a tile. A `path` prefix can be extracted with zero copy if the storage uses multiple-of-8 strata.

Potential extensions:
- Make the `last` field `uint64` (`bits` is now between 1 and 64), so that for log trees `path` will always be empty (hence no allocations at all).

pavelkalinnikov:
Quick test shows 3x speedup of the `Siblings` function:

```
BenchmarkNodeIDSiblings-12     	   51264	     23553 ns/op
BenchmarkNodeID2Siblings-12    	  154784	      7631 ns/op
```

PR #1856, opened 2019-09-23
Add NodeID2, the fast NodeID with zero allocations

pavelkalinnikov:
This change introduces a new experimental `NodeID2` type.

The new type has multiple advantages over the old `NodeID`:

- Immutability is enforced by design.
- It doesn't require multiple-of-8 prefixes.
- It can be used as a Go map key directly.
- The `Sibling` method is cheap.
- It takes less space than the current `[]byte + int` implementation.
- Client code never sees the internal structure of this type.

Benchmarks show a 3x speedup of the `Siblings` function (runs `Siblings` for 512 x 32byte IDs):

```
BenchmarkNodeIDSiblings-12     	     100	  11361849 ns/op
BenchmarkNodeID2Siblings-12    	     338	   3428912 ns/op
```

Addresses #1855

### DIFF
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -65,19 +65,6 @@ func (n NodeID2) Prefix(bits uint) NodeID2 {
 	return NodeID2{path: n.path[:bytes], last: last, bits: tail}
 }
 
-// Suffix returns the suffix of NodeID2 after the given number of bits.
-func (n NodeID2) Suffix(bits uint) NodeID2 {
-	if mx := n.BitLen(); bits > mx {
-		panic(fmt.Sprintf("Suffix: bits %d > %d", bits, mx))
-	} else if bits == mx {
-		return NodeID2{}
-	} else if bits%8 == 0 {
-		return NewNodeID2(n.path[bits/8:], mx-bits)
-	}
-	// TODO(pavelkalinnikov): Support arbitrary lengths.
-	panic("Suffix: only multiples of 8 are supported")
-}
-
 // Sibling returns the NodeID2 of the nodes's sibling in a binary tree. If the
 // node is the root then the returned ID is the same.
 func (n NodeID2) Sibling() NodeID2 {
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -35,7 +35,7 @@ func NewNodeID2(path string, bits int) NodeID2 {
 	} else if mx := len(path) * 8; bits > mx {
 		panic(fmt.Sprintf("NewNodeID2: bits %d > %d", bits, mx))
 	}
-	bytes, tail, mask := decompose(bits)
+	bytes, tail, mask := split(bits)
 	last := path[bytes] & mask
 	return NodeID2{path: path[:bytes], last: last, bits: byte(tail)}
 }
@@ -53,7 +53,7 @@ func (n NodeID2) Prefix(bits int) NodeID2 {
 		panic(fmt.Sprintf("Prefix: bits %d > %d", bits, mx))
 	}
 	last := n.last
-	bytes, tail, mask := decompose(bits)
+	bytes, tail, mask := split(bits)
 	if bytes != len(n.path) {
 		last = n.path[bytes]
 	}
@@ -72,12 +72,6 @@ func (n NodeID2) Suffix(bits int) NodeID2 {
 	}
 	// TODO(pavelkalinnikov): Support arbitrary lengths.
 	panic("Suffix: only multiples of 8 are supported")
-}
-
-// PrefixBytes returns a prefix of bytes*8 bits, as bytes. Must always be
-// called with a prefix shorter than BitLen().
-func (n NodeID2) PrefixBytes(bytes int) string {
-	return n.path[:bytes]
 }
 
 // Sibling returns the NodeID2 of the nodes's sibling in a binary tree. If the
@@ -100,13 +94,13 @@ func (n NodeID2) String() string {
 	return fmt.Sprintf("[%s%0*b]", path, n.bits, n.last>>(8-n.bits))
 }
 
-// decompose returns decomposition of a NodeID2 with the given number of bits.
+// split returns decomposition of a NodeID2 with the given number of bits.
 //
 // The first int of the returned triple is the number of full bytes stored in
 // the dynamically allocated part. The second one is the number of bits in the
 // tail byte (between 1 and 8). The third value is a mask with the
 // corresponding number of higher bits set.
-func decompose(bits int) (int, int, byte) {
+func split(bits int) (int, int, byte) {
 	bytes := (bits - 1) / 8
 	tailBits := 1 + (bits-1)%8
 	mask := ^byte(1<<uint(8-tailBits) - 1)
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -23,7 +23,7 @@ import "fmt"
 type NodeID2 struct {
 	path string
 	last byte
-	bits byte
+	bits uint8
 }
 
 // NewNodeID2 creates a NodeID2 from the given path bytes truncated to the
@@ -37,7 +37,7 @@ func NewNodeID2(path string, bits int) NodeID2 {
 	}
 	bytes, tail, mask := split(bits)
 	last := path[bytes] & mask
-	return NodeID2{path: path[:bytes], last: last, bits: byte(tail)}
+	return NodeID2{path: path[:bytes], last: last, bits: uint8(tail)}
 }
 
 // BitLen returns the length of the NodeID2 in bits.
@@ -58,7 +58,7 @@ func (n NodeID2) Prefix(bits int) NodeID2 {
 		last = n.path[bytes]
 	}
 	last &= mask
-	return NodeID2{path: n.path[:bytes], last: last, bits: byte(tail)}
+	return NodeID2{path: n.path[:bytes], last: last, bits: uint8(tail)}
 }
 
 // Suffix returns the suffix of NodeID2 after the given number of bits.
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -65,8 +65,9 @@ func (n NodeID2) Prefix(bits uint) NodeID2 {
 	return NodeID2{path: n.path[:bytes], last: last, bits: tail}
 }
 
-// Sibling returns the NodeID2 of the nodes's sibling in a binary tree. If the
-// node is the root then the returned ID is the same.
+// Sibling returns the NodeID2 of the nodes's sibling in a binary tree, i.e.
+// the ID of the parent node's other child. If the node is the root then the
+// returned ID is the same.
 func (n NodeID2) Sibling() NodeID2 {
 	last := n.last ^ byte(1<<(8-n.bits))
 	return NodeID2{path: n.path, last: last, bits: n.bits}
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -37,6 +37,7 @@ func NewNodeID2(path string, bits uint) NodeID2 {
 	}
 	bytes, tail, mask := split(bits)
 	last := path[bytes] & mask
+	// Note: Getting the substring is cheap because strings are immutable in Go.
 	return NodeID2{path: path[:bytes], last: last, bits: tail}
 }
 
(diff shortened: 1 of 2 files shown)


storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	bytes, tail, mask := decompose(bits)
+	if bytes != len(n.path) {
+		last = n.path[bytes]
+	}
+	last &= mask
+	return NodeID2{path: n.path[:bytes], last: last, bits: byte(tail)}
+}
+
+// Suffix returns the suffix of NodeID2 after the given number of bits.
+func (n NodeID2) Suffix(bits int) NodeID2 {
```
pavelkalinnikov: For `Suffix`, we probably can have a similar type which has a partial first byte&mask (potentially empty, if the suffix is aligned), plus the internal bytes string, plus the tail byte&mask.

Martin2112:
Have we got any benchmarks for this? Just to look at risk v. rewards.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+
+package tree
+
+import "fmt"
+
+// NodeID2 is a faster NodeID that does zero memory allocations.
+//
+// TODO(pavelkalinnikov): Rename to NodeID and document it properly when the
+// code has migrated.
+type NodeID2 struct {
```
Martin2112: If `NodeID` was an interface (which has been considered before) doing that change first might make this one easier. Also specializations could enforce appropriate constraints for logs v. maps if that makes sense. It's not obviously way better so hasn't been done before.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	path string
```
Martin2112: I really hate this about Go, storing binary data in strings is against nature and God. No good can come of it.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	last byte
+	bits byte
+}
+
+// NewNodeID2 creates a NodeID2 from the given path bytes truncated to the
+// specified number of bits if necessary. Panics if the number of bits is
+// negative, or is more than the byte string contains.
+func NewNodeID2(path string, bits int) NodeID2 {
```
Martin2112: Reducing the number of constructors is good but this might not cover all the uses.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+
+// BitLen returns the length of the NodeID2 in bits.
+func (n NodeID2) BitLen() int {
+	return len(n.path)*8 + int(n.bits)
+}
+
+// Prefix returns the prefix of NodeID2 with the given number of bits.
+func (n NodeID2) Prefix(bits int) NodeID2 {
+	if bits == 0 {
+		return NodeID2{}
```
Martin2112: They're immutable so no need to reallocate here.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	path := fmt.Sprintf("%08b", []byte(n.path))
+	path = path[1 : len(path)-1] // Trim the brackets.
+	if len(path) > 0 {
+		path += " "
+	}
+	return fmt.Sprintf("[%s%0*b]", path, n.bits, n.last>>(8-n.bits))
+}
+
+// decompose returns decomposition of a NodeID2 with the given number of bits.
```
Martin2112: Maybe rename this `split` or `tail`? Just to avoid any confusion with compact range code.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Prefix returns the prefix of NodeID2 with the given number of bits.
+func (n NodeID2) Prefix(bits int) NodeID2 {
+	if bits == 0 {
+		return NodeID2{}
+	} else if mx := n.BitLen(); bits > mx {
+		panic(fmt.Sprintf("Prefix: bits %d > %d", bits, mx))
+	}
+	last := n.last
```
Martin2112: If you `bits` is a multiple of 8 the call to `decompose` can be avoided. Possibly a premature optimization.

storage/tree/node_id2.go
Martin2112: Possibly. I did some work optimizing for the log case, which avoids allocating them. Might be more to do and not sure what makes sense for the map.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	last &= mask
+	return NodeID2{path: n.path[:bytes], last: last, bits: byte(tail)}
+}
+
+// Suffix returns the suffix of NodeID2 after the given number of bits.
+func (n NodeID2) Suffix(bits int) NodeID2 {
+	if mx := n.BitLen(); bits > mx {
+		panic(fmt.Sprintf("Suffix: bits %d > %d", bits, mx))
+	} else if bits == mx {
+		return NodeID2{}
```
Martin2112: Again, can use a shared empty node.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	} else if bits%8 == 0 {
+		return NewNodeID2(n.path[bits/8:], mx-bits)
+	}
+	// TODO(pavelkalinnikov): Support arbitrary lengths.
```
Martin2112: I'm not sure we need to worry about this. We don't have any use cases for it at the moment. I suppose we might in the map?

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+	panic("Suffix: only multiples of 8 are supported")
+}
+
+// PrefixBytes returns a prefix of bytes*8 bits, as bytes. Must always be
```
Martin2112: "as bytes" doesn't match with `string` type. Suggest something like "as a `string` containing the bytes"?

storage/tree/node_id2.go
```diff
@@ -0,0 +1,114 @@
...
+// called with a prefix shorter than BitLen().
+func (n NodeID2) PrefixBytes(bytes int) string {
+	return n.path[:bytes]
+}
+
+// Sibling returns the NodeID2 of the nodes's sibling in a binary tree. If the
+// node is the root then the returned ID is the same.
+func (n NodeID2) Sibling() NodeID2 {
```
Martin2112: Is this used anywhere? I see the map calls `Siblings`.

storage/tree/node_id2.go
Martin2112: To clarify - there's no point in making it an interface unless there's actual value in having multiple implementations. I'm not sure there is.

pavelkalinnikov:
> Have we got any benchmarks for this? Just to look at risk v. rewards.

Yes, see the `Benchmark` tests added in this PR. Added the results to the PR description.

storage/tree/node_id2.go
Martin2112: OK.

storage/tree/node_id2.go
Martin2112: Yeah an immutable byte string would be great.

storage/tree/node_id2.go
Martin2112: It must get allocated somewhere but I don't have time to look at what assembly code gets generated and it's probably not worth worrying about.

Martin2112:
Re: benchmarking. Can we do a bit of profiling to see how much time the server spends on node ID manipulation. I don't think I kept the results from my previous runs.

storage/tree/node_id2.go
Martin2112: That sounds OK but I can only review stuff based on what I can see in use now.

Martin2112:
This looks basically OK to me but I'll give other reviewers a chance to comment.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+// code has migrated.
+type NodeID2 struct {
+	path string
+	last byte
+	bits byte
+}
+
+// NewNodeID2 creates a NodeID2 from the given path bytes truncated to the
+// specified number of bits if necessary. Panics if the number of bits is
+// negative, or is more than the byte string contains.
```
mhutchinson: Consider taking a uint if you want to statically prevent negative numbers.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+func NewNodeID2(path string, bits int) NodeID2 {
+	if bits == 0 {
+		return NodeID2{}
+	} else if mx := len(path) * 8; bits > mx {
+		panic(fmt.Sprintf("NewNodeID2: bits %d > %d", bits, mx))
+	}
+	bytes, tail, mask := split(bits)
+	last := path[bytes] & mask
+	return NodeID2{path: path[:bytes], last: last, bits: byte(tail)}
```
mhutchinson: It *may* be worth a comment here on the fact that go creates substrings without the need for further allocation because the underling arrays are immutable.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+import "fmt"
+
+// NodeID2 is a faster NodeID that does zero memory allocations.
+//
+// TODO(pavelkalinnikov): Rename to NodeID and document it properly when the
+// code has migrated.
+type NodeID2 struct {
+	path string
+	last byte
+	bits byte
```
mhutchinson: Replace this byte with uint8. It's clearer what's going on and the same memory usage.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+	return NodeID2{path: path[:bytes], last: last, bits: byte(tail)}
+}
+
+// BitLen returns the length of the NodeID2 in bits.
+func (n NodeID2) BitLen() int {
+	return len(n.path)*8 + int(n.bits)
+}
+
+// Prefix returns the prefix of NodeID2 with the given number of bits.
+func (n NodeID2) Prefix(bits int) NodeID2 {
```
mhutchinson: This looks to duplicate NewNodeID2 a lot. I think it can be replaced with:
{
  return NewNodeID2(n.path, bits)
}
?

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+	bytes, tail, mask := split(bits)
+	if bytes != len(n.path) {
+		last = n.path[bytes]
+	}
+	last &= mask
+	return NodeID2{path: n.path[:bytes], last: last, bits: byte(tail)}
+}
+
+// Suffix returns the suffix of NodeID2 after the given number of bits.
+func (n NodeID2) Suffix(bits int) NodeID2 {
```
mhutchinson: I would consider dropping this for now. If this is to stay here, I suggest making it take a NodeID2 as its parameter and make this method effectively be "Reroot this node ID within the subtree specified by the argument ID".

storage/tree/node_id2.go
```diff
@@ -0,0 +1,108 @@
...
+		return NodeID2{}
+	} else if bits%8 == 0 {
+		return NewNodeID2(n.path[bits/8:], mx-bits)
+	}
+	// TODO(pavelkalinnikov): Support arbitrary lengths.
+	panic("Suffix: only multiples of 8 are supported")
+}
+
+// Sibling returns the NodeID2 of the nodes's sibling in a binary tree. If the
+// node is the root then the returned ID is the same.
```
mhutchinson: I know this is a ridiculous addition to the doc, but I think it's worth stating that "the sibling is the parent's other child". I've seen Siblings used in the Trillian codebase to mean "all of the other nodes that would be needed for an inclusion proof".

storage/tree/node_id2.go
pavelkalinnikov: Removed.

storage/tree/node_id2.go
pavelkalinnikov: Done.

storage/tree/node_id2.go
```diff
@@ -0,0 +1,100 @@
...
+
+// String returns a human-readable bit string.
+func (n NodeID2) String() string {
+	if n.BitLen() == 0 {
+		return "[]"
+	}
+	path := fmt.Sprintf("%08b", []byte(n.path))
+	path = path[1 : len(path)-1] // Trim the brackets.
+	if len(path) > 0 {
+		path += " "
```
mhutchinson: Pedantic nit: the class level comment is misleading: this seems to be allocating and the class doc says "NodeID2 is a faster NodeID that does zero memory allocations."

pavelkalinnikov: Hehe, well yeah. Does the comment look better now?

mhutchinson: Thank you for entertaining my pedantry. LGTM.

Issue #1850, opened 2019-09-19
Travis doesn't run "go test" without the short flag


mhutchinson:
`go test github.com/google/trillian/quota/etcd/quotaapi`

This test command reliably fails for me. It doesn't on Travis. Looking into why, it's because Travis doesn't run the TestServer_UpdateConfig_Race test because we always provide the short test flag.

PR #1859, opened 2019-09-24
Fix data race by passing wanted config by value

mhutchinson:
This means that this test now passes, which unblocks running it in Travis - #1850.

### DIFF
--- a/quota/etcd/quotaapi/quota_server_test.go
+++ b/quota/etcd/quotaapi/quota_server_test.go
@@ -372,7 +372,7 @@ func TestServer_UpdateConfig_ResetQuota(t *testing.T) {
 	}
 }
 
-func TestServer_UpdateConfig_Race(t *testing.T) {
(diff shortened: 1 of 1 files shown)


quota/etcd/quotaapi/quota_server_test.go
```diff
@@ -372,7 +372,7 @@ func TestServer_UpdateConfig_ResetQuota(t *testing.T) {
 	}
 }
 
-func TestServer_UpdateConfig_Race(t *testing.T) {
+func TestServer_UpdateConfig_ConcurrentUpdates(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping race test due to short mode")
```
Martin2112: delete "race" from Skip message.

mhutchinson: Done.

PR #1860, opened 2019-09-24
Run all the tests in the presubmit scripts

### LINKED ISSUE #1850: Travis doesn't run "go test" without the short flag

mhutchinson:
This script is invoked by Travis and by good developers that want to ensure their changes work properly. Only running those tests that don't self identify as not short doesn't work well for either of these use cases. This fixes #1850 .

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -87,7 +87,7 @@ main() {
     echo 'running go build'
     go build ./...
 
-    export TEST_FLAGS="-short -timeout=${GO_TEST_TIMEOUT:-5m}"
+    export TEST_FLAGS="-timeout=${GO_TEST_TIMEOUT:-5m}"
 
     if [[ ${coverage} -eq 1 ]]; then
       TEST_FLAGS+=" -covermode=atomic -coverprofile=coverage.txt"


scripts/presubmit.sh
```diff
@@ -87,7 +87,7 @@ main() {
     echo 'running go build'
     go build ./...
 
-    export TEST_FLAGS="-short -timeout=${GO_TEST_TIMEOUT:-5m}"
+    export TEST_FLAGS="-timeout=${GO_TEST_TIMEOUT:-5m}"
```
RJPercival: It might be useful to be able to supplement these flags from outside of the script, e.g.
```suggestion
    export TEST_FLAGS="${TEST_FLAGS} -timeout=${GO_TEST_TIMEOUT:-5m}"
```

RJPercival: That might be best done as a separate change though, and would call for a CHANGELOG entry since there's a chance someone already has that environment variable set and it suddenly starts interacting with this script.

mhutchinson: Will leave as-is. If anyone wants this feature they can add it. It's orthogonal to the thrust of this PR.

RJPercival: The relevance was that it'd allow people to keep running the tests with the -short flag if they so desired.

PR #1861, opened 2019-09-24
Removed redundant test for ETCD in presubmit tests

mhutchinson:
Setting WITH_ETCD causes ETCD_DIR to be set, which is used in the integration tests. This env variable is not read in the presubmit tests though, which collapses the row to be equivalent to one of the existing flagsets. This will save Travis from running the presubmit tests with race detection twice per run, which is ~7m30s of execution time.


mhutchinson:
I had a mid-air collision with an additional change and Al's approval. Will wait for another approval from Al, or a nod from @gdbelvin before submitting this.

PR #1862, opened 2019-09-24
Factored test coverage generation into a Travis job

mhutchinson:
The integration tests weren't actually generating a coverage report anyway. The presubmit checks were, but the new job is equivalent to that. The two entries have been removed from the env configuration because they are now equivalent to the line under each of them, modulo the race flag. If we're going to run the tests with the race flag anyway, then there is no advantage to running the tests without it too; it's just a waste of compute time.

This PR tidies up the after_success which was running after all of our 16 tests, but only doing something for one of them. That's an unneccessary load on the codecov website, which I have observed timeouts in the travis logs for. Let's be good citizens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,9 @@ cache:
     - "$HOME/gopath/pkg/mod"
 
 env: #TODO: Merge the test matrix into individual jobs
-  - PRESUB_TESTS=true                                      PRESUBMIT_OPTS="--coverage --no-linters --no-generate"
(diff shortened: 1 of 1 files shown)


gdbelvin:
Rebase on race condition fix?

mhutchinson:
The race condition detected in this PR is not fixed and is tracked by #1863 

This PR reduces the amount of load on Travis by ~1 min dev-perceived per build by shrinking the long pole, and ~15mins total compute time per build (1hr30m -> 1hr15m).

PR #1868, opened 2019-09-25
Refactor: remove Travis common script, which wasn't that common at all

mhutchinson:
The script was just switching and invoking different commands for different environment setups. This PR inlines these into the specific job targets which improves readability and allows for more flexibility per target. Next up is to inline PRESUBMIT_OPTS and the etcd installation which always happens, even though it's only needed for one test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,25 +100,39 @@ jobs:
        - popd
     - name: "presubmit"
       env:
-       - PRESUB_TESTS=true GOFLAGS='-race' PRESUBMIT_OPTS="--no-linters --no-generate"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -100,25 +100,39 @@ jobs:
        - popd
     - name: "presubmit"
       env:
-       - PRESUB_TESTS=true GOFLAGS='-race' PRESUBMIT_OPTS="--no-linters --no-generate"
+       - GOFLAGS='-race' PRESUBMIT_OPTS="--no-linters --no-generate" GO_TEST_TIMEOUT=20m
```
gdbelvin: You can put these one line versions all on the same line like so:

```
script: ./scripts/presubmit.sh ${PRESUBMIT_OPTS}
```

Same for env

mhutchinson: Done.

PR #1869, opened 2019-09-25
Skip large tests when --race is passed

gdbelvin:
--race slows things down significantly and I'm not sure it really provides
value for larger tests.

### DIFF
--- a/integration/maptest/map.go
+++ b/integration/maptest/map.go
@@ -21,9 +21,6 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/golang/glog"
-	"github.com/kylelemons/godebug/pretty"
-
 	"github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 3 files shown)


mhutchinson:
This feels too mysterious to me. I can imagine a workflow like:

```
go test -count 5 ./... # fails 2/5 times on some test
go test -race -count 5 ./... # assume data race so test for this, fails 0/5 times because tests are now being skipped
```

My preferred solution would be that we tag these tests as short, and only ever run (during dev, and in Travis) -race with -short. What's missing from this workflow is the opposite of -short so we can have two Travis configs such as:
```
go test -short -race ./...
go test -long ./...
```

For now, perhaps sucking up the duplication is fine and have:
```
go test -short -race ./...
go test ./...
```

Seems better than this approach where we will have tests not running in Travis, which we just fixed - #1850.

gdbelvin:
I agree that the following approach is best:
```
go test -short -race ./...
go test ./...
```

The challenge is that in Bazil test targets have one set of flags which then get invoked with and without `--race`.

mhutchinson:
Can you get approval from someone like Al or Martin for this? I'm genuinely not sure how much we want to encourage this kind of magical enablement/disablement in our codebase.

If this is approved, we need to have a Travis test target which explicitly does not pass in the -race argument. All of the presubmit & integration tests currently do that, which means these tests can be broken and not detected.

Martin2112:
I'm not sure about this one. There's been a number of cases where tests haven't been running and we didn't notice.

There is only real value in having the race detection if it runs regularly and automatically.

gdbelvin:
Pulled the obvious bits out of this CL into #1884.

gdbelvin:
Closing for now

PR #1870, opened 2019-09-26
Disable go build caching in Travis

mhutchinson:
TBD if this makes it faster or slower. It was taking ~20s after each build target to update the cache, and as all the builds for a revision happen in parallel and with different build tags, there are a lot of good reasons this might cost more to move bytes than it saves us in compile time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ go_import_path: github.com/google/trillian
 
 cache:
   directories:
-    - "$HOME/.cache/go-build"
-    - "$HOME/.cache/go-build-race"
     - "$HOME/google-cloud-sdk/"
     - "$HOME/gopath/pkg/mod"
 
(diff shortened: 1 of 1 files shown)


mhutchinson:
This seems to make approximately zero difference when looking at the results of this run in Travis.

mhutchinson:
I'm on the fence about this change. Both of these arguments seem equally compelling:
a) if it ain't broke, don't fix it; and
b) if you aren't using it, get rid of it

Will let the jury decide.

Martin2112:
It definitely used to speed up builds, but that's a long time ago.

Martin2112:
Maybe for some reason it no longer uses the cache and we haven't noticed?

mhutchinson:
Unclear. It's apparent that this PR fits in with the total compute time distribution observable on https://travis-ci.org/google/trillian/pull_requests. I'd err on the side of removing this and then adding it when we can demonstrate its value. As it stands, anyone trying to speed up our builds might look at this and assume the optimization is already working.

PR #1871, opened 2019-09-26
Refactor: extracted SMR stash out of hammerState

mhutchinson:
There is a lot going on in the hammer. This is the start of teasing out understandable black boxes that are unit tested instead of functionally tested as part of a large and complex system.

### DIFF
--- a/testonly/hammer/record.go
+++ b/testonly/hammer/record.go
@@ -1,0 +1,61 @@
+package hammer
+
+import (
+	"fmt"
+	"reflect"
+	"sync"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian/types"
+)
+
+const (
+	// How many SMRs to hold on to.
+	smrCount = 30
+)
+
+// smrStash provides thread-safe access to an ordered, bounded queue of previously
+// witnessed SMRs
+type smrStash struct {
+	mu sync.RWMutex
+
+	// SMRs are arranged from later to earlier (so [0] is the most recent), and the
+	// discovery of new SMRs will push older ones off the end. No SMR for a revision
+	// will be stored more than once.
+	smr [smrCount]*types.MapRootV1
+}
+
+// pushSMR ensures that the SMR is the latest revision and adds it to the queue of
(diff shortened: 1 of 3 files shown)


testonly/hammer/record_test.go
```diff
@@ -0,0 +1,58 @@
...
+			seq:     []types.MapRootV1{r2, r1},
+			wantErr: true,
+		},
+		{
+			desc: "same revision with same contents",
+			seq:  []types.MapRootV1{r0, r0},
+		},
+		{
+			desc:    "same revision but different contents",
+			seq:     []types.MapRootV1{r0, types.MapRootV1{Revision: 0, RootHash: testonly.MustHexDecode("FFFF")}},
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			seq:     []types.MapRootV1{r0, {Revision: 0, RootHash: testonly.MustHexDecode("FFFF")}},
```

testonly/hammer/record.go
```diff
@@ -0,0 +1,61 @@
...
+
+// smrStash provides thread-safe access to an ordered, bounded queue of previously
+// witnessed SMRs
+type smrStash struct {
+	mu sync.RWMutex
+
+	// SMRs are arranged from later to earlier (so [0] is the most recent), and the
+	// discovery of new SMRs will push older ones off the end. No SMR for a revision
+	// will be stored more than once.
+	smr [smrCount]*types.MapRootV1
```
gdbelvin: I don't know if more refactoring is a waste of time, but this makes me think of 
https://golang.org/pkg/container/list/

testonly/hammer/record_test.go
mhutchinson: Done.

testonly/hammer/record.go
mhutchinson: I don't want to spend more time on this than needed. If we were going to do other work on this stash then this might be worth changing over. Thanks for the pointer though.

PR #1872, opened 2019-09-26
Moved content checking from setLeaves to getSMR methods

mhutchinson:
SetLeaves is being deprecated in #1842 and replaced with WriteLeaves, which does not return an SMR upon writing. This PR moves the checks that were performed on the map contents being applied properly from the write to the reader. This adds a PickRevision method and tests it.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -681,7 +681,7 @@ leafloop:
 	if err != nil {
 		return err
 	}
-	glog.V(2).Infof("%d: set %d leaves, rev=%d", s.cfg.MapID, len(leaves), rev)
+	glog.V(2).Infof("%d: set %d leaves, rev=%d", s.cfg.MapID, len(leaves), writeRev)
 	return nil
 }
 
@@ -727,7 +727,7 @@ func (s *hammerState) getSMR(ctx context.Context, prng *rand.Rand) error {
 	}
 	glog.V(2).Infof("%d: got SMR(time=%q, rev=%d)", s.cfg.MapID, time.Unix(0, int64(root.TimestampNanos)), root.Revision)
 
-	if err := s.validateSMRMatchesWrittenContents(root); err != nil {
+	if err := s.verify(root); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


testonly/hammer/hammer.go
```diff
@@ -671,24 +671,17 @@ leafloop:
...
 	}
-	wantRootHash, err := newContents.RootHash(s.cfg.MapID, s.vc.Hasher)
-	if err != nil {
-		return err
-	}
-	if !bytes.Equal(root.RootHash, wantRootHash) {
-		return fmt.Errorf("failed to check root hash: got %x, want %x", root.RootHash, wantRootHash)
-	}
-	glog.V(2).Infof("%d: set %d leaves, new SMR(time=%q, rev=%d)", s.cfg.MapID, len(leaves), time.Unix(0, int64(root.TimestampNanos)), root.Revision)
+	glog.V(2).Infof("%d: set %d leaves, rev=%d", s.cfg.MapID, len(leaves), rev)
```
pavelkalinnikov: Should this print `writeRev` instead?

mhutchinson: This method inside the hammer is still called setLeaves so I decided to leave it as-is. In a future life I will come back and rename everything in the hammer, but it would be better to be a standalone PR as it will cause a large line delta.

testonly/hammer/hammer.go
```diff
@@ -761,6 +758,23 @@ func (s *hammerState) getSMRRev(ctx context.Context, prng *rand.Rand) error {
...
 	}
+
+	if err := s.validateSMRMatchesWrittenContents(root); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *hammerState) validateSMRMatchesWrittenContents(root *types.MapRootV1) error {
```
pavelkalinnikov: I'd call this a bit shorter, e.g. just `validate[Match]` or `verify[Match]`.

testonly/hammer/hammer.go
```diff
@@ -761,6 +758,23 @@ func (s *hammerState) getSMRRev(ctx context.Context, prng *rand.Rand) error {
...
+	mapContents := s.prevContents.PickRevision(root.Revision)
+	wantRootHash, err := mapContents.RootHash(s.cfg.MapID, s.vc.Hasher)
```
pavelkalinnikov: Same idea here, there is not need for such a long name in such a short function. Something like `want` would do.

testonly/mapcontents_test.go
```diff
@@ -156,3 +157,98 @@ func TestRootHash(t *testing.T) {
...
+			},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			var vmc VersionedMapContents
+			var gotErr error
+			for _, u := range test.updates {
+				_, e := vmc.UpdateContentsWith(u.rev, u.leaves)
+				if e != nil {
```
pavelkalinnikov: Can merge this line with the previous.

testonly/hammer/hammer.go
pavelkalinnikov: I mean, previously this line printed `root.Revision` which corresponded to `writeRev`, not `rev`.

mhutchinson: Good call. Thanks for clarifying,

mhutchinson:
Comments addressed.

PR #1874, opened 2019-09-27
HStar3: A faster non-recursive HStar2

pavelkalinnikov:
This change introduces the new `HStar3` algorithm, a future replacement for `HStar2`. It takes advantage of the faster `NodeID2` type operations like `Prefix` and `Sibling`, and doesn't use the heavy `big.Int` calculations.

Benchmarks compare with a 10-15% speed-up on the same dataset:

```
BenchmarkHStar2Root-12    	       2	 517407218 ns/op
BenchmarkHStar3Root-12    	       3	 443897809 ns/op
```

`HStar3` code also claims to be more readable, as it is not recursive.

It is also more generic. For example, it doesn't depend on `MapHasher` interface and has no "empty hash" notion. The idea is that the caller is the one who provides empty hashes. It allows more flexibility: for example, the caller can error if some hash is not present (e.g. they used the `GetReadSet` method to prefetch all hashes, but `HStar3` tries to read something different), or return an empty hash (e.g. if they did not use the prefetching option).

The lack of empty hash notion allows (in principle) using `HStar3` even with logs. If `HashChildrenFn` returns `nil` unless both children are non-`nil`, and otherwise returns a "log" `HashChildren` then we can basically get similar functionality as `compact.Range` provides.

### DIFF
--- a/merkle/hstar2_test.go
+++ b/merkle/hstar2_test.go
@@ -239,7 +239,7 @@ func TestHStar2RootGolden(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Root: %v", err)
 	}
-	want := "d01bd540dc8b4a3ca3ac8deb485b431e9ce1290becb36838c4463a811d15c7f6"
+	want := "daf17dc2c83f37962bae8a65d294ef7fca4ffa02c10bdc4ca5c4dec408001c98"
 	if got := hex.EncodeToString(hash); got != want {
 		t.Errorf("Root: got %x, want %v", hash, want)
 	}
@@ -255,10 +255,12 @@ func BenchmarkHStar2Root(b *testing.B) {
 	}
 }
 
+// leafHashes generates n leaf updates at depth 256. The function is
+// pseudo-random, and the returned data depends only on n.
 func leafHashes(t testing.TB, n int) []*HStar2LeafHash {
 	t.Helper()
-	// Use a fixed sequence to ensure runs are comparable
-	r := rand.New(rand.NewSource(42424242))
+	// Use a random sequence that depends on n.
+	r := rand.New(rand.NewSource(int64(n)))
 	lh := make([]*HStar2LeafHash, 0, n)
 
 	for l := 0; l < n; l++ {
--- a/merkle/hstar2_test.go
+++ b/merkle/hstar2_test.go
@@ -17,6 +17,7 @@ package merkle
 import (
 	"bytes"
 	"crypto"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"math/rand"
@@ -232,6 +233,18 @@ func TestHStar2NegativeTreeLevelOffset(t *testing.T) {
 	}
 }
 
+func TestHStar2RootGolden(t *testing.T) {
+	hs2 := NewHStar2(42, coniks.New(crypto.SHA256))
+	hash, err := hs2.HStar2Root(256, leafHashes(t, 500))
+	if err != nil {
+		t.Fatalf("Root: %v", err)
+	}
+	want := "d01bd540dc8b4a3ca3ac8deb485b431e9ce1290becb36838c4463a811d15c7f6"
+	if got := hex.EncodeToString(hash); got != want {
+		t.Errorf("Root: got %x, want %v", hash, want)
+	}
+}
+
 func BenchmarkHStar2Root(b *testing.B) {
 	hs2 := NewHStar2(42, coniks.New(crypto.SHA256))
 	for i := 0; i < b.N; i++ {
@@ -242,8 +255,8 @@ func BenchmarkHStar2Root(b *testing.B) {
 	}
 }
 
-func leafHashes(b *testing.B, n int) []*HStar2LeafHash {
-	b.Helper()
+func leafHashes(t testing.TB, n int) []*HStar2LeafHash {
+	t.Helper()
 	// Use a fixed sequence to ensure runs are comparable
 	r := rand.New(rand.NewSource(42424242))
 	lh := make([]*HStar2LeafHash, 0, n)
@@ -251,11 +264,11 @@ func leafHashes(b *testing.B, n int) []*HStar2LeafHash {
 	for l := 0; l < n; l++ {
 		h := make([]byte, 32)
 		if _, err := r.Read(h); err != nil {
-			b.Fatalf("Failed to make random leaf hashes: %v", err)
+			t.Fatalf("Failed to make random leaf hashes: %v", err)
 		}
 		path := make([]byte, 32)
 		if _, err := r.Read(path); err != nil {
-			b.Fatalf("Failed to make random path: %v", err)
+			t.Fatalf("Failed to make random path: %v", err)
 		}
 		lh = append(lh, &HStar2LeafHash{
 			LeafHash: h,
--- a/merkle/smt/hstar3.go
+++ b/merkle/smt/hstar3.go
@@ -95,7 +95,8 @@ func (h HStar3) GetReadSet(updates []NodeUpdate, top uint) map[tree.NodeID2][]by
 			ids[sib] = nil
 		}
 	}
-	// Delete the nodes that don't contribute to the updated hashes because
+	// Delete the sibling nodes that are both in the set. Their original hashes
+	// don't contribute to the updated hashes because they are both changed.
 	for id := range ids {
 		sib := id.Sibling()
 		if _, ok := ids[sib]; ok {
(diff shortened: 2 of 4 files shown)


pavelkalinnikov:
I recommend reviewing this commit-by-commit.

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,186 @@
...
+func (h HStar3) updateAt(updates []NodeUpdate, depth uint, ns NodeStorage) ([]NodeUpdate, error) {
+	// Apply the updates.
+	for _, upd := range updates {
+		ns.Set(upd.ID, upd.Hash)
+	}
+	// Calculate the updates that propagate to one level above.
+	newLen := 0
+	for i, ln := 0, len(updates); i < ln; i, newLen = i+1, newLen+1 {
+		sib := updates[i].ID.Sibling()
+		left, right := updates[i].Hash, []byte{}
```
golangcibot: ineffectual assignment to `right` (from `ineffassign`)

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,186 @@
...
+		for id, d := upd.ID, h.depth; d > top; d-- {
+			sib := id.Prefix(d).Sibling()
+			if _, ok := ids[sib]; ok {
+				// All the upper siblings have been added already, so skip them.
+				break
+			}
+			ids[sib] = nil
+		}
+	}
+	// Delete the nodes that don't contribute to the updated hashes because
```
pavelkalinnikov: oups, need to end this comment

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,187 @@
...
+func (h HStar3) updateAt(updates []NodeUpdate, depth uint, ns NodeStorage) ([]NodeUpdate, error) {
+	// Apply the updates.
+	for _, upd := range updates {
+		ns.Set(upd.ID, upd.Hash)
+	}
+	// Calculate the updates that propagate to one level above.
+	newLen := 0
+	for i, ln := 0, len(updates); i < ln; i, newLen = i+1, newLen+1 {
+		sib := updates[i].ID.Sibling()
+		left, right := updates[i].Hash, []byte(nil)
```
golangcibot: ineffectual assignment to `right` (from `ineffassign`)

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,183 @@
...
+// GetReadSet returns the set of all the node IDs that HStar3 requires to read
+// in order to apply the passed-in updates and propagate node hash changes up
+// to the root(top) node(s) at the specified depth. The updates must have been
+// processed by Prepare.
+//
+// Note: The returned map could have bool value type, but []byte allows the
+// caller to reuse this map for filling in the hashes for the Update method.
+//
+// TODO(pavelkalinnikov): Return only tile IDs.
+func (h HStar3) GetReadSet(updates []NodeUpdate, top uint) map[tree.NodeID2][]byte {
```
mhutchinson: This name is properly confusing. It's a bit like Reading Festival - is it a festival with lots of books, or is it a festival in Reading? Or both?

I initially read this as 3 verbs: "Get x, Read x, Set x", but it's intended to be parsed as "Get the {nodes} to read".

This API surface area is a bit large for what is required here. I would propose an inversion of control - have a single method which takes objects that allow these nodes to be fetched in batches. That also avoids the strange artifact where this function is creating an overly ambitious return type because it can predict what its clients are going to do next.

Rough idea of how my thoughts could be implemented
```
NodeStorageBuilder {
  nodes map[tree.NodeID2][]byte
}
func (nsb NodeStorageBuilder) Prefetch(n tree.NodeID2) {
  nodes[n] = nil
}
func (nsb NodeStorageBuilder) BuildStorage() NodeStorage {
  return storageDependantFetchMethod(nsb.nodes)
}


UberUpdate(updates []NodeUpdate, top uint, nsb NodeStorageBuilder) {
  prepare(updates)
  getReadSet(nsb)
  update(updates, top, nsb.BuildStorage())
}
```
In the case that you don't want to prefetch, the NSB would do nothing for Prefetch, and then return something with a direct line to the DB for BuildStorage. For the (sane) case where you want to prefetch things, the only place that would talk to the DB is what I've called storageDependantFetchMethod above, and the NodeStorage returned wouldn't have any connection to the DB which prevents prefetch misses from silently succeeding.

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,183 @@
...
+
+// updateAt applies the given node updates at the specified tree level.
+// Returns the updates that propagated to the level above.
+func (h HStar3) updateAt(updates []NodeUpdate, depth uint, ns NodeStorage) ([]NodeUpdate, error) {
+	// Apply the updates.
+	for _, upd := range updates {
+		ns.Set(upd.ID, upd.Hash)
+	}
+	// Calculate the updates that propagate to one level above.
+	newLen := 0
```
mhutchinson: I think a comment on newLen that it will be in the range [length(updates)/2, length(updates)] & it shrinks whenever two updates have the same parent.

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,183 @@
...
+			var err error
+			if right, err = ns.Get(sib); err != nil {
+				return nil, err
+			}
+			if isLeftChild(sib) {
+				left, right = right, left
+			}
+		}
+		hash := h.hash(left, right)
+		updates[newLen] = NodeUpdate{ID: sib.Prefix(depth - 1), Hash: hash}
```
mhutchinson: For me, it would be clearer if newLen was incremented here instead of in the loop post section.

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,183 @@
...
+func (h HStar3) updateAt(updates []NodeUpdate, depth uint, ns NodeStorage) ([]NodeUpdate, error) {
+	// Apply the updates.
+	for _, upd := range updates {
+		ns.Set(upd.ID, upd.Hash)
+	}
+	// Calculate the updates that propagate to one level above.
+	newLen := 0
+	for i, ln := 0, len(updates); i < ln; i, newLen = i+1, newLen+1 {
+		sib := updates[i].ID.Sibling()
+		left := updates[i].Hash
```
mhutchinson: This could actually be the rightmost node at first, and only becomes left after the if/else below. It's fine when you notice the switcheroo with isLeftChild()... below, but until then it hurt my brain. Suggest leaving left and right both uninitialized here, and initialize both of them in the block below.

merkle/hstar2_test.go
```diff
@@ -242,20 +255,20 @@ func BenchmarkHStar2Root(b *testing.B) {
 	}
 }
 
-func leafHashes(b *testing.B, n int) []*HStar2LeafHash {
-	b.Helper()
+func leafHashes(t testing.TB, n int) []*HStar2LeafHash {
```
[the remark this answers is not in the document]
pavelkalinnikov: Done. Both added comments explaining the pseudo-randomness, and made the function more random.

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,186 @@
...
+// NodeStorage reads and writes sparse Merkle tree node hashes.
+type NodeStorage interface {
+	// Get returns the hash of the given node.
+	Get(id tree.NodeID2) ([]byte, error)
+	// Set sets the hash of the given node.
+	Set(id tree.NodeID2, hash []byte)
+}
+
+// HashChildrenFn computes a node hash based on its child nodes' hashes.
+type HashChildrenFn func(left, right []byte) []byte
```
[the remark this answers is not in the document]
pavelkalinnikov: Not necessarily. It depends on how the thing overarching `HStar3` is set up. If it's parallel then `HashChildren` must be thread-safe, but if not then it doesn't have to. If I added such a comment, it would be 1 level above in the sparse Merkle tree writer.

merkle/smt/hstar3.go
pavelkalinnikov: Great idea. Done.

merkle/smt/hstar3.go
Martin2112: `Prepare` or similar?

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,197 @@
...
+		return res < 0
+	}
+	aLast, _ := a.LastByte()
+	bLast, _ := b.LastByte()
+	return aLast < bLast
+}
+
+// sortUpdates sorts the updates slice for it to be usable by HStar3. It also
+// verifies that the nodes are placed at the required depth, and there are no
+// ID duplicates amongh them.
```
mhutchinson: Typo "amongh".

Suggest: "there are no duplicate IDs"

merkle/smt/hstar3.go
```diff
@@ -0,0 +1,197 @@
...
+	top   uint
+}
+
+// NewHStar3 returns a new instance of HStar3 for the given set of node hash
+// updates at the specified tree depth. This HStar3 is capable of propagating
+// these updates up to the passed-in top level of the tree.
+//
+// Warning: This call and other HStar3 methods modify the updates slice
+// in-place, so the caller must ensure to not reuse it.
+func NewHStar3(updates []NodeUpdate, hash HashChildrenFn, depth, top uint) (HStar3, error) {
```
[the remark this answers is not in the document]
pavelkalinnikov: Done. Technically, top == depth works just as fine, so I'm allowing it.

merkle/smt/hstar3.go
pavelkalinnikov: `Prepare` sounds good. Done.

PR #1875, opened 2019-09-30
Switch the map hammer to use WriteLeaves instead of SetLeaves

mhutchinson:
This will allow #1842 to be submitted without needing to disable linter checks for using deprecated methods.

This is based on top of #1872, which is required so that the missing SMR in WriteLeaves (compared to SetLeaves response) is not an issue.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -139,6 +139,7 @@ type MapConfig struct {
 	MapID                int64 // 0 to use an ephemeral tree
 	MetricFactory        monitoring.MetricFactory
 	Client               trillian.TrillianMapClient
+	Write                trillian.TrillianMapWriteClient
 	Admin                trillian.TrillianAdminClient
 	RandSource           rand.Source
 	EPBias               MapBias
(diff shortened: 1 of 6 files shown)


testonly/hammer/hammer.go
```diff
@@ -761,6 +761,23 @@ func (s *hammerState) getSMRRev(ctx context.Context, prng *rand.Rand) error {
...
+}
+
+func (s *hammerState) validateSMRMatchesWrittenContents(root *types.MapRootV1) error {
+	mapContents := s.prevContents.PickRevision(root.Revision)
+	wantRootHash, err := mapContents.RootHash(s.cfg.MapID, s.vc.Hasher)
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(root.RootHash, wantRootHash) {
+		return fmt.Errorf("failed to check root hash: got %x, want %x", root.RootHash, wantRootHash)
```
AlCutter: /failed to check/unexpected/?

mhutchinson: I'll update the error message in the base PR (#1872). If I had a dollar for every suggestion made on code that was moved as part of a refactoring, I'd be a rich man :-)

PR #1878, opened 2019-09-30
Fixed flakiness introduced in #1872

mhutchinson:
Verifying root correctness at old revisions isn't guaranteed to work because we only keep a finite cache of committed state. Trying to check SMRs for revisions > 10 in the past would fail. The previous code only checked this immediately on write, where #1872 performed the check on both a) examining the latest SMR; and b) examining historic SMRs. This change only keeps a) and drops the check in b), because the old state may have been purged. This gives less coverage, but is closer to what we had before. As work progresses on this hammer, I intend to increase coverage over historic roots. The priority for now is removing the flake though.

This caused flakiness because of the randomness of the hammer and the number of iterations. This change also increases the number of operations performed in Travis, which would turn this from a flakey failure to a persistent failure, if this PR didn't also include the medicine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,8 +101,8 @@ jobs:
       env: GOFLAGS='-race --tags=pkcs11' GO_TEST_TIMEOUT=20m
       script: ./scripts/presubmit.sh --no-linters --no-generate
     - name: "integration"
-      env: GOFLAGS='-race' GO_TEST_TIMEOUT=20m
-      script: ./integration/integration_test.sh && HAMMER_OPTS="--operations=50" ./integration/maphammer.sh 3
+      env: GO_TEST_TIMEOUT=20m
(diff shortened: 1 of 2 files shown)


pavelkalinnikov:
Or an `Errorf` even, to make it more visible.

mhutchinson:
As mentioned in the PR description, including this verification in the checking of older SMRs was a new feature that "seemed like a good idea at the time" when doing #1872. The idea still has merit, but printing any kind of a warning or error is creating noise. This feature needs to be thought through more. I think making the stash of previously seen SMRs have a logarithmic decay, and then only requesting ones we already have to re-verify might be an approach. But requesting random SMRs and printing spam log entries if we don't have a copy to verify against seems like fundamentally the wrong approach.

PR #1879, opened 2019-10-01
Handle case where no SMR found yet

mhutchinson:
This eliminates a potential nil pointer error if String is called before any SMR has been found.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -387,8 +387,11 @@ func (s *hammerState) String() string {
 		totalInvalidReqs += int(invalidReqs.Value(s.label(), string(ep)))
 		totalErrs += int(errs.Value(s.label(), string(ep)))
 	}
-	smr := s.previousSMR(0)
-	return fmt.Sprintf("%d: lastSMR.rev=%d ops: total=%d (%f ops/sec) invalid=%d errs=%v%s", s.cfg.MapID, smr.Revision, totalReqs, float64(totalReqs)/interval.Seconds(), totalInvalidReqs, totalErrs, details)
+	var latestRev uint64
+	if smr := s.previousSMR(0); smr != nil {
+		latestRev = smr.Revision
+	}
+	return fmt.Sprintf("%d: lastSMR.rev=%d ops: total=%d (%f ops/sec) invalid=%d errs=%v%s", s.cfg.MapID, latestRev, totalReqs, float64(totalReqs)/interval.Seconds(), totalInvalidReqs, totalErrs, details)
 }
 
 func (s *hammerState) pushSMR(smr *types.MapRootV1) error {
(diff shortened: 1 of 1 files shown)


testonly/hammer/hammer.go
```diff
@@ -387,8 +387,11 @@ func (s *hammerState) String() string {
 		totalInvalidReqs += int(invalidReqs.Value(s.label(), string(ep)))
 		totalErrs += int(errs.Value(s.label(), string(ep)))
 	}
-	smr := s.previousSMR(0)
-	return fmt.Sprintf("%d: lastSMR.rev=%d ops: total=%d (%f ops/sec) invalid=%d errs=%v%s", s.cfg.MapID, smr.Revision, totalReqs, float64(totalReqs)/interval.Seconds(), totalInvalidReqs, totalErrs, details)
+	var latestRev uint64
+	if smr := s.previousSMR(0); smr != nil {
```
gdbelvin: This feels a bit hacky.  
- Do you want to do this check inside `previousSMR`?
- Should `previousSMR` return (SMR, OK) like go maps do?
- Should the map of previous SMRs be a map of SMRs rather than pointers so that the zero value is meaningful rather than nil?

mhutchinson: It's common practice in other languages to return an object if it exists, or nil/null if it doesn't. I'm unclear on how this is hacky. Maybe a Go thing? Though Go does this with errors all of the time, so 🤷‍♂

Worth noting that this previousSMR method has had the same interface and behaviour for 2 years: https://github.com/google/trillian/pull/824. This PR is simply attempting to fix a latent bug that has existed that whole time, which actually turned up for me when I made the MaxKeys in the hammer huge locally. This nil behaviour is relied on in other places in the code to determine that we don't have such a revision.

If we did want to improve it, I'm not convinced that returning a boolean which is effectively "is the response nil" provides value. But we can do that if it's idiomatic.

Returning a non-nil value of an SMR would be misleading. The purpose of the previousSMR method is to return which SMRs have been found thus far. Effectively synthesizing them locally from default values feels like making up responses. It also then prevents the hammer from clearly signaling whether it received a valid SMR for root 0, or whether we found no valid SMRs at all. To be fair this PR hides that from the response here, so I've changed this code to print a negative value if no SMR has been found.

PTAL?

gdbelvin: Relying on `nil` always makes me nervous. I personally prefer things that are more semantically explicit. 
We don't have to make the change here, but I think I'd prefer to see something like (SMR, OK).
It's more canonical, and we can avoid null pointer exceptions at the same time. 

Add a TODO?

mhutchinson: I feel like nil is quite semantically explicit to indicate "what you asked for doesn't exist", especially in the context of an internal helper method in testonly code. But I've added a TODO to reconsider this.

PR #1881, opened 2019-10-01
Added operation on MapClient to get root by revision

mhutchinson:
This operation is unused in this PR, but a follow up PR will make use of it. This operation is likely to be used heavily in the real world.

### DIFF
--- a/client/map_client.go
+++ b/client/map_client.go
@@ -52,6 +52,16 @@ func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRo
 	return c.VerifySignedMapRoot(rootResp.GetMapRoot())
 }
 
+// GetAndVerifyMapRootByRevision verifies and returns the map root with the given revision.
+func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context, revision int64) (*types.MapRootV1, error) {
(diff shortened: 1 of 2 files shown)


mhutchinson:
Travis has actually completed successfully but doesn't seem to have propagated success back to Github. Idunnowhy.

client/map_client.go
```diff
@@ -52,6 +52,16 @@ func (c *MapClient) GetAndVerifyLatestMapRoot(ctx context.Context) (*types.MapRo
...
 }
 
+// GetAndVerifyMapRootByRevision verifies and returns the map root with the given revision.
+func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context, revision int64) (*types.MapRootV1, error) {
+	rootResp, err := c.Conn.GetSignedMapRootByRevision(ctx, &trillian.GetSignedMapRootByRevisionRequest{MapId: c.MapID, Revision: revision})
+	if err != nil {
+		s := status.Convert(err)
+		return nil, status.Errorf(s.Code(), "GetSignedMapRootByRevision(%v): %v", c.MapID, s.Message())
+	}
+	return c.VerifySignedMapRoot(rootResp.GetMapRoot())
```
gdbelvin: Might want to include a check that the returned revision was indeed the one that we asked for.

mhutchinson: Done.

PR #1882, opened 2019-10-01
Implement sparse Merkle tree writer on HStar3

pavelkalinnikov:
This change introduces a sparse Merkle tree `Writer` that uses `HStar3` algorithm.

Key design decisions:
- Use `NodeID2` and `HStar3` because they are fast.
- Use 2-level sharding for simplicity, but allow easy extension if needed at some point.
- Throw away the transaction concept, as it belongs to `storage` / application layers.
- Shift control to the user of this type. This allows, for example, writing different tree shards in different transactions or even machines. See `TestWriterBigBatch` for instance.

In-memory benchmarks show almost a 2x speedup compared to the old `SparseMerkleTreeWriter`:

```
Before:	BenchmarkSparseMerkleTreeWriterBigBatch-12	2	569661060 ns/op
After:	BenchmarkWriterBigBatch-12			4	312185125 ns/op
```

### DIFF
--- a/merkle/smt/writer.go
+++ b/merkle/smt/writer.go
@@ -1,0 +1,162 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package smt
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/google/trillian/merkle/hashers"
+	"github.com/google/trillian/storage/tree"
+)
+
+// NodeBatchAccessor reads and writes batches of Merkle tree node hashes.
+type NodeBatchAccessor interface {
+	// Get returns the hashes of the given nodes, as a map keyed by their IDs.
+	// The returned hashes may be missing or be nil for empty subtrees.
+	Get(ids []tree.NodeID2) (map[tree.NodeID2][]byte, error)
+	// Set applies the given node hash updates.
+	Set(upd []NodeUpdate) error
+}
+
+// Writer handles sharded writes to a sparse Merkle tree.
+type Writer struct {
+	treeID int64
+	hasher hashers.MapHasher
+	// depths maps the depth of "leaf" nodes of a shard to its root depth.
+	depths map[uint]uint
+}
+
+// NewWriter creates a new Writer for the specified tree of the given height,
+// with 2 levels of sharding. One shard spans the the upper levels down to
+// split point, and all the other shards span the levels below.
+func NewWriter(treeID int64, hasher hashers.MapHasher, height, split uint) *Writer {
+	if split > height {
+		panic(fmt.Errorf("NewWriter: split(%d) > height(%d)", split, height))
+	}
+	depths := map[uint]uint{height: split, split: 0}
+	return &Writer{treeID: treeID, hasher: hasher, depths: depths}
+}
+
+// Split sorts and splits the given list of node hash updates into shards, i.e.
+// the subsets belonging to different subtrees. The updates must belong to the
+// same tree level, one of the sharding split points. For example, with a
+// 2-level sharding it only makes sense to Split leaf updates.
+func (w *Writer) Split(upd []NodeUpdate) ([][]NodeUpdate, error) {
+	if len(upd) == 0 { // Nothing to split.
+		return nil, nil
+	}
+	depth := upd[0].ID.BitLen()
+	if err := sortUpdates(upd, depth); err != nil {
+		return nil, err
+	}
+	top, found := w.depths[depth]
+	if !found {
+		return nil, fmt.Errorf("unexpected depth %d", depth)
+	}
+	// TODO(pavelkalinnikov): Try estimating the capacity for this slice.
+	var shards [][]NodeUpdate
+	// The updates are sorted, so we can split them by prefix.
+	for begin, i := 0, 0; i < len(upd); i++ {
+		pref := upd[i].ID.Prefix(top)
+		next := i + 1
+		// Check if this ID ends the shard.
+		if next == len(upd) || upd[next].ID.Prefix(top) != pref {
+			shards = append(shards, upd[begin:next])
+			begin = next
+		}
+	}
+	return shards, nil
+}
+
+// Write applies the given list of updates to a single shard, and returns the
+// resulting update of the shard root. It uses the given node accessor for
+// reading and writing tree nodes. Typically, the input updates have been
+// obtained from Split method.
+func (w *Writer) Write(upd []NodeUpdate, acc NodeBatchAccessor) (NodeUpdate, error) {
+	if len(upd) == 0 {
+		return NodeUpdate{}, errors.New("nothing to write")
+	}
+	depth := upd[0].ID.BitLen()
+	top, found := w.depths[depth]
+	if !found {
+		return NodeUpdate{}, fmt.Errorf("unexpected depth %d", depth)
+	}
+
+	hs, err := NewHStar3(upd, w.hasher.HashChildren, depth, top)
+	if err != nil {
+		return NodeUpdate{}, err
+	}
+	ids := hs.Prepare()
+	nodes, err := acc.Get(ids)
+	if err != nil {
+		return NodeUpdate{}, err
+	}
+	sa := w.newAccessor(nodes)
+	topUpd, err := hs.Update(sa)
+	if err != nil {
+		return NodeUpdate{}, err
+	} else if ln := len(topUpd); ln != 1 {
+		return NodeUpdate{}, fmt.Errorf("writing across %d shards, want 1", ln)
+	}
+	if err := acc.Set(sa.writes); err != nil {
+		return NodeUpdate{}, err
+	}
+
+	return topUpd[0], nil
+}
+
+// newAccessor returns a NodeAccessor for HStar3 algorithm based on the set of
+// preloaded node hashes.
+func (w *Writer) newAccessor(nodes map[tree.NodeID2][]byte) *shardAccessor {
+	// For any node that HStar3 reads, it also writes its sibling. Therefore we
+	// can pre-allocate this many items for the writes slice.
+	// TODO(pavelkalinnikov): The actual number of written nodes will be slightly
+	// bigger by at most the number of written leaves. Try allocating precisely.
+	writes := make([]NodeUpdate, 0, len(nodes))
+	return &shardAccessor{w: w, reads: nodes, writes: writes}
+}
+
+// shardAccessor provides read and write access to nodes used by HStar3.
+type shardAccessor struct {
+	w      *Writer
+	reads  map[tree.NodeID2][]byte
+	writes []NodeUpdate
+}
+
+// Get returns the hash of the given node from the preloaded map, or a hash of
+// an empty subtree at this position if such node is not found.
+func (s *shardAccessor) Get(id tree.NodeID2) ([]byte, error) {
+	if hash, ok := s.reads[id]; ok && hash != nil {
+		return hash, nil
+	}
+	return hashEmpty(s.w.hasher, s.w.treeID, id), nil
+}
+
+// Set adds the given node hash update to the list of writes.
+func (s *shardAccessor) Set(id tree.NodeID2, hash []byte) {
+	s.writes = append(s.writes, NodeUpdate{ID: id, Hash: hash})
+}
+
+// TODO(pavelkalinnikov): Make MapHasher.HashEmpty method take the id directly.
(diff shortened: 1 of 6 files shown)


merkle/smt/writer_test.go
```diff
@@ -0,0 +1,121 @@
...
+			upd = append(upd, u)
+		}
+	}
+
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+	splitUpd := make([]NodeUpdate, 0, 256)
+
+	w := NewWriter(treeID, hasher, 256, 8)
+	err := w.Split(upd, func(upd []NodeUpdate) error {
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

merkle/smt/writer_test.go
```diff
@@ -0,0 +1,121 @@
...
+		wg.Add(1)
+		go func() {
```
golangcibot: SA2002: the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (from `staticcheck`)

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/google/trillian/merkle/hashers"
+	"github.com/google/trillian/storage/tree"
+)
+
+// NodeBatchAccessor reads and writes batches of Merkle tree node hashes.
+type NodeBatchAccessor interface {
+	// Get returns the hashes of the given nodes, as the corresponding values of
+	// the passed-in map. The initial map values must be nil, and will remain nil
+	// for nodes that root empty subtrees.
+	Get(nodes map[tree.NodeID2][]byte) error
```
gdbelvin: Modifying arguments is a bit of an anti-pattern in go.  Is there a strong reason to not use a return value?

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// NewWriter creates a new Writer for the specified tree of the given height,
+// with 2 levels of sharding. One shard spans the the upper levels down to
+// split point, and all the other shards span the levels below.
+func NewWriter(treeID int64, hasher hashers.MapHasher, height, split uint) *Writer {
+	if split > height {
+		panic(fmt.Errorf("NewWriter: split(%d) > height(%d)", split, height))
+	}
+	shards := make(map[uint]shardWriter, 2)
```
gdbelvin: It would be nice if this were a named constant

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	w := &Writer{treeID: treeID, hasher: hasher, shards: shards}
+	shards[height] = shardWriter{w: w, depth: height, top: split}
```
gdbelvin: Perhaps add a comment here that if `2` were changed,  this would need to be a loop of some kind

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	Get(nodes map[tree.NodeID2][]byte) error
+	// Set applies the given node hash updates.
+	Set(upd []NodeUpdate) error
+}
+
+// Writer handles sharded writes to a sparse Merkle tree.
+type Writer struct {
+	treeID int64
+	hasher hashers.MapHasher
+	shards map[uint]shardWriter
```
gdbelvin: Add a comment about how the shards are indexed. 
It appears from below that the integer is the height in the tree?

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+// with 2 levels of sharding. One shard spans the the upper levels down to
+// split point, and all the other shards span the levels below.
+func NewWriter(treeID int64, hasher hashers.MapHasher, height, split uint) *Writer {
+	if split > height {
+		panic(fmt.Errorf("NewWriter: split(%d) > height(%d)", split, height))
+	}
+	shards := make(map[uint]shardWriter, 2)
+	w := &Writer{treeID: treeID, hasher: hasher, shards: shards}
+	shards[height] = shardWriter{w: w, depth: height, top: split}
+	shards[split] = shardWriter{w: w, depth: split, top: 0}
```
gdbelvin: Almost feels like `top` is behaving like a linked list

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Split sorts and splits the given list of node hash updates into shards, i.e.
+// the subsets belonging to different subtrees. It calls the visit function for
+// each of the shards, and stops if the latter returns an error. The updates
+// must belong to the same tree level, one of the sharding split points. For
+// example, with a 2-level sharding it only makes sense to Split leaf updates.
+func (w *Writer) Split(upd []NodeUpdate, visit func([]NodeUpdate)) error {
+	if len(upd) == 0 {
+		return errors.New("nothing to split")
```
gdbelvin: Not sure if this is really an error condition

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	depth := upd[0].ID.BitLen()
```
gdbelvin: rather than deriving this from the first item, should this be passed in and then asserted that every update contains the same depth?

gdbelvin: ie. Depth should almost be a tree level property.

gdbelvin: sortUpdates also seems to be combining validation with sorting.  Should these be separate and just have a lexographic sorting method?

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	if err := sortUpdates(upd, depth); err != nil {
+		return err
+	}
+	shard, found := w.shards[depth]
```
gdbelvin: Indexing shards by depth seems non-intuitive to me.  
Wouldn't we want multiple shards at the same depth?

merkle/smt/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	shards[split] = shardWriter{w: w, depth: split, top: 0}
+	return w
+}
+
+// Split sorts and splits the given list of node hash updates into shards, i.e.
+// the subsets belonging to different subtrees. It calls the visit function for
+// each of the shards, and stops if the latter returns an error. The updates
+// must belong to the same tree level, one of the sharding split points. For
+// example, with a 2-level sharding it only makes sense to Split leaf updates.
+func (w *Writer) Split(upd []NodeUpdate, visit func([]NodeUpdate)) error {
```
gdbelvin: I would expect a function called `Split` to return splits. 
Should we separate responsibilities here for simplicity? 
Returning `[][]NodeUpdate` wouldn't waste memory since they're all indexed out of the same parent array. The caller could then do what it wants to with the split nodes.

merkle/smt/writer.go
pavelkalinnikov: We do have multiple shards at the same level. The `w.shards` is more like metadata for shards, which depends only on the level. Maybe I should call these "tiers" or "shardMeta" etc.

It also looks like I can bake the `shardWriter` type into the `Writer`.

merkle/smt/writer.go
pavelkalinnikov: Yeah, the `top` of one shard is the `depth` of another. A list is not necessarily suitable though, I need quick indexing by `depth`. Could be done with:
- a binary search over a sorted slice of depths
- a `switch` statement (we only have 2 depths)
- a `map` keyed by `depth`

merkle/smt/writer.go
pavelkalinnikov: Done.

merkle/smt/writer.go
pavelkalinnikov: Done.

merkle/smt/writer.go
pavelkalinnikov: > rather than deriving this from the first item, should this be passed in and then asserted that every update contains the same depth?

This wouldn't add any more safety, only increase the number of parameters. Part of the `sortUpdates` is making sure that the IDs have the same length, because otherwise the ordering that it defines doesn't make sense. So I thought it was a good idea to combine the two. Not sure about duplicate detection though being in the right place there, but why not... Both `HStar3` and `Writer` use it in the same way.

Still on the fence with that, let's see what other reviewers think.

merkle/smt/writer.go
pavelkalinnikov: Fixed.

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+func (w *Writer) newAccessor(nodes map[tree.NodeID2][]byte) *shardAccessor {
+	// For any node that HStar3 reads, it also writes its sibling. Therefore we
+	// can pre-allocate this many items for the writes slice.
+	// TODO(pavelkalinnikov): The actual number of written nodes will be slightly
+	// bigger by at most the number of written leaves. Try allocating precisely.
+	writes := make([]NodeUpdate, 0, len(nodes))
+	return &shardAccessor{w: w, reads: nodes, writes: writes}
+}
+
+// shardAccessor provides read and write access to nodes used by HStar3.
```
mhutchinson: Please make it crystal clear that this operates entirely in-memory.

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// TODO(pavelkalinnikov): Make MapHasher.HashEmpty method take the id directly.
+func hashEmpty(hasher hashers.MapHasher, treeID int64, id tree.NodeID2) []byte {
+	index := make([]byte, hasher.Size())
+	copy(index, id.FullBytes())
+	if last, bits := id.LastByte(); bits != 0 {
+		index[len(id.FullBytes())] = last
+	}
+	height := hasher.BitLen() - int(id.BitLen())
```
gdbelvin: This looks like it's doing what `FullBytes` should be doing? 
Should this be inside of NodeID2?

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+type Writer struct {
+	treeID int64
+	hasher hashers.MapHasher
+	// depths maps the depth of "leaf" nodes of a shard to its root depth.
+	depths map[uint]uint
+}
+
+// NewWriter creates a new Writer for the specified tree of the given height,
+// with 2 levels of sharding. One shard spans the the upper levels down to
+// split point, and all the other shards span the levels below.
```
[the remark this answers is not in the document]
pavelkalinnikov: For the first question, there is no such requirement as this type doesn't know about tiles (yet). The caller is responsible for layout alignment checks.

For the second paragraph, I think the `Writer` type comment does the job of explaining this.

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, nil
+	}
+	depth := upd[0].ID.BitLen()
+	if err := sortUpdates(upd, depth); err != nil {
+		return nil, err
+	}
+	top, found := w.depths[depth]
+	if !found {
+		return nil, fmt.Errorf("unexpected depth %d", depth)
+	}
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+			shards = append(shards, upd[begin:next])
+			begin = next
+		}
+	}
+	return shards, nil
+}
+
+// Write applies the given list of updates to a single shard, and returns the
+// resulting update of the shard root. It uses the given node accessor for
+// reading and writing tree nodes. Typically, the input updates have been
```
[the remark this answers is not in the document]
pavelkalinnikov: Elaborated the 2 typical ways of calling this method.

merkle/smt/writer.go
```diff
@@ -0,0 +1,162 @@
...
+// obtained from Split method.
+func (w *Writer) Write(upd []NodeUpdate, acc NodeBatchAccessor) (NodeUpdate, error) {
```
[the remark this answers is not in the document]
pavelkalinnikov: The top-level shard updates needs manual construction by the client, so having a separate type makes it a bit less convenient. Also, if `split == 0`, then the `Split` call is not necessary, so again the client would need to convert the slice. I tried explaining the use-cases in the comments.

Leaving as is for now. I wonder though if slice appends would work on `type ShardUpdates []NodeUpdate` without conversion. Maybe it's not that bad actually.

merkle/smt/writer.go
pavelkalinnikov: Good news. After baking the only-2-levels assumption into this type, there is only one `depth` that we can accept here (the leaves), so I made this method no longer look at `upd[0]`.

merkle/smt/writer.go
pavelkalinnikov: `FullBytes` does not return the last byte, even if it's full, so it's not the same thing.
I don't want to put this to `NodeID2` yet, until it's clear that there is more than one use-case. Also I hope to avoid this conversion entirely when `storage` and `hashers.MapHasher` support `NodeID2`.

merkle/smt/writer.go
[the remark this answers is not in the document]
mhutchinson: I'm happy with the new comments to explain the intended uses. It's also a lot clearer now there are explicitly only 2 levels to the shard structure. Resolving this.

merkle/smt/writer.go
pavelkalinnikov: Ended up with a `switch`, as more than 2 levels seems YAGNI.

pavelkalinnikov:
The tests coverage seems decent now, so I'm submitting this PR.

PR #1883, opened 2019-10-02
Hstar3: Implement a faster Prepare algorithm

pavelkalinnikov:
This change implements a faster `Prepare` algorithm that generates the slice of nodes straight away without using a Golang `map`.

The benchmark introduced in this PR shows a 10x speedup of `Prepare`:

```
Before:	BenchmarkHStar3Prepare-12    	      19	  59717482 ns/op
After:	BenchmarkHStar3Prepare-12    	     189	   5793267 ns/op
```

### DIFF
--- a/merkle/smt/hstar3.go
+++ b/merkle/smt/hstar3.go
@@ -79,20 +79,31 @@ func NewHStar3(updates []NodeUpdate, hash HashChildrenFn, depth, top uint) (HSta
 // TODO(pavelkalinnikov): Return only tile IDs.
 func (h HStar3) Prepare() []tree.NodeID2 {
 	empty := tree.NodeID2{}
-	// Start with a dummy sentinel value.
+
+	// Start with a single "sentinel" empty ID, which helps maintaining the loop
+	// invariants below. Preallocate enough memory to store all the node IDs.
 	ids := make([]tree.NodeID2, 1, len(h.upd)*int(h.depth-h.top)+1)
 	pos := make([]int, h.depth-h.top)
 
 	// For each node, add all its ancestors' siblings, down to the given depth.
+	// Avoid duplicate IDs, and possibly remove already added ones if they become
+	// unnecessary as more updates are added.
+	//
+	// Loop invariants:
+	// 1. pos[idx] < len(ids), for each idx.
+	// 2. ids[pos[idx]] is the ID of the rightmost sibling at depth idx+h.top+1,
+	//    or an empty ID if there is none at this depth yet.
+	//
+	// Note: The algorithm works because the list of updates is sorted.
 	for _, upd := range h.upd {
 		for id, d := upd.ID, h.depth; d > h.top; d-- {
 			pref := id.Prefix(d)
 			idx := d - h.top - 1
 			if p := pos[idx]; ids[p] == pref {
-				// Delete that node because its original hash does not contribute to
-				// the updates, so should not be read.
+				// Delete that node because its original hash will be overridden, so it
+				// does not contribute to hash updates anymore.
 				ids[p] = empty
-				// All the upper siblings have been added already, so skip them.
+				// Skip the upper siblings as they have been added already.
 				break
 			}
 			pos[idx] = len(ids)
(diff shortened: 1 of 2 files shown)


merkle/smt/hstar3_test.go
```diff
@@ -144,14 +144,66 @@ func TestHStar3Prepare(t *testing.T) {
...
+		got := hs.Prepare()
+		for _, id := range got {
+			if !ids[id] {
+				t.Fatalf("not found: %v", id)
+			}
+			delete(ids, id)
+		}
+		if ln := len(ids); ln != 0 {
+			t.Fatalf("%d IDs left", ln)
+		}
```
pavelkalinnikov: Should just call `idsToMap` and check that they are `DeepEqual`.

merkle/smt/hstar3.go
```diff
@@ -70,32 +70,43 @@ func NewHStar3(updates []NodeUpdate, hash HashChildrenFn, depth, top uint) (HSta
...
+// the top level specified in the constructor. It may be useful for creating a
+// a NodeAccessor, e.g. by batch-reading the nodes from elsewhere.
 //
 // TODO(pavelkalinnikov): Return only tile IDs.
-func (h HStar3) Prepare() map[tree.NodeID2][]byte {
-	ids := make(map[tree.NodeID2][]byte)
+func (h HStar3) Prepare() []tree.NodeID2 {
+	empty := tree.NodeID2{}
+	ids := []tree.NodeID2{empty} // Start with a dummy sentinel value.
+	pos := make([]int, h.depth-h.top)
```
mhutchinson: The relationship between pos and ids is not immediately clear. Its meaning can be derived, but I'd like to see the loop invariant (https://en.wikipedia.org/wiki/Loop_invariant) stated. Things that I had to work out:
1) All elements in `pos` are effectively pointing at the empty sentinel value in `ids` to begin
2) You fill out `pos` from back-to-front
3) There will always be at least one element in `pos` which equals `len(ids)-1` (i.e. points to the last valid index where `ids[i]` is valid)
4) The whole reason that this trick with `pos` works is precisely because the nodes are already sorted left-to-right

merkle/smt/hstar3.go
```diff
@@ -70,32 +70,43 @@ func NewHStar3(updates []NodeUpdate, hash HashChildrenFn, depth, top uint) (HSta
...
+// load in order to compute node hash updates from the initial tree depth up to
+// the top level specified in the constructor. It may be useful for creating a
+// a NodeAccessor, e.g. by batch-reading the nodes from elsewhere.
 //
 // TODO(pavelkalinnikov): Return only tile IDs.
-func (h HStar3) Prepare() map[tree.NodeID2][]byte {
-	ids := make(map[tree.NodeID2][]byte)
+func (h HStar3) Prepare() []tree.NodeID2 {
+	empty := tree.NodeID2{}
+	ids := []tree.NodeID2{empty} // Start with a dummy sentinel value.
```
mhutchinson: May be worth considering initializing `ids` with a capacity >>> 1. In the worst case you need `len(h.upd) * (h.depth-h.top)` elements. May be worth pre-allocating that?

pavelkalinnikov: Yes! This decreased the benchmark running time from 30ms to 6ms.

merkle/smt/hstar3.go
```diff
@@ -70,32 +70,43 @@ func NewHStar3(updates []NodeUpdate, hash HashChildrenFn, depth, top uint) (HSta
...
 }
 
-// Prepare returns the set of all the node IDs that the Update method will load
-// in order to compute node hash updates from the initial tree depth up to the
-// top level specified in the constructor. It may be useful for constructing a
-// NodeAccessor, e.g. by batch-reading the nodes from elsewhere.
-//
-// Note: The returned map could have bool value type, but []byte allows the
-// caller to reuse this map for filling in the hashes for the Update method.
+// Prepare returns the list of all the node IDs that the Update method will
```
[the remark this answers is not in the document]
pavelkalinnikov: Modified the comment.

merkle/smt/hstar3.go
pavelkalinnikov: Done. From your list, the items 2 is irrelevant, and 3 needed a bit of generalization. LMK how that looks to you now.

Issue #1880, opened 2019-10-01
Proposed Map Hammer Model


mhutchinson:
The Map hammer (/testonly/hammer/hammer.go) kicks off two flavours of thread:
A) read-only checker thread
B) read/write thread

Any number of type A can be kicked off, and one of B is kicked off. In principle this shape will be kept, but some of the operations in B will be moved to A; currently A only performs GetAndVerifyMapLeavesByRevision with random revisions. Type B will perform a number of operations from SetLeaves to fetching SMRs.

More specifically, the proposal is to make:
- Type A use only the Map Read API and exercise the operations on there according to the biases in the configuration
- Type B to use only the Map Write API and perform work on there which more closely represents typical sequencer/write personality workflows

This change more accurately reflects what entities will do in the real world, and allows for scaling and testing the load of the write vs read characteristics far more cleanly.

mhutchinson:
TBD whether the main thread should only use operations from the write API. This better represents what happens in the real world, but it means that a hammer invocation with no checkers doesn't exercise many of the methods. I suggest changing the default checkers to 1 and making the main thread use only methods from the Write API.

PR #1886, opened 2019-10-03
Refactored the valid Map operations into a separate type

mhutchinson:
This is the first step to encapsulating many of the different things going on in the hammer, including:
 - definition of valid vs invalid operations
 - read vs write

This moves things in the direction of #1880

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -309,11 +309,12 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
 	var prevContents testonly.VersionedMapContents
 	var smrs smrStash
 	validReadOps := validReadOps{
-		mc:              mc,
-		extraSize:       cfg.ExtraSize,
-		chooseLeafCount: func(prng *rand.Rand) int { return pickIntInRange(cfg.MinLeaves, cfg.MaxLeaves, prng) },
-		prevContents:    &prevContents,
-		smrs:            &smrs,
+		mc:           mc,
+		extraSize:    cfg.ExtraSize,
+		minLeaves:    cfg.MinLeaves,
+		maxLeaves:    cfg.MaxLeaves,
+		prevContents: &prevContents,
+		smrs:         &smrs,
 	}
 
 	return &hammerState{
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -280,7 +280,7 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
 		return nil, fmt.Errorf("failed to get tree information: %v", err)
 	}
 	glog.Infof("%d: hammering tree with configuration %+v", cfg.MapID, tree)
-	vc, err := client.NewMapClientFromTree(cfg.Client, tree)
+	mc, err := client.NewMapClientFromTree(cfg.Client, tree)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get tree verifier: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


testonly/hammer/record.go
```diff
@@ -68,6 +68,7 @@ func (s *smrStash) pushSMR(smr types.MapRootV1) error {
 	return nil
 }
 
+// TODO(mhutchinson): consider making this return a bool or an error to indicate missing result.
```
pavelkalinnikov: nit: select(`consider making this `) -> Trash.

testonly/hammer/reader.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/golang/glog"
+	"github.com/google/trillian"
+	"github.com/google/trillian/client"
+	"github.com/google/trillian/testonly"
+	"github.com/google/trillian/types"
+)
+
+type intSelector func(prng *rand.Rand) int
+
+// validReadOps performs valid read operations against the map.
```
pavelkalinnikov: Are there invalid ones? Maybe rename to `readOps`?

testonly/hammer/hammer.go
```diff
@@ -309,11 +306,23 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
 		cfg.OperationDeadline = 60 * time.Second
 	}
 
+	var prevContents testonly.VersionedMapContents
+	var smrs smrStash
+	validReadOps := validReadOps{
+		vc:              vc,
+		extraSize:       cfg.ExtraSize,
+		chooseLeafCount: func(prng *rand.Rand) int { return pickIntInRange(cfg.MinLeaves, cfg.MaxLeaves, prng) },
```
pavelkalinnikov: Is this random int selector over-engineered? Both `hammerState` and `validReadOps` have a copy of this function (with the implementation that appears to be the same). Also, there is a type definition for it. But in the end, there is only one implementation that calls `pickIntInRange`. Maybe remove the type/method/function, and call `pickIntInRange` directly? For example, you could throw away the `extraSize` field in favor of a full `cfg`, and then use `cfg.Min/MaxLeaves` in `validReadOps`.

testonly/hammer/reader.go
```diff
@@ -0,0 +1,145 @@
...
+type validReadOps struct {
+	vc              *client.MapClient
```
pavelkalinnikov: This name is a bit outdated probably, as there is no "v" in the type name. How about `mc`?

testonly/hammer/hammer.go
```diff
@@ -412,8 +412,12 @@ func (s *hammerState) chooseInvalid(ep MapEntrypointName, prng *rand.Rand) bool
 }
 
 func (s *hammerState) chooseLeafCount(prng *rand.Rand) int {
-	delta := 1 + s.cfg.MaxLeaves - s.cfg.MinLeaves
-	return s.cfg.MinLeaves + prng.Intn(delta)
+	return pickIntInRange(s.cfg.MinLeaves, s.cfg.MaxLeaves, prng)
```
pavelkalinnikov: Since this method now seems to be used only once in `hammerState`, I suggest inlining it.

testonly/hammer/record.go
mhutchinson: Agree.

testonly/hammer/reader.go
mhutchinson: There will be invalid ones! See performInvalidOp on hammer.go. The plan is to have one interface for performing readOps, and two implementations (one that is valid, and one that is invalid). We could even have a third implementation which randomly chooses between valid vs invalid, which then sucks even more of the logic of how to do things out of the main hammer.

This invalidReadOps will be the next PR.

testonly/hammer/reader.go
mhutchinson: Done.

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: Part of the motivation behind this PR is to encapsulate better. Anything with a reference to a `MapConfig` effectively has access to the whole world. That is too big of a handbag and makes it hard to reason about what code is up to.

I could pass in a min/max instead of this function, but this way kinda naturally happened as a way of making the intent of the read operations as clear as possible. Happy to pass in min/max and copy the impl of pickIntInRange into validReadOps if that is clearer.

pavelkalinnikov: Yeah, I think it would be clearer.

In the future maybe it would make sense to create sub-configs of `cfg`, and copy the corresponding sub-config into `validReadOps`.

PR #1888, opened 2019-10-03
Refactor invalid ops into separate class

mhutchinson:
This finishes off what was hinted at by the work started in #1886. This removes logic from the hammer class, and removes duplication of switching.

The next stage will be to break out the write functionality so that it is only called by the main thread, and the read methods will only be used by the checkers threads. This heads towards the design in #1880.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -498,9 +498,9 @@ type readOps interface {
 	getSMRRev(context.Context, *rand.Rand) error
 }
 
-type setLeaves func(context.Context, *rand.Rand) error
+type setLeavesFn func(context.Context, *rand.Rand) error
 
-func performOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand, read readOps, write setLeaves) error {
+func performOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand, read readOps, write setLeavesFn) error {
 	switch ep {
 	case GetLeavesName:
 		return read.getLeaves(ctx, prng)
@@ -514,7 +514,7 @@ func performOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand, read 
 		// TODO(mhutchinson): This mutation method needs to be removed from here.
 		return write(ctx, prng)
 	default:
-		return fmt.Errorf("internal error: unknown entrypoint %s selected for operation", ep)
+		return fmt.Errorf("internal error: unknown operation %s", ep)
 	}
 }
 
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -511,13 +511,13 @@ func (s *hammerState) performOp(ctx context.Context, ep MapEntrypointName, prng 
 func (s *hammerState) performInvalidOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand) error {
 	switch ep {
 	case GetLeavesName:
-		return s.invalidReadOps.getLeavesInvalid(ctx, prng)
+		return s.invalidReadOps.getLeaves(ctx, prng)
 	case GetLeavesRevName:
-		return s.invalidReadOps.getLeavesRevInvalid(ctx, prng)
+		return s.invalidReadOps.getLeavesRev(ctx, prng)
 	case GetSMRName:
-		return s.invalidReadOps.getSMRInvalid(ctx, prng)
+		return s.invalidReadOps.getSMR(ctx, prng)
 	case GetSMRRevName:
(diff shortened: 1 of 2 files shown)


mhutchinson:
@pavelkalinnikov I recommend reviewing commit by commit. I learned this from you and I think it works very well for compound refactoring such as this.

testonly/hammer/hammer.go
```diff
@@ -254,8 +254,9 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 
 // hammerState tracks the operations that have been performed during a test run.
 type hammerState struct {
-	cfg          *MapConfig
-	validReadOps *validReadOps
+	cfg            *MapConfig
+	validReadOps   *validReadOps
```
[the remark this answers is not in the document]
pavelkalinnikov: Oh I see the `SetLeaves` thing below, so that's probably yes?

testonly/hammer/hammer.go
```diff
@@ -491,37 +491,30 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
...
-		return s.validReadOps.getSMR(ctx, prng)
-	case GetSMRRevName:
-		return s.validReadOps.getSMRRev(ctx, prng)
-	case SetLeavesName:
-		return s.setLeaves(ctx, prng)
-	default:
-		return fmt.Errorf("internal error: unknown entrypoint %s selected for valid request", ep)
-	}
+type readOps interface {
+	getLeaves(context.Context, *rand.Rand) error
```
pavelkalinnikov: I wonder if `rand.Rand` should be baked into the implementation of the type rather than the interface. I'm guessing that the caller uses the same instance of `Rand` for all these calls anyway. The fact of re-creating it before each call complicates this a bit maybe...

testonly/hammer/hammer.go
```diff
@@ -491,37 +491,30 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
...
+	getLeavesRev(context.Context, *rand.Rand) error
+	getSMR(context.Context, *rand.Rand) error
+	getSMRRev(context.Context, *rand.Rand) error
 }
 
-func (s *hammerState) performInvalidOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand) error {
+type setLeaves func(context.Context, *rand.Rand) error
```
pavelkalinnikov: nit: `setLeavesFn` or smth, to make it not look like an actual function name.

testonly/hammer/hammer.go
```diff
@@ -514,7 +514,7 @@ func performOp(ctx context.Context, ep MapEntrypointName, prng *rand.Rand, read
 		// TODO(mhutchinson): This mutation method needs to be removed from here.
 		return write(ctx, prng)
 	default:
-		return fmt.Errorf("internal error: unknown entrypoint %s selected for valid request", ep)
+		return fmt.Errorf("internal error: unknown entrypoint %s selected for operation", ep)
```
pavelkalinnikov: nit to the pre-existing code: Why such a long error text? `"unknown entrypoint %s"` would be totally enough.

testonly/hammer/hammer.go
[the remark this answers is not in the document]
mhutchinson: It is indeed a yes. It could be that further refactoring will actually separate this such that write and read ops are never near each other and thus the "read" token in here seems spurious, but that's not the case at the moment so will leave for now.

testonly/hammer/hammer.go
mhutchinson: I initially thought this too, but in the event of a retry being required it actually passes in a prng that returns the same sequence as before for predictability during the retry. This needs to stay as a parameter for now. See `retryOneOp` implementation for where this prng regeneration is performed.

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: Agree. Shortened more, but changing this too much will make it inconsistent with the other long errors and then we'll be on a magical whirlwind adventure to get consistency again.

testonly/hammer/reader.go
```diff
@@ -155,3 +156,91 @@ func (o *validReadOps) verify(root *types.MapRootV1) error {
 	}
 	return nil
 }
+
+// invalidReadOps performs invalid read operations against the map.
+type invalidReadOps struct {
+	mapID        int64
+	client       trillian.TrillianMapClient
```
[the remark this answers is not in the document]
mhutchinson: This client is the raw gRPC client. The other client (previously called vc to hint at validating client) does validation of the inputs and responses. This makes it particularly bad for performing invalid requests.

PR #1889, opened 2019-10-03
Refactor: Choose the operation separately to invoking it

mhutchinson:
This allows the retryOp method to be more easily reused between the write and read.

The next step is to move the main thread to only perform writes and the checkers to only perform reads, according to a random selection. This requires even more refactoring though, as the hammerState.prng is only supposed to be used by the main thread, but it can be accessed from anywhere that has hammerState. If I switch this over now, the tests will stop being deterministic given a seed.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -443,28 +443,33 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
 	}
 
 	glog.V(3).Infof("%d: perform %s operation", s.cfg.MapID, ep)
-	defer func(start time.Time) {
-		rspLatency.Observe(time.Since(start).Seconds(), s.label(), string(ep))
-	}(time.Now())
 
-	deadline := time.Now().Add(s.cfg.OperationDeadline)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	var firstErr error
+	return s.retryOp(ctx, op, string(ep))
+}
+
+func (s *hammerState) retryOp(ctx context.Context, fn mapOperationFn, opName string) error {
+	defer func(start time.Time) {
+		rspLatency.Observe(time.Since(start).Seconds(), s.label(), opName)
+	}(time.Now())
+
+	deadline := time.Now().Add(s.cfg.OperationDeadline)
 	seed := s.prng.Int63()
 	done := false
+	var firstErr error
 	for !done {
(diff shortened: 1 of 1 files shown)


testonly/hammer/hammer.go
```diff
@@ -430,32 +430,46 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
...
 	defer cancel()
 
-	var firstErr error
+	return s.retryOp(ctx, op, string(ep))
+}
+
+func (s *hammerState) retryOp(ctx context.Context, fn mapOperationFn, opName string) error {
+	defer func(start time.Time) {
+		rspLatency.Observe(time.Since(start).Seconds(), s.label(), opName)
+	}(time.Now())
```
pavelkalinnikov: What a nice hack (that somebody else gets the credit for :D).

testonly/hammer/hammer.go
```diff
@@ -430,32 +430,46 @@ func (s *hammerState) retryOneOp(ctx context.Context) (err error) {
...
+		return err
 	}
 
 	glog.V(3).Infof("%d: perform %s operation", s.cfg.MapID, ep)
-	defer func(start time.Time) {
-		rspLatency.Observe(time.Since(start).Seconds(), s.label(), string(ep))
-	}(time.Now())
 
-	deadline := time.Now().Add(s.cfg.OperationDeadline)
 	ctx, cancel := context.WithCancel(ctx)
```
pavelkalinnikov: This canceling code looks unreasonable here. Probably it should be moved to `retryOp`, as there is some manual deadlinery over there.

mhutchinson: Done

testonly/hammer/hammer.go
mhutchinson: :-D

testonly/hammer/hammer.go
```diff
@@ -455,6 +451,9 @@ func (s *hammerState) retryOp(ctx context.Context, fn mapOperationFn, opName str
 		rspLatency.Observe(time.Since(start).Seconds(), s.label(), opName)
 	}(time.Now())
 
+	ctx, cancel := context.WithCancel(ctx)
```
pavelkalinnikov: Hm, by the look of it, `cancel()` in the end of this function doesn't make any difference, does it?

mhutchinson: I don't think it makes any difference, but wanted to keep this a pure refactoring PR.

I'll follow up with a wee tidy.

PR #1891, opened 2019-10-04
Refactor: removed PRNG from hammerState

mhutchinson:
This PR introduces a mapWorker class which has its own PRNG and will be scoped to a single threaded worker. I took the opportunity to move a number of methods from hammerState to mapWorker too. The goal is eventually that each worker would only have the local state it needs, plus the explicit thread-safe objects for sharing information - effectively an in-memory gossip network.

This takes us closer to the design in #1880.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -220,7 +220,7 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		w := newWorker(s.cfg, rand.New(s.cfg.RandSource))
+		w := newWorker(&cfg, rand.New(cfg.RandSource))
 		glog.Infof("%d: start main goroutine", cfg.MapID)
 		count, err := w.performOperations(ctx, done, s)
 		errs <- err // may be nil for the main goroutine completion
@@ -253,14 +253,19 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 	return firstErr
 }
 
-// mapWorker allows each thread to have its own PRNG, which makes it easy to
-// reason about the behaviour of each thread being internally consistent and
-// reproducible given a seed. Obviously threads that depend on the state of
-// the map may still have non-deterministic behaviour even when replayed from
-// the same seed.
+// mapWorker represents a single entity in the Verifiable Map ecosystem.
+// The worker may be a read-only client, or a writer which adds new entries to
+// the map. Each worker should be as independent as possible (i.e. share little
+// to no state), though through well defined interfaces this guideline may be
+// ignored, which will allow effectively an in-memory gossip network to develop
+// between workers, which makes the validation more significant.
+//
+// Each worker should has its own PRNG, which makes the sequence of operations
+// that it will perform deterministic.
 type mapWorker struct {
 	prng  *rand.Rand
 	mapID int64
+	label string
 
 	bias MapBias // Each worker can have its own customized map bias
 
@@ -272,14 +277,11 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
 	return &mapWorker{
 		prng:              prng,
 		mapID:             cfg.MapID,
+		label:             strconv.FormatInt(cfg.MapID, 10),
 		bias:              cfg.EPBias,
 		retryErrors:       cfg.RetryErrors,
 		operationDeadline: cfg.OperationDeadline,
 	}
-}
-
-func (w *mapWorker) label() string {
-	return strconv.FormatInt(w.mapID, 10)
 }
 
 // hammerState tracks the operations that have been performed during a test run.
@@ -360,8 +362,8 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
 	}, nil
 }
 
-// TODO(mhutchinson): Remove hammerState from here - it allows access to too much global info
-// which makes reasoning about the behaviour too difficult.
+// TODO(mhutchinson): Remove hammerState from here - it allows access to global info
+// which makes reasoning about the behaviour difficult.
 func (w *mapWorker) performOperations(ctx context.Context, done <-chan struct{}, s *hammerState) (uint64, error) {
 	count := uint64(0)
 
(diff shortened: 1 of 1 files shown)


testonly/hammer/hammer.go
```diff
@@ -220,8 +220,9 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		glog.Infof("%d: start main goroutine", s.cfg.MapID)
-		count, err := s.performOperations(ctx, done)
+		w := newWorker(s.cfg, rand.New(s.cfg.RandSource))
+		glog.Infof("%d: start main goroutine", cfg.MapID)
```
pavelkalinnikov: Is `s.cfg` the same thing as `cfg`? Use one or the other consistently?

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,35 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
...
+	return &mapWorker{
+		prng:              prng,
+		mapID:             cfg.MapID,
+		bias:              cfg.EPBias,
+		retryErrors:       cfg.RetryErrors,
+		operationDeadline: cfg.OperationDeadline,
+	}
+}
+
+func (w *mapWorker) label() string {
```
pavelkalinnikov: Consider storing this as `mapWorker.label` field in the constructor.

testonly/hammer/hammer.go
```diff
@@ -327,23 +353,25 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
...
 		prevContents:   &prevContents,
 		smrs:           &smrs,
 		validReadOps:   &validReadOps,
 		invalidReadOps: &invalidReadOps,
 	}, nil
 }
 
-func (s *hammerState) performOperations(ctx context.Context, done <-chan struct{}) (uint64, error) {
+// TODO(mhutchinson): Remove hammerState from here - it allows access to too much global info
+// which makes reasoning about the behaviour too difficult.
```
pavelkalinnikov: nit: This sentence is a bit dramatic :D Nothing should be **too** difficult to Martin :)

testonly/hammer/hammer.go
```diff
@@ -412,60 +440,60 @@ func (s *hammerState) String() string {
...
-func (s *hammerState) chooseOp(prng *rand.Rand) MapEntrypointName {
-	return s.cfg.EPBias.choose(prng)
+func pickIntInRange(min, max int, prng *rand.Rand) int {
+	delta := 1 + max - min
+	return min + prng.Intn(delta)
 }
 
-func (s *hammerState) chooseInvalid(ep MapEntrypointName, prng *rand.Rand) bool {
-	return s.cfg.EPBias.invalid(ep, prng)
+func (w *mapWorker) chooseOp() MapEntrypointName {
```
pavelkalinnikov: These one-line methods are a bit annoying, esp. given that they are used **once** and from within the same type.
Consider inlining them here or in a separate PR.

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,35 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 	return firstErr
 }
 
+// mapWorker allows each thread to have its own PRNG, which makes it easy to
```
pavelkalinnikov: nit: "thread" sounds a bit too generic, and is also not a Golang term, so triggers "what's that"s. Do you have a mental model of what `mapWorker` will do in follow-ups? The "worker" term might be good enough already.

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,35 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
...
+// reason about the behaviour of each thread being internally consistent and
+// reproducible given a seed. Obviously threads that depend on the state of
+// the map may still have non-deterministic behaviour even when replayed from
+// the same seed.
```
pavelkalinnikov: So far this comment looks like the only purpose of `mapWorker` is this PRNG thing. Consider framing the comment in terms of what the worker does.

testonly/hammer/hammer.go
mhutchinson: I thought I'd already fixed all these occurrences. I did a lot of back and forth at around this point of the refactoring so I probably undid it. Good spot.

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: Redone this comment. I wrote this first mostly to justify it to myself I guess :-)

testonly/hammer/hammer.go
mhutchinson: There's nothing like a bit of **DRAMA** to make a comment more interesting. No?

Toned it down.

testonly/hammer/hammer.go
mhutchinson: Agree. I see your refactoring style uses a big hammer than mine. I kinda like it.

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,37 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
...
 }
 
+// mapWorker represents a single entity in the Verifiable Map ecosystem.
+// The worker may be a read-only client, or a writer which adds new entries to
+// the map. Each worker should be as independent as possible (i.e. share little
+// to no state), though through well defined interfaces this guideline may be
+// ignored, which will allow effectively an in-memory gossip network to develop
+// between workers, which makes the validation more significant.
+//
+// Each worker should has its own PRNG, which makes the sequence of operations
```
pavelkalinnikov: delete `should`

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,37 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
...
+// that it will perform deterministic.
```
pavelkalinnikov: nit: `will perform` -> `performs`

testonly/hammer/hammer.go
```diff
@@ -252,6 +253,37 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
...
+type mapWorker struct {
+	prng  *rand.Rand
+	mapID int64
+	label string
+
+	bias MapBias // Each worker can have its own customized map bias
```
pavelkalinnikov: nit: Add period.

testonly/hammer/hammer.go

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: That was close. Thanks!

PR #1893, opened 2019-10-07
Refactor: extracted readWorker and writeWorker

mhutchinson:
Reads and writes are logically distinct entities in the Map ecosystem. This refactor makes that explicit.

The next step is to make the read worker perform random read operations instead of just getLeavesRev. The next next step will be to make the write thread only perform writes.

This shapes things up in line with #1880.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -210,7 +210,7 @@ func HitMap(ctx context.Context, cfg MapConfig) error {
 			defer wg.Done()
 			w := newReadWorker(s, i)
 			glog.Infof("%d: start checker %d", s.cfg.MapID, i)
-			err := w.loopUntilDone(ctx, done)
+			err := w.loop(ctx, done)
 			if err != nil {
 				errs <- err
 			}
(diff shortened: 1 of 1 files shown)


testonly/hammer/hammer.go
```diff
@@ -284,6 +285,75 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
...
+		select {
+		case <-done:
+			return count, nil
+		default:
+		}
+		if err := w.retryOneOp(ctx, w.s); err != nil {
+			return count, err
+		}
+	}
+	return count, nil
```
pavelkalinnikov: It seems that we get to this line only when `count == w.operations`. How about `return w.operations, nil` and inlining `count := uint64(0)` to the `for` statement?

testonly/hammer/hammer.go
```diff
@@ -284,6 +285,75 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
...
+func newWriteWorker(s *hammerState) *writeWorker {
+	return &writeWorker{
+		mapWorker: newWorker(s.cfg, rand.New(s.cfg.RandSource)),
+
+		operations: s.cfg.Operations,
+		s:          s,
+	}
+}
+
+func (w *writeWorker) loop(ctx context.Context, done <-chan struct{}) (uint64, error) {
```
pavelkalinnikov: How about `run`?

testonly/hammer/hammer.go
```diff
@@ -284,6 +285,75 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
...
+func newReadWorker(s *hammerState, idx int) *readWorker {
+	return &readWorker{
+		mapWorker: newWorker(s.cfg, rand.New(rand.NewSource(int64(idx)))),
+
+		validReadOps:   s.validReadOps,
+		invalidReadOps: s.invalidReadOps,
+	}
+}
+
+func (w *readWorker) loopUntilDone(ctx context.Context, done <-chan struct{}) error {
```
pavelkalinnikov: How about just `run`, same as in `writeWorker`?

testonly/hammer/hammer.go
```diff
@@ -284,6 +285,75 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
 	}
 }
 
+// readWorker performs read-only operations on a fixed map.
+type readWorker struct {
+	*mapWorker
+
```
pavelkalinnikov: optional nit: I'd not bother with empty lines here and in the constructor. Passim for `writeWorker`.

testonly/hammer/hammer.go
```diff
@@ -284,6 +285,75 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
...
+func (w *readWorker) loopUntilDone(ctx context.Context, done <-chan struct{}) error {
+	for {
+		select {
+		case <-done:
+			return nil
+		default:
+		}
+		if err := w.validReadOps.getLeavesRev(ctx, w.prng); err != nil {
+			if _, ok := err.(errSkip); ok {
```
pavelkalinnikov: nit to the previous code: I'd define `var errSkip = errors.New("skip something something")` and just compare it for equality.

testonly/hammer/hammer.go
```diff
@@ -338,6 +416,9 @@ func newWriteWorker(s *hammerState) *writeWorker {
 	}
 }
 
+// loop continuously performs mutation operations on the map until the done signal
+// is received via th channel, an error is encountered, or the maximum number of
```
pavelkalinnikov: simply: `... until the done channel is closed, an error ...`
Same for `readWorker`.

testonly/hammer/hammer.go
```diff
@@ -285,6 +285,82 @@ func newWorker(cfg *MapConfig, prng *rand.Rand) *mapWorker {
 	}
 }
 
+func (w *mapWorker) retryOneOp(ctx context.Context, s *hammerState) (err error) {
```
pavelkalinnikov: Just to double check: Have `retry[One]Op` methods been moved without modification here?

mhutchinson: Yup, they are just a move. I did them in a second commit so that would be clearer, but unfortunately the diff tool here isn't smart enough to show a reordering move clearly.

testonly/hammer/hammer.go
mhutchinson: I'm exercising the "do nothing" branch of "optional" here - it makes it clearer to me what is going on when the the linebreak is kept. My attitude will likely change as the Go embedding pattern instead of a more classical inheritance pattern becomes normal to me.

testonly/hammer/hammer.go
mhutchinson: errSkip is used in quite a few places and is very orthogonal to this PR. Will leave for now. I may well come through and change up `errSkip` and `testonly.ErrInvariant` which have limited/zero uses. But this is not the time nor place.

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: Yup.

testonly/hammer/hammer.go
mhutchinson: I think I prefer it the way it is. This way is clear - I like the consistency that every return is `return count, ...` - and I'm not too concerned with holding onto a single uint64 for an extra beat.

testonly/hammer/hammer.go
mhutchinson: Resolved.

PR #1896, opened 2019-10-07
Add and use NodeID/NodeID2 conversion

pavelkalinnikov:
This PR adds conversion between `NodeID` and `NodeID2`. It is necessary to integrate new sparse Merkle tree `Writer` on `HStar3` without breaking `storage` API (yet).

### DIFF
--- a/storage/tree/node.go
+++ b/storage/tree/node.go
@@ -201,6 +201,23 @@ loop:
 	}
 }
 
+// NewNodeIDFromID2 constructs a NodeID from a NodeID2.
+func NewNodeIDFromID2(id NodeID2) NodeID {
+	bytes := id.FullBytes()
+	last, bits := id.LastByte()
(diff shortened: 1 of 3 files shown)


storage/tree/node.go
```diff
@@ -201,6 +201,23 @@ loop:
 	}
 }
 
+// NewNodeIDFromID2 constructs a NodeID from a NodeID2.
+func NewNodeIDFromID2(id NodeID2) NodeID {
+	bytes := id.FullBytes()
+	last, bits := id.LastByte()
+	path := make([]byte, len(bytes)+int(bits+7)/8)
+	copy(path, bytes)
```
mhutchinson: You *could* use the result of copy to determine whether the last byte needs filling in. Or not. I'm not a cop.

pavelkalinnikov: I found a shorter way.

PR #1897, opened 2019-10-07
map_rpc_server: Simplify makeSignedMapRoot

pavelkalinnikov:
This change:
- Removes the unused `mapID` parameter from `makeSignedMapRoot`.
- Inlines `time.Now` computation into it, because all the callers do that.

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -311,10 +311,9 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
 
-	mapID := req.MapId
-	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(mapID, 10))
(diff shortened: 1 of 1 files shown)


server/map_rpc_server.go
```diff
@@ -310,10 +310,9 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
 
-	mapID := req.MapId
-	t.setLeafCounter.Add(float64(len(req.Leaves)), string(mapID))
+	t.setLeafCounter.Add(float64(len(req.Leaves)), string(req.MapId))
```
pavelkalinnikov: This line is weird. I don't think that `string` conversion of `int64` works as intended here.

pavelkalinnikov: https://play.golang.org/p/SnORcuFf2XU

pavelkalinnikov: Fixed.

pavelkalinnikov: Factored out #1898

PR #1899, opened 2019-10-07
Added extra checks for bad states

mhutchinson:
This makes the code do what the comment implied was going to happen. Any missing revisions would be bad news given how these previous contents are used for checking map correctness.

None of this is triggered as the hammer is currently set up, but you may be surprised to learn that making the hammer properly concurrent will trigger some of these.

Also took the opportunity to simplify a little code that impaired readability.

### DIFF
--- a/testonly/hammer/reader.go
+++ b/testonly/hammer/reader.go
@@ -115,10 +115,7 @@ func (o *validReadOps) getSMR(ctx context.Context, prng *rand.Rand) error {
 	}
 	glog.V(2).Infof("%d: got SMR(time=%q, rev=%d)", o.mc.MapID, time.Unix(0, int64(root.TimestampNanos)), root.Revision)
 
-	if err := o.verify(root); err != nil {
-		return err
-	}
-	return nil
+	return o.verify(root)
 }
 
 // getSMRRev randomly chooses a previously seen SMR from the queue and checks that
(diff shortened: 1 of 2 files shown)


testonly/hammer/reader.go
```diff
@@ -115,10 +115,7 @@ func (o *validReadOps) getSMR(ctx context.Context, prng *rand.Rand) error {
 	}
 	glog.V(2).Infof("%d: got SMR(time=%q, rev=%d)", o.mc.MapID, time.Unix(0, int64(root.TimestampNanos)), root.Revision)
 
-	if err := o.verify(root); err != nil {
-		return err
-	}
-	return nil
+	return o.verify(root)
```
pavelkalinnikov: All kinds of thumbs.

testonly/mapcontents.go
```diff
@@ -299,11 +299,16 @@ func (p *VersionedMapContents) UpdateContentsWith(rev uint64, leaves []*trillian
...
 		return nil, ErrInvariant{fmt.Sprintf("got rev %d, want >=1 when trying to update hammer state with contents", rev)}
 	}
-	if p.contents[0] != nil && int64(rev) <= p.contents[0].Rev {
-		return nil, ErrInvariant{fmt.Sprintf("got rev %d, want >%d when trying to update hammer state with new contents", rev, p.contents[0].Rev)}
+	if p.contents[0] == nil {
+		if rev != 1 {
+			return nil, ErrInvariant{fmt.Sprintf("got rev %d, want 1 when trying to update hammer state with new contents", rev)}
+		}
+	} else if want := p.contents[0].Rev + 1; int64(rev) != want {
+		return nil, ErrInvariant{fmt.Sprintf("got rev %d, want %d when trying to update hammer state with new contents", rev, want)}
```
pavelkalinnikov: To avoid the long `ErrInvariant` duplicate, could do this:
```
nextRev := 1
if c := p.contents[0]; c != nil {
	nextRev = c.Rev + 1
}
if rev != nextRev {
	return nil, ErrInvariant{...}
}
```

PR #1901, opened 2019-10-08
Refactor: read worker uses MapBias, but maintains same behaviour

mhutchinson:
It is clear from this point how the read worker will be changed to perform other read-only operations. This is a refactor and does not change behaviour except in one regard: the counter metrics for total operations performed are now incremented by the checker threads too.

Surely this is the last refactor before #1880 can be reached? :-)

In seriousness, there is a change that needs to happen to the data structures which maintain previous contents of the map and the SMRs to make them thread-safe. This will happen in a coming PR.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -17,6 +17,7 @@ package hammer
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"strconv"
@@ -80,7 +81,11 @@ const (
 	GetSMRRevName    = MapEntrypointName("GetSMRRev")
 )
 
-var mapEntrypoints = []MapEntrypointName{GetLeavesName, GetLeavesRevName, SetLeavesName, GetSMRName, GetSMRRevName}
+// Read-only map entry points
+var roMapEntrypoints = []MapEntrypointName{GetLeavesName, GetLeavesRevName, GetSMRName, GetSMRRevName}
+
+// All map entry points
+var mapEntrypoints = append(roMapEntrypoints, SetLeavesName)
 
 // Choice is a readable representation of a choice about how to perform a hammering operation.
 type Choice string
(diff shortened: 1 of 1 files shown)


testonly/hammer/hammer.go
```diff
@@ -80,7 +81,11 @@ const (
 	GetSMRRevName    = MapEntrypointName("GetSMRRev")
 )
 
-var mapEntrypoints = []MapEntrypointName{GetLeavesName, GetLeavesRevName, SetLeavesName, GetSMRName, GetSMRRevName}
+// Read-only map entry points
```
pavelkalinnikov: nit: Period. Passim.

testonly/hammer/hammer.go
```diff
@@ -370,8 +375,14 @@ type readWorker struct {
 }
 
 func newReadWorker(s *hammerState, idx int) *readWorker {
+	readBias := MapBias{
+		Bias:          make(map[MapEntrypointName]int),
+		InvalidChance: make(map[MapEntrypointName]int),
+	}
+	// TODO(mhutchinson): populate readBias by iterating over roMapEntrypoints
```
pavelkalinnikov: This seems like 3 lines of code, so maybe do it in this PR?

testonly/hammer/hammer.go
```diff
@@ -409,7 +443,7 @@ type writeWorker struct {
 
 func newWriteWorker(s *hammerState) *writeWorker {
 	return &writeWorker{
-		mapWorker: newWorker(s.cfg, rand.New(s.cfg.RandSource)),
+		mapWorker: newWorker(s.cfg, s.cfg.EPBias, rand.New(s.cfg.RandSource)),
```
pavelkalinnikov: Does this bias not need filtering out read-only stuff?

testonly/hammer/hammer.go

testonly/hammer/hammer.go
mhutchinson: No, because it changes behaviour in such a way that it causes races on the data structures, and we will start querying for non-existent revisions of the map, or finding revisions of the map which we haven't stored the local copy of yet.

For reference, I did first have it in the branch this came from, but had to separate this functional change out to identify the cause of the new failures.

testonly/hammer/hammer.go
mhutchinson: It will eventually. This PR is a refactoring that enables these functional changes. There are still underlying broken things about the local copies of the data structures that need resolving before we can change this though.

PR #1902, opened 2019-10-08
Replace legacy sparse Merkle tree writer

pavelkalinnikov:
This change removes `SparseMerkleTreeWriter`, and replaces it with a faster and clearer `smt.Writer`.

The key design decision is to keep sharding/concurrency of `smt.Writer` on the highest possible level (the application). It is unlikely that we would want to move it to `storage`, as we might want to extend `map_tree_updater.go` to a distributed setup, which is easier to do on the application level. Another advantage is that `map_tree_updater.go` is not exported, which means we can change this logic more easily.


pavelkalinnikov:
Please take a high-level look at this.

server/map_tree_updater.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	writeRev, err := tx.WriteRevision(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(pavelkalinnikov): Return an error when no updates are given.
+	if len(upd) == 0 {
+		// Get the root node for the read revision.
+		nodes, err := tx.GetMerkleNodes(ctx, readRev, []tree.NodeID{tree.NodeID{}})
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		nodes, err := tx.GetMerkleNodes(ctx, readRev, []tree.NodeID{{}})
```

server/map_rpc_server.go
```diff
@@ -346,8 +347,15 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 			return err
 		}
 
-		newRoot, err = t.updateTree(ctx, tree, hasher, tx, hkv, req.Metadata, writeRev)
-		return err
+		updater := &mapTreeUpdater{tree: tree, hasher: hasher, opts: t.opts, reg: t.registry}
```
pavelkalinnikov: This could be created outside of the TX.

server/map_tree_updater.go
```diff
@@ -0,0 +1,198 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
pavelkalinnikov: 2019

pavelkalinnikov:
I can probably split this into more phases. The phase 1 would be factoring out the `mapTreeWriter` type without swapping the SMT implementation.

server/map_tree_updater.go
```diff
@@ -0,0 +1,198 @@
...
+type mapTreeUpdater struct {
+	tree   *trillian.Tree
+	hasher hashers.MapHasher
+	opts   TrillianMapServerOptions
+	reg    extension.Registry
+}
+
+// update updates the sparse Merkle tree at the write revision of tx with the
+// given leaf changes, and writes it to the storage. Returns the new root hash.
+func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, upd []smt.NodeUpdate) ([]byte, error) {
```
gdbelvin: Wish list: the ability to be able to express a that these updates will be happening at an expected revision, and that no prior writes have occurred at this revision.

server/map_tree_updater.go
```diff
@@ -0,0 +1,198 @@
...
+	readRev, err := tx.ReadRevision(ctx)
+	if err != nil {
+		return nil, err
+	}
+	writeRev, err := tx.WriteRevision(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(pavelkalinnikov): Return an error when no updates are given.
```
gdbelvin: It is a desired feature to be able to advance the revision number with no changes. 
IE, this updates the signed map root timestamp

pavelkalinnikov: Right, but I wonder whether this deserves a different endpoint. Root signing seems like an orthogonal thing to tree building. It feels artificial to reuse `SetLeaves` RPC for that purpose.

mhutchinson: The API is staying as-is for the moment. We're working on improving performance within the bounds of the current API. If this causes performance problems then we can consider changing it, but otherwise we'll revisit this when we consider what shape the Map API should take when we move this from experimental.

server/map_tree_updater.go
pavelkalinnikov: Done to an extent: Now this accepts a `writeRev`, which is also verified to the bone in #1904.

server/map_tree_updater.go
```diff
@@ -38,120 +38,152 @@ type mapTreeUpdater struct {
 
 // update updates the sparse Merkle tree at the passed-in revision with the
 // given leaf updates, and writes it to the storage. Returns the new root hash.
-func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, hkv []merkle.HashKeyValue, rev int64) ([]byte, error) {
-	// Work around a performance issue when using the map in
-	// single-transaction mode by preloading all the nodes we know the
-	// sparse Merkle writer is going to need.
+func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, upd []smt.NodeUpdate, writeRev int64) ([]byte, error) {
+	// TODO(pavelkalinnikov): Return an error when no updates are given.
```
pavelkalinnikov: Another option is to move this [annoying] clause to `map_rpc_server.go`. The closer it is to the API level the better.

pavelkalinnikov: Done.

gdbelvin:
Is this still WIP?  If so, please remove the reviewer attention set

gdbelvin:
In fact, could I suggest closing the PR and re-opening it with Github's new "WIP" mode for PRs?

pavelkalinnikov:
This is in a final shape. To be merged after another PR with tests ensuring that it doesn't break stuff.

server/map_tree_updater.go
```diff
@@ -38,120 +38,140 @@ type mapTreeUpdater struct {
...
+	}
+
+	var mu sync.Mutex // Guards splitUpd.
+	splitUpd := make([]smt.NodeUpdate, 0, 256)
+
+	g, gCtx := errgroup.WithContext(ctx)
+	for _, upd := range shards {
+		upd := upd
+		g.Go(func() error {
+			rootUpd, err := update(gCtx, upd)
```
AlCutter: rootUpd -> shardRootUpd?

pavelkalinnikov: Done.

pavelkalinnikov:
This PR is now well covered by tests after #1937 and #1938.

server/map_tree_updater.go
```diff
@@ -38,114 +38,135 @@ type mapTreeUpdater struct {
...
+			shardUpds = append(shardUpds, shardRootUpd)
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
-	rootHash, err := smtWriter.CalculateRoot(ctx)
+	rootUpd, err := update(ctx, shardUpds)
```
AlCutter: Strictly, I suppose you should lock `mu` for the `shardUpds` read.

server/map_tree_updater.go
```diff
@@ -38,114 +38,135 @@ type mapTreeUpdater struct {
...
+			acc := &txAccessor{tx: tx, rev: writeRev}
+			var err error
+			rootUpd, err = w.Write(ctx, updCopy, acc)
+			return err
+		})
+		return rootUpd, err
+	}
+
+	var mu sync.Mutex // Guards shardUpds.
+	shardUpds := make([]smt.NodeUpdate, 0, 256)
```
gdbelvin: 256 should be tied to the split variable

server/map_tree_updater.go
```diff
@@ -38,114 +38,135 @@ type mapTreeUpdater struct {
...
 	if t.singleTX && t.preload {
-		if err := doPreload(ctx, tx, t.hasher.BitLen(), hkv, rev-1); err != nil {
+		if err := doPreload(ctx, tx, uint(t.hasher.BitLen()), upd, writeRev-1); err != nil {
 			return nil, err
 		}
 	}
 
-	smtWriter, err := merkle.NewSparseMerkleTreeWriter(ctx, t.tree.TreeId, rev, t.hasher, t.newTXRunner(t.tree, tx))
+	// TODO(pavelkalinnikov): Make the layout configurable.
+	w := smt.NewWriter(t.tree.TreeId, t.hasher, uint(t.hasher.BitLen()), 8)
```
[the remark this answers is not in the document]
gdbelvin: make 8 a defined variable

server/map_tree_updater.go
```diff
@@ -38,114 +38,135 @@ type mapTreeUpdater struct {
...
+	shards, err := w.Split(upd)
 	if err != nil {
 		return nil, err
 	}
 
-	if err = smtWriter.SetLeaves(ctx, hkv); err != nil {
+	runTX := t.newTXFunc(tx)
+	update := func(ctx context.Context, upd []smt.NodeUpdate) (smt.NodeUpdate, error) {
```
gdbelvin: This deserves a comment about writing one shard
Return vars could also be named, this would save a few var declarations

server/map_tree_updater.go
pavelkalinnikov: Nice argument! Although I think there is already a memory barrier in `g.Wait()` which guarantees happens-before between this line and everything that's in `g.Go(...)`.

pavelkalinnikov: I put a corresponding comment about that.

PR #1903, opened 2019-10-08
server: Factor out mapTreeUpdater

pavelkalinnikov:
This change factors out a type responsible for transactional map tree updates. This is in preparation for a bigger refactoring in #1902.

### DIFF
--- a/server/map_tree_updater.go
+++ b/server/map_tree_updater.go
@@ -1,0 +1,148 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/merkle"
+	"github.com/google/trillian/merkle/hashers"
+	"github.com/google/trillian/storage"
+	"github.com/google/trillian/storage/tree"
+)
+
+// updateTree updates the sparse Merkle tree at the specified revision based on the passed-in
+// leaf changes, and writes it to the storage. Returns the new signed map root, which is also
+// submitted to storage.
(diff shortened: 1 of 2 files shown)


pavelkalinnikov:
I recommend to review this commit by commit.

server/map_tree_updater.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
pavelkalinnikov: FYI: I will change the year to 2019 when a bigger refactoring takes place.

server/map_tree_updater.go
```diff
@@ -20,26 +20,35 @@ import (
...
-// leaf changes, and writes it to the storage. Returns the new signed map root, which is also
-// submitted to storage.
-func (t *TrillianMapServer) updateTree(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, hkv []merkle.HashKeyValue, rev int64) ([]byte, error) {
+// mapTreeUpdater updates the sparse Merkle tree of the map in one or multiple
+// map tree transactions.
+type mapTreeUpdater struct {
+	tree   *trillian.Tree
+	hasher hashers.MapHasher
+	opts   TrillianMapServerOptions
+	reg    extension.Registry
```
mhutchinson: Recommend swapping this for MapStorage. I'm a big fan of giving dependencies the smallest possible surface area so we can reason about what's going on. It also makes it clearer what the intent is.

server/map_tree_updater.go
```diff
@@ -20,26 +20,35 @@ import (
...
-// updateTree updates the sparse Merkle tree at the specified revision based on the passed-in
-// leaf changes, and writes it to the storage. Returns the new signed map root, which is also
-// submitted to storage.
-func (t *TrillianMapServer) updateTree(ctx context.Context, tree *trillian.Tree, hasher hashers.MapHasher, tx storage.MapTreeTX, hkv []merkle.HashKeyValue, rev int64) ([]byte, error) {
+// mapTreeUpdater updates the sparse Merkle tree of the map in one or multiple
+// map tree transactions.
+type mapTreeUpdater struct {
+	tree   *trillian.Tree
+	hasher hashers.MapHasher
+	opts   TrillianMapServerOptions
```
mhutchinson: Similar to the below comment, please explicitly pass in the two flag values for this instead of a giant handbag of stuff that is the options for MapServer, which this is not.

server/map_tree_updater.go
pavelkalinnikov: Done.

server/map_tree_updater.go
pavelkalinnikov: Done.

pavelkalinnikov:
My sincere gratitude to Mr Hutchinson.

PR #1907, opened 2019-10-15
Hammer supports writer getting ahead of published SMR

mhutchinson:
This will be required when WriteLeaves publishes SMRs asynchronously to writing the leaf values. This change makes VersionedMapContents updated with a revision X only when an SMR for a revision >= X is found. This allows the readers to keep using VersionedMapContents as they did before, without risking the writer getting so far ahead of the latest SMR that we push out all 10 of the revisions stored in memory that could be used to validate what the readers are getting from the map.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -473,12 +473,10 @@ type hammerState struct {
 	cfg            *MapConfig
 	validReadOps   *validReadOps
 	invalidReadOps *invalidReadOps
+	gossipHub      *gossipHub
+	head           *testonly.MapContents // TODO(mhutchinson): This should move to writeWorker.
 
 	start time.Time
-
-	// copies of earlier contents of the map
-	prevContents *testonly.VersionedMapContents
-	smrs         *smrStash
 
 	mu sync.RWMutex // Protects everything below
 
@@ -519,28 +517,26 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
 		cfg.OperationDeadline = 60 * time.Second
 	}
 
-	var prevContents testonly.VersionedMapContents
-	var smrs smrStash
+	gossipHub := newGossipHub()
 	validReadOps := validReadOps{
 		mc:           mc,
 		extraSize:    cfg.ExtraSize,
 		minLeaves:    cfg.MinLeaves,
 		maxLeaves:    cfg.MaxLeaves,
-		prevContents: &prevContents,
-		smrs:         &smrs,
+		prevContents: gossipHub.contents,
+		gossipHub:    gossipHub,
 	}
 	invalidReadOps := invalidReadOps{
 		mapID:        cfg.MapID,
 		client:       cfg.Client,
-		prevContents: &prevContents,
-		smrs:         &smrs,
+		prevContents: gossipHub.contents,
+		gossipHub:    gossipHub,
 	}
 
 	return &hammerState{
 		cfg:            cfg,
 		start:          time.Now(),
-		prevContents:   &prevContents,
-		smrs:           &smrs,
+		gossipHub:      gossipHub,
 		validReadOps:   &validReadOps,
 		invalidReadOps: &invalidReadOps,
 	}, nil
@@ -581,11 +577,13 @@ func (s *hammerState) String() string {
 		totalInvalidReqs += int(invalidReqs.Value(s.label(), string(ep)))
 		totalErrs += int(errs.Value(s.label(), string(ep)))
 	}
-	var latestRev int64 = -1
-	if smr := s.smrs.previousSMR(0); smr != nil {
-		latestRev = int64(smr.Revision)
+	var revStr string
+	if latestRev, found := s.gossipHub.getLastReadRev(); found {
+		revStr = strconv.FormatUint(latestRev, 10)
+	} else {
+		revStr = "N/A"
 	}
-	return fmt.Sprintf("%d: lastSMR.rev=%d ops: total=%d (%f ops/sec) invalid=%d errs=%v%s", s.cfg.MapID, latestRev, totalReqs, float64(totalReqs)/interval.Seconds(), totalInvalidReqs, totalErrs, details)
+	return fmt.Sprintf("%d: lastSMR.rev=%v ops: total=%d (%f ops/sec) invalid=%d errs=%v%s", s.cfg.MapID, revStr, totalReqs, float64(totalReqs)/interval.Seconds(), totalInvalidReqs, totalErrs, details)
 }
 
 func pickIntInRange(min, max int, prng *rand.Rand) int {
@@ -628,15 +626,14 @@ func (s *hammerState) setLeaves(ctx context.Context, prng *rand.Rand) error {
 		n = 1
 	}
 	leaves := make([]*trillian.MapLeaf, 0, n)
-	contents := s.prevContents.LastCopy()
 	rev := int64(0)
-	if contents != nil {
-		rev = contents.Rev
+	if s.head != nil {
+		rev = s.head.Rev
 	}
 leafloop:
 	for i := 0; i < n; i++ {
 		choice := choices[prng.Intn(len(choices))]
-		if contents.Empty() {
+		if s.head.Empty() {
 			choice = CreateLeaf
 		}
 		switch choice {
@@ -650,7 +647,7 @@ leafloop:
 			})
 			glog.V(3).Infof("%d: %v: data[%q]=%q", s.cfg.MapID, choice, key, string(value))
 		case UpdateLeaf, DeleteLeaf:
-			key := contents.PickKey(prng)
+			key := s.head.PickKey(prng)
 			// Not allowed to have the same key more than once in the same request
 			for _, leaf := range leaves {
 				if bytes.Equal(leaf.Index, key) {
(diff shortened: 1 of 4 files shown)


testonly/hammer/record.go
```diff
@@ -28,48 +30,99 @@ const (
...
+	if lastRoot := g.previousSMR(0); lastRoot != nil {
+		return lastRoot.Revision, true
+	}
+	return 0, false
+}
+
+// proposeLeaves should be called *before* writing a new map revision. The revision
+// will be available to readers only once advertiseSMR has been called with a root
+// that is >= the revision of this write.
+func (g *gossipHub) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) error {
```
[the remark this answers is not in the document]
pavelkalinnikov: Systems like KT solve this by making writes idempotent, i.e. for a revision N they always propose the same leaves. Should we have this property in the hammer?

testonly/hammer/hammer.go
```diff
@@ -519,28 +517,26 @@ func newHammerState(ctx context.Context, cfg *MapConfig) (*hammerState, error) {
...
+		prevContents: gossipHub.contents,
+		gossipHub:    gossipHub,
 	}
 
 	return &hammerState{
 		cfg:            cfg,
 		start:          time.Now(),
-		prevContents:   &prevContents,
-		smrs:           &smrs,
+		gossipHub:      gossipHub,
```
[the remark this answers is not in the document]
mhutchinson: Yes.

testonly/hammer/record.go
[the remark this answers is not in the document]
mhutchinson: That possibility might exist. This error already existed in the old code anyway and we haven't yet encountered it so I'm not too worried.

In the situation you describe the hammer would try again with a prng seeded with the same value, so would actually generate the same set of leaves (`retryOp` has `seed := w.prng.Int63()` before the retry logic). But then again it will likely fail on `WriteLeaves` because the revision we're trying to write already exists and the `map_rpc_server` will object, even if the update is logically the same. Modifying that is a feature request far beyond the scope of this PR though.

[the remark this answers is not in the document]
mhutchinson: Yup, that's part of the logic for using a map here. In reality it won't matter as we generate the same update in the case that we retry, but it's good to have some separation of concerns here.

testonly/hammer/record.go
```diff
@@ -28,48 +30,99 @@ const (
...
+	if head, found := g.getLastReadRev(); found && rev <= head {
```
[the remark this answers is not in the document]
mhutchinson: I threw this away. It was only trying to be defensive and catch a programming error, but it complicates the readability of the code. If this check would have failed here, it will fail when the writer proceeds to write this proposal to the map.

testonly/hammer/record.go
```diff
@@ -28,48 +30,99 @@ const (
...
+		return fmt.Errorf("Cannot create revision %d, latest SMR is %d", rev, head)
+	}
+
```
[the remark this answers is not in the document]
mhutchinson: It's gone.

testonly/hammer/record.go
```diff
@@ -28,48 +30,99 @@ const (
...
+// gossipHub allows details of what has been written to and read from the Map
+// to be shared between different workers running in the same process. The same
+// interface could be implemented by something which shared these details with
+// a remote process via gRPC if that was desirable.
+// The goal is thatVersionedMapContents can be subsumed into this class - there
+// should be no reason to address it directly. For now, this class coordinates
+// writes, reads, and local state such that VersionedMapContents always contains
+// revisions which are no further ahead than the most recent SMR published by the
+// Map.
+type gossipHub struct {
```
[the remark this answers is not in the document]
mhutchinson: sharedState it is.

testonly/hammer/hammer.go
```diff
@@ -581,11 +577,13 @@ func (s *hammerState) String() string {
...
 	}
-	var latestRev int64 = -1
-	if smr := s.smrs.previousSMR(0); smr != nil {
-		latestRev = int64(smr.Revision)
+	var revStr string
+	if latestRev, found := s.sharedState.getLastReadRev(); found {
+		revStr = strconv.FormatUint(latestRev, 10)
+	} else {
+		revStr = "N/A"
 	}
```
pavelkalinnikov: You can remove this clause by initializing `revStr := "N/A"`.

testonly/hammer/hammer.go
```diff
@@ -677,16 +674,16 @@ leafloop:
...
-		return fmt.Errorf("failed to set-leaves(count=%d): %v", len(leaves), err)
-	}
 
-	_, err = s.prevContents.UpdateContentsWith(writeRev, leaves)
-	if err != nil {
+	if err := s.sharedState.proposeLeaves(writeRev, leaves); err != nil {
 		return err
 	}
+	if _, err := s.cfg.Write.WriteLeaves(ctx, &req); err != nil {
+		return fmt.Errorf("failed to set-leaves(count=%d): %v", len(leaves), err)
```
pavelkalinnikov: nit: WriteLeaves?

testonly/hammer/reader.go
```diff
@@ -90,6 +92,10 @@ func (o *validReadOps) doGetLeaves(ctx context.Context, prng *rand.Rand, latest
 		if err != nil {
 			return fmt.Errorf("failed to GetAndVerifyMapLeaves: %v", err)
 		}
+		err = o.sharedState.advertiseSMR(*root)
+		if err != nil {
```
pavelkalinnikov: nit: Merge the two lines?

testonly/hammer/reader.go
```diff
@@ -120,7 +126,7 @@ func (o *validReadOps) getSMR(ctx context.Context, prng *rand.Rand) error {
 		return fmt.Errorf("failed to get-smr: %v", err)
 	}
 
-	err = o.smrs.pushSMR(*root)
+	err = o.sharedState.advertiseSMR(*root)
 	if err != nil {
```
pavelkalinnikov: nit: Merge the two lines?

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
 	smrCount = 30
 )
 
-// smrStash provides thread-safe access to an ordered, bounded queue of previously
-// witnessed SMRs
-type smrStash struct {
-	mu sync.RWMutex
+// sharedState allows details of what has been written to and read from the Map
```
pavelkalinnikov: simplify: "allows details ... to be shared between ..." -> "shares the details ... between ..."

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+// to be shared between different workers running in the same process. The same
+// interface could be implemented by something which shared these details with
+// a remote process via gRPC if that was desirable.
+// The goal is that VersionedMapContents can be subsumed into this class - there
+// should be no reason to address it directly. For now, this class coordinates
+// writes, reads, and local state such that VersionedMapContents always contains
+// revisions which are no further ahead than the most recent SMR published by the
+// Map.
```
pavelkalinnikov: nit: "SMR published by the Map." -> "published SMR."

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+type sharedState struct {
+	contents *testonly.VersionedMapContents
+
+	mu            sync.RWMutex
```
pavelkalinnikov: Could you add a comment on what fields this mutex guards?

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+	return 0, false
+}
+
+// proposeLeaves should be called *before* writing a new map revision. The revision
+// will be available to readers only once advertiseSMR has been called with a root
+// that is >= the revision of this write.
+func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
```
pavelkalinnikov: supernit: "meh" to this empty line.

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+// TODO(mhutchinson): Make this way more tolerant so it accepts older SMRs and
+// checks they are equivalent to previously seen versions if applicable.
+func (g *sharedState) advertiseSMR(smr types.MapRootV1) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	var prevRev uint64 // The last revision committed to contents.
+	if g.smrs[0] != nil {
+		if g.smrs[0].Revision > smr.Revision {
+			return fmt.Errorf("pushSMR called with stale root. Received revision %d, already had revision %d", smr.Revision, g.smrs[0].Revision)
```
pavelkalinnikov: nit: Simplify this to `"stale root: got revision %d, already had %d"`.

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+		} else if g.smrs[0].Revision == smr.Revision {
+			if !reflect.DeepEqual(g.smrs[0], &smr) {
+				return fmt.Errorf("pushSMR witnessed different SMRs for revision=%d. Had %+v, received %+v", smr.Revision, g.smrs[0], smr)
```
pavelkalinnikov: nit: `"SMR mismatch at revision %d: had %+v, got %+v"`.

testonly/hammer/record.go
```diff
@@ -28,48 +30,95 @@ const (
...
+	g.smrs[0] = &smr
+
+	for i := prevRev + 1; i <= smr.Revision; i++ {
+		if leaves, ok := g.pendingWrites[i]; ok {
+			if _, err := g.contents.UpdateContentsWith(i, leaves); err != nil {
+				return err
+			}
+			delete(g.pendingWrites, i)
+		} else {
+			return fmt.Errorf("Found SMR(r=%d), but failed to find pending write for r=%d", smr.Revision, i)
```
pavelkalinnikov: Typically we start `Errorf` messages with lower case.

testonly/hammer/hammer.go

testonly/hammer/reader.go
mhutchinson: Yes.

testonly/hammer/reader.go
mhutchinson: Making the codebase shorter one line at a time.

PR #1911, opened 2019-10-16
Hammer allows SMRs from the past to be checked

mhutchinson:
This prevents things blowing up if there are race conditions where two workers find new SMRs and advertize them slightly out of order.

This is more comprehensive, because old SMRs will be checked whenver we find them to make sure history isn't being switched around.

This is faster because more operations in advertiseSMR are done with read locks instead of write locks.

### DIFF
--- a/testonly/hammer/record.go
+++ b/testonly/hammer/record.go
@@ -84,7 +84,8 @@ func (g *sharedState) advertiseSMR(smr types.MapRootV1) error {
 		if g.smrs[0] == nil || g.smrs[0].Revision < smr.Revision {
 			return true, nil
 		}
-		pos := sort.Search(smrCount, func(i int) bool {
+		// Search to smrCount-1 so that pos will always be a valid index into the array.
+		pos := sort.Search(smrCount-1, func(i int) bool {
 			return g.smrs[i] == nil || g.smrs[i].Revision <= smr.Revision
 		})
 		if known := g.smrs[pos]; known != nil && known.Revision == smr.Revision && !reflect.DeepEqual(known, &smr) {
--- a/testonly/hammer/record.go
+++ b/testonly/hammer/record.go
@@ -17,6 +17,7 @@ package hammer
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"sync"
 
 	"github.com/golang/glog"
@@ -83,13 +84,11 @@ func (g *sharedState) advertiseSMR(smr types.MapRootV1) error {
 		if g.smrs[0] == nil || g.smrs[0].Revision < smr.Revision {
 			return true, nil
 		}
-
-		for i := 0; i < smrCount && g.smrs[i] != nil && g.smrs[i].Revision >= smr.Revision; i++ {
-			if g.smrs[0].Revision == smr.Revision {
-				if !reflect.DeepEqual(g.smrs[i], &smr) {
-					return false, fmt.Errorf("SMR mismatch at revision %d: had %+v, got %+v", smr.Revision, g.smrs[0], smr)
-				}
-			}
+		pos := sort.Search(smrCount, func(i int) bool {
+			return g.smrs[i] == nil || g.smrs[i].Revision <= smr.Revision
+		})
+		if known := g.smrs[pos]; known != nil && known.Revision == smr.Revision && !reflect.DeepEqual(known, &smr) {
+			return false, fmt.Errorf("SMR mismatch at revision %d: had %+v, got %+v", smr.Revision, known, smr)
 		}
 		return false, nil
 	}()
--- a/testonly/hammer/record.go
+++ b/testonly/hammer/record.go
@@ -75,44 +75,62 @@ func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) erro
 	return nil
 }
 
-// TODO(mhutchinson): Make this way more tolerant so it accepts older SMRs and
-// checks they are equivalent to previously seen versions if applicable.
 func (g *sharedState) advertiseSMR(smr types.MapRootV1) error {
-	g.mu.Lock()
-	defer g.mu.Unlock()
+	update, err := func() (bool, error) {
+		g.mu.RLock()
+		defer g.mu.RUnlock()
 
-	var prevRev uint64 // The last revision committed to contents.
-	if g.smrs[0] != nil {
-		if g.smrs[0].Revision > smr.Revision {
-			return fmt.Errorf("stale root: got revision %d, already had %d", smr.Revision, g.smrs[0].Revision)
-		} else if g.smrs[0].Revision == smr.Revision {
-			if !reflect.DeepEqual(g.smrs[0], &smr) {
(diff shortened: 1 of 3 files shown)


testonly/hammer/record.go
```diff
@@ -75,44 +75,62 @@ func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) erro
...
-	// Shuffle earlier SMRs along.
-	for i := smrCount - 1; i > 0; i-- {
-		g.smrs[i] = g.smrs[i-1]
-	}
+	if update {
+		g.mu.Lock()
+		defer g.mu.Unlock()
+
+		// update is a suggestion. We need to confirm the condition holds now we have write lock.
+		if g.smrs[0] != nil && g.smrs[0].Revision >= smr.Revision {
```
pavelkalinnikov: This check could be avoided if the lock wasn't released. I wonder if this `if update` case can be just inlined to line 83. And the you'd also not need the temporary function.

testonly/hammer/record.go
```diff
@@ -75,44 +75,62 @@ func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) erro
...
 
-	var prevRev uint64 // The last revision committed to contents.
-	if g.smrs[0] != nil {
-		if g.smrs[0].Revision > smr.Revision {
-			return fmt.Errorf("stale root: got revision %d, already had %d", smr.Revision, g.smrs[0].Revision)
-		} else if g.smrs[0].Revision == smr.Revision {
-			if !reflect.DeepEqual(g.smrs[0], &smr) {
-				return fmt.Errorf("SMR mismatch at revision %d: had %+v, got %+v", smr.Revision, g.smrs[0], smr)
+		for i := 0; i < smrCount && g.smrs[i] != nil && g.smrs[i].Revision >= smr.Revision; i++ {
+			if g.smrs[0].Revision == smr.Revision {
```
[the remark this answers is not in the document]
mhutchinson: Well that's embarrassing. Good catch. I had a test for this, but it got an error for a different reason which is hilarious with hindsight.

testonly/hammer/record.go
```diff
@@ -75,44 +75,62 @@ func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) erro
...
+		}
 
-	var prevRev uint64 // The last revision committed to contents.
-	if g.smrs[0] != nil {
-		if g.smrs[0].Revision > smr.Revision {
-			return fmt.Errorf("stale root: got revision %d, already had %d", smr.Revision, g.smrs[0].Revision)
-		} else if g.smrs[0].Revision == smr.Revision {
-			if !reflect.DeepEqual(g.smrs[0], &smr) {
-				return fmt.Errorf("SMR mismatch at revision %d: had %+v, got %+v", smr.Revision, g.smrs[0], smr)
+		for i := 0; i < smrCount && g.smrs[i] != nil && g.smrs[i].Revision >= smr.Revision; i++ {
```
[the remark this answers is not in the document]
mhutchinson: Done.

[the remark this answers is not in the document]
mhutchinson: Good call.

testonly/hammer/record.go
mhutchinson: This would involve trying to upgrade the read lock to a write lock, or making all operations take a write lock. I'm not particularly excited by that. Did you miss the subtle RLock vs Lock, or am I missing something that makes this a good idea?

testonly/hammer/record.go
```diff
@@ -83,13 +84,11 @@ func (g *sharedState) advertiseSMR(smr types.MapRootV1) error {
...
 			return true, nil
 		}
-
-		for i := 0; i < smrCount && g.smrs[i] != nil && g.smrs[i].Revision >= smr.Revision; i++ {
-			if g.smrs[0].Revision == smr.Revision {
-				if !reflect.DeepEqual(g.smrs[i], &smr) {
-					return false, fmt.Errorf("SMR mismatch at revision %d: had %+v, got %+v", smr.Revision, g.smrs[0], smr)
-				}
-			}
+		pos := sort.Search(smrCount, func(i int) bool {
```
pavelkalinnikov: Note that there is a "not found" `pos == smrCount` situation, if all `smrs` have `Revision > smr.Revision`. That's why I put "minus 1" in my suggestion, which helped to mitigate this subtlety. It's a bit ninja though, so feel free to comment it /  make it readable in other way.

testonly/hammer/record.go
pavelkalinnikov: I see, yeah, I missed that bit. SGTM.

testonly/hammer/record.go
pavelkalinnikov: Alternatively, you could allocate an extra "sentinel" item in the `smrs` slice.

pavelkalinnikov:
I feel like this SMR stack is quite a complicated data structure, and should be factored out / tested separately.

pavelkalinnikov:
Also, since all the workers progress through the SMRs in increasing order, it could be made so that SMR stack contains revisions between min{worker[i].SMR} and max{worker[i].SMR}. It would be a deque essentially, in which the tail is cleared as soon as all the workers are done with this revision.

testonly/hammer/record.go
```diff
@@ -75,44 +76,61 @@ func (g *sharedState) proposeLeaves(rev uint64, leaves []*trillian.MapLeaf) erro
...
+			prevRev = g.smrs[0].Revision
+		}
 
-	for i := prevRev + 1; i <= smr.Revision; i++ {
-		if leaves, ok := g.pendingWrites[i]; ok {
-			if _, err := g.contents.UpdateContentsWith(i, leaves); err != nil {
-				return err
+		glog.V(2).Infof("adding new SMR: %+v", smr)
+		// Shuffle earlier SMRs along.
+		for i := smrCount - 1; i > 0; i-- {
```
pavelkalinnikov: I nicer implementation in terms of moving date etc is having this slice sorted the other way around, and then this jiggery-pokery is transformed into:

- Initialization: ```smrs = make([]SMR, smrCount)```
- Push: ```g.smrs = append(g.smrs, &smr)[1:]``` (quite memory efficient)
- Latest: ```g.smrs[smrCount-1]```
- Iterate/Search the other way around.

pavelkalinnikov:
> Workers hit a random range of SMRs. They aren't just reading from head. This means keeping only "live" SMRs doesn't make sense.

Does it mean the read workers can try to read anything between 0 and current revision? And the longer the test runs the higher the probability of wasting cycles on trying to read non-existing revisions? I guess, the same filtering effect could be achieved if the read workers only read 
 revisions between `[current-historySize, current]`, only with fewer "misses".

And then an auto-clearing `sharedState` makes sense to me if we wanted to eliminate "misses". The workers would need to call something like `sharedState.forget(current-historySize)` each time they see a current SMR update. Or this could even be built into the `sharedState.announceSMR`.

Anyway, this probably won't happen given priorities :)

mhutchinson:
The readers only check revisions which are present in the SMRs in sharedState. This is effectively your suggestion.

```
which := prng.Intn(smrCount)
smrRoot := o.sharedState.previousSMR(which)
```

PR #1912, opened 2019-10-16
WriteLeaves: Always require the expected revision

pavelkalinnikov:
This change removes the feature of `Set/WriteLeaves` API calls, which allowed to commit leaves to a new "latest" revision rather than a fixed one. This feature is dangerous because it can cause races etc, and it was not used anyway. Additionally, the `revision == 0` meaning was overloaded (depending on the RPC being called): it could mean the "empty tree" revision, or "the latest" revision. Now it has only one meaning.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ Not yet released; provisionally v1.3.3 (may change).
 
 ### Map Changes
 
-The verifiable map is still experimental. APIs have been deprecated and will be
-deleted in the near future. These changes will not affect the Trillian module
-semantic version due to the experimental status of the Map.
+The verifiable map is still experimental. APIs, such as SetLeaves, have been
+deprecated and will be deleted in the near future. The semantic of WriteLeaves
+has become stricter: now it always requires the caller to specify the write
+revision. These changes will not affect the Trillian module semantic version due
+to the experimental status of the Map.
 
 The map client has been updated so that GetAndVerifyMapLeaves and
 GetAndVerifyMapLeavesByRevision return the MapRoot for the revision at which the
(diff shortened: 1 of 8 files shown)


server/map_rpc_server.go
```diff
@@ -384,8 +380,8 @@ func (t *TrillianMapServer) getWriteRevision(ctx context.Context, tree *trillian
 	if err != nil {
 		return 0, err
 	}
-	if assertRev != 0 && writeRev != assertRev {
-		return 0, status.Errorf(codes.FailedPrecondition, "can't write to revision %v", assertRev)
+	if writeRev != assertRev {
+		return 0, status.Errorf(codes.FailedPrecondition, "can't write to revision %v: only %d", assertRev, writeRev)
```
pavelkalinnikov: Oups, this was added during the debugging phase, I'll restore the original line.

CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ Not yet released; provisionally v1.3.3 (may change).
 
 ### Map Changes
 
-The verifiable map is still experimental. APIs have been deprecated and will be
-deleted in the near future. These changes will not affect the Trillian module
-semantic version due to the experimental status of the Map.
+The verifiable map is still experimental. APIs, such as SetLeaves, have been
+deprecated and will be deleted in the near future. The semantic of WriteLeaves
```
mhutchinson: nit: in English saying "the semantics of X are" is more common than "the semantic of X is".

server/map_rpc_server.go
mhutchinson: Reminder to please address.

gdbelvin:
Please verify this against Key Transparency locally first

pavelkalinnikov:
KT tests pass locally.

PR #1913, opened 2019-10-17
client: Remove unused SetLeaves method

pavelkalinnikov:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ to the experimental status of the Map.
 The map client has been updated so that GetAndVerifyMapLeaves and
 GetAndVerifyMapLeavesByRevision return the MapRoot for the revision at which the
 leaves were fetched. Without this callers of GetAndVerifyMapLeaves in particular
-were unable to reason about which map revision they were seeing. The unused
+were unable to reason about which map revision they were seeing. The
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ to the experimental status of the Map.
 The map client has been updated so that GetAndVerifyMapLeaves and
 GetAndVerifyMapLeavesByRevision return the MapRoot for the revision at which the
 leaves were fetched. Without this callers of GetAndVerifyMapLeaves in particular
-were unable to reason about which map revision they were seeing.
+were unable to reason about which map revision they were seeing. The unused
```
mhutchinson: Drop the unused. It was used in the last version of Trillian we published.

pavelkalinnikov:
> LGTM. My one concern is whether we should create a map_write_client.go before we remove this?

I don't think it should be strictly before. I suggest to add it when it's actually needed, as right now things seem to be working without it.

Martin2112:
I think we can probably do this before the map_write_client.

Martin2112:
Travis doesn't seem happy.

pavelkalinnikov:
Oups, I mixed it with another branch.

PR #1914, opened 2019-10-17
map_rpc_server: Refactor empty revision creation

pavelkalinnikov:
This change puts empty revision creation for `SetLeaves` and `InitMap` in the same function, and tests that `SetLeaves` with no updates works as intended.

This also makes #1902 simpler by not requiring the tree updater to handle the empty updates case.

### DIFF
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -313,8 +313,8 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	}
 	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
 
-	// TODO(pavelkalinnikov): Factor out a CreateRevision RPC, and merge it with
-	// InitMap which creates the 0th revision.
+	// TODO(pavelkalinnikov): Factor out a CreateRevision RPC in the MapWrite
+	// service, and use it instead of InitMap to create the 0th revision.
 	if len(req.Leaves) == 0 {
 		newSMR, err := t.addRevision(ctx, req.MapId, req.Revision, req.Metadata)
 		if err != nil {
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -308,10 +308,11 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapLeavesRequest) (*trillian.SetMapLeavesResponse, error) {
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
+	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
+
 	if req.Revision <= 0 {
 		return nil, status.Error(codes.FailedPrecondition, "revision must be > 0")
 	}
-	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
 
 	if len(req.Leaves) == 0 {
 		newSMR, err := t.addRevision(ctx, req.MapId, req.Revision, req.Metadata)
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -41,7 +41,6 @@ const (
 )
 
 var (
-	optsMapInit  = trees.NewGetOpts(trees.Admin, trillian.TreeType_MAP)
 	optsMapRead  = trees.NewGetOpts(trees.Query, trillian.TreeType_MAP)
 	optsMapWrite = trees.NewGetOpts(trees.UpdateMap, trillian.TreeType_MAP)
 )
--- a/server/map_rpc_server.go
+++ b/server/map_rpc_server.go
@@ -313,8 +313,6 @@ func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapL
 	}
 	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
 
-	// TODO(pavelkalinnikov): Factor out a CreateRevision RPC in the MapWrite
-	// service, and use it instead of InitMap to create the 0th revision.
 	if len(req.Leaves) == 0 {
 		newSMR, err := t.addRevision(ctx, req.MapId, req.Revision, req.Metadata)
 		if err != nil {
@@ -521,6 +519,10 @@ func (t *TrillianMapServer) InitMap(ctx context.Context, req *trillian.InitMapRe
 }
 
 // addRevision adds a new revision without changing the tree contents.
+//
+// TODO(pavelkalinnikov): Consider making it a CreateRevision RPC of the
+// MapService, a more generic version of InitMap that allows creating empty
+// revisions other than the 0th.
 func (t *TrillianMapServer) addRevision(ctx context.Context, treeID, rev int64, meta []byte) (*trillian.SignedMapRoot, error) {
 	tree, hasher, err := t.getTreeAndHasher(ctx, treeID, optsMapWrite)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/map_rpc_server.go
```diff
@@ -309,9 +309,18 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapLeavesRequest) (*trillian.SetMapLeavesResponse, error) {
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
-
 	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
 
+	// TODO(pavelkalinnikov): Factor out a CreateRevision RPC, and merge it with
+	// InitMap which creates the 0th revision.
+	if len(req.Leaves) == 0 {
+		newSMR, err := t.addRevision(ctx, req.MapId, req.Revision, req.Metadata)
```
pavelkalinnikov: If `Revision == 0` then this will have the same effect as `InitMap`. We should probably disallow it for now?

pavelkalinnikov:
PTAL.

server/map_rpc_server.go
```diff
@@ -309,9 +308,21 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
 func (t *TrillianMapServer) SetLeaves(ctx context.Context, req *trillian.SetMapLeavesRequest) (*trillian.SetMapLeavesResponse, error) {
 	ctx, spanEnd := spanFor(ctx, "SetLeaves")
 	defer spanEnd()
-
+	if req.Revision <= 0 {
+		return nil, status.Error(codes.FailedPrecondition, "revision must be > 0")
+	}
 	t.setLeafCounter.Add(float64(len(req.Leaves)), strconv.FormatInt(req.MapId, 10))
```
mhutchinson: Can we decide a general approach for when to use spans and increment counters in relations to precondition checks? It seems odd that in this PR we perform this check in between these operations.

My view is that context-free input errors (e.g. missing/invalid parameters) should be checked before counters and spans. Validation which requires context (e.g. tree with given ID can't be found) can happen after.

server/map_rpc_server.go
```diff
@@ -309,9 +308,21 @@ func (t *TrillianMapServer) getLeavesByRevision(ctx context.Context, mapID int64
...
 
+	// TODO(pavelkalinnikov): Factor out a CreateRevision RPC in the MapWrite
```
mhutchinson: I don't agree with the TODO necessarily. In the commonly assumed case that each write batch will map to a published SMR revision then the sequencer/writer needs to make two calls to make a batch be published. When WriteLeaves becomes async, this adds the complexity of whether CreateRevision can be called for a revision which has not had its tree calculated yet.

I'm of the view that having a published revision per WriteLeaves batch massively simplifies the logical model of how the map works. The only downside to this is that WriteLeaves must be called with an empty batch to publish an SMR with a new timestamp but committing to the same contents. I don't think that's going to be very common, and it's still conceptually simple.

server/map_rpc_server_test.go
```diff
@@ -74,7 +74,7 @@ func TestInitMap(t *testing.T) {
 		wantCode   codes.Code
 	}{
 		{desc: "init new map", getRootErr: storage.ErrTreeNeedsInit, wantInit: true, root: nil, wantCode: codes.OK},
-		{desc: "init new map, no err", getRootErr: nil, wantInit: true, root: nil, wantCode: codes.OK},
```
mhutchinson: For clarity, this test was wrong before? It looks like it returned no error and a nil root, which never happens in reality?

server/map_rpc_server_test.go
```diff
@@ -336,6 +336,63 @@ func TestGetSignedMapRootByRevision(t *testing.T) {
 	}
 }
 
+func TestSetLeavesEmpty(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	ctx := context.Background()
+
+	root := &trillian.SignedMapRoot{
+		MapRoot:   []byte("\000\001 \306h\237\020\201*\t\200\227m\2253\3308u(!f\025\225g\3545\025W\026\301A:\365=j\025\315\333#\364\266\341\026\000\000\000\000\000\000\000\000\000\000"),
```
mhutchinson: Where did you generate these bytes from? Given the code below I've inferred that it must commit to an empty tree at revision 0, but can we make this clearer?

server/map_rpc_server_test.go
```diff
@@ -336,6 +336,63 @@ func TestGetSignedMapRootByRevision(t *testing.T) {
...
+	}{
+		{root: root, wantErr: false},
+		{root: root, revShift: -1, wantErr: true},
+		{root: root, revShift: -10, wantErr: true},
+		{root: root, revShift: 1, wantErr: true},
+		{root: nil, wantErr: true},
+		{root: &trillian.SignedMapRoot{MapRoot: []byte("invalid")}, wantErr: true},
+		{root: nil, smrErr: errors.New("you shall not pass"), wantErr: true},
+	} {
+		t.Run("", func(t *testing.T) {
```
mhutchinson: The empty test description here seems suboptimal. On failure is it easy to identify which of the cases failed?

pavelkalinnikov: Yeah, the framework auto-adds a suffix looking something like `#01`, which is logically equivalent to manually adding `strconv.Itoa(i)`. I just find this description bit unnecessary, the index should be enough.

server/map_rpc_server_test.go
pavelkalinnikov: Yeah, never happens. But if it does then it's probably an error, so the test was improved a bit.

server/map_rpc_server.go
pavelkalinnikov: > I don't agree with the TODO necessarily. In the commonly assumed case that each write batch will map to a published SMR revision then the sequencer/writer needs to make two calls to make a batch be published. When WriteLeaves becomes async, this adds the complexity of whether CreateRevision can be called for a revision which has not had its tree calculated yet.

I'm not proposing to call `CreateRevision` for every revision, only for empty ones. So it's still 1 RPC per revision.

Before:
1. `InitMap`
2. `WriteLeaves(rev=1, leaves=10)`, `WriteLeaves(rev=2, leaves=0)`, ....

After:
1. `CreateRevision(rev=0)`
2. `WriteLeaves(rev=1, leaves=10)`, `CreateRevision(rev=2)`, ...

 
> I'm of the view that having a published revision per WriteLeaves batch massively simplifies the logical model of how the map works. The only downside to this is that WriteLeaves must be called with an empty batch to publish an SMR with a new timestamp but committing to the same contents. I don't think that's going to be very common, and it's still conceptually simple.

The reason why someone would want to create a new SMR with 0 updates is refreshing the timestamp (I think that is the only one so far?). The reason for wanting to update the timestamp is that there haven't been any updates for quite a long time. In this case it makes sense to have a synchronous call like `CreateRevision(rev=N)` which immediately creates/returns an SMR.

The `WriteLeaves` call, if I understand the idea/design correctly, is not meant to be synchronous. It seems unreasonable allowing to create an empty SMR "at some point in the future": it's either needed "right now", or not needed at all (in case `WriteLeaves` was called in the meantime, and the new SMR construction is ongoing anyway).

server/map_rpc_server_test.go
pavelkalinnikov: Extracted the "for testing" SMR creation into a separate function. Plus added tests that not necessarily need revision==1.

server/map_rpc_server.go
pavelkalinnikov: We could decide this in a follow-up PR. Right now there are methods that either start a span right in the beginning (regardless of the checks), or not start it at all (as in `GetLeavesByRevisionNoProof`). The PR keeps this consistent.

pavelkalinnikov: In the frame of this PR, I can put the `Revision <= 0` check below the counter increase if you prefer?

server/map_rpc_server.go
pavelkalinnikov: The purpose of this TODO was to express my "meh" to the API that we have right now. I made it a "Consider" type of TODO, because there are other options, which include decoupling SMRs from revisions (I don't see a strong reason why signing should be part of updating the tree), or allowing multiple SMRs per revision. Since the next gen map is in the design phase, all options should be on the table.

Moved this TODO to `addRevision` method comment.

server/map_rpc_server.go
mhutchinson: I think incrementing the counter consistently at the start of the method has some merits. Let's do that for now.

PR #1916, opened 2019-10-18
Writer only performs write operations

mhutchinson:
Increased the number of checkers from 0 to 1 by default so that callers relying on the defaults don't see a large change in behaviour.

Also included a bunch of refactors to clean up the hammer in this PR now that the read and write workers are separated. Recommend reviewing commit-by-commit.

### DIFF
--- a/testonly/hammer/hammer.go
+++ b/testonly/hammer/hammer.go
@@ -465,6 +465,104 @@ func (w *writeWorker) writeOnce(ctx context.Context) (err error) {
 	return w.retryOp(ctx, w.setLeaves, string(ep))
 }
 
+func (w *writeWorker) setLeaves(ctx context.Context, prng *rand.Rand) error {
+	choices := []Choice{CreateLeaf, UpdateLeaf, DeleteLeaf}
+
+	cfg := w.s.cfg
+	n := pickIntInRange(cfg.MinLeaves, cfg.MaxLeaves, prng)
+	if n == 0 {
+		n = 1
+	}
+	leaves := make([]*trillian.MapLeaf, 0, n)
+	rev := int64(0)
+	if w.head != nil {
+		rev = w.head.Rev
+	}
+leafloop:
+	for i := 0; i < n; i++ {
+		choice := choices[prng.Intn(len(choices))]
+		if w.head.Empty() {
+			choice = CreateLeaf
+		}
+		switch choice {
+		case CreateLeaf:
+			key := w.nextKey()
+			value := w.nextValue()
+			leaves = append(leaves, &trillian.MapLeaf{
+				Index:     testonly.TransparentHash(key),
+				LeafValue: value,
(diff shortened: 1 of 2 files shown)


testonly/hammer/hammer.go
```diff
@@ -454,10 +459,22 @@ func (w *writeWorker) writeOnce(ctx context.Context) (err error) {
...
+
+func (w *writeWorker) nextKey() string {
+	w.keyIdx++
+	return fmt.Sprintf("key-%08d", w.keyIdx)
+}
+
+func (w *writeWorker) nextValue() []byte {
+	w.valueIdx++
+	result := make([]byte, w.s.cfg.LeafSize)
+	copy(result, fmt.Sprintf(valueFormat, w.valueIdx))
```
pavelkalinnikov: Could you please inline the `valueFormat` here? It's only used once, and having it as a separate constant only harms readability. Also, it would be consistent with `nextKey()` function which does inline the format string.

testonly/hammer/hammer.go
```diff
@@ -466,15 +483,8 @@ type hammerState struct {
 	validReadOps   *validReadOps
 	invalidReadOps *invalidReadOps
 	sharedState    *sharedState
-	head           *testonly.MapContents // TODO(mhutchinson): This should move to writeWorker.
 
 	start time.Time
-
-	mu sync.RWMutex // Protects everything below
```
pavelkalinnikov: Yay, mutex go away.

testonly/hammer/hammer.go
```diff
@@ -608,36 +602,37 @@ func getOp(ep MapEntrypointName, read readOps, write mapOperationFn) (mapOperati
...
-		if s.head.Empty() {
+		if w.head.Empty() {
 			choice = CreateLeaf
 		}
 		switch choice {
 		case CreateLeaf:
-			key := s.nextKey()
-			value := s.nextValue()
+			key := w.nextKey()
+			value := w.nextValue()
```
pavelkalinnikov: optional: Would it be nice as `key, value := w.nextKey(), w.nextValue()`?

testonly/hammer/hammer.go
mhutchinson: The reason this is still in the constants section is so that `minValueLen` can be defined alongside it. Separating them increases the probability of drift there. I'm not happy with this inconsistency here either, but I'll err on the side of reduced readability over probability of introducing a logic bug.

testonly/hammer/hammer.go
```diff
@@ -450,24 +450,14 @@ func (w *writeWorker) run(ctx context.Context, done <-chan struct{}) (uint64, er
 }
 
 func (w *writeWorker) writeOnce(ctx context.Context) (err error) {
-	ep := w.bias.choose(w.prng)
+	ep := SetLeavesName
```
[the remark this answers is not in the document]
mhutchinson: Added TODO

PR #1920, opened 2019-10-21
Move StorageProvider interfaces to storage

pavelkalinnikov:
This change moves `StorageProvider` interfaces/functions to `storage` package, and adds aliases in `server` package for backward compatibility.

This is the first step of #1919. What follows is the move of the existing providers to the corresponding packages.

### Checklist

### DIFF
--- a/server/storage_provider.go
+++ b/server/storage_provider.go
@@ -22,28 +22,28 @@ import "github.com/google/trillian/storage"
 // NewStorageProviderFunc is the signature of a function which can be
 // registered to provide instances of storage providers.
 //
-// Deprecated: storage.NewProviderFunc must be used directly.
+// Deprecated: storage.NewProviderFunc should be used directly.
 type NewStorageProviderFunc = storage.NewProviderFunc
 
 // StorageProvider is an interface which allows Trillian binaries to use
 // different storage implementations.
 //
-// Deprecated: storage.Provider must be used directly.
+// Deprecated: storage.Provider should be used directly.
 type StorageProvider = storage.Provider
 
 // RegisterStorageProvider registers the provided StorageProvider.
 //
-// Deprecated: storage.RegisterProvider must be used directly.
+// Deprecated: storage.RegisterProvider should be used directly.
 var RegisterStorageProvider = storage.RegisterProvider
 
 // NewStorageProviderFromFlags returns a new StorageProvider instance of the
 // type specified by flag.
 //
-// Deprecated: storage.NewProviderFromFlags must be used directly.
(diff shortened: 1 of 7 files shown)


RJPercival:
Could you mark the aliases as deprecated please? That should discourage people from using them. A TODO to remove them in the next major version would also be a good idea.

RJPercival:
Also deserves a CHANGELOG entry.

pavelkalinnikov:
> Alternatively, we were considering using a GitHub Issue to track all of the cleanup required for the next major release - could do that?

I think this is what we should do. There is another "aliases" file already for `NodeID` etc. We need to consolidate these kinds of things.

server/storage_provider.go
```diff
@@ -14,87 +14,36 @@
...
-// NewStorageProviderFunc is the signature of a function which can be registered
-// to provide instances of storage providers.
-type NewStorageProviderFunc func(monitoring.MetricFactory) (StorageProvider, error)
-
-var (
-	storageSystem = flag.String("storage_system", "mysql", fmt.Sprintf("Storage system to use. One of: %v", storageProviders()))
+// NewStorageProviderFunc is the signature of a function which can be
+// registered to provide instances of storage providers.
+//
+// Deprecated: storage.NewProviderFunc must be used directly.
```
RJPercival: nit: Instead of "must", rephrase as "should" - it'll still work with the aliases (for now), so it's not strictly a "must".

storage/provider.go
```diff
@@ -0,0 +1,99 @@
...
+	spByName map[string]NewProviderFunc
+)
+
+// RegisterProvider registers the given storage Provider.
+func RegisterProvider(name string, sp NewProviderFunc) error {
+	spMu.Lock()
+	defer spMu.Unlock()
+
+	spOnce.Do(func() {
+		spByName = make(map[string]NewProviderFunc)
```
RJPercival: I know this is how the code already was, so you shouldn't change this in this PR, but I was wondering why `spByName` is initialized in this way, rather than just having `spByName = make(map[string]NewProviderFunc)` in the var block at top of the file. Any idea?

pavelkalinnikov: No idea, looks like a poor man's optimization to me. Refactored it.

pavelkalinnikov:
Done the version bump in CHANGELOG.

pavelkalinnikov:
Well, adding the "Deprecated" comments caused linter failures :) Will fix it in this PR then.

RJPercival:
At least we know the deprecation warnings work!

PR #1922, opened 2019-10-22
Fix MySQL quota provider initialization bug

pavelkalinnikov:
The order in which MySQL storage and quota providers are created is undefined. MySQL quota provider previously relied on the database field having been created by the storage provider. Now it's not required.

### DIFF
--- a/server/mysql_storage_provider.go
+++ b/server/mysql_storage_provider.go
@@ -33,9 +33,10 @@ var (
 	maxConns = flag.Int("mysql_max_conns", 0, "Maximum connections to the database")
 	maxIdle  = flag.Int("mysql_max_idle_conns", -1, "Maximum idle database connections in the connection pool")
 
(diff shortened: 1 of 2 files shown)


server/mysql_storage_provider.go
```diff
@@ -33,9 +33,10 @@ var (
 	maxConns = flag.Int("mysql_max_conns", 0, "Maximum connections to the database")
 	maxIdle  = flag.Int("mysql_max_idle_conns", -1, "Maximum idle database connections in the connection pool")
 
-	mysqlOnce            sync.Once
-	mysqlOnceErr         error
-	mySQLstorageInstance *mysqlProvider
+	mysqlMu              sync.Mutex
+	mysqlErr             error
```
golangcibot: var `mysqlErr` is unused (from `unused`)

PR #1923, opened 2019-10-22
Read and write batch sizes are configured separately

mhutchinson:
There is no reason in the real world that the number of leaves fetched by clients in a batch would have any relationship to the number of leaves written at a time. The single parameter was an artefact of the way the hammer was written. Allowing these to be configured separately better approximates the real world and allows the hammer to hit the map more flexibly.

Making either of the read or write flags be above 100 triggers the issue in #1845, though it does so differently for each of the batch sizes.

### DIFF
--- a/testonly/hammer/hammer_test.go
+++ b/testonly/hammer/hammer_test.go
@@ -73,8 +73,10 @@ func TestRetryExposesDeadlineError(t *testing.T) {
 		LeafSize:      1000,
 		ExtraSize:     100,
 		// TODO(mhutchinson): Increase these when #1845 is understood & fixed.
-		MinLeaves:   100,
-		MaxLeaves:   150,
+		MinLeavesR:  100,
+		MaxLeavesR:  150,
+		MinLeavesW:  100,
+		MaxLeavesW:  150,
 		Operations:  *operations,
 		NumCheckers: 1,
 	}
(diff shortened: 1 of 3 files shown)


testonly/hammer/hammer_test.go
```diff
@@ -124,8 +126,10 @@ func TestInProcessMapHammer(t *testing.T) {
 		LeafSize:      1000,
 		ExtraSize:     100,
 		// TODO(mhutchinson): Increase these when #1845 is understood & fixed.
-		MinLeaves:   100,
-		MaxLeaves:   150,
+		MinLeavesR:  100,
```
pavelkalinnikov: async nit: The 2 configs seem the same. Maybe consolidate its creation into a shared helper?

testonly/hammer/maphammer/main.go
```diff
@@ -53,8 +53,10 @@ var (
 	outLog          = flag.String("log_to", "", "File to record operations in")
 	seed            = flag.Int64("seed", -1, "Seed for random number generation")
 	operations      = flag.Uint64("operations", ^uint64(0), "Number of operations to perform")
-	minLeaves       = flag.Int("min_leaves", 0, "Minimum count of leaves to affect per-operation")
-	maxLeaves       = flag.Int("max_leaves", 10, "Maximum count of leaves to affect per-operation")
+	minLeaves       = flag.Int("min_leaves", 0, "Minimum count of leaves to read per-operation")
```
pavelkalinnikov: Is there any reason to reuse the old names? Maybe name them explicitly as `min/max_leaves_read`?

PR #1927, opened 2019-10-23
Include full details if leaf inclusion proof fails

mhutchinson:
Including so much information seems a little non-standard, but this given how serious an error this is, having the relevant info seems like a must.  Useful when diagnosing #1845.

### DIFF
--- a/client/map_verifier.go
+++ b/client/map_verifier.go
@@ -18,7 +18,7 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/gogo/protobuf/proto"
+	"github.com/golang/protobuf/proto"
 	"github.com/google/trillian"
 	"github.com/google/trillian/maps"
 	"github.com/google/trillian/merkle"
--- a/client/map_verifier.go
+++ b/client/map_verifier.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/google/trillian"
 	"github.com/google/trillian/maps"
 	"github.com/google/trillian/merkle"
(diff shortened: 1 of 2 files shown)


client/map_verifier.go
```diff
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/gogo/protobuf/proto"
```
golangcibot: `github.com/gogo/protobuf/proto` is in the blacklist (from `depguard`)

client/map_verifier.go
```diff
@@ -81,7 +82,10 @@ func (m *MapVerifier) VerifyMapLeafInclusion(smr *trillian.SignedMapRoot, leafPr
 
 // VerifyMapLeafInclusionHash verifies a MapLeafInclusion object against a root hash.
 func (m *MapVerifier) VerifyMapLeafInclusionHash(rootHash []byte, leafProof *trillian.MapLeafInclusion) error {
-	return merkle.VerifyMapInclusionProof(m.MapID, leafProof.GetLeaf(), rootHash, leafProof.GetInclusion(), m.Hasher)
+	if err := merkle.VerifyMapInclusionProof(m.MapID, leafProof.GetLeaf(), rootHash, leafProof.GetInclusion(), m.Hasher); err != nil {
+		return status.Errorf(status.Code(err), "VerifyMapLeafInclusionHash(): %v, trillian.MapLeafInclusion=\n%v", err, proto.MarshalTextString(leafProof))
```
pavelkalinnikov: Why not make the caller of this function do the error wrapping?

gdbelvin: Sometimes I prefer an error format more like
`VerifyMapLeafInclusionHash(<args>): err`

Agreed that putting this in the client library would make it more useful.

client/map_verifier.go
mhutchinson: Thanks

pavelkalinnikov:
There is also an option of logging instead of returning a huge error btw.

Issue #1738, opened 2019-06-27
Fix proto buffer pass by value, shallow copying or unsafe comparisons


Martin2112:
We need to update proto buffer usages where they are passed by value or compared by generic code like `reflect.DeepEquals()`. Passing them by value creates shallow copies that can share internal state. As the generated structs contain additional exported `XXX_` fields managed by the proto implementation generic comparisons using all fields can produce incorrect results.

Typical problem cases:

* Passing a proto as a value parameter to a func
* Using a proto as a map key
* Struct assignment of a proto in cases like `*proto1= proto2`
* Using a proto as a value field in another struct (commonly test structs)
* Using `json.Marshal`, `reflect.DeepEquals` or similar on protos

The scope of this should be limited as we don't use protos heavily outside gRPC interfaces and storage.

Martin2112:
Can't 100% guarantee that I've fixed all of them but I think I've got most of the ones in our repositories that can be found by `go vet`.

PR #1928, opened 2019-10-23
Fix improper gomock comparisons

dsnet:
The gomock framework uses reflect.DeepEqual by default unless the
test passes an object that implements gomock.Matcher.
Use of reflect.DeepEqual on protocol buffer messages is problematic
because generated messages contain unexported fields, mutexes,
and other internal data structures that reflect.DeepEqual does not
know how to handle. Fix the tests by defining a custom gomock.Matcher
to properly compare the inputs.

### Checklist

### DIFF
--- a/log/sequencer_manager_test.go
+++ b/log/sequencer_manager_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keys"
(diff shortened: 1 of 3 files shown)


Martin2112:
Hi,

Thanks for the PR, we've been trying to fix these but have obviously not found them all.

I think you can perhaps use `cmp.Comparer` for this and avoid having to create a matcher. Example in `log_rpc_server_test.go` is something like:

`... cmp.Equal(got, test.want, cmp.Comparer(proto.Equal)) ... `

dsnet:
You can't `cmp.Options` are not compatible with the `gomock` framework.

Martin2112:
That's quite annoying.

Martin2112:
Work towards #1738.

PR #1930, opened 2019-10-25
Added tests for MySQL large batch operations

mhutchinson:
These were intended to show that #1845 could be easily reproduced with batches over 1k but they run fine. More investigation is needed, but more tests is always good.

### DIFF
--- a/storage/mysql/log_storage_test.go
+++ b/storage/mysql/log_storage_test.go
@@ -371,6 +371,41 @@ func TestQueueLeaves(t *testing.T) {
 	}
 }
 
+func TestQueueLeavesDuplicateBigBatch(t *testing.T) {
+	ctx := context.Background()
+
+	cleanTestDB(DB)
+	as := NewAdminStorage(DB)
+	tree := mustCreateTree(ctx, t, as, testonly.LogTree)
+	s := NewLogStorage(DB, nil)
+
+	const leafCount = 999 + 1
+	leaves := createTestLeaves(leafCount, 20)
+
+	runLogTX(s, tree, t, func(ctx context.Context, tx storage.LogTreeTX) error {
+		if _, err := tx.QueueLeaves(ctx, leaves, fakeQueueTime); err != nil {
+			t.Fatalf("Failed to queue leaves: %v", err)
+		}
+		return nil
+	})
+
+	runLogTX(s, tree, t, func(ctx context.Context, tx storage.LogTreeTX) error {
+		if _, err := tx.QueueLeaves(ctx, leaves, fakeQueueTime); err != nil {
+			t.Fatalf("Failed to queue leaves: %v", err)
+		}
+		return nil
+	})
+
+	// Should see the leaves in the database. There is no API to read from the unsequenced data.
+	var count int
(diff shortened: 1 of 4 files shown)


pavelkalinnikov:
If you manage to hack into the map transaction creation code and set the `UseLargePreload` flag, you should be able to reproduce this bug.

pavelkalinnikov:
I mean, in the place where `server.NewTrillianMapServer` is invoked, the `TrillianMapServerOptions` should have both bools set to true.

pavelkalinnikov:
Ignore my last comment, it's a storage layer, there is no `MapServer` here. But I see another reason why this does not reproduce itself. Will comment inline.

storage/mysql/map_storage_test.go
```diff
@@ -372,6 +372,81 @@ func TestMapSetGetMultipleRevisions(t *testing.T) {
 	}
 }
 
+func TestMapSetGetBigBatch(t *testing.T) {
```
pavelkalinnikov: This test won't reproduce the problem because it sits in the `GetMerkleNodes` call.

mhutchinson: Yup, I wrote this one before the latest finding with the preloader. Figured it was still worth having anyway.

storage/mysql/storage_test.go
```diff
@@ -154,6 +154,77 @@ func TestLogNodeRoundTripMultiSubtree(t *testing.T) {
 	}
 }
 
+func TestMultiRevisionFetchBigBatch(t *testing.T) {
+	nodes1 := createSomeNodes(10000)
```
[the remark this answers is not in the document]
mhutchinson: I'll generate these with a wider distribution and see if the behaviour changes!

storage/mysql/map_storage_test.go
```diff
@@ -570,3 +721,29 @@ func mapTree(mapID int64) *trillian.Tree {
 		HashStrategy: trillian.HashStrategy_TEST_MAP_HASHER,
 	}
 }
+
+func createMapNodes(count int) []stree.Node {
+	r := make([]stree.Node, count)
+	bs := make([]byte, 32, 32) // Default value of 0 index in 2^256 keyspace.
```
golangcibot: S1019: should use make([]byte, 32) instead (from `gosimple`)

golangcibot: S1019: should use make([]byte, 32) instead (from `gosimple`)

storage/mysql/log_storage_test.go
```diff
@@ -371,6 +371,42 @@ func TestQueueLeaves(t *testing.T) {
 	}
 }
 
+func TestQueueLeavesDuplicateBigBatch(t *testing.T) {
+	t.Skip("Known Issue: https://github.com/google/trillian/issues/1845")
```
pavelkalinnikov: Is there a way to skip these conditionally, depending on the environment?

mhutchinson: It's possible but more work. It would involve skipping if the DB is not present, then querying it for the version, parsing it, and then skipping based on MariaDB maj=10, min>=3.

I can do that and test it on MariaDB versions, but I've spent enough time on this and would need to test such code works for MySQL and I have little appetite for that.

PR #1931, opened 2019-10-25
Run the MySQL storage tests in MariaDB integration tests

mhutchinson:
This will flush out issues like #1845 faster and more precisely than relying on the hammer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,6 +122,7 @@ jobs:
       env: GO_TEST_TIMEOUT=20m
       script:
         - mysql -u root -e 'SHOW VARIABLES LIKE "%version%";'
+        - go test ./storage/mysql/...
         - ./integration/integration_test.sh && HAMMER_OPTS="--operations=50" ./integration/maphammer.sh 3
 
 services:


.travis.yml
```diff
@@ -122,6 +122,7 @@ jobs:
       env: GO_TEST_TIMEOUT=20m
       script:
         - mysql -u root -e 'SHOW VARIABLES LIKE "%version%";'
+        - go test ./storage/mysql/...
```
pavelkalinnikov: Eh? Is it not run?

pavelkalinnikov: It seems like only the "coverage" target runs the unit-tests. It makes sense though to run the specific ones here.

pavelkalinnikov:
Do you expect #1930 failing in Travis after this PR?

.travis.yml
mhutchinson: The tests are run in other targets, but it makes sense to run the DB-specific tests again with different DB implementations installed.

mhutchinson:
> Do you expect #1930 failing in Travis after this PR?

Absolutely. I'll be sad if they don't.

PR #1932, opened 2019-10-25
Make faster Map effective

pavelkalinnikov:
This change makes the new Map writer design effective.

The tiling structure of Map is now explicit in `storage` API. The new `Layout`
method of `MapStorage` interface returns the layout, and `MapTX`'s `GetTiles` and
`SetTiles` methods allow reading and writing tiles. Most of the tile processing
logic, e.g. computing intermediate hashes, now resides in the application layer.

The new code supports both single- and multi- transaction writes, and is fully
compatible with the previous implementation (true only for MySQL storage).

`SubtreeCache` is no longer used by the Map, and so there are no longer hidden
races/locking between parallel `GetTiles`/`SetTiles` calls.

As a result, MySQL map integration tests running time is reduced by 20%.

### Checklist

### DIFF
--- a/server/map_tree_updater.go
+++ b/server/map_tree_updater.go
@@ -33,20 +33,21 @@ type mapTreeUpdater struct {
 	layout   *tree.Layout
 	hasher   hashers.MapHasher
 	ms       storage.MapStorage
+	writeRev int64
 	singleTX bool
 	preload  bool
 }
 
-// update updates the sparse Merkle tree at the passed-in revision with the
+// update updates the sparse Merkle tree at the current write revision with the
 // given leaf node updates, and writes it to the storage. Returns the new root
 // hash. Requires nodes slice to be non-empty.
-func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes []smt.Node, writeRev int64) ([]byte, error) {
+func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes []smt.Node) ([]byte, error) {
 	// Work around a performance issue when using the map in single-transaction
 	// mode by preloading all the tiles we know the Writer is going to need.
 	var preloaded *smt.TileSet
 	if t.singleTX && t.preload {
 		var err error
-		if preloaded, err = t.doPreload(ctx, tx, nodes, writeRev-1); err != nil {
+		if preloaded, err = t.doPreload(ctx, tx, nodes); err != nil {
 			return nil, err
 		}
 	}
@@ -66,7 +67,7 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
 		err = runTX(ctx, func(ctx context.Context, tx storage.MapTreeTX) error {
 			nodesCopy := make([]smt.Node, len(nodes))
 			copy(nodesCopy, nodes) // Protect from TX restarts.
-			acc := &txAccessor{updater: t, tiles: preloaded, tx: tx, rev: writeRev}
+			acc := &txAccessor{updater: t, tiles: preloaded, tx: tx}
 			var err error
 			root, err = w.Write(ctx, nodesCopy, acc)
 			return err
@@ -116,7 +117,6 @@ type txAccessor struct {
 	// (and the only) Get call occurs. A missing tile is interpreted as empty.
 	tiles *smt.TileSet
 	tx    storage.MapTreeTX
-	rev   int64
 }
 
 // TODO(pavelkalinnikov): Get should take a list of tile IDs.
@@ -124,7 +124,7 @@ func (t *txAccessor) Get(ctx context.Context, ids []tree.NodeID2) (map[tree.Node
 	// TODO(pavelkalinnikov): Factor out preload into another accessor.
 	if t.tiles == nil {
 		var err error
-		if t.tiles, err = t.updater.load(ctx, t.tx, ids, t.rev-1); err != nil {
+		if t.tiles, err = t.updater.load(ctx, t.tx, ids); err != nil {
 			return nil, err
 		}
 	}
@@ -166,7 +166,7 @@ func (t *mapTreeUpdater) newTXFunc(tx storage.MapTreeTX) txFunc {
 
 // doPreload loads all Merkle tree tiles that will be involved when updating
 // the given set of nodes at the specified revision.
-func (t *mapTreeUpdater) doPreload(ctx context.Context, tx storage.MapTreeTX, nodes []smt.Node, rev int64) (*smt.TileSet, error) {
+func (t *mapTreeUpdater) doPreload(ctx context.Context, tx storage.MapTreeTX, nodes []smt.Node) (*smt.TileSet, error) {
 	ctx, spanEnd := spanFor(ctx, "doPreload")
 	defer spanEnd()
 	// TODO(pavelkalinnikov): Avoid using HStar3 directly.
@@ -175,13 +175,13 @@ func (t *mapTreeUpdater) doPreload(ctx context.Context, tx storage.MapTreeTX, no
 	if err != nil {
 		return nil, err
 	}
-	return t.load(ctx, tx, hs.Prepare(), rev)
+	return t.load(ctx, tx, hs.Prepare())
 }
 
 // load fetches Merkle tree tiles containing the given node IDs.
-func (t *mapTreeUpdater) load(ctx context.Context, tx storage.MapTreeTX, ids []tree.NodeID2, rev int64) (*smt.TileSet, error) {
+func (t *mapTreeUpdater) load(ctx context.Context, tx storage.MapTreeTX, ids []tree.NodeID2) (*smt.TileSet, error) {
 	tileIDs := toTileIDs(ids, t.layout)
-	tiles, err := tx.GetTiles(ctx, rev, tileIDs)
+	tiles, err := tx.GetTiles(ctx, t.writeRev-1, tileIDs)
 	if err != nil {
 		return nil, err
 	}
--- a/server/map_tree_updater.go
+++ b/server/map_tree_updater.go
@@ -66,7 +66,7 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
 		err = runTX(ctx, func(ctx context.Context, tx storage.MapTreeTX) error {
 			nodesCopy := make([]smt.Node, len(nodes))
 			copy(nodesCopy, nodes) // Protect from TX restarts.
-			acc := &txAccessor{mtu: t, read: preload, tx: tx, rev: writeRev}
+			acc := &txAccessor{updater: t, read: preload, tx: tx, rev: writeRev}
 			var err error
 			root, err = w.Write(ctx, nodesCopy, acc)
 			return err
@@ -109,7 +109,7 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
 }
 
 type txAccessor struct {
-	mtu *mapTreeUpdater
+	updater *mapTreeUpdater
 	// read is a cache of tree tiles that Get returns directly instead of calling
 	// GetTiles. It is initialized by the first Get call if not specified.
 	read *smt.TileSet
@@ -122,7 +122,7 @@ func (t *txAccessor) Get(ctx context.Context, ids []tree.NodeID2) (map[tree.Node
 	// TODO(pavelkalinnikov): Factor out preload into another accessor.
 	if t.read == nil {
 		var err error
-		if t.read, err = t.mtu.load(ctx, t.tx, ids, t.rev-1); err != nil {
+		if t.read, err = t.updater.load(ctx, t.tx, ids, t.rev-1); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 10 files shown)


storage/cloudspanner/map_storage.go
```diff
@@ -263,6 +265,14 @@ func (tx *mapTX) Set(ctx context.Context, index []byte, value *trillian.MapLeaf)
 	return stx.BufferWrite([]*spanner.Mutation{m})
 }
 
+func (m *mapTX) GetHashes(ctx context.Context, rev int64, ids []tree.NodeID2) (*smt.TileSet, error) {
```
golangcibot: receiver name m should be consistent with previous receiver name tx for mapTX (from `golint`)

storage/cloudspanner/map_storage.go
```diff
@@ -263,6 +265,14 @@ func (tx *mapTX) Set(ctx context.Context, index []byte, value *trillian.MapLeaf)
...
+	return nil, errors.New("not implemented")
+}
+
+func (m *mapTX) SetHashes(ctx context.Context, ts *smt.TileSet, upd []smt.NodeUpdate) error {
```
golangcibot: receiver name m should be consistent with previous receiver name tx for mapTX (from `golint`)

storage/tree/node_id2.go
```diff
@@ -37,10 +37,28 @@ func NewNodeID2(path string, bits uint) NodeID2 {
...
+		panic(fmt.Sprintf("NewNodeID2WithLast: bits %d > 8", bits))
+	} else if bits == 0 && len(path) != 0 {
+		panic("NewNodeID2WithLast: bits=0, but path is not empty")
+	}
+	return newMaskedNodeID2(path, last, bits)
+}
+
+// newMaskedNodeID2 constructs a NodeID ensuring its invariants are met.
+func newMaskedNodeID2(path string, last byte, bits uint8) NodeID2 {
+	last &= ^byte(1<<(8-bits) - 1) // Unset the unsed bits.
```
golangcibot: `unsed` is a misspelling of `used` (from `misspell`)
```suggestion

```

pavelkalinnikov:
This PR is now ready for a proper review.

server/map_tree_updater.go
```diff
@@ -110,44 +109,36 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
...
+		}
 	}
-	return res, nil
+	return t.read.Hashes(), nil
 }
 
-func (t txAccessor) Set(ctx context.Context, nodes []smt.Node) error {
-	stNodes := make([]tree.Node, 0, len(nodes))
+func (t *txAccessor) Set(ctx context.Context, nodes []smt.Node) error {
+	mut := smt.NewTileSetMutation(t.read)
```
AlCutter: `mut` -> `m`?
(there's already a `t.mtu` and `mut` is very similar)

server/map_tree_updater.go
```diff
@@ -67,7 +66,7 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
 		err = runTX(ctx, func(ctx context.Context, tx storage.MapTreeTX) error {
 			nodesCopy := make([]smt.Node, len(nodes))
 			copy(nodesCopy, nodes) // Protect from TX restarts.
-			acc := &txAccessor{hashes: hashes, preload: preload, tx: tx, rev: writeRev}
+			acc := &txAccessor{mtu: t, read: preload, tx: tx, rev: writeRev}
```
gdbelvin: give mtu a slightly more descriptive name?

gdbelvin: treeUpdater?

server/map_tree_updater.go
```diff
@@ -110,44 +109,36 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
...
-	// calling GetMerkleNodes. Used only if preload is true.
-	hashes  map[tree.NodeID2][]byte
-	preload bool
-
-	tx  storage.MapTreeTX
-	rev int64
+	mtu *mapTreeUpdater
+	// read is a cache of tree tiles that Get returns directly instead of calling
+	// GetTiles. It is initialized by the first Get call if not specified.
+	read *smt.TileSet
```
gdbelvin: read a verb which makes it sound more like a function. 
Could we make it a noun like readCache?

gdbelvin: Could you make it clearer what happens if the readCache is missing a row? Does that make it a cache miss or does that produce an error or some kind?

server/map_tree_updater.go
```diff
@@ -110,44 +109,36 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
...
-	if t.preload {
-		return t.hashes, nil
-	}
-
-	// TODO(pavelkalinnikov): Pass NodeID2 directly to storage.
-	convIDs := make([]tree.NodeID, 0, len(ids))
-	for _, id := range ids {
-		convIDs = append(convIDs, tree.NewNodeIDFromID2(id))
-	}
-	nodes, err := t.tx.GetMerkleNodes(ctx, t.rev-1, convIDs)
```
gdbelvin: Am I correct in understanding that `GetMerkleNodes` has no callers now?

gdbelvin: Could you mark that API as deprecated?

server/map_tree_updater.go
```diff
@@ -110,44 +109,36 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
...
+			return nil, err
+		}
 	}
-	return res, nil
+	return t.read.Hashes(), nil
 }
 
-func (t txAccessor) Set(ctx context.Context, nodes []smt.Node) error {
-	stNodes := make([]tree.Node, 0, len(nodes))
+func (t *txAccessor) Set(ctx context.Context, nodes []smt.Node) error {
```
gdbelvin: Is set for leaf nodes or interior nodes? It would be nice to have that documented.

gdbelvin: Maybe it's the leaf nodes of interior tiles?

server/map_tree_updater.go
```diff
@@ -168,20 +159,49 @@ func (t *mapTreeUpdater) newTXFunc(tx storage.MapTreeTX) txFunc {
...
+}
+
+// load fetches Merkle tree tiles containing the given node IDs.
+func (t *mapTreeUpdater) load(ctx context.Context, tx storage.MapTreeTX, ids []tree.NodeID2, rev int64) (*smt.TileSet, error) {
+	tileIDs := toTileIDs(ids, t.layout)
+	tiles, err := tx.GetTiles(ctx, rev, tileIDs)
+	if err != nil {
+		return nil, err
+	}
+	ts := smt.NewTileSet(t.tree.TreeId, t.hasher, t.layout)
```
gdbelvin: Is a TileSet supposed to be an immutable object? I'm surprised to see an Add operation

storage/mysql/map_storage.go
```diff
@@ -22,8 +22,12 @@ import (
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/merkle/hashers"
+	"github.com/google/trillian/merkle/smt"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/cache"
+	"github.com/google/trillian/storage/storagepb"
+	"github.com/google/trillian/storage/storagepb/convert"
+	stree "github.com/google/trillian/storage/tree"
```
gdbelvin: group renames below

storage/mysql/tree_storage.go
```diff
@@ -170,6 +169,7 @@ type treeTX struct {
 	treeType      trillian.TreeType
 	hashSizeBytes int
 	subtreeCache  *cache.SubtreeCache
+	dirty         []*storagepb.SubtreeProto
```
gdbelvin: I don't follow why we now have dirty / flush semantics in the storage layer.

server/map_tree_updater.go
pavelkalinnikov: It is still used by logs (sequencer and proofs), so it's not deprecated yet. We will delete this API in `v2` (more specifically, it will be moved to a log-specific storage API).

server/map_tree_updater.go
pavelkalinnikov: There is a TODO for `TileSet` to make it immutable.

storage/mysql/tree_storage.go
pavelkalinnikov: We did have it previously. `SubtreeCache` accumulated all writes, which were then flushed on `treeTX.Commit`.

I found it a bit cheaper to batch-write everything in one RPC rather than one RPC per shard. Also, after some long debugging of MySQL storage I found out that stuff can't be put to a MySQL transaction object concurrently, so this means I either need to lock `t.mu` and write (that's again a lot of lock contention, plus notice that it's shared with reads as well), or do a faster lock `t.mu` + append-to-slice (and only write once in the end).

server/map_tree_updater.go
pavelkalinnikov: Renamed to `updater`.

server/map_tree_updater.go
pavelkalinnikov: Renamed to `tiles`, and improved commenting.

server/map_tree_updater.go
pavelkalinnikov: Added method comment.

server/map_tree_updater.go
```diff
@@ -110,44 +109,41 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, nodes
...
 type txAccessor struct {
-	// hashes is a cache of node hashes that Get returns directly instead of
-	// calling GetMerkleNodes. Used only if preload is true.
-	hashes  map[tree.NodeID2][]byte
-	preload bool
-
-	tx  storage.MapTreeTX
-	rev int64
+	updater *mapTreeUpdater
+	// tiles is a cache of tree tiles that Get uses to fetch node hashes from. It
```
pavelkalinnikov: The comment still needs a bit more clarity.

storage/cloudspanner/map_storage.go
```diff
@@ -110,6 +112,11 @@ func (ms *mapStorage) SnapshotForTree(ctx context.Context, tree *trillian.Tree)
 	return ms.begin(ctx, tree, true, ms.ts.client.ReadOnlyTransaction())
 }
 
+// Layout is not implemented.
+func (ms *mapStorage) Layout(tree *trillian.Tree) (*tree.Layout, error) {
+	return nil, errors.New("not implemented")
```
mhutchinson: Does this (& similar stubs, below) mean that this PR removes cloudspanner support for the Map?

pavelkalinnikov:
> How much faster is this on MySQL?

See the PR description: 20% faster, for some definition of "faster". It is also significantly simpler.

> I'm concerned that this appears to stop the map working at HEAD for cloudspanner. We need to have some story for how to get that functionality back before we can merge this into master.

I can add a note to `CHANGELOG`, and also update the feature matrix in this PR. It will be rather simple to add it back, we need to add implementation of the new 3 method. However, I don't think it is strictly necessary: a) the Map API is "experimental", and so are the implementations; b) nobody uses the `couldspanner` Map; c) we are redesigning the Map anyway; d) The `cloudspanner` Map is not tested.

storage/cloudspanner/map_storage.go
pavelkalinnikov: Yes.

server/map_tree_updater.go
```diff
@@ -168,20 +164,49 @@ func (t *mapTreeUpdater) newTXFunc(tx storage.MapTreeTX) txFunc {
...
-func doPreload(ctx context.Context, tx storage.MapTreeTX, depth uint, upd []smt.Node, rev int64) (map[tree.NodeID2][]byte, error) {
+// doPreload loads all Merkle tree tiles that will be involved when updating
+// the given set of nodes at the specified revision.
+func (t *mapTreeUpdater) doPreload(ctx context.Context, tx storage.MapTreeTX, nodes []smt.Node, rev int64) (*smt.TileSet, error) {
 	ctx, spanEnd := spanFor(ctx, "doPreload")
 	defer spanEnd()
-
 	// TODO(pavelkalinnikov): Avoid using HStar3 directly.
-	hs, err := smt.NewHStar3(upd, nil, depth, 0)
+	hs, err := smt.NewHStar3(nodes, nil, uint(t.hasher.BitLen()), 0)
```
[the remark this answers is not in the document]
pavelkalinnikov: I like this idea. There are a bunch of places throughout the code where this could be done, so let me put a TODO for now and address it in a follow-up.

server/map_tree_updater.go
```diff
@@ -168,20 +164,49 @@ func (t *mapTreeUpdater) newTXFunc(tx storage.MapTreeTX) txFunc {
...
 	if err != nil {
 		return nil, err
 	}
-	acc := txAccessor{tx: tx, rev: rev + 1}
-	return acc.Get(ctx, hs.Prepare())
+	return t.load(ctx, tx, hs.Prepare(), rev)
+}
+
+// load fetches Merkle tree tiles containing the given node IDs.
+func (t *mapTreeUpdater) load(ctx context.Context, tx storage.MapTreeTX, ids []tree.NodeID2, rev int64) (*smt.TileSet, error) {
```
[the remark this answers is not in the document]
pavelkalinnikov: Good idea. Done.

pavelkalinnikov:
Addressed the comments and updated the changelog.

storage/cloudspanner/map_storage.go
mhutchinson: This is awkward because of the 2 customers we need to support with the map, 1/2 of them want to use CloudSpanner. I've raised #2001 to allow this to move ahead.

PR #1933, opened 2019-10-25
Move irrelevant types from storage/tree

pavelkalinnikov:
The callback types are being moved to their original place in `storage` package. This a) helps eliminating circular dependencies in #1932, and b) makes `tree` package agnostic of `SubtreeCache` machinery.

### DIFF
--- a/storage/aliases.go
+++ b/storage/aliases.go
@@ -14,16 +14,25 @@
 
 package storage
 
-import tree "github.com/google/trillian/storage/tree"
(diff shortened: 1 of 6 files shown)


gdbelvin:
Not sure I have all the context. Could you add a bit of a description in the PR description about the current and future desired state of the tree and storage libraries?

pavelkalinnikov:
Updated the PR comment.

PR #1936, opened 2019-10-28
NodeID2: Introduce a low-level constructor

pavelkalinnikov:
This change adds a `NodeID2` constructor that allows the caller to take advantage of `NodeID2` internal structure. For example, `storage` implementations that have their tree tiles aligned to 8-bit heights will be able to construct node IDs of small tiles without any additional memory allocations except for the root.

This change also improves documentation of `NodeID2` in general.

Example of future usage in #1932.

### DIFF
--- a/storage/tree/node_id2.go
+++ b/storage/tree/node_id2.go
@@ -16,15 +16,34 @@ package tree
 
 import "fmt"
 
-// NodeID2 is a faster NodeID that does zero memory allocations in transforming
-// methods like Prefix and Sibling. NodeID2 can be used as a key in Golang maps
-// directly, as well as compared equal.
+// NodeID2 identifies a node of a Merkle tree. It is a bit string that counts
+// the node down from the tree root, i.e. 0 and 1 bits correspond to going to
+// the left or right child correspondingly.
 //
-// TODO(pavelkalinnikov): Rename to NodeID and document it properly when the
-// code has migrated.
+// NodeID2 is immutable, comparable, and can be used as a Golang map key. It
+// also incurs zero memory allocations in transforming methods like Prefix and
+// Sibling.
+//
+// The internal structure of NodeID2 is driven by its use-cases:
+// - To make NodeID2 objects immutable and comparable, the Golang string type
+//   is used for storing the bit string bytes.
+// - To make Sibling and Prefix operations fast, the last byte is stored
+//   separately from the rest of the bytes, so that it can be "amended".
+// - To make NodeID2 objects comparable, there is only one (canonical) way to
+//   encode an ID. For example, if the last byte is used partially, its unused
+//   bits are always unset. See invariants next to field definitions below.
+//
+// Constructors and methods of NodeID2 make sure its invariants are always met.
+//
+// For example, an 11-bit node ID [1010,1111,001] is structured as follows:
+// - path string contains 1 byte, which is [1010,1111].
+// - last byte is [0010,0000]. Note the unset lower 5 bits.
+// - bits is 3, so effectively only the upper 3 bits [001] of last are used.
+//
+// TODO(pavelkalinnikov, v2): Replace NodeID with this type.
 type NodeID2 struct {
 	path string
-	last byte  // Invariant: Lowest (8-bits) bits are unset.
+	last byte  // Invariant: Lowest (8-bits) bits of the last byte are unset.
 	bits uint8 // Invariant: 1 <= bits <= 8, or bits == 0 if len(path) == 0.
 }
 
@@ -55,7 +74,9 @@ func NewNodeID2WithLast(path string, last byte, bits uint8) NodeID2 {
 	return newMaskedNodeID2(path, last, bits)
 }
 
-// newMaskedNodeID2 constructs a NodeID ensuring its invariants are met.
+// newMaskedNodeID2 constructs a NodeID ensuring its invariants are met. The
+// last byte is masked so that the given number of upper bits are in use, and
+// the others are unset.
 func newMaskedNodeID2(path string, last byte, bits uint8) NodeID2 {
 	last &= ^byte(1<<(8-bits) - 1) // Unset the unused bits.
 	return NodeID2{path: path, last: last, bits: bits}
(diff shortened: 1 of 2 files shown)


storage/tree/node_id2.go
```diff
@@ -24,8 +24,8 @@ import "fmt"
 // code has migrated.
 type NodeID2 struct {
 	path string
-	last byte
-	bits uint8
+	last byte  // Invariant: Lowest (8-bits) bits are unset.
```
gdbelvin: last is `byte` which is 8 bits, making this comment confusing.  is `last` always 0 ?

storage/tree/node_id2.go
```diff
@@ -37,10 +37,28 @@ func NewNodeID2(path string, bits uint) NodeID2 {
 	} else if mx := uint(len(path)) * 8; bits > mx {
 		panic(fmt.Sprintf("NewNodeID2: bits %d > %d", bits, mx))
 	}
-	bytes, tail, mask := split(bits)
-	last := path[bytes] & mask
+	bytes, tailBits := split(bits)
 	// Note: Getting the substring is cheap because strings are immutable in Go.
-	return NodeID2{path: path[:bytes], last: last, bits: tail}
+	return newMaskedNodeID2(path[:bytes], path[bytes], tailBits)
```
gdbelvin: What are the semantics of the unset bits of `last`.  The comment above makes it sound like they should be unset, but here they look set.

storage/tree/node_id2.go
```diff
@@ -37,10 +37,28 @@ func NewNodeID2(path string, bits uint) NodeID2 {
...
+func NewNodeID2WithLast(path string, last byte, bits uint8) NodeID2 {
+	if bits > 8 {
+		panic(fmt.Sprintf("NewNodeID2WithLast: bits %d > 8", bits))
+	} else if bits == 0 && len(path) != 0 {
+		panic("NewNodeID2WithLast: bits=0, but path is not empty")
+	}
+	return newMaskedNodeID2(path, last, bits)
+}
+
+// newMaskedNodeID2 constructs a NodeID ensuring its invariants are met.
```
gdbelvin: Clarify what happens to `last`.  Is it masked? forwarded untouched to the struct?

storage/tree/node_id2.go
gdbelvin: Does `path[bytes]` really auto convert from string to byte??

storage/tree/node_id2.go
```diff
@@ -101,15 +118,9 @@ func (n NodeID2) String() string {
...
-// corresponding number of higher bits set.
-func split(bits uint) (uint, uint8, byte) {
-	bytes := (bits - 1) / 8
-	tailBits := uint8(1 + (bits-1)%8)
-	mask := ^byte(1<<(8-tailBits) - 1)
-	return bytes, tailBits, mask
+// split returns the decomposition of a NodeID2 with the given number of bits.
+// The first int returned is the number of full bytes stored in the dynamically
+// allocated part. The second one is the number of bits in the tail byte.
+func split(bits uint) (uint, uint8) {
```
gdbelvin: For increased clarity consider naming the return variables 
```suggestion
func split(bits uint) (bytes uint, tailBits uint8) {
```

storage/tree/node_id2.go
pavelkalinnikov: `last` is not always 0.

Example: An 11-bit ID `1010,1010,101` is interpreted as follows:

- `path` contains 1 byte, which is `1010,1010`.
- `last` is `1010,0000`.
- `bits` is 3, so effectively the upper 3 bits of `last` are used (`101`).

pavelkalinnikov: I might put this example straight into the comments if it looks good to you :)

storage/tree/node_id2.go
pavelkalinnikov: > What are the semantics of the unset bits of `last`. The comment above makes it sound like they should be unset, but here they look set.

The bits are unset in order to have a canonical representation of IDs. This helps using `NodeID2` as a Golang map key, for instance, so that 2 IDs constructed independently can compare equal regardless of the way they were created.

> Does `path[bytes]` really auto convert from string to byte??

`path[bytes]` returns a byte straight away. Golang strings contain bytes.

storage/tree/node_id2.go
pavelkalinnikov: Good one, I will add it.

storage/tree/node_id2.go
pavelkalinnikov: Done.

storage/tree/node_id2.go
pavelkalinnikov: Yes, it's masked. Done.

storage/tree/node_id2.go
pavelkalinnikov: Done.

storage/tree/node_id2.go
pavelkalinnikov: The `last` byte is masked in `newMaskedNodeID2`.

storage/tree/node_id2.go
```diff
@@ -16,16 +16,35 @@ package tree
 
 import "fmt"
 
-// NodeID2 is a faster NodeID that does zero memory allocations in transforming
-// methods like Prefix and Sibling. NodeID2 can be used as a key in Golang maps
-// directly, as well as compared equal.
+// NodeID2 identifies a node of a Merkle tree. It is a bit string that counts
+// the node down from the tree root, i.e. 0 and 1 bits correspond to going to
+// the left or right child correspondingly.
```
pavelkalinnikov: English speakers, does "correspond to ... correspondingly" sound ok?

pavelkalinnikov: suggestion: `0 and 1 bits represent going to the left or right child correspondingly`.

storage/tree/node_id2.go
```diff
@@ -16,16 +16,35 @@ package tree
...
+// - last byte is [0010,0000]. Note the unset lower 5 bits.
+// - bits is 3, so effectively only the upper 3 bits [001] of last are used.
+//
+// TODO(pavelkalinnikov, v2): Replace NodeID with this type.
 type NodeID2 struct {
 	path string
-	last byte
-	bits uint8
+	last byte  // Invariant: Lowest (8-bits) bits of the last byte are unset.
+	bits uint8 // Invariant: 1 <= bits <= 8, or bits == 0 if len(path) == 0.
```
pavelkalinnikov: This comment is a bit inaccurate: `bits` can be 0 **only if** `len(path) == 0` (for the "empty" root ID), but can be non-zero if the ID is 1 to 7 bits long.

pavelkalinnikov:
This PR now combines documentation edits, a refactoring, and introducing a new constructor. Let me know if you would like to see it split.

PR #1937, opened 2019-10-28
map_rpc_server: Add test for non-empty SetLeaves

pavelkalinnikov:
This change adds a higher level test for the map RPC server's `SetLeaves` call with non-empty number of updates. It will also help validating #1902 to a greater level of confidence.

### DIFF
--- a/server/map_rpc_server_test.go
+++ b/server/map_rpc_server_test.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"bytes"
 	"context"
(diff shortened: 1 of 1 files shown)


gdbelvin:
Please also add this test to the map integration tests where there is a real storage layer backing the operation

pavelkalinnikov:
There is sufficient coverage in the integration tests already. This test is here to improve unit-test coverage for this core bit.

PR #1943, opened 2019-11-01
Layout: Get tile height independently of NodeID

pavelkalinnikov:
This change makes `Layout.TileHeight` method more generic, in order for it to be usable with `NodeID2`. A prereq for #1932.

### DIFF
--- a/storage/tree/layout.go
+++ b/storage/tree/layout.go
@@ -14,9 +14,7 @@
 
 package tree
 
-import (
-	"fmt"
-)
+import "fmt"
 
 const (
 	// depthQuantum defines the smallest supported tile height, which all tile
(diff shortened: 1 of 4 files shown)


storage/tree/layout.go
```diff
@@ -97,9 +95,11 @@ func (l *Layout) Split(id NodeID) (TileID, *Suffix) {
 	return TileID{Root: NodeID{Path: []byte{}}}, EmptySuffix
 }
 
-// GetTileHeight returns the height of the tile with the passed-in ID.
-func (l *Layout) GetTileHeight(id TileID) int {
-	return l.getStratumAt(id.Root.PrefixLenBits).height
+// TileHeight returns the height of a tile with its root located at the
+// specified depth from the tree root. The result is not defined if rootDepth
+// is not a tile boundary.
```
mhutchinson: Add TODO to move this back to a state where it takes a structured argument instead of an int.

pavelkalinnikov: Done.

PR #1946, opened 2019-11-01
Do not treat a failure to identify a license as a terminal error

RJPercival:
Simply list that license as "unknown".

### Checklist


RJPercival:
The "save" subcommand will still return an error if the license can't be identified (or is forbidden), and that's what we use in the Docker manifests, so the build will fail. The "csv" subcommand is just used for building a report, which you don't really want failing because of an unknown license (since then you can't report it!). There's also a new "check" subcommand that explicitly checks for forbidden/unknown licenses, which leaves even less reason for this subcommand to complain.

PR #1949, opened 2019-11-04
Suffix: Make parsing stricter

pavelkalinnikov:
This change makes `ParseSuffix` to check that the `base64`-decoded bytes slice has the appropriate size before attempting to decompose/use it. A few more tests and TODOs are added as well.

### DIFF
--- a/storage/tree/suffix_test.go
+++ b/storage/tree/suffix_test.go
@@ -29,36 +29,46 @@ const (
 	// storage/cache when merkle no longer depends on storage.NodeID
 )
 
-//h2b6 takes a hex string and emits a base64 string
+// h2b6 takes a hex string and emits a base64 string.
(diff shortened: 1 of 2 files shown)


storage/tree/suffix_test.go
```diff
@@ -255,7 +265,6 @@ func Test8BitSuffixCache(t *testing.T) {
 		// 8 bits suffix should be cached.
 		{b: 8, path: []byte{0x76}, wantCache: true},
 		// above 8 bits should not be cached.
-		{b: 9, path: []byte{0x40}, wantCache: false},
```
pavelkalinnikov: This was an incorrect `Suffix` as `path` has to have 2 bytes instead of 1.

Martin2112: Well spotted.

PR #1950, opened 2019-11-05
smt: Add conversion between SubtreeProto and Tile

pavelkalinnikov:
This change introduces a new non-proto `Tile` type which will be used for exchanging tiles through the map storage API boundary. Using `SubtreeProto` directly seems like a bad practice for reasons: a) implementations might want to not use proto; b) the internal structure of `SubtreeProto` has `base64`-encoded node IDs which has no reason to be used on the application layer.

This change also adds functions for converting between `Tile` and the legacy `SubtreeProto`.

This is part of #1932 which uses the `Tile` structure.

### DIFF
--- a/merkle/smt/tile.go
+++ b/merkle/smt/tile.go
@@ -1,0 +1,30 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package smt
+
+import "github.com/google/trillian/storage/tree"
+
+// Tile represents a sparse Merkle tree tile, i.e. a dense set of tree nodes
+// located under a single "root" node at a distance not exceeding the tile
+// height. A tile is identified by the ID of its root node. The Tile struct
+// contains the list of non-empty leaf nodes, which can be used to reconstruct
+// all the remaining inner nodes of the tile.
+//
+// TODO(pavelkalinnikov): Introduce invariants on the order/content of Leaves.
+// TODO(pavelkalinnikov): Rename NodeUpdate to a more generic Node or NodeHash.
+type Tile struct {
+	ID     tree.NodeID2
+	Leaves []NodeUpdate
+}
--- a/merkle/smt/tile.go
+++ b/merkle/smt/tile.go
@@ -22,6 +22,7 @@ import "github.com/google/trillian/storage/tree"
 // contains the list of non-empty leaf nodes, which can be used to reconstruct
 // all the remaining inner nodes of the tile.
 //
+// TODO(pavelkalinnikov): Make Tile immutable.
 // TODO(pavelkalinnikov): Introduce invariants on the order/content of Leaves.
 // TODO(pavelkalinnikov): Rename NodeUpdate to a more generic Node or NodeHash.
 type Tile struct {
--- a/storage/storagepb/convert.go
+++ b/storage/storagepb/convert.go
@@ -1,0 +1,88 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


merkle/smt/tile.go
```diff
@@ -0,0 +1,30 @@
...
+// located under a single "root" node at a distance not exceeding the tile
+// height. A tile is identified by the ID of its root node. The Tile struct
+// contains the list of non-empty leaf nodes, which can be used to reconstruct
+// all the remaining inner nodes of the tile.
+//
+// TODO(pavelkalinnikov): Introduce invariants on the order/content of Leaves.
+// TODO(pavelkalinnikov): Rename NodeUpdate to a more generic Node or NodeHash.
+type Tile struct {
+	ID     tree.NodeID2
+	Leaves []NodeUpdate
```
gdbelvin: Unrelated comment, but is there a reason for `[]NodeUpdate` vs `map[NodeID]Hash`?

storage/storagepb/convert.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/google/trillian/merkle/smt"
+	"github.com/google/trillian/storage/tree"
+)
+
+// Unmarshal converts the given SubtreeProto to a Merkle tree tile.
+func Unmarshal(sp *SubtreeProto) (smt.Tile, error) {
+	if d := sp.Depth; d <= 0 {
+		return smt.Tile{}, fmt.Errorf("wrong depth %d, want > 0", d)
+	}
+	height := uint(sp.Depth)
```
gdbelvin: All proto fields should be accessed via the `Get*` functions for API compatibility with future proto changes.

storage/storagepb/convert.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/trillian/merkle/smt"
+	"github.com/google/trillian/storage/tree"
+)
+
+// Unmarshal converts the given SubtreeProto to a Merkle tree tile.
+func Unmarshal(sp *SubtreeProto) (smt.Tile, error) {
```
gdbelvin: Love the Unmarshal semantics. Typically, however the Unmarshal signatures look more like:
```suggestion
func Unmarshal(sp *SubtreeProto, tile *smt.Tile) error {
```
eg. https://golang.org/pkg/encoding/json/#Unmarshal

On the contrary though, those type of signatures are only relevant for general functions where they are doing reflection on the destination type - so disregard. 

The second question about the signature is on the pros/cons of returning a pointer or not. 
Pointer receivers are standard, and allow returning `nil` on error conditions, but have the downside of making escape analysis slightly harder. Is there are reason you've chosen an immediate return type?

merkle/smt/tile.go
mhutchinson: NodeUpdate looks like the wrong name for the type if you want to use it in this context. It's being used as NodeState.

merkle/smt/tile.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Tile represents a sparse Merkle tree tile, i.e. a dense set of tree nodes
+// located under a single "root" node at a distance not exceeding the tile
+// height. A tile is identified by the ID of its root node. The Tile struct
+// contains the list of non-empty leaf nodes, which can be used to reconstruct
+// all the remaining inner nodes of the tile.
+//
+// TODO(pavelkalinnikov): Introduce invariants on the order/content of Leaves.
+// TODO(pavelkalinnikov): Rename NodeUpdate to a more generic Node or NodeHash.
+type Tile struct {
```
mhutchinson: I think it's worth considering making Tile immutable, unless specifically designed to be mutable. At the moment, its fields are public and thus anyone that sees it can fiddle with its state.

storage/storagepb/convert.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, fmt.Errorf("tile root unaligned: %d", prefBits)
+	}
+	prefBytes := prefBits / 8
+	bits := prefBits + height
+
+	leaves := make(map[string][]byte, len(t.Leaves))
+	for _, upd := range t.Leaves {
+		id := upd.ID
+		// TODO(pavelkalinnikov): These must move to be part of Tile contract.
+		if bl := id.BitLen(); bl != bits {
```
gdbelvin: Thanks for doing these checks

storage/storagepb/gen.go
```diff
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package storagepb contains protobuf definitions and helpers used by various
+// storage implementations.
+//
+// TODO(pavelkalinnikov, v2): It's a bad practice to reuse the same protobuf
+// across multiple repositories/implementations. Remove it, and operate with
+// regular in-memory Golang structs through the storage API boundary.
```
mhutchinson: I don't think using protos as the definition of the data to be stored is necessarily a bad thing. There are clear guidelines for versioning protos and checking backwards compatibility. As it is "just" data, there are no methods to also worry about versioning on these structures either.

Having protos for transport between the storage and application layers doesn't force anyone to use them for storage. My advice would be that a storage layer which wants to store protos should define their own proto and copy between the formats. Is this what the comment means about reusing the protobuf in other repos?

gdbelvin:
LGTM from me.  @mhutchinson over to you

pavelkalinnikov:
@mhutchinson I've sneakily submitted this to iterate forward. Feel free to leave more comments, and I will address them in follow-up PRs.

mhutchinson:
This is fine. Thanks.

Also, your sneakiness needs more work.

PR #1951, opened 2019-11-05
mapTreeUpdater: Use the preloaded hashes

pavelkalinnikov:
This change modifies `mapTreeUpdater` so that it uses the result of `GetMerkleNodes` of the preloading phase. It eliminates a bunch of redundant `GetMerkleNodes` calls that previously followed the preload. It also allows avoiding extra `SubtreeCache` coordination happening behind the scenes.

### DIFF
--- a/server/map_tree_updater.go
+++ b/server/map_tree_updater.go
@@ -44,6 +44,7 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, upd [
 	// mode by preloading all the nodes we know the Writers are going to need.
 	var hashes map[tree.NodeID2][]byte
 	preload := t.singleTX && t.preload
+	// Note: It's fine if hashes == nil, it only happens if preload == false.
 	if preload {
 		var err error
 		if hashes, err = doPreload(ctx, tx, uint(t.hasher.BitLen()), upd, writeRev-1); err != nil {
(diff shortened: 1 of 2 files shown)


server/map_tree_updater.go
```diff
@@ -106,11 +109,17 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, upd [
 }
 
 type txAccessor struct {
-	tx  storage.MapTreeTX
-	rev int64
+	hashes  map[tree.NodeID2][]byte
```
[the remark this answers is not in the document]
pavelkalinnikov: Added a comment.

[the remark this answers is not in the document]
pavelkalinnikov: Not checking `nil` cause that seems dangerous. Technically, `doPreload` could return a nil map at some point, even though it's not happening now. Anyway, this `bool`+`map` pair is going to be replaced by a `*TileSet` pointer in a follow-up PR, which has a more natural `nil` value.

server/map_tree_updater.go
```diff
@@ -106,11 +110,20 @@ func (t *mapTreeUpdater) update(ctx context.Context, tx storage.MapTreeTX, upd [
...
+	// hashes is a cache of node hashes that Get returns directly instead of
+	// calling GetMerkleNodes. Used only if preload is true.
+	hashes  map[tree.NodeID2][]byte
+	preload bool
+
 	tx  storage.MapTreeTX
 	rev int64
 }
 
 func (t txAccessor) Get(ctx context.Context, ids []tree.NodeID2) (map[tree.NodeID2][]byte, error) {
```
gdbelvin: This feels like a completely different way of doing Gets.  
Would it make sense to have a *different* txAccessor for the cached reads?
The one off - ifs make me nervous that this will be harder to test.

pavelkalinnikov: Don't worry about the tests, both cases are covered already (see codecov report).

Could do the splitting, but the `Set` method is shared, so I would need some extra machinery around it to have 2 half-identical implementations. Also, there is a high chance that the preload mode will be removed.

Plus, this accessor is not in the final shape, I'm changing it in #1932 to use tileset. How about leaving a TODO to potentially break this type down when some unclear bits are settled?

pavelkalinnikov: Added TODO.

gdbelvin:
This LGTM to me -- what future plans do we have for preloading?

pavelkalinnikov:
> This LGTM to me -- what future plans do we have for preloading? Are we going to nix it?

Yeah, it's likely not needed - my plan is to A/B test it after the refactoring and decide.

PR #1957, opened 2019-11-06
Fix generic comparisons on protobuf messages

dsnet:
Generated protobuf messages contain internal data structures
that general purpose comparison functions (e.g., reflect.DeepEqual,
pretty.Compare, etc) do not properly compare. It is already the case
today that these functions may report a difference when two messages
are actually semantically equivalent.

Fix all usages by either calling proto.Equal directly if
the top-level types are themselves proto.Message, or by calling
cmp.Equal with the cmp.Comparer(proto.Equal) option specified.
This option teaches cmp to use proto.Equal anytime it encounters
proto.Message types.

### Checklist

### DIFF
--- a/integration/maptest/map.go
+++ b/integration/maptest/map.go
@@ -29,7 +29,6 @@ import (
 	"github.com/google/trillian/examples/ct/ctmapper/ctmapperpb"
 	"github.com/google/trillian/testonly"
 	"github.com/google/trillian/types"
-	"github.com/kylelemons/godebug/pretty"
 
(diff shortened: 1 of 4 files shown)


Martin2112:
For #1738.

Martin2112:
Seems to be a problem with travis build:

```
# github.com/google/trillian/server/errors [github.com/google/trillian/server/errors.test]
server/errors/errors_test.go:51:52: undefined: diff
```

dsnet:
Tests fixed. PTAL.

PR #1959, opened 2019-11-07
Move go code out from proto package

AlCutter:
Move golang code out from storagepb proto package. 


pavelkalinnikov:
Is this an anti-pattern to have both the proto and helpers in one package?

Could you also add package comment, and fix the `storagepb` package comment in `gen.go`?

AlCutter:
> Is this an anti-pattern to have both the proto and helpers in one package?
> 
> Could you also add package comment, and fix the `storagepb` package comment in `gen.go`?

This can cause build problems in some environments, e.g. BAZEL.

Added package comment & trimmed down gen.go commentary.

Issue #1952, opened 2019-11-05
Fully declarative deployment configs for Google Cloud


RJPercival:
Our deployment process currently involves running shell scripts in examples/deployment/kubernetes, which use a combination of tools to build Docker images, setup a Kubernetes cluster on GCP, configure it and deploy Trillian. This could be improved by using a tool such as [Terraform](https://www.terraform.io/) to replace the shell scripts and either relying on [our published Docker images](https://gcr.io/trillian-opensource-ci) or using Google Cloud Build to create the images. [Kustomize](https://kustomize.io/) could be used to support a wider variety of deployments (currently, we use tenvsubst for a similar purpose, but it is less flexible).

RJPercival:
We now have a Terraform config for deploying Trillian to GCP with a Cloud Spanner database. However, it's still necessary to use the shell scripts if you want to use a MySQL database instead.

Issue #1820, opened 2019-09-11
"etcdclusters.etcd.database.coreos.com" not found when running create.sh


tweksteen:
Following the steps from https://github.com/google/trillian/tree/master/examples/deployment/kubernetes/README.md, the create.sh scripts ends with:
```
[...]
clusterrolebinding.rbac.authorization.k8s.io/etcd-cluster-admin-binding created
clusterrolebinding.rbac.authorization.k8s.io/etcd-operator created
clusterrole.rbac.authorization.k8s.io/etcd-operator created
deployment.extensions/trillian-etcd-operator created
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "etcdclusters.etcd.database.coreos.com" not found
```

This might be more a warning than an error, as I am able to continue with the next steps.

paulmattei:
Expected. Documented in Deployment [Readme](https://github.com/google/trillian/blob/master/deployment/README.md)

More info:
https://github.com/google/trillian/pull/1962#discussion_r353918274

PR #1962, opened 2019-11-11
Add supported deployment with Terraform

tweksteen:
Create a root directory with a default supported deployment (based on
previous resources from examples/deployment). The intent is to have
this directory for all the supported deployment types and keep examples/
for deployments that are considered community-supported.

The first supported deployment is Trillian on Google Cloud using Cloud
Spanner as a backend.

This is a follow up from the discussions in #1942 and contributes to #1952.

### Checklist

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -1,0 +1,46 @@
+# Trillian supported deployments
+
+This directory contains the supported deployment types for Trillian. Currently,
+this includes:
+
+- [Kubernetes on GCP using Cloud
+  Spanner](#kubernetes-on-gcp-using-cloud-spanner)
+
+Further deployment types (community-supported) can be found in the
+examples/deployment directory.
+
+# Kubernetes on GCP using Cloud Spanner
+
+## Architecture
+
+The infrastructure is created using Terraform. It consists of a Cloud Spanner
+database, a Trillian-specific etcd cluster, the Trillian logserver and
+logsigner. A Workload Identity is set up by Terraform to give the services access
+to Cloud Spanner. The container images are based on v1.3.3.
+
+## Prerequisites
+1. [Terraform](https://www.terraform.io/), [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/),
+   and [gcloud](https://cloud.google.com/sdk/gcloud/) are installed.
+1. You have a Google account with billing configured.
+1. Terraform should have access to the credentials to manage your Cloud Project. Follow the
+   instructions from the [Terraform Google Provider documentation](https://www.terraform.io/docs/providers/google/guides/provider_reference.html#full-reference) to set this up.
+
+## Installation
+1. Create a new project and copy its [Project
+   ID](https://cloud.google.com/resource-manager/docs/creating-managing-projects#identifying_projects).
+1. Create the infrastructure: `terraform init && terraform apply -var="gcp_project=PROJECT_ID"`.
+   It is ok to re-execute this command at any time.
+1. The script will output a gcloud command similar to:
+   ```shell
+   gcloud config set project PROJECT_ID && \
+   gcloud container clusters get-credentials cluster --region=us-west1
+   ```
+   Execute this set of commands to get the credentials for your new cluster.
+1. Deploy the containers: `kubectl apply -k .` This last command may fail with
+   an [error](https://github.com/google/trillian/issues/1820) similar to:
+   `unable to recognize ".": no matches for kind "EtcdCluster" in version "etcd.database.coreos.com/v1beta2"`.
+   In this case, simply re-run it until the error is not reported anymore.
+
+## Uninstall
+1. Destroy the infrastructure: `terraform destroy`.
+
--- a/deployment/deploy-config.yaml
+++ b/deployment/deploy-config.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: deploy-config
+data:
+  STORAGE_SYSTEM: cloud_spanner
+  STORAGE_FLAG: --cloudspanner_uri=projects/PROJECT_ID/instances/trillian-spanner/databases/trillian-db
+  SIGNER_DEQUEUE_BUCKET_FRACTION: "--cloudspanner_dequeue_bucket_fraction=0.0078"
+  SIGNER_BATCH_SIZE: "--batch_size=1800"
+  SIGNER_INTERVAL: "--sequencer_interval=2ms"
+  SIGNER_NUM_SEQUENCERS: "--num_sequencers=10"
+  SIGNER_MASTER_HOLD_JITTER: "--master_hold_jitter=7200s"
--- a/deployment/log-signer-deployment.yaml
+++ b/deployment/log-signer-deployment.yaml
@@ -1,0 +1,76 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/name: trillian
+    app.kubernetes.io/component: logsigner
+  name: trillian-log-signer-deployment
+spec:
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: trillian
+        app.kubernetes.io/component: logsigner
+    spec:
+      serviceAccountName: trillian-svc
+      restartPolicy: Always
+      containers:
+      - name: trillian-log-signer
+        args: [
+        "$(STORAGE_FLAG)",
+        "--storage_system=$(STORAGE_SYSTEM)",
+        "--etcd_servers=trillian-etcd-cluster-client:2379",
+        "--quota_system=etcd",
+        "--etcd_http_service=trillian-logsigner-http",
+        "--http_endpoint=0.0.0.0:8091",
+        "--sequencer_guard_window=1s",
+        "$(SIGNER_INTERVAL)",
+        "$(SIGNER_NUM_SEQUENCERS)",
+        "$(SIGNER_BATCH_SIZE)",
+        "$(SIGNER_DEQUEUE_BUCKET_FRACTION)",
+        "$(SIGNER_MASTER_HOLD_JITTER)",
+        "--alsologtostderr"
+        ]
+        envFrom:
+        - configMapRef:
+            name: deploy-config
+        image: gcr.io/trillian-opensource-ci/log_signer:v1.3.3
+        imagePullPolicy: Always
+        resources:
+          limits:
+            cpu: "1"
+          requests:
+            cpu: "1"
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8091
+          failureThreshold: 3
+          periodSeconds: 30
+          timeoutSeconds: 5
+        ports:
+        - containerPort: 8091
+          name: http-metrics
+      - name: prometheus-to-sd
+        image: gcr.io/google-containers/prometheus-to-sd:v0.5.2
+        ports:
+          - name: profiler
+            containerPort: 6060
+        command:
+          - /monitor
+          - --stackdriver-prefix=custom.googleapis.com
+          - --source=logsigner:http://localhost:8091/metrics
+          - --pod-id=$(POD_NAME)
+          - --namespace-id=$(POD_NAMESPACE)
+          - --scrape-interval=5s
+          - --export-interval=60s
+        env:
+          - name: POD_NAME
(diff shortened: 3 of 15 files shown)


deployment/deploy-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: deploy-config
+data:
+  STORAGE_SYSTEM: cloud_spanner
+  STORAGE_FLAG: --cloudspanner_uri=projects/PROJECT_ID/instances/trillian-spanner/databases/trillian-db
+  SIGNER_DEQUEUE_BUCKET_FRACTION: "0.0078"
+  SIGNER_BATCH_SIZE: "1800"
+  SIGNER_INTERVAL: "2ms"
+  SIGNER_NUM_SEQUENCERS: "10"
+  SIGNER_MASTER_HOLD_JITTER: "7200s"
```
RJPercival: These SIGNER_* fields should include the flag names, e.g. https://github.com/google/trillian/blob/a6efc69d95ffe1465f7efa256d2fd294de588787/examples/deployment/kubernetes/trillian-ci-spanner.yaml#L16-L20

deployment/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Further deployment types (community-supported) can be found in the
+examples/deployment directory.
+
+# Kubernetes on GCP using Cloud Spanner
+
+## Architecture
+
+The infrastructure is created using Terraform. It consists of a Cloud Spanner
+database, a Trillian-specific etcd cluster, the Trillian logservice and
```
RJPercival:
```suggestion
database, a Trillian-specific etcd cluster, the Trillian logserver and
```

deployment/README.md
```diff
@@ -0,0 +1,41 @@
...
+logsigner. A Workload Identity is setup by Terraform to give the services access
```
RJPercival:
```suggestion
logsigner. A Workload Identity is set up by Terraform to give the services access
```

deployment/README.md
```diff
@@ -0,0 +1,41 @@
...
+to Cloud Spanner. The container images are the latest official release.
```
RJPercival: "the latest official release" - this could easily become inaccurate if someone creates a new release but doesn't update this doc. Probably better to just say "v1.3.3" here.

tweksteen:
I should point out the known issue of failure due to timeout when enabling the different API within the Cloud Project. This can be solved by re-running `terraform apply` in case of failure.

There is also a similar issue with `kubectl` when adding the roles related to etcd. Again, re-running `kubectl` will eventually reach the desired state.

deployment/log-signer-deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+        "$(SIGNER_NUM_SEQUENCERS)",
+        "$(SIGNER_BATCH_SIZE)",
+        "$(SIGNER_DEQUEUE_BUCKET_FRACTION)",
+        "$(SIGNER_MASTER_HOLD_JITTER)",
+        "--alsologtostderr"
+        ]
+        envFrom:
+        - configMapRef:
+            name: deploy-config
+        image: gcr.io/trillian-opensource-ci/log_signer:v1.3.3
```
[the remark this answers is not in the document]
RJPercival: Ok, I mostly just wanted to know how we'd make it configurable, since we'd need to do so if we wanted to use this for CI deployments. It's fine to defer that for now though.

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+  type = string
+}
+
+variable "region" {
+  type    = string
+  default = "us-west1"
+}
+
+provider "google" {
+  project = var.gcp_project
```
RJPercival: nit: Terraform recommends adding `version = "~> 3.0"` here.

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+}
+
+provider "google-beta" {
+  project = var.gcp_project
```
RJPercival: nit: Terraform recommends adding `version = "~> 3.0"` here.

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+# Main cluster definition
+
+## Force recent version for Kubernetes master
+data "google_container_engine_versions" "gke-ver" {
+  location       = var.region
+  version_prefix = "1.14."
+}
+
+resource "google_container_cluster" "trillian-cluster" {
+  provider           = "google-beta"
```
RJPercival:
```
Warning: Quoted references are deprecated

  on main.tf line 37, in resource "google_container_cluster" "trillian-cluster":
  37: 	provider = "google-beta"

In this context, references are expected literally rather than in quotes.
Terraform 0.11 and earlier required quotes, but quoted references are now
deprecated and will be removed in a future version of Terraform. Remove the
quotes surrounding this reference to silence this warning.
```

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+  name               = "cluster"
+  location           = var.region
+  node_version       = "${data.google_container_engine_versions.gke-ver.latest_node_version}"
```
RJPercival:
```
Warning: Interpolation-only expressions are deprecated

  on main.tf line 40, in resource "google_container_cluster" "trillian-cluster":
  40:   node_version = "${data.google_container_engine_versions.gke-ver.latest_node_version}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
```

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+  min_master_version = "${data.google_container_engine_versions.gke-ver.latest_node_version}"
```
RJPercival:
```
Warning: Interpolation-only expressions are deprecated

  on main.tf line 41, in resource "google_container_cluster" "trillian-cluster":
  41:   min_master_version = "${data.google_container_engine_versions.gke-ver.latest_node_version}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
```

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+
+resource "google_spanner_instance" "trillian-spanner" {
+  name         = "trillian-spanner"
+  display_name = "Trillian Spanner Instance"
+  config       = "regional-${var.region}"
+  num_nodes    = 1
+}
+
+resource "google_spanner_database" "trillian-db" {
+  instance = "${google_spanner_instance.trillian-spanner.name}"
```
RJPercival:
```
Warning: Interpolation-only expressions are deprecated

  on main.tf line 78, in resource "google_spanner_database" "trillian-db":
  78:   instance = "${google_spanner_instance.trillian-spanner.name}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
```

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+
+resource "google_project_iam_binding" "trillian-sc-metricwriter" {
+  role = "roles/monitoring.metricWriter"
+  members = [
+    "serviceAccount:${google_service_account.trillian.email}",
+  ]
+}
+
+resource "google_service_account_iam_binding" "trillian-sc-identity" {
+  service_account_id = "${google_service_account.trillian.name}"
```
RJPercival:
```
Warning: Interpolation-only expressions are deprecated

  on main.tf line 114, in resource "google_service_account_iam_binding" "trillian-sc-identity":
 114:   service_account_id = "${google_service_account.trillian.name}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
```

deployment/README.md
```diff
@@ -0,0 +1,43 @@
...
+1. [Terraform](https://www.terraform.io/), [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/),
+   and [gcloud](https://cloud.google.com/sdk/gcloud/) are installed.
+1. You have a Google account with billing configured.
+1. Terraform should have access to the credentials to manage your Cloud Project. Follow the
+   instructions from the [Terraform Google Provider documentation](https://www.terraform.io/docs/providers/google/guides/provider_reference.html) to set this up.
+
+## Installation
+1. Create a new project and copy its [Project
+   ID](https://cloud.google.com/resource-manager/docs/creating-managing-projects#identifying_projects).
+1. Create the infrastructure: `terraform init && terraform apply -var="gcp_project=PROJECT_ID"`.
```
RJPercival: When executing this step, I get the following error:
```
Error: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

  on main.tf line 10, in provider "google":
  10: provider "google" {
```
It's not completely obvious which instructions, if any, I should follow on the linked page.

RJPercival: I found that running ` gcloud auth application-default login` fixed this. The complete instructions are here: https://www.terraform.io/docs/providers/google/guides/provider_reference.html#full-reference. I think it'd be helpful to add this as a step above.

deployment/main.tf
```diff
@@ -0,0 +1,135 @@
...
+locals {
+  trillian_ddl = file("${path.module}/../storage/cloudspanner/spanner.sdl")
+}
+
+resource "google_spanner_instance" "trillian-spanner" {
+  name         = "trillian-spanner"
+  display_name = "Trillian Spanner Instance"
+  config       = "regional-${var.region}"
+  num_nodes    = 1
+}
```
RJPercival: I think you need to add a dependency on `google_project_service.spanner-api` here:
```terraform
depends_on = [
  google_project_service.spanner-api,
]
```
Without it, this resource won't wait for the Spanner API to be enabled, resulting in errors the first time to run `terraform apply`. I think there are a few other resources that require APIs to be enabled here that don't express that dependency.

deployment/README.md
RJPercival: Ah I see this was linked to from the prerequisites section. I'd add that #full-reference anchor to that link, since that takes the reader straight to the relevant instructions.

deployment/README.md
```diff
@@ -0,0 +1,43 @@
...
+   It is ok to re-execute this command at any time.
+1. The script will output a gcloud command similar to:
+   ```shell
+   gcloud config set project PROJECT_ID && \
+   gcloud container clusters get-credentials cluster --region=us-west1
+   ```
+   Execute this set of commands to get the credentials for your new cluster.
+1. Deploy the containers: `kubectl apply -k .`
```
RJPercival: This command fails with an error the first time you run it:
```
error: unable to recognize ".": no matches for kind "EtcdCluster" in version "etcd.database.coreos.com/v1beta2"
```
This is because etcd-operator takes a short time to install its CRD. Consider splitting this step up into a "deploy etcd-operator" step and then a "deploy everything else" step.

tweksteen: Right, this is the same error I reported in #1820. By splitting the two commands, there is still a chance that the CRD is not yet created. From the examples/ directory, we know that the `kubectl wait for` solution do not work (NotFound may be returned instantly). 
I extracted the CRD from etcd-operator and explicitly added it to kustomize. This should decrease the time required to create the CRD (as we don't wait for the operator to be up). Either way, I modified README.md to describe this issue. I don't think there is a better solution at this point.

PR #1963, opened 2019-11-11
Pick safe values for batch sizes #1845

mhutchinson:
These tests would likely fail if they were run for more operations. We have found that batch sizes of <= 100 is reliable, so we should stick to that for stability and clarity.


gdbelvin:
Is it worth adding a runtime check to ensure batch sizes are within safe bounds?

PR #1972, opened 2019-11-15
smt: Introduce MapHasher wrapper

pavelkalinnikov:
This change introduces an unexported map hasher wrapper in `smt` package, to simplify the usage of (hasher, treeID) pair throughout the package and make it safer.

### DIFF
--- a/merkle/smt/hasher.go
+++ b/merkle/smt/hasher.go
@@ -25,8 +25,8 @@ type mapHasher struct {
 	treeID int64
 }
 
-// wrapHasher returns a mapHasher binding the given hasher to a tree ID.
-func wrapHasher(hasher hashers.MapHasher, treeID int64) mapHasher {
+// bindHasher returns a mapHasher binding the given hasher to a tree ID.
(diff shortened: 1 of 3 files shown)


merkle/smt/hasher.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// mapHasher is a wrapper around MapHasher bound to a specific tree ID.
+type mapHasher struct {
+	mh     hashers.MapHasher
+	treeID int64
+}
+
+// wrapHasher returns a mapHasher binding the given hasher to a tree ID.
+func wrapHasher(hasher hashers.MapHasher, treeID int64) mapHasher {
```
pavelkalinnikov: Maybe `bindHasher`?

pavelkalinnikov:
PTAL.

merkle/smt/hasher.go
Martin2112: I think `bindHasher` would be clearer. `forTree` etc. doesn't quite work as you pass the hasher in.

PR #1978, opened 2019-11-26
Set golangci-lint --exclude-use-default=false

RJPercival:
This enables some checks that golangci-lint disables by default, such as checking for package comments. If it turns out that any of these checks are too noisy, we can re-enable them. However, right now, it's causing lint checks to fail when we try to import our code into Google (where these checks are not disabled).

See https://github.com/golangci/golangci-lint/blob/v1.21.0/README.md#command-line-options

### Checklist

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -32,3 +32,7 @@ linters:
     - varcheck
     # TODO(gbelvin): write license linter and commit to upstream.
     # ./scripts/check_license.sh is run by ./scripts/presubmit.sh
+
+issues:
+  # Don't turn off any checks by default. We can do this explicitly if needed.
+  exclude-use-default: false
--- a/integration/maptest/map.go
+++ b/integration/maptest/map.go
@@ -581,6 +581,7 @@ func RunInclusion(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdm
 	}
 }
 
+// RunGetLeafByRevisionNoProof fails the test if the map server does not respond correctly to a GetLeavesByRevision request.
 func RunGetLeafByRevisionNoProof(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient, twrite trillian.TrillianMapWriteClient) {
 	tree, err := newTreeWithHasher(ctx, tadmin, tmap, trillian.HashStrategy_TEST_MAP_HASHER)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


RJPercival:
This results in the following new lint errors:
```
storage/testdb/testdb.go:57:1: comment on exported function `SetFDLimit` should be of the form `SetFDLimit ...` (golint)
// SetFDULimit sets the soft limit on the maximum number of open file descriptors.
^
storage/aliases.go:25:6: exported type `NodeID` should have comment or be unexported (golint)
type NodeID = tree.NodeID
     ^
storage/aliases.go:26:6: exported type `Node` should have comment or be unexported (golint)
type Node = tree.Node
     ^
storage/aliases.go:27:6: exported type `Suffix` should have comment or be unexported (golint)
type Suffix = tree.Suffix
     ^
storage/aliases.go:38:2: exported var `NewNodeIDFromHash` should have comment or be unexported (golint)
	NewNodeIDFromHash         = tree.NewNodeIDFromHash
	^
integration/storagetest/maptestrunner.go:26:6: exported type `MapStorageFactory` should have comment or be unexported (golint)
type MapStorageFactory func(ctx context.Context, t *testing.T) (storage.MapStorage, storage.AdminStorage)
     ^
integration/storagetest/maptestrunner.go:27:6: exported type `MapStorageTest` should have comment or be unexported (golint)
type MapStorageTest func(ctx context.Context, t *testing.T, s storage.MapStorage, as storage.AdminStorage)
     ^
integration/storagetest/maptests.go:31:1: exported method `MapTests.TestCheckDatabaseAccessible` should have comment or be unexported (golint)
func (*MapTests) TestCheckDatabaseAccessible(ctx context.Context, t *testing.T, s storage.MapStorage, _ storage.AdminStorage) {
^
integration/storagetest/maptests.go:37:1: exported method `MapTests.TestMapSnapshot` should have comment or be unexported (golint)
func (*MapTests) TestMapSnapshot(ctx context.Context, t *testing.T, s storage.MapStorage, as storage.AdminStorage) {
^
integration/maptest/map.go:584:1: exported function `RunGetLeafByRevisionNoProof` should have comment or be unexported (golint)
func RunGetLeafByRevisionNoProof(ctx context.Context, t *testing.T, tadmin trillian.TrillianAdminClient, tmap trillian.TrillianMapClient, twrite trillian.TrillianMapWriteClient) {
^
```

gdbelvin:
Could I suggest a slightly alternate approach? That of explicitly listing the linters you want enabled. 
This approach will cause new linters to be added when we update golangci-lint. Explicitly listing the linters we use would be slightly more stable and hermetic.

RJPercival:
This doesn't change the linters that are enabled. It simply removes a set of default exclusion regexps that are used to filter out error messages produced by the linters we do have enabled.

gdbelvin:
Ah. That’s neat!

PR #1979, opened 2019-11-26
Add a presubmit to prevent .pb.txt and .proto.txt files being added

Martin2112:
Add a section to `presubmit.sh` that will fail if there are any `.pb.txt` or `.proto.txt` files present. These should be named `.textproto`.

### Checklist

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -77,7 +77,7 @@ main() {
   # Prevent the creation of proto files with .txt extensions.
   bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
   if [[ "${#bad_protos}" -ne 0 ]]; then
-    echo "Text based protos must use the textproto extension:"
+    echo "Text-based protos must use the .textproto extension:"
     echo $bad_protos
     exit 1
   fi
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -74,6 +74,14 @@ main() {
     grep -v _string.go | \
     tr '\n' ' ')"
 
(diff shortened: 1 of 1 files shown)


scripts/presubmit.sh
```diff
@@ -74,6 +74,14 @@ main() {
     grep -v _string.go | \
     tr '\n' ' ')"
 
+  # Prevent the creation of proto files with .txt extensions.
+  bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
+  if [[ "${#bad_protos}" -ne 0 ]]; then
+    echo "Text based protos must use the textproto extension:"
```
RJPercival: Just to make it obvious that ".textproto" *is* the textproto extension...

```suggestion
    echo "Text-based protos must use the .textproto extension:"
```

scripts/presubmit.sh
```diff
@@ -74,6 +74,14 @@ main() {
     grep -v _string.go | \
     tr '\n' ' ')"
 
+  # Prevent the creation of proto files with .txt extensions.
+  bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
```
RJPercival: I imagine people could very well come up with other variations, but this is a good starting point.

RJPercival:
I think you'll need to rebase on master and let Travis retrigger in order for it to pass - sorry about that.

scripts/presubmit.sh
Martin2112: Yep, these are the two that could cause import problems. I'll do the other repo script similarly next.

PR #1983, opened 2019-11-26
[treetex] Add more control over the generated tree

AlCutter:
Adds a little more control over the tree rendered by `treetex`:
 - allow the setting of leaf data (rather than simply indices)
 - allow some of the node styles to be overridden

### Checklist

### DIFF
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -183,7 +183,7 @@ func perfect(prefix string, height uint, index uint64, nodeText nodeTextFunc) {
 // drawLeaf emits TeX code to render a leaf.
 func drawLeaf(prefix string, index uint64, text string) {
 	a := nInfo[compact.NewNodeID(0, index)]
-	fmt.Printf("%s [%s, %s, tier=leaf]\n", prefix, text, a.String())
+	fmt.Printf("%s [%s, %s, align=center, tier=leaf]\n", prefix, text, a.String())
 }
 
 // openInnerNode renders TeX code to open an internal node.
@@ -328,7 +328,7 @@ var nodeFormats = map[string]nodeTextFunc{
 	"hash": func(id compact.NodeID) string {
 		childLevel := id.Level - 1
 		leftChild := id.Index * 2
-		return fmt.Sprintf("{$H_{%d.%d} = \\\\ H(H_{%d.%d} || H_{%d.%d})$}", id.Level, id.Index, childLevel, leftChild, childLevel, leftChild+1)
+		return fmt.Sprintf("{$H_{%d.%d} =$ \\\\ $H(H_{%d.%d} || H_{%d.%d})$}", id.Level, id.Index, childLevel, leftChild, childLevel, leftChild+1)
 	},
 }
 
@@ -345,7 +345,7 @@ func main() {
 
 	var ldf nodeTextFunc = func(id compact.NodeID) string {
 		if id.Level == 0 {
-			return fmt.Sprintf("%d", id.Index)
+			return fmt.Sprintf("{$H_{0.%d} =$ \\\\ $H(leaf_{%[1]d})$}", id.Index)
 		}
 		return innerNodeText(id)
 	}
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -81,6 +81,7 @@ const (
 
 var (
 	treeSize  = flag.Uint64("tree_size", 23, "Size of tree to produce")
+	leafData  = flag.String("leaf_data", "", "Comma separated list of leaf data text (setting this overrides --tree_size")
 	inclusion = flag.Int64("inclusion", -1, "Leaf index to show inclusion proof")
 	megaMode  = flag.Uint("megamode_threshold", 4, "Treat perfect trees larger than this many layers as a single entity")
 	ranges    = flag.String("ranges", "", "Comma-separated Open-Closed ranges of the form L:R")
@@ -99,6 +100,8 @@ type nodeInfo struct {
 	leaf         bool
 	rangeIndices []int
 }
+
+type nodeTextFunc func(id compact.NodeID) string
 
 // String returns a string containing Forest attributes suitable for
 // rendering the node, given its type.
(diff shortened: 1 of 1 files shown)


docs/merkletree/treetex/main.go
```diff
@@ -185,16 +192,20 @@ func drawLeaf(prefix string, index uint64) {
 // Returns a func to be called to close the node.
 func openInnerNode(prefix string, id compact.NodeID) func() {
 	attr := nInfo[id].String()
-	fmt.Printf("%s [%d.%d, %s, tier=%d\n", prefix, id.Level, id.Index, attr, id.Index-1)
+	fmt.Printf("%s [%d.%d, %s, tier=%d\n", prefix, id.Level, id.Index, attr, id.Level)
```
pavelkalinnikov: I saw this change in a different PR. Should it be deleted from this one?

docs/merkletree/treetex/main.go
```diff
@@ -315,6 +330,15 @@ func main() {
 	flag.Parse()
 	height := uint(bits.Len64(*treeSize-1)) + 1
 
+	var ldf leafDataFunc = func(idx uint64) string { return fmt.Sprintf("%d", idx) }
+
+	if len(*leafData) > 0 {
+		leaves := strings.Split(*leafData, ",")
+		*treeSize = uint64(len(leaves))
+		log.Printf("Overriding treeSize to %d since --leafData was set", *treeSize)
```
pavelkalinnikov: `leaf_data`?

docs/merkletree/treetex/main.go
```diff
@@ -185,16 +192,20 @@ func drawLeaf(prefix string, index uint64) {
...
 
 // perfectInner renders the nodes of a perfect internal subtree.
-func perfectInner(prefix string, level uint, index uint64, top bool) {
+func perfectInner(prefix string, level uint, index uint64, top bool, leafData leafDataFunc) {
 	id := compact.NewNodeID(level, index)
 	modifyNodeInfo(id, func(n *nodeInfo) {
 		n.leaf = id.Level == 0
 		n.perfectRoot = top
+		if id.Level == 0 {
+			n.leaf = true
```
pavelkalinnikov: This line is redundant (see 3 lines above).

docs/merkletree/treetex/main.go
AlCutter: yeah, it was just stacked on top - the other one is merged and this one rebased now.

docs/merkletree/treetex/main.go
AlCutter: done

docs/merkletree/treetex/main.go
AlCutter: I meant to remove the one above - dinged it now.

docs/merkletree/treetex/main.go
```diff
@@ -169,54 +175,56 @@ func perfectMega(prefix string, height uint, leafIndex uint64) {
...
 
 // openInnerNode renders TeX code to open an internal node.
 // The caller may emit any number of child nodes before calling the returned
 // func to close the node.
 // Returns a func to be called to close the node.
-func openInnerNode(prefix string, id compact.NodeID) func() {
+func openInnerNode(prefix string, id compact.NodeID, text string) func() {
 	attr := nInfo[id].String()
-	fmt.Printf("%s [%d.%d, %s, tier=%[2]d\n", prefix, id.Level, id.Index, attr)
+	fmt.Printf("%s [%s, %s, tier=%d\n", prefix, text, attr, id.Level)
```
pavelkalinnikov: Does it look better without `id.Index`?

docs/merkletree/treetex/main.go
```diff
@@ -318,6 +326,32 @@ func main() {
 	flag.Parse()
 	height := uint(bits.Len64(*treeSize-1)) + 1
 
+	var innerNodeText = func(id compact.NodeID) string {
+		if id.Level == 0 {
+			panic("attempting to use inner node text func for leaf text")
+		}
```
pavelkalinnikov: I don't think these 3 lines are necessary.

docs/merkletree/treetex/main.go
```diff
@@ -318,6 +326,32 @@ func main() {
...
+
+	var ldf nodeTextFunc = func(id compact.NodeID) string {
+		if id.Level == 0 {
+			return fmt.Sprintf("%d", id.Index)
+		}
+		return innerNodeText(id)
+	}
+
+	if len(*leafData) > 0 {
+		leaves := strings.Split(*leafData, ",")
```
pavelkalinnikov: How about pre-computing `leaves` for the other case? Then the node function can be shared, and you can merge all 3 function into 1 nice little function.

docs/merkletree/treetex/main.go
AlCutter: That's built in (or not, depending on the formatting chosen) to the `text` label now

docs/merkletree/treetex/main.go
AlCutter: They got dinged already (although it looks like I got distracted and didn't push the updated branch :/)

docs/merkletree/treetex/main.go
```diff
@@ -312,12 +321,47 @@ func modifyRangeNodeInfo() error {
...
+
 // Whee - here we go!
 func main() {
 	// TODO(al): check flag validity.
 	flag.Parse()
 	height := uint(bits.Len64(*treeSize-1)) + 1
 
+	innerNodeText := nodeFormats[*nodeFormat]
+	if innerNodeText == nil {
+		log.Fatalf("unknown --node_format %s", *nodeFormat)
```
pavelkalinnikov: `Exitf`?

docs/merkletree/treetex/main.go
AlCutter: My thinking was that if you're specifying `--leaf_data` then there aren't going to be that many leaves, so keeping them in a slice was fine, but otherwise you can set `--tree_size` to 100,000 (and it'll still work because of the "perfect tree collapsing"), but it'll only need a handful of _actual_ leaves to display.

e.g.:
image

pavelkalinnikov: SNTBAATM (sounds not too bad at all to me)

docs/merkletree/treetex/main.go
AlCutter: For better or worse this is golang `log` rather than `glog` which doesn't have `Exitf` :/

docs/merkletree/treetex/main.go
```diff
@@ -312,12 +321,47 @@ func modifyRangeNodeInfo() error {
...
+	}
+
+	var ldf nodeTextFunc = func(id compact.NodeID) string {
+		if id.Level == 0 {
+			return fmt.Sprintf("%d", id.Index)
```
pavelkalinnikov: Is it okay that this `Sprintf` does not depend on node format?

docs/merkletree/treetex/main.go
pavelkalinnikov: Ah, indeed. Ok, leave it.

docs/merkletree/treetex/main.go
```diff
@@ -312,12 +321,47 @@ func modifyRangeNodeInfo() error {
 	return nil
 }
 
+var nodeFormats = map[string]nodeTextFunc{
```
pavelkalinnikov: I feel like the whole formatting thing could be factored out. There is probably a bigger issue with the `main.go` file containing too much stuff already.

docs/merkletree/treetex/main.go
AlCutter: Yup - that's for the leaves, which defaults to indices unless the user specifies literals with `--leaf_data`

docs/merkletree/treetex/main.go
AlCutter: That's probably true, but maybe for another time...

PR #1985, opened 2019-11-27
[treetex] Add leaf data nodes too

AlCutter:
Add box nodes representing leaf data on `treetex` images.

e.g.:

image

### Checklist

### DIFF
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -362,11 +362,8 @@ func main() {
 		leaves := strings.Split(*leafData, ",")
 		*treeSize = uint64(len(leaves))
 		log.Printf("Overriding treeSize to %d since --leaf_data was set", *treeSize)
-		nodeText = func(id compact.NodeID) string {
-			if id.Level == 0 {
-				return leaves[id.Index]
-			}
-			return innerNodeText(id)
+		dataFormat = func(id compact.NodeID) string {
+			return leaves[id.Index]
 		}
 	}
 
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -334,6 +334,8 @@ var nodeFormats = map[string]nodeTextFunc{
 		return fmt.Sprintf("%d.%d", id.Level, id.Index)
 	},
 	"hash": func(id compact.NodeID) string {
+		// For "hash" format node text, levels >=1 need a different format
+		// [H=H(childL|childR)]from the base level (H=H(leafN)].
 		if id.Level >= 1 {
 			childLevel := id.Level - 1
 			leftChild := id.Index * 2
--- a/docs/merkletree/treetex/main.go
+++ b/docs/merkletree/treetex/main.go
@@ -87,7 +87,7 @@ var (
 	megaMode   = flag.Uint("megamode_threshold", 4, "Treat perfect trees larger than this many layers as a single entity")
 	ranges     = flag.String("ranges", "", "Comma-separated Open-Closed ranges of the form L:R")
 
-	attrPerfectRoot   = flag.String("attr_perfect_root", "line width=4pt", "Latex treatment for perfect root nodes")
+	attrPerfectRoot   = flag.String("attr_perfect_root", "", "Latex treatment for perfect root nodes (e.g. 'line width=3pt')")
 	attrEphemeralNode = flag.String("attr_ephemeral_node", "draw, dotted", "Latex treatment for ephemeral nodes")
 
 	// nInfo holds nodeInfo data for the tree.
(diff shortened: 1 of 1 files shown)


docs/merkletree/treetex/main.go
```diff
@@ -176,14 +176,19 @@ func perfectMega(prefix string, height uint, leafIndex uint64) {
...
 // drawLeaf emits TeX code to render a leaf.
-func drawLeaf(prefix string, index uint64, text string) {
+func drawLeaf(prefix string, index uint64, leafText, dataText string) {
 	a := nInfo[compact.NewNodeID(0, index)]
-	fmt.Printf("%s [%s, %s, align=center, tier=leaf]\n", prefix, text, a.String())
+
+	// First render the leaf node of the Merkle tree
+	fmt.Printf("%s [%s, %s, align=center, tier=leaf\n", prefix, leafText, a.String())
+	// and then a child-node representing the leaf data itself:
+	a.leaf = true
```
pavelkalinnikov: How come it was false before this change?

pavelkalinnikov: I'm pretty sure it should be `true` already. Can you assert it locally?

docs/merkletree/treetex/main.go
```diff
@@ -197,31 +202,32 @@ func openInnerNode(prefix string, id compact.NodeID, text string) func() {
...
 	id := compact.NewNodeID(level, index)
 	modifyNodeInfo(id, func(n *nodeInfo) {
-		n.leaf = id.Level == 0
 		n.perfectRoot = top
 	})
 
-	text := nodeText(id)
 	if level == 0 {
-		drawLeaf(prefix, index, text)
+		leafNodeText := nodeText(id)
```
pavelkalinnikov: You could keep this line before `if level == 0 {`, and not duplicate it.

docs/merkletree/treetex/main.go
```diff
@@ -343,18 +356,15 @@ func main() {
 		log.Fatalf("unknown --node_format %s", *nodeFormat)
 	}
 
-	var ldf nodeTextFunc = func(id compact.NodeID) string {
-		if id.Level == 0 {
-			return fmt.Sprintf("{$H_{0.%d} =$ \\\\ $H(leaf_{%[1]d})$}", id.Index)
-		}
+	var nodeText nodeTextFunc = func(id compact.NodeID) string {
```
pavelkalinnikov: `nodeText := innerNodeText`?

docs/merkletree/treetex/main.go
```diff
@@ -321,14 +327,21 @@ func modifyRangeNodeInfo() error {
...
 	"hash": func(id compact.NodeID) string {
-		childLevel := id.Level - 1
-		leftChild := id.Index * 2
-		return fmt.Sprintf("{$H_{%d.%d} =$ \\\\ $H(H_{%d.%d} || H_{%d.%d})$}", id.Level, id.Index, childLevel, leftChild, childLevel, leftChild+1)
+		if id.Level >= 1 {
+			childLevel := id.Level - 1
+			leftChild := id.Index * 2
+			return fmt.Sprintf("{$H_{%d.%d} =$ \\\\ $H(H_{%d.%d} || H_{%d.%d})$}", id.Level, id.Index, childLevel, leftChild, childLevel, leftChild+1)
+		}
+		return fmt.Sprintf("{$H_{%d.%d} =$ \\\\ $H(leaf_{%[2]d})$}", id.Level, id.Index)
```
pavelkalinnikov: What does `%[2]d` do? Can't find it in documentation.

pavelkalinnikov:
A bit of whining: This tool gets harder to review.

Some ideas for future:
- Maybe don't split formatters into internal/leaf nodes and data nodes, but instead do one function taking a `nodeInfo`.
- Separate the tree model from printing.

docs/merkletree/treetex/main.go
AlCutter: Ah, that's because I removed the bit where it was set to true inside `perfectInner` - arguably it makes more sense to just handle this here since it's really about presentation.

wdyt?

docs/merkletree/treetex/main.go
AlCutter: I don't _think_ it's duplicated? We're passing text through here, and the funcs below - but your suggestion also indirectly suggests that this is suboptimal for grokking, so I've made `drawLeaf` take the funcs directly too.

docs/merkletree/treetex/main.go
AlCutter: ta, done.

docs/merkletree/treetex/main.go
AlCutter: This makes the fmt specifier take from the `nth` (1-based index) parameter, so in this case it's saying _"re-use the 2nd parameter here (even though it was 'consumed' earlier in the format string)"_

AlCutter:
Whining - not wrong, maybe one day :)

docs/merkletree/treetex/main.go
pavelkalinnikov: Yeah, I think it's better to handle presentation stuff here, as opposed to scattering it between functions.

docs/merkletree/treetex/main.go
pavelkalinnikov: There were 2 branches, each calling `nodeText(id)` for the same ID. Now it looks better.

docs/merkletree/treetex/main.go
pavelkalinnikov: Wow, good to know.

Can't say I like this feature too much: "After processing a bracketed expression [n], subsequent verbs will use arguments n+1, n+2, etc. unless otherwise directed."

docs/merkletree/treetex/main.go
```diff
@@ -321,14 +325,21 @@ func modifyRangeNodeInfo() error {
...
 
+var dataFormat = func(id compact.NodeID) string {
+	return fmt.Sprintf("{$leaf_{%d}$}", id.Index)
+}
+
 var nodeFormats = map[string]nodeTextFunc{
 	"address": func(id compact.NodeID) string {
 		return fmt.Sprintf("%d.%d", id.Level, id.Index)
 	},
 	"hash": func(id compact.NodeID) string {
```
pavelkalinnikov: How come the "hash" function has to handle leaf nodes, but "address" doesn't?

AlCutter: Because the "hash" format nodes where level >=1 is of the form H=H(childL|childR), but the Level=0 needs to be H=H(leafN) [and there will be a "-1" level with [leafN]:

image

Whereas the "address" format doesn't have this differentiation:

image

pavelkalinnikov: SGTM. Put a comment?

AlCutter: done!

PR #1987, opened 2019-11-29
smt: Introduce tile scanning helper

pavelkalinnikov:
This change adds an unexported `scan` helper to `Tile` which helps to calculate
internal hashes of a tile. Will be used in #1932.

### DIFF
--- a/merkle/smt/tile.go
+++ b/merkle/smt/tile.go
@@ -29,25 +29,27 @@ type Tile struct {
 	Leaves []Node
 }
 
-// scan visits all non-empty nodes of the tile except the root.
-func (t Tile) scan(lo *tree.Layout, h mapHasher, visit func(tree.NodeID2, []byte)) error {
+// scan visits all non-empty nodes of the tile except the root. The order of
+// node visits is arbitrary.
+func (t Tile) scan(l *tree.Layout, h mapHasher, visit func(Node)) error {
 	top := t.ID.BitLen()
-	height := uint(lo.TileHeight(int(top)))
+	height := uint(l.TileHeight(int(top)))
 	// TODO(pavelkalinnikov): Remove HStar3 side effects, to avoid copying here.
+	// Currently, the Update method modifies the nodes given to NewHStar3.
 	leaves := make([]Node, len(t.Leaves))
 	copy(leaves, t.Leaves)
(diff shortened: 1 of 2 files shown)


merkle/smt/tile.go
```diff
@@ -28,3 +28,34 @@ type Tile struct {
 	ID     tree.NodeID2
 	Leaves []Node
 }
+
+// scan visits all non-empty nodes of the tile except the root.
+func (t Tile) scan(lo *tree.Layout, h mapHasher, visit func(tree.NodeID2, []byte)) error {
```
AlCutter: nitty: "Layout" feels like one word to me, so 'l'?

pavelkalinnikov: `visit` can take `Node` as input

merkle/smt/tile_test.go
```diff
@@ -0,0 +1,75 @@
...
+	lo := tree.NewLayout([]int{8, 24})
+	h := bindHasher(maphasher.Default, 1)
+
+	ids := []tree.NodeID2{
+		tree.NewNodeID2("\x00", 8),
+		tree.NewNodeID2("\x02", 8),
+		tree.NewNodeID2("\xF0", 8),
+		tree.NewNodeID2("\x00\x01\x02\x03", 32),
+	}
+	prefs := func(high, low uint, ids ...tree.NodeID2) []tree.NodeID2 {
```
AlCutter: `prefixes`? 

(prefs is quite hard-coded to 'preferences' in my old brain - _Likely reason for that:_  
)

merkle/smt/tile.go
```diff
@@ -28,3 +28,34 @@ type Tile struct {
...
+	top := t.ID.BitLen()
+	height := uint(lo.TileHeight(int(top)))
+	// TODO(pavelkalinnikov): Remove HStar3 side effects, to avoid copying here.
```
mhutchinson: Is there a link to context here? It's OK if this makes sense to you and you plan to resolve it soon, but if you got abducted by aliens this weekend I would need to put in some work to find what side-effect this was pointing at and how/why it might be changed.

merkle/smt/tile.go
pavelkalinnikov: Done.

PR #1991, opened 2019-11-29
Treetex proofs

AlCutter:
Make proofs in treetex output a bit clearer.

### Checklist


docs/merkletree/treetex/main.go
```diff
@@ -186,9 +186,16 @@ func drawLeaf(prefix string, index uint64, leafText, dataText nodeTextFunc) {
 	a := nInfo[id]
 
 	// First render the leaf node of the Merkle tree
+	rangeIndices := a.rangeIndices
```
pavelkalinnikov: Could copy the whole `a` here, and use it for the second `Printf`.

AlCutter: good call, done.

docs/merkletree/treetex/main.go
```diff
@@ -295,11 +300,18 @@ func modifyRangeNodeInfo() error {
 		// Set leaves:
 		for i := l; i < r; i++ {
 			id := compact.NewNodeID(0, i)
-			modifyNodeInfo(id, func(n *nodeInfo) { n.rangeIndices = append(n.rangeIndices, ri) })
+			modifyNodeInfo(id, func(n *nodeInfo) {
+				n.incPath = true
```
pavelkalinnikov: It's not an inclusion path, so maybe add a TODO to fix this?

AlCutter: I think this was actually unnecessary, dinged it.

PR #1992, opened 2019-11-29
Treetex colouring

AlCutter:
Fix up some colouring issues with proofs.

### Checklist


AlCutter:
For reference, this is what it does for overlaps:

image

pavelkalinnikov:
Looks good. Is it possible to remove the red tint from the compact range nodes? Also, for node 0.7 it would make more sense to blend green and blue.

AlCutter:
> Looks good. Is it possible to remove the red tint from the compact range nodes? Also, for node 0.7 it would make more sense to blend green and blue.

The red is to indicate "proof" node (like it does for inclusion proofs), and node 0.7 is only a proof for the `blue` range I think.

pavelkalinnikov:
> The red is to indicate "commit" node (like it does for inclusion proofs)

To me the red bit is misleading and in fact redundant. Compact-range-colored node already implies its commitness.

> and node 0.7 is only valid for the `blue` range I think.

Indeed.

docs/merkletree/treetex/README.md
```diff
@@ -79,7 +79,7 @@ To highlight nodes which make up (multiple) range proofs, use the `--ranges`
 flag.
 
 ```bash
-go run github.com/google/trillian/docs/merkletree/treetex --tree_size=18 --ranges=1:3,7:14
+go run github.com/google/trillian/docs/merkletree/treetex --tree_size=18 --ranges=1:4,7:14
```
pavelkalinnikov: This doesn't seem to match the picture.

docs/merkletree/treetex/main.go
```diff
@@ -96,13 +99,14 @@ var (
...
-	leaf         bool
-	rangeIndices []int
+	proof             bool
+	incPath           bool
+	target            bool
+	perfectRoot       bool
+	ephemeral         bool
+	leaf              bool
+	rangeIndices      []int
+	proofRangeIndices []int
```
pavelkalinnikov: Why do you distinguish `rangeIndices` and `proofRangeIndices`? As I understand it, the source of confusion is that the leaf nodes are in fact storing 2 nodes, so one array is used by "upper" leaf nodes, and the other one - by "lower" ones.

Consider separating "data node info" to make this division explicit.

pavelkalinnikov: It's not really a proof. A range proof is what **complements** a compact range.

How about renaming `rangeIndices` to `dataRangeIndices`, and `proofRangeIndices` to `rangeIndices`?

AlCutter:
> > The red is to indicate "commit" node (like it does for inclusion proofs)
> 
> To me the red bit is misleading and in fact redundant.

okaaaay, done :)

And overlaps now look like this:
image

docs/merkletree/treetex/main.go
```diff
@@ -96,13 +99,14 @@ var (
...
 // nodeInfo represents the style to be applied to a tree node.
 type nodeInfo struct {
-	incProof     bool
-	incPath      bool
-	target       bool
-	perfectRoot  bool
-	ephemeral    bool
-	leaf         bool
-	rangeIndices []int
+	proof             bool
```
pavelkalinnikov: This type mixes "global" node properties (e.g. `ephemeral` and `perfectRoot`) with query-dependent ones (proofs, ranges, etc). Consider storing a slice of styles/modifiers/classes instead of a bunch of bools for the query dependent properties - this would allow you to also have multiple inclusion proofs at the same time etc.

pavelkalinnikov:
The new picture looks great!

I will do a final review on Monday.

docs/merkletree/treetex/main.go
AlCutter: That sounds like a good idea, but it's a job for another day.

docs/merkletree/treetex/main.go
AlCutter: Yup, that's right.
I'll add a TODO for that.

AlCutter: done

docs/merkletree/treetex/README.md
AlCutter: What would you expect to see in `compactrange.png` for these arguments?

pavelkalinnikov: Looks correct now.

docs/merkletree/treetex/main.go
```diff
@@ -117,11 +122,33 @@ func (n nodeInfo) String() string {
...
-		fill = "inclusion"
+
+	if n.proof {
 		if n.ephemeral {
-			fill = "inclusion_ephemeral"
-			attr = append(attr, *attrEphemeralNode)
+			fill = "proof_ephemeral"
+		} else {
+			fill = "proof"
+		}
```
pavelkalinnikov:
```golang
fill = "proof"
if n.ephemeral {
	fill = "proof_ephemeral"
}
```

docs/merkletree/treetex/main.go
```diff
@@ -185,15 +204,20 @@ func drawLeaf(prefix string, index uint64, leafText, dataText nodeTextFunc) {
...
-	// First render the leaf node of the Merkle tree
-	a.rangeIndices = nil
-	a.incPath = false
+	// First render the leaf node of the Merkle tree.
+	if len(a.dataRangeIndices) > 0 {
+		a.incPath = false
+	}
+	if !a.proof {
+		a.dataRangeIndices = nil
+	}
```
pavelkalinnikov: I don't follow the connection between ranges and proofs here. Maybe comment a bit?

docs/merkletree/treetex/main.go
AlCutter: done

docs/merkletree/treetex/main.go
AlCutter: I think this was left-over from when there was only the one range, I've dinged it.

pavelkalinnikov:
It would be cool to have Travis generate all the "example" images in presubmit, and check that they are correct. In absence of unit-tests this would at least provide some assurance.

AlCutter:
I agree! :)

PR #1996, opened 2019-12-03
Set HTTP endpoint to disabled by default 

paulmattei:
### Checklist

- [N/A] I have updated [documentation](docs/) accordingly (including the [feature implementation matrix](docs/Feature_Implementation_Matrix.md)).

### DIFF
--- a/server/trillian_log_server/main.go
+++ b/server/trillian_log_server/main.go
@@ -58,7 +58,7 @@ import (
 
 var (
 	rpcEndpoint     = flag.String("rpc_endpoint", "localhost:8090", "Endpoint for RPC requests (host:port)")
-	httpEndpoint    = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP metrics and REST requests on (host:port, empty means disabled)")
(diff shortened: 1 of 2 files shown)


server/trillian_log_server/main.go
```diff
@@ -58,7 +58,7 @@ import (
 
 var (
 	rpcEndpoint     = flag.String("rpc_endpoint", "localhost:8090", "Endpoint for RPC requests (host:port)")
-	httpEndpoint    = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP metrics and REST requests on (host:port, empty means disabled)")
+	httpEndpoint    = flag.String("http_endpoint", "", "Endpoint for HTTP metrics and REST requests on (host:port, empty means disabled)")
```
AlCutter: This endpoint is also used for serving metrics (e.g. in Prometheus), so we'd nobble folks who are using that.

I suspect that what you're really after is disabling the REST API by default?

It _looks_ like the REST API might be being registered at line 162 in this file - you could add a boolean flag to control whether or not the REST api is registered on the mux?

PR #2004, opened 2019-12-04
Made the hammer internal-only

mhutchinson:
This was never intended to be referenced elsewhere in this codebase, and especially not from any other repositories.


mhutchinson:
> Am I right in thinking that the entire "trillian/testonly" directory could be made internal, but you're trying to constrain the size of the PR?

Ideally that would be true, but `quota/cacheqm/cache_test.go` uses `testonly/matchers/atleast.go` so it's not as clear cut as that. But you are right that one goal of this is to carve off as little as possible as a test bite to make sure this approach of tucking things in `internal` directories doesn't cause other unintended consequences.

pavelkalinnikov:
This PR LGTM, but I'm still a bit skeptical about the whole `internal` thing. To me the idea of moving almost everything to `internal` seems like a nightmare. Also, even in this PR we already have this "but": We want to move the whole package to `internal`, but there is one particular type/file that is being used in a way that's not internal. If we end up creating parallel directories like `storage/foo/bar` and `internal/storage/foo/bar`, I will be unhappy.

PR #2006, opened 2019-12-05
Fix some warnings

Martin2112:
### Checklist

### DIFF
--- a/quota/etcd/storage/quota_storage_test.go
+++ b/quota/etcd/storage/quota_storage_test.go
@@ -393,7 +393,7 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 
 	want := &storagepb.Configs{} // default cfgs is empty
 	if _, err := qs.UpdateConfigs(ctx, true /* reset */, updater(want)); err != nil {
-		t.Fatalf("UpdateConfigs() returned err = %v", err)
+		t.Fatalf("UpdateConfigs()= _,%v want: _,nil", err)
 	}
 
 	for _, test := range tests {
--- a/quota/etcd/storage/quota_storage_test.go
+++ b/quota/etcd/storage/quota_storage_test.go
@@ -389,14 +389,14 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 	}
 
 	ctx := context.Background()
-	qs := &QuotaStorage{Client: client}
-
-	want := &storagepb.Configs{} // default cfgs is empty
-	if _, err := qs.UpdateConfigs(ctx, true /* reset */, updater(want)); err != nil {
-		t.Fatalf("UpdateConfigs()= (_,%q), want (_,nil)", err)
-	}
-
 	for _, test := range tests {
+		qs := &QuotaStorage{Client: client}
+
+		want := &storagepb.Configs{} // default cfgs is empty
+		if _, err := qs.UpdateConfigs(ctx, true /* reset */, updater(want)); err != nil {
+			t.Fatalf("UpdateConfigs()= (_,%q), want (_,nil)", err)
+		}
+
 		t.Run(test.desc, func(t *testing.T) {
 			_, err := qs.UpdateConfigs(ctx, false /* reset */, test.update)
 			// All the test cases should result in an error.
@@ -404,7 +404,6 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 				t.Fatalf("UpdateConfigs()=_, %v, want: _, %v", err, test.wantErr)
 			}
 			if !strings.Contains(err.Error(), test.wantErr) {
-				// Fatal because the config has been changed, which will break all following tests.
 				t.Fatalf("UpdateConfigs()=_,%v, want substring %q", err, test.wantErr)
 			}
 
--- a/quota/etcd/storage/quota_storage_test.go
+++ b/quota/etcd/storage/quota_storage_test.go
@@ -406,12 +406,12 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 			// Then if we got an error check that it was the expected one.
 			if err != nil && !strings.Contains(err.Error(), test.wantErr) {
 				// Fatal because the config has been changed, which will break all following tests.
-				t.Fatalf("UpdateConfigs() returned err = %v, want substring %q", err, test.wantErr)
+				t.Fatalf("UpdateConfigs()=_,%v, want substring %q", err, test.wantErr)
 			}
 
 			stored, err := qs.Configs(ctx)
 			if err != nil {
-				t.Errorf("Configs() returned err = %v", err)
+				t.Errorf("Configs()=_,%v want: _,nil", err)
 				return
 			}
 			if got := stored; !proto.Equal(got, want) {
--- a/quota/etcd/storage/quota_storage_test.go
+++ b/quota/etcd/storage/quota_storage_test.go
@@ -397,7 +397,12 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		if _, err := qs.UpdateConfigs(ctx, false /* reset */, test.update); !strings.Contains(err.Error(), test.wantErr) {
+		var errStr string
+		_, err := qs.UpdateConfigs(ctx, false /* reset */, test.update)
+		if err != nil {
+			errStr = err.Error()
+		}
+		if !strings.Contains(errStr, test.wantErr) {
 			// Fatal because the config has been changed, which will break all following tests.
 			t.Fatalf("%v: UpdateConfigs() returned err = %v, want substring %q", test.desc, err, test.wantErr)
 		}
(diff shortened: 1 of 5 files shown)


Martin2112:
Some potential fixes. Commits are independent.

crypto/keys/der/der_test.go
```diff
@@ -68,9 +68,9 @@ func TestFromProto(t *testing.T) {
 			t.Parallel()
 
 			signer, err := FromProto(test.keyProto)
-			if gotErr := err != nil; gotErr != test.wantErr {
+			if (err != nil) != test.wantErr {
 				t.Fatalf("FromProto(%#v) = (_, %q), want (_, nil)", test.keyProto, err)
```
RJPercival: This error message looks wrong - it assumes that we wanted `err == nil`, but that's not the case if `test.wantErr == true`.

crypto/keys/pem/pem_test.go
```diff
@@ -91,10 +91,10 @@ func TestFromProto(t *testing.T) {
 		},
 	} {
 		signer, err := FromProto(test.keyProto)
-		if gotErr := err != nil; gotErr != test.wantErr {
+		if (err != nil) != test.wantErr {
 			t.Errorf("%v: FromProto(%#v) = (_, %q), want (_, nil)", test.desc, test.keyProto, err)
```
RJPercival: This error message looks wrong - it assumes that we wanted `err == nil`, but that's not the case if `test.wantErr == true`.

crypto/keys/der/der_test.go
```diff
@@ -68,9 +68,9 @@ func TestFromProto(t *testing.T) {
...
-			} else if gotErr {
+			} else if err != nil {
```
RJPercival: Can get rid of the `else` now that there's no shared `gotErr` var.

crypto/keys/der/der_test.go
```diff
@@ -68,9 +68,9 @@ func TestFromProto(t *testing.T) {
 			t.Parallel()
 
 			signer, err := FromProto(test.keyProto)
-			if gotErr := err != nil; gotErr != test.wantErr {
```
RJPercival: What's the warning about using `gotErr`?

crypto/keys/pem/pem_test.go
```diff
@@ -91,10 +91,10 @@ func TestFromProto(t *testing.T) {
...
 			continue
-		} else if gotErr {
+		} else if err != nil {
```
RJPercival: Can get rid of the `else` now that there's no shared `gotErr` var.

quota/etcd/storage/quota_storage_test.go
```diff
@@ -397,7 +397,12 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		if _, err := qs.UpdateConfigs(ctx, false /* reset */, test.update); !strings.Contains(err.Error(), test.wantErr) {
+		var errStr string
+		_, err := qs.UpdateConfigs(ctx, false /* reset */, test.update)
+		if err != nil {
+			errStr = err.Error()
+		}
+		if !strings.Contains(errStr, test.wantErr) {
```
RJPercival: This doesn't seem right. If `test.wantErr == ""`, the test won't fail when `err != nil`.

crypto/keys/der/der_test.go
Martin2112: It's confusing the IntelliJ nilness inspector into thinking that values are being used without checking for errors.

`Error corresponding to variable 'signer' may not be nil`

I could suppress it with `/noinspection GoNilness` but didn't really want to have tool specific comments in the code.

Martin2112: Maybe I'll just revert the error handling changes and turn off this inspection.

crypto/keys/der/der_test.go
Martin2112: Reverted this change.

crypto/keys/pem/pem_test.go
Martin2112: Reverted this change.

crypto/keys/der/der_test.go
Martin2112: Reverted.

crypto/keys/der/der_test.go
Martin2112: Reverted.

crypto/keys/pem/pem_test.go
Martin2112: Reverted.

quota/etcd/storage/quota_storage_test.go
Martin2112: Reverted.

Martin2112:
I've dropped the commit that had the error handling changes.

quota/etcd/storage/quota_storage_test.go
RJPercival: It doesn't look very reverted...

Martin2112: Ah that one was part of a different commit. I'll try again.

Martin2112:
PTAL

quota/etcd/storage/quota_storage_test.go
```diff
@@ -393,24 +393,32 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
 
 	want := &storagepb.Configs{} // default cfgs is empty
 	if _, err := qs.UpdateConfigs(ctx, true /* reset */, updater(want)); err != nil {
-		t.Fatalf("UpdateConfigs() returned err = %v", err)
+		t.Fatalf("UpdateConfigs()= _,%v want: _,nil", err)
```
RJPercival:
```suggestion
		t.Fatalf("UpdateConfigs()= (_,%q), want (_,nil)", err)
```

quota/etcd/storage/quota_storage_test.go
```diff
@@ -393,24 +393,32 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
...
-		stored, err := qs.Configs(ctx)
-		if err != nil {
-			t.Errorf("%v:Configs() returned err = %v", test.desc, err)
-			continue
-		}
-		if got := stored; !proto.Equal(got, want) {
-			diff := pretty.Compare(got, want)
-			t.Fatalf("%v: post-Configs() diff (-got +want)\n%v", test.desc, diff)
-		}
+			stored, err := qs.Configs(ctx)
```
RJPercival: nit: I wouldn't have thought we'd want to do this if UpdateConfigs() failed, but the test passes so I guess it's fine.

quota/etcd/storage/quota_storage_test.go
```diff
@@ -393,24 +393,32 @@ func TestQuotaStorage_UpdateConfigsErrors(t *testing.T) {
...
-		}
+		t.Run(test.desc, func(t *testing.T) {
+			_, err := qs.UpdateConfigs(ctx, false /* reset */, test.update)
+			// First check that an error did / did not occur as the test expects.
+			if (err != nil) != (test.wantErr != "") {
+				t.Fatalf("UpdateConfigs()=_, %v, want: _, %v", err, test.wantErr)
+			}
+			// Then if we got an error check that it was the expected one.
+			if err != nil && !strings.Contains(err.Error(), test.wantErr) {
+				// Fatal because the config has been changed, which will break all following tests.
```
RJPercival: This comment suggests that all subsequent test cases should be aborted, but this no longer happens since changing this to use t.Run().

quota/etcd/storage/quota_storage_test.go
Martin2112: My assumption was that in the error case we still want to ensure the quota state hasn't changed. There are no non-error test cases so I tweaked it slightly.

This test was pretty unclear, I suspect it isn't testing all the possible combinations and could probably be made more readable.

quota/etcd/storage/quota_storage_test.go
Martin2112: Yes, but it seemed acceptable as it's unlikely that we'll get a bug that mutates the state and returns a different error and it fails safe if we do, just with more noise.

RJPercival: Could you update the comment then, since it's misleading now. Removing the comment would be fine.

I wonder why it shares the `QuotaStorage` between tests? Seems to unnecessarily expose test cases to breaking each other.

Martin2112: Done. 

I suspect it used the same structure as the non-error test, where it accumulates mutations into the storage. Makes sense there but doesn't seem to be any reason do it here so I've rearranged it so the tests are independent.

RJPercival: Oh I see why - there's a TestMain() that starts an embedded etcd server that is shared by all the tests, so any test case can affect any other.

Martin2112: In this test there should be no changes to quota state as all the updates are expected to fail. So if it passes it should leave it untouched.

PR #2010, opened 2019-12-16
Move Proof proto definition to trillian.proto

tweksteen:
The Proof message may be used by a personality or client[1] without
requiring knowledge of the complete log API. Move the Proof definition
from trillian_log_api.proto to trillian.proto. The other modifications
are the result of running `go generate -x ./...`.

[1] https://android-review.googlesource.com/c/platform/external/avb/+/1190237

### Checklist


Martin2112:
I think this is a logical change as the other similar entities are in the main proto. 

Fortunately, they're in the same generated Go package. For other languages this might not be the case as the containing class will change, to take a Java example.

@mhutchinson do you have any comments on how we should make this change?

mhutchinson:
> I think this is a logical change as the other similar entities are in the main proto.

Agreed.

> Fortunately, they're in the same generated Go package. For other languages this might not be the case as the containing class will change, to take a Java example.
> 
> @mhutchinson do you have any comments on how we should make this change?

I think Java is fine. See the imports at https://github.com/mhutchinson/trillian-java-3p/blob/master/src/com/github/mhutchinson/trillianj/AdminClient.java for example; those imports have consistent names even though Tree is defined in a separate file to the Admin API.

For this change, anyone importing the original proto file would have needed the new destination file on the classpath anyway, due to the transitive dependency. I'd score this 4/5 for safety. I don't have any good ideas on how to get that final piece of comfort right now.

Martin2112:
Ah yes, the `java_multiple_files` option is set. The outer class just holds descriptors and stuff in that case so imports don't change.

PR #2011, opened 2019-12-17
Fix up some tests to do better error handling.

Martin2112:
Also log an error for a failed close in operation manager but no behavioural changes.

### Checklist

### DIFF
--- a/storage/mysql/admin_storage_test.go
+++ b/storage/mysql/admin_storage_test.go
@@ -91,7 +91,7 @@ func TestAdminTX_TreeWithNulls(t *testing.T) {
 	}
 	treeID := tree.TreeId
 
-	if err := setNulls(t, ctx, DB, treeID); err != nil {
+	if err := setNulls(ctx, t, DB, treeID); err != nil {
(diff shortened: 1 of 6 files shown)


storage/mysql/admin_storage_test.go
```diff
@@ -241,12 +241,17 @@ func TestCheckDatabaseAccessible_OK(t *testing.T) {
 	}
 }
 
-func setNulls(ctx context.Context, db *sql.DB, treeID int64) error {
+func setNulls(ctx context.Context, t *testing.T, db *sql.DB, treeID int64) error {
```
RJPercival: nit: Since you're making this a test helper, could simply it further by removing the `error` return value and just calling `t.Fatal()` instead.

mhutchinson:
Is this PR still desired? If so, can we get it merged? :-)

Martin2112:
I wasn't sure the benefits were really worth it. As it now has conflicts and requires rework I think I'll close it unless anyone disagrees.

PR #2014, opened 2019-12-18
metrics: Add helper for arbitrary histogram buckets

pavelkalinnikov:


pavelkalinnikov:
Actually, I think we could increase the default multiplier from `1.05` to `1.07`. It would raise the latencies range from [40ms, 24h] to [40ms, 283d], with no significant loss in precision in the lower buckets.

gdbelvin:
A unit test would help document some of the expected behavior buckets

pavelkalinnikov:
Added some tests.

PR #2015, opened 2019-12-18
monitoring: Increase the default latency range

pavelkalinnikov:
This change does not harm precision of lower buckets in a significant way.

### DIFF
--- a/monitoring/buckets_test.go
+++ b/monitoring/buckets_test.go
@@ -63,9 +63,9 @@ func TestPercentileBuckets(t *testing.T) {
 func TestLatencyBuckets(t *testing.T) {
 	// Just do some probes on the result to make sure it looks sensible.
 	buckets := monitoring.LatencyBuckets()
-	checkExpBuckets(t, buckets, 0.04, 1.05, 300)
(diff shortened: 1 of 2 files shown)


monitoring/buckets_test.go
```diff
@@ -63,9 +63,9 @@ func TestPercentileBuckets(t *testing.T) {
 func TestLatencyBuckets(t *testing.T) {
 	// Just do some probes on the result to make sure it looks sensible.
 	buckets := monitoring.LatencyBuckets()
-	checkExpBuckets(t, buckets, 0.04, 1.05, 300)
-	// Highest bucket should be about 86400 sec = 1 day (allow some leeway).
-	if got, want := math.Abs(buckets[len(buckets)-1]-86400.0), 300.0; got > want {
+	checkExpBuckets(t, buckets, 0.04, 1.07, 300)
+	// Highest bucket should be about 282 days (allow some leeway).
```
gdbelvin: use `cmp` to compare floats with some leeway?
This unit test is difficult to parse

pavelkalinnikov: Improved readability a bit. How does it look?

gdbelvin: LGTM

PR #2018, opened 2019-12-19
Add go-licenses to tools.go

RJPercival:
This tool is used when building Docker images, in order to check dependencies have compatible licenses and to comply with those licenses. Including it in tools.go will optimise Docker image builds by fetching the tool when downloading Trillian's dependencies, which are then cached.

### Checklist


RJPercival:
I can see an argument for making the Docker files explicitly grab this tool during the cacheable phase, rather than making it a declared dependency of Trillian. Thoughts?

NatalieDoduc:
(nit: would you mind re-wording your PR description s/This too/go-licences/? Given you also modify tools.go, it's slightly confusing which you're referring to at a quick read.)

What would your alternate proposal look like? `go-licences` is not a dependency, so to speak, more of a helper tool to validate correctness, so your proposal makes more sense to me. That being said, adding it to tools.go is a nice and simple solution.

gdbelvin:
I slightly prefer explicitly referencing this by version during the docker builds
The reason for not automatically putting everything in `tools.go` is that it bloats the dependency list, adding overhead to every `go {build,test,list,install}` operation.

gdbelvin:
bump?

RJPercival:
I'll add an explicit reference in the Dockerfiles instead.

PR #2023, opened 2020-01-09
Export a metric of total number of leaves fetched

AlCutter:
Adds a simple metric to track the number of individual entries fetched by calls to the `GetEntriesBy.*` API calls.

### Checklist

### DIFF
--- a/server/log_rpc_server.go
+++ b/server/log_rpc_server.go
@@ -505,6 +505,7 @@ func (t *TrillianLogRPCServer) GetLeavesByIndex(ctx context.Context, req *trilli
 	}
 	defer t.closeAndLog(ctx, tree.TreeId, tx, "GetLeavesByIndex")
 
+	t.fetchedLeaves.Add(float64(len(req.LeafIndex)))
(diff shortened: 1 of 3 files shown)


server/log_rpc_server.go
```diff
@@ -541,6 +548,8 @@ func (t *TrillianLogRPCServer) GetLeavesByRange(ctx context.Context, req *trilli
 	}
 	defer t.closeAndLog(ctx, tree.TreeId, tx, "GetLeavesByRange")
 
+	t.fetchedLeaves.Add(float64(req.Count))
```
mhutchinson: The others are incremented after the leaves are fetched, but this one is before. It's also possible that this is incremented but we skip the fetch altogether because of the `req.StartIndex < int64(root.TreeSize)` check below.

I suggest always incrementing the counter immediately after fetching the leaves successfully, XOR incrementing it immediately before making the call regardless of whether it succeeds.

AlCutter: Thanks, that was a mistake - there were all supposed to be after having fetched - but you've made me think that having it before making the actual call would give a better visibility of what we're asking storage to do.

PR #2028, opened 2020-01-29
Move quota providers into packages under /quota

RJPercival:
Makes them optional to import, which is useful if you want to build a server binary without one or more of them.

### Checklist

### DIFF
--- a/quota/mysqlqm/quota_provider.go
+++ b/quota/mysqlqm/quota_provider.go
@@ -12,28 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package server
+package mysqlqm
 
 import (
 	"flag"
 
 	"github.com/golang/glog"
 	"github.com/google/trillian/quota"
-	"github.com/google/trillian/quota/mysqlqm"
+	"github.com/google/trillian/server"
 	"github.com/google/trillian/storage/mysql"
 )
 
-// QuotaMySQL represents the MySQL quota implementation.
-const QuotaMySQL = "mysql"
+// Quota represents the MySQL quota implementation.
+const Quota = "mysql"
 
 var (
-	maxUnsequencedRows = flag.Int("max_unsequenced_rows", mysqlqm.DefaultMaxUnsequenced, "Max number of unsequenced rows before rate limiting kicks in. "+
+	maxUnsequencedRows = flag.Int("max_unsequenced_rows", DefaultMaxUnsequenced, "Max number of unsequenced rows before rate limiting kicks in. "+
 		"Only effective for quota_system=mysql.")
 )
 
 func init() {
-	if err := RegisterQuotaManager(QuotaMySQL, newMySQLQuotaManager); err != nil {
-		glog.Fatalf("Failed to register quota manager %v: %v", QuotaMySQL, err)
+	if err := server.RegisterQuotaManager(Quota, newMySQLQuotaManager); err != nil {
+		glog.Fatalf("Failed to register quota manager %v: %v", Quota, err)
 	}
 }
 
@@ -42,7 +42,7 @@ func newMySQLQuotaManager() (quota.Manager, error) {
 	if err != nil {
 		return nil, err
 	}
-	qm := &mysqlqm.QuotaManager{
+	qm := &QuotaManager{
 		DB:                 db,
 		MaxUnsequencedRows: *maxUnsequencedRows,
 	}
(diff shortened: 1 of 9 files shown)


pavelkalinnikov:
Have you considered moving the provider to `/quota/etcd` instead?

quota/mysqlqm/quota_provider.go
```diff
@@ -12,28 +12,28 @@
...
 var (
-	maxUnsequencedRows = flag.Int("max_unsequenced_rows", mysqlqm.DefaultMaxUnsequenced, "Max number of unsequenced rows before rate limiting kicks in. "+
+	maxUnsequencedRows = flag.Int("max_unsequenced_rows", DefaultMaxUnsequenced, "Max number of unsequenced rows before rate limiting kicks in. "+
 		"Only effective for quota_system=mysql.")
 )
 
 func init() {
-	if err := RegisterQuotaManager(QuotaMySQL, newMySQLQuotaManager); err != nil {
-		glog.Fatalf("Failed to register quota manager %v: %v", QuotaMySQL, err)
+	if err := server.RegisterQuotaManager(Quota, newMySQLQuotaManager); err != nil {
```
pavelkalinnikov: For storage providers, I remember moving the "register" function from `server` to `storage`, to avoid weird dependencies. Did you check if moving `RegisterQuotaManager` and its guts to `/quota` makes sense?

pavelkalinnikov: In my mental model, almost nothing should depend on `server`.

RJPercival: Looks like that works, done.

quota/provider_test.go
```diff
@@ -12,12 +12,10 @@
...
 // limitations under the License.
 
-package server
+package quota
 
 import (
 	"testing"
-
-	"github.com/google/trillian/quota"
 )
```
pavelkalinnikov: Squash to ```import "testing"```?

quota/provider.go
```diff
@@ -12,49 +12,35 @@
...
-		glog.Fatalf("Failed to register %v: %v", QuotaNoop, err)
-	}
-}
+// NewManagerFunc is the signature of a function which can be registered
+// to provide instances of a quota manager.
+type NewManagerFunc func() (Manager, error)
 
-// RegisterQuotaManager registers the provided QuotaManager.
-func RegisterQuotaManager(name string, qp NewQuotaManagerFunc) error {
+// RegisterManager registers the provided Manager.
```
pavelkalinnikov:
```suggestion
// RegisterManager registers the provided Manager function.
```

quota/provider.go
```diff
@@ -78,13 +64,13 @@ func quotaSystems() []string {
...
-	return NewQuotaManager(*QuotaSystem)
+// NewManagerFromFlags returns a Manager implementation as speficied by flag.
+func NewManagerFromFlags() (Manager, error) {
+	return NewManager(*System)
 }
 
-// NewQuotaManager returns a quota.Manager implementation.
-func NewQuotaManager(name string) (quota.Manager, error) {
+// NewManager returns a Manager implementation.
+func NewManager(name string) (Manager, error) {
```
pavelkalinnikov: For consistency, consider renaming `Manager` bits to `Provider` throughout this PR.

quota/noop.go
```diff
@@ -17,10 +17,23 @@ package quota
 import (
 	"context"
 	"fmt"
+
+	"github.com/golang/glog"
 )
 
+// NoopManagerName represents the noop quota implementation.
+const NoopManagerName = "noop"
```
pavelkalinnikov: Does this need to be exported at this point? Nothing seems to be using it from outside this file.

quota/provider.go
RJPercival: A `Manager` is actually akin to a storage implementation - it's not a provider. In this case, `NewManagerFunc` is the provider. I could rename that to `ProviderFunc` if you think that'd be an improvement?

quota/noop.go
RJPercival: It might be helpful if someone wants to check whether the current quota system is "noop" though, and is consistent with the other quota implementations (which both have a `QuotaManagerName` constant).

quota/noop.go
```diff
@@ -17,10 +17,23 @@ package quota
 import (
 	"context"
 	"fmt"
+
+	"github.com/golang/glog"
 )
 
+// noopManagerName represents the noop quota implementation.
+const noopManagerName = "noop"
```
pavelkalinnikov: Could you make sure `storage` also uses "manager name" rather than "provider name"?

RJPercival: It doesn't have an equivalent constant - it just uses the strings inline when calling `storage.RegisterProvider()`.

PR #2030, opened 2020-02-13
Remove dependency on ct-go/x509

gdbelvin:
There is no reason Trillian should be depending on the lax x509
certificate parsing provided by the certificate-transparency-go/x509
library.


rolandshoemaker:
In a similar vein, why is ct-go/asn1 being used?

RJPercival:
I don't think it should be - feel free to send a PR fixing that.

PR #2033, opened 2020-02-14
Remove stdlibs from blacklist

gdbelvin:
Remove stdlibs from blacklist

Reverts f92eb83a0581d00739856551ff59a1ccecb3e2c4
Follows #2032  #2030

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -15,8 +15,6 @@ linters-settings:
       - github.com/google/certificate-transparency-go/trillian
       - github.com/google/certificate-transparency-go/x509
       - github.com/google/certificate-transparency-go/asn1
-      - encoding/asn1
-      - crypto/x509
 
 linters:
   disable-all: true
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -14,6 +14,7 @@ linters-settings:
       - github.com/gogo/protobuf/proto
       - github.com/google/certificate-transparency-go/trillian
       - github.com/google/certificate-transparency-go/x509
+      - github.com/google/certificate-transparency-go/asn1
       - encoding/asn1
       - crypto/x509
```
RJPercival: Should these stdlib packages be on the block list?

RJPercival: I wonder why they weren't actually being blocked? You've recently added crypto/x509 imports without complaint from the linter right?

gdbelvin: The linter only runs against *new* code.

RJPercival: Those imports you added were new though, weren't they?

PR #2053, opened 2020-02-21
Protect key handler map with a mutex

gdbelvin:
This fixes a race condition when multiple threads access methods in the
keys package.

### DIFF
--- a/crypto/keys/handlers.go
+++ b/crypto/keys/handlers.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto"
 	"fmt"
+	"sync"
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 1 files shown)


crypto/keys/handlers.go
```diff
@@ -27,8 +28,11 @@ import (
 // For example, the protobuf message may contain a key or identify where a key can be found.
 type ProtoHandler func(context.Context, proto.Message) (crypto.Signer, error)
 
-// handlers convert a protobuf message into a crypto.Signer.
-var handlers = make(map[string]ProtoHandler)
+var (
+	// handlers convert a protobuf message into a crypto.Signer.
+	handlers   = make(map[string]ProtoHandler)
+	handlersMu sync.RWMutex
```
NatalieDoduc: nit: might as well call this handlersMutex?

gdbelvin: `Mu` is a pretty standard prefix for Mutex in Go code

PR #2065, opened 2020-03-09
Update README's instructions for generated code dependencies

NatalieDoduc:
Rather than enumerating and linking to the target dependency page,
explicitly list the command to run to get the required dependencies.
This allows the use of versions as specified in our `go.mod` file, and
avoids discrepancies between development environment and what is tested
by Travis.

The changes included do not affect the CHANGELOG nor the features matrix.


README.md
```diff
@@ -182,11 +182,17 @@ Some of the Trillian Go code is autogenerated from other files:
...
 
-  - `mockgen` tool from https://github.com/golang/mock
-  - `stringer` tool from https://golang.org/x/tools/cmd/stringer
-  - `protoc`, [Go support for protoc](https://github.com/golang/protobuf),
-     [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) and
-     [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc).
+  - a series of tools, using `go install` to ensure that the versions are
+    compatible and tested:
+
+    ```
```
gdbelvin: Should we make it clear that these commands need to be executed from the `trillian` dir?
```
cd $(go list -f '{{ .Dir }}' github.com/google/trillian)
```

gdbelvin:
Looks like `travis_setup_go` is broken at the moment.

README.md
NatalieDoduc: Excellent suggestion. Done!

PR #2087, opened 2020-04-21
Make the map strata configurable per-tree.

mhutchinson:
Previously there was a default strata configuration for all maps, which had 10 layers of index strata of height 1 byte. This is overkill for most maps; 4 is sufficient for maps with billions of entries.

Sending this early for discussion. This approach comes with the limitation that what I'm calling the index strata (the upper layers) need to be 1-byte each. The previous approach would allow different heights at different strata.

This change is the first step to making it configurable per-tree. TODO:
 - allow this to be set at tree creation time
 - write & read this from the DB


mhutchinson:
Motivation behind this is that I've generated maps with billions of key/value entries, and having 4 index layers is enough to have the leaf strata having only a handful of values. Having 10 index layers means that we need to pull out a lot more tiles that contain only a single value, which is wasted work.

Martin2112:
Seems to be an issue with proto generation. Probably the version doesn't match the one Travis is using.

Overall I think this is worth trying as we previously didn't have time to evaluate the effect of changing strata depths on performance.

I don't think the 1 byte upper index levels will be a significant limitation given we have that now. 

I'll do a proper review later.

docs/api.md
```diff
@@ -1368,6 +1368,7 @@ not created dynamically.
 | update_time | [google.protobuf.Timestamp](#google.protobuf.Timestamp) |  | Time of last tree update. Readonly (automatically assigned on updates). |
 | deleted | [bool](#bool) |  | If true, the tree has been deleted. Deleted trees may be undeleted during a certain time window, after which they&#39;re permanently deleted (and unrecoverable). Readonly. |
 | delete_time | [google.protobuf.Timestamp](#google.protobuf.Timestamp) |  | Time of tree deletion, if any. Readonly. |
+| index_bytes | [int32](#int32) |  | For a fixed-size data structure, the number of prefix bytes assigned to single-byte index shards. Not applicable to variable-size logs. Example, if the tree is a fixed-size data structure with 256 bits (32 bytes), and index_bytes=3: there will be 4 tiles to describe the path for any leaf; * 3 index tiles of 1 byte each; and * 1 final leaf tile of 29 bytes height. Readonly. |
```
AlCutter, line 1371: I find this name & description difficult to grok, "single-byte index shards".
I think it might be because it's jumping between the related topics of shards/indices/tiles.
Since this moves to a new map storage model of prefix strata and a single suffix mega-strata, perhaps could adopt those (or similar) terms and document (elsewhere) that a prefix stratum is an 8bit depth stratum and the suffix stratum is the remaining tree depth after the num_prefix_strata are taken into consideration?

trillian.proto
```diff
@@ -201,6 +201,15 @@ message Tree {
   // Time of tree deletion, if any.
   // Readonly.
   google.protobuf.Timestamp delete_time = 20;
+
+  // For a fixed-size data structure, the number of prefix bytes assigned to single-byte index shards.
```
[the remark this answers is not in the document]
mhutchinson: Addressed.

docs/api.md
mhutchinson: Addressed. PTAL.

mhutchinson:
Thanks for the reviews. Looks like there are no huge objections to this approach. I've updated the terminology and will now wire this up properly.

mhutchinson:
I've convinced myself that this approach isn't quite right. This may be an even bigger adventure, but `storage_settings` on the `trillian.Tree` seems to be the right abstraction layer for this. The tiles/strata aren't exposed outside of Trillian, so having this information raw on `Tree` seems to be a layering violation. Tree is expected to be used outside of Trillian as it describes the hash strategies and key information for the tree, so I've convinced myself it shouldn't be raw within this object.

pavelkalinnikov:
Good observation. Yeah, `storage_settings` seems to be a better place. It seems like we only use it internally, and MySQL storage does not support that field yet?

mhutchinson:
@pavelkalinnikov it's used in cloudspanner https://github.com/google/trillian/blob/5b133ccb6cff318a0fee3a87c3bd927d51b47241/storage/cloudspanner/admin.go#L725, but mysql/postgres/memory implementations don't support this yet.

gdbelvin:
Just a quick request that when this gets picked up again to pop a test in `integration/storage` for good coverage

mhutchinson:
#2284 - this is no longer going to be needed.

PR #2091, opened 2020-04-25
Remove trailing foreward slash from clone url

lukehinds:
The example/README.md has a trailing `/` after the clone URL
which if used in a terminal will result in error expressing
the repo cannot be found.


gdbelvin:
Thanks for submitting this PR!

PR #2110, opened 2020-05-14
Migrate from kylelemons/godebug to google/cmp

gdbelvin:
kylelemons does not support proto v1.4

`cmp` is stricter than `pretty`:
- `cmp` warns about unexported fields, which `pretty` would ignore. In some cases (context.Context) the structs being compared had no public fields, resulting in an always true test.
- `nil` is not equal to `{}`

### DIFF
--- a/merkle/compact/range_test.go
+++ b/merkle/compact/range_test.go
@@ -26,7 +26,6 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/trillian/merkle/rfc6962"
 	"github.com/google/trillian/merkle/testonly"
 )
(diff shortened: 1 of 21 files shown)


merkle/compact/range_test.go
```diff
@@ -213,7 +214,8 @@ func TestGoldenRanges(t *testing.T) {
 			if want := roots[size]; !bytes.Equal(hash, want) {
 				t.Errorf("root hash mismatch: got %x, want %x", hash, want)
 			}
-			if diff := pretty.Compare(cr.Hashes(), hashes[size]); diff != "" {
+			// TODO(): Remove equate empty
+			if diff := cmp.Diff(cr.Hashes(), hashes[size], cmpopts.EquateEmpty()); diff != "" {
```
gdbelvin, line 218: @pavelkalinnikov - this line worries me. Could you double check the accuracy of compact range?

pavelkalinnikov: The compact range code is fine: a default constructed `compact.Range` has a default constructed `hashes` slice, i.e. `null`. So, if you set [this](https://github.com/google/trillian/blob/737aff5a2e691fdd515617d362887afb2e63effe/merkle/testonly/constants.go#L86) line to `nil`, `EquateEmpty` shouldn't be needed. Let me know if that breaks anything else.

PR #2115, opened 2020-05-18
Migrate off of deprecated proto APIs

gdbelvin:
proto v1.4 linter fixes
- Mark `github.com/golang/proto` imports. with. ` //nolint`
- Regenerate protos with `go generate ./...`
- Use `proto.Clone` for copying. (The new generated protos have mutexes inside of them)
- Migrate off deprecated `proto.EnumValueMap` and `proto.MessageName`


client/log_client_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/proto" //nolint:staticcheck
```
Martin2112, line 24: Is the nolint temporary or related to some other thing we need to address or something else?

quota/etcd/quotaapi/quota_server_test.go
```diff
@@ -436,12 +436,12 @@ func TestServer_UpdateConfig_ConcurrentUpdates(t *testing.T) {
...
 					want.CurrentTokens = got.CurrentTokens // Not important for this test
 					want.MaxTokens = tokens
-					if !proto.Equal(got, &want) {
+					if !proto.Equal(got, want) {
 						diff := cmp.Diff(got, &want, cmp.Comparer(proto.Equal))
 						t.Errorf("%v: post-UpdateConfig() diff (-got +want):\n%v", want.Name, diff)
 					}
 				}
-			}(num, *cfg)
+			}(num, proto.Clone(cfg).(*quotapb.Config))
```
Martin2112, line 444: Not sure how this got missed when I went looking for them last year.

client/log_client_test.go
gdbelvin: The nolint will be there until we do a full migration to proto v2. 

See full explanation here:
https://blog.golang.org/protobuf-apiv2

quota/etcd/quotaapi/quota_server_test.go
gdbelvin: They are easy to miss

client/log_client_test.go
Martin2112: Thanks.

PR #2117, opened 2020-05-20
Allow storage of Ed25519 Signatures

matheusd:
Following up on #1536, while Ed25519 keys are internally supported, signatures can't currently be stored due to the use of an Enum in the appropriate field.

This fixes the issue for new installations. Please advise if some form of upgrade script is needed (I couldn't find any place to add them to).

### Checklist


matheusd:
@googlebot I signed it!

marcopeereboom:
This is great. I would love to see this go in.

crypto/signatures.go
```diff
@@ -31,7 +31,7 @@ func SignatureAlgorithm(k gocrypto.PublicKey) sigpb.DigitallySigned_SignatureAlg
 		return sigpb.DigitallySigned_ECDSA
 	case *rsa.PublicKey:
 		return sigpb.DigitallySigned_RSA
-	case *ed25519.PublicKey:
+	case ed25519.PublicKey:
```
Martin2112, line 34: Why is this different to the other?

gdbelvin: The underlying type is `[]byte`. https://golang.org/src/crypto/ed25519/ed25519.go?s=1385:1406#L31

The other types are `struct`s that have pointers. 

This change looks correct

Martin2112:
Looks like the travis failure is the lint issue that Gary is fixing so this is ok to merge.

Martin2112:
@matheusd if you rebase on master and let travis run & pass you should be able to merge.

matheusd:
The only thing failing now is code coverage (which I'm not sure it's applicable since there are no added lines). Do I have to take any other actions?

Martin2112:
Merged. Thanks for the contribution.

PR #2128, opened 2020-06-04
smt: Formalise Tile struct semantics

pavelkalinnikov:
This change:

- Clarifies the semantics of `Tile` in the type comment.
- Refactors the `TileSet` type so that the invariants are preserved.
- Adds a couple of tests that catch situations where the invariants weren't previously held.

### DIFF
--- a/merkle/smt/tiles.go
+++ b/merkle/smt/tiles.go
@@ -17,6 +17,7 @@ package smt
 import (
 	"bytes"
 	"fmt"
+	"sort"
 
 	"github.com/google/trillian/merkle/hashers"
 	"github.com/google/trillian/storage/tree"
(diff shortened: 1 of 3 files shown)


merkle/smt/tiles.go
```diff
@@ -97,6 +98,9 @@ func (t *TileSetMutation) Build() ([]Tile, error) {
 	for id, upd := range t.tiles {
 		had, ok := t.read.tiles[id]
 		if !ok {
+			sort.Slice(upd, func(i, j int) bool {
```
pavelkalinnikov, line 101: This `Sort` can go away if we specify the order in which nodes can be `Set` (it is implicitly assumed that they are sorted now), or only allow inserting a whole tile at once.

Will add a TODO.

pavelkalinnikov: Done.

PR #2129, opened 2020-06-05
smt: Introduce type for sorted list of nodes

pavelkalinnikov:
This change introduces `NodesRow` type which encapsulates a list of nodes with unique and sorted IDs. This type is used for storing nodes in `Tile` type, and for passing in updates to its `Merge` method. More usages will be added in follow-up PRs.

So far `NodesRow` is backwards compatible with the previously used plain slice `[]Node`.

### DIFF
--- a/merkle/smt/nodes.go
+++ b/merkle/smt/nodes.go
@@ -28,6 +28,39 @@ type Node struct {
 	Hash []byte
 }
 
+// NodesRow contains nodes at the same tree level sorted by ID from left to
+// right. The IDs of the nodes are unique.
+//
+// TODO(pavelkalinnikov): Hide nodes so that only this package can modify them.
+type NodesRow []Node
+
+// NewNodesRow creates a NodesRow from the given list of nodes. The nodes are
+// reordered in-place if not already sorted.
+func NewNodesRow(nodes []Node) (NodesRow, error) {
+	if len(nodes) == 0 {
+		return nodes, nil
+	}
+	if err := Prepare(nodes, nodes[0].ID.BitLen()); err != nil {
+		return nil, err
+	}
+	return nodes, nil
+}
+
+// inSubtree returns whether all the nodes in this row are strictly under the
+// node with the given ID. Panics if the row is empty.
+func (n NodesRow) inSubtree(root tree.NodeID2) bool {
+	rootLen := root.BitLen()
+	if n[0].ID.BitLen() <= rootLen {
+		return false
+	}
+	if n[0].ID.Prefix(rootLen) != root {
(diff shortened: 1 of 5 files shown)


merkle/smt/tile.go
```diff
@@ -33,28 +34,34 @@ import (
...
-//
-// TODO(pavelkalinnikov): Add SortedNodes type to statically "assert" order.
+// hold. Use NewNodesRow function for ordering nodes correctly.
 type Tile struct {
 	ID     tree.NodeID2
-	Leaves []Node
+	Leaves NodesRow
 }
 
 // Merge returns a new tile which is a combination of this tile with the given
```
pavelkalinnikov, line 43: This comment needs an update.

merkle/smt/nodes.go
```diff
@@ -28,6 +28,39 @@ type Node struct {
...
+		return nodes, nil
+	}
+	if err := Prepare(nodes, nodes[0].ID.BitLen()); err != nil {
+		return nil, err
+	}
+	return nodes, nil
+}
+
+// inSubtree returns whether all the nodes in this row are strictly under the
+// node with the given ID. Panics if the row is empty.
```
mhutchinson, line 50: I wonder if returning true if the row is empty is better than panicking? This is internal though so we can always revisit.

merkle/smt/nodes.go
```diff
@@ -28,6 +28,39 @@ type Node struct {
...
+func (n NodesRow) inSubtree(root tree.NodeID2) bool {
+	rootLen := root.BitLen()
+	if n[0].ID.BitLen() <= rootLen {
+		return false
+	}
+	if n[0].ID.Prefix(rootLen) != root {
+		return false
+	}
+	// Note: It is enough to check only the first and the last node ID because
+	// the list is sorted.
```
mhutchinson, line 60: This is a Good Comment™.

merkle/smt/nodes.go
pavelkalinnikov: Yeah, I did that first, but then I realised that the wrapping code checks emptiness anyway.

Another thought visited me that an empty set has "all" the properties. I.e. `inSubtree` is a predicate "for all nodes n in this row, n is in the subtree with the given ID" which is true for an empty set, however empty set also satisfies "for all nodes n in this row, n is **not** in the subtree with the given ID". Having at least one node in the list is a "proof" that the updates were indeed intended for this subtree, whereas an empty update could be created for any other subtree and passed to this one by accident.

So, given that, I decided to force the caller of this method to at least **think** about the empty case.

PR #2132, opened 2020-06-09
In-memory cloudspanner testing

gdbelvin:
This PR also bumps the Go version to 1.14 to take advantage of the new `t.Cleanup` function.


storage/cloudspanner/gen.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+cat > spanner.sdl.go <<EOF
+// Code generated by storage/cloudspanner/gen.sh DO NOT EDIT.
+
+package cloudspanner
+
+const ddlString = \`
+EOF
+cat spanner.sdl >> spanner.sdl.go
```
AlCutter, line 9: This will break if anyone adds a ` char to the SDL file, a possible workaround might be something like:

```
const ddlBase64 = \"
EOF

base64 < spanner.sdl >> spanner.sdl.go

echo '\"' >> spanner.sdl.go
```

and add a test helper to unbase64 and return the the DDL to tests?

storage/cloudspanner/map_storage_test.go
```diff
@@ -33,7 +45,11 @@ var cloudDBPath = flag.String("test_cloud_spanner_database", "", "eg: projects/m
 func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
 	t.Helper()
 	if *cloudDBPath == "" {
-		t.Skip("-test_cloud_spanner_database flag is unset")
+		ddl, err := readDDL()
```
AlCutter, line 48: Worth extending the flag description string above (L43) to mention that leaving empty means use the in-mem test DB?

storage/cloudspanner/map_storage_test.go
```diff
@@ -42,12 +58,68 @@ func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
 	return client
 }
 
+// Unique per test binary invocation
+var timestamp = time.Now().UTC().Format("jan-02-15-04-05")
```
AlCutter, line 62:
```
var (
  ...
)
```

Also, should these be inside the `uniqueDBName` func?

storage/cloudspanner/map_storage_test.go
```diff
@@ -42,12 +58,68 @@ func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
...
+var testBinary = strings.ToLower(strings.Replace(path.Base(os.Args[0]), ".test", "", 1))
+var invocationID = fmt.Sprintf("%s-%s", timestamp, testBinary)
+var dbCount uint32 // Unique per test invocation
+
+func uniqueDBName(project, instance string) string {
+	database := fmt.Sprintf("%s-%d", invocationID, atomic.AddUint32(&dbCount, 1))
```
AlCutter, line 68: Did you consider using `sync.Mutex` rather than `atomic`?

It'll be an extra line or two, but I think it'll be more clear why it's there;`atomic` is pretty low level and, e.g., I could see someone extending this function later and referencing `dbCount` directly without using `LoadInt32` - would that be a bug/introduce potential for undefined behaviour?

I don't really know for sure without going and doing some reading on memory barriers and stuff, and there are probably others out there like me :)

storage/cloudspanner/sdl_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+package cloudspanner
+
+import "testing"
+
+func TestDDL(t *testing.T) {
+	_, err := readDDL()
+	if err != nil {
+		t.Fatal(err)
+	}
```
cindylindeed, line 23: Add comment that readDDL functionality is tested in map_storage_test.go.

storage/cloudspanner/map_storage_test.go
```diff
@@ -42,12 +58,68 @@ func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
...
+	}
+	op, err := adminClient.UpdateDatabaseDdl(ctx, &databasepb.UpdateDatabaseDdlRequest{
+		Database:   dbName,
+		Statements: statements,
+	})
+	if err != nil {
+		t.Fatalf("Starting DDL update: %v", err)
+	}
+	if err := op.Wait(ctx); err != nil {
+		t.Fatal(err)
```
AlCutter, line 111: Maybe `Fatalf("Failed to update DB: %v", err)`?

storage/cloudspanner/gen.sh
gdbelvin: I think this is a bit YAGNI.  The TestDDL wil lcatch this. I also haven't seen `\`` in any SDLs

AlCutter: Backtick is perfectly valid DDL in the right place:

https://cloud.google.com/spanner/docs/data-definition-language#ddl_syntax
> Some words, like type names, are reserved in Cloud Spanner's DDL. If you need to use a reserved keyword as an identifier in your schema, enclose it in backticks (`).

storage/cloudspanner/map_storage_test.go
gdbelvin: Added `:memory:` option

storage/cloudspanner/map_storage_test.go
gdbelvin: some, yes

storage/cloudspanner/map_storage_test.go
gdbelvin: I started off with sync.Mutex, but other team members thought that atomic was better.

storage/cloudspanner/gen.sh
gdbelvin: Fair point - done.

storage/cloudspanner/sdl_test.go
gdbelvin: Testing DDL isn't really the primary purpose of `map_storage_test`, so if you don't mind I'll defer that.

storage/cloudspanner/map_storage_test.go
gdbelvin: Switched to sync.Mutex

storage/cloudspanner/map_storage_test.go
gdbelvin: done

storage/cloudspanner/map_storage_test.go
AlCutter: Cool, thanks - I think that's the right approach.

FWIW, here's [Ian Lance Taylor talking on atomic vs. mutexes](https://groups.google.com/d/msg/golang-nuts/0uPDCRiBWqc/X11KP_IaocYJ):
>The behaviour of mutexes is guaranteed by the Go memory model
(http://golang.org/ref/mem).  The behaviour of the sync/atomic
functions is not.
>
>Perhaps the memory model should discuss the sync/atomic functions.  A
downside is that that would just encourage people to use them, which
is almost certainly a bad idea.  Another downside is that I don't
think it's possible to describe the sync/atomic functions in terms of
a happens-before relationship.  I think you have to resort to odd
circumlocutions like "if a LoadInt32 happens to follow a StoreInt32 to
the same memory address, then anything that happens before the
StoreInt32 also happens before the LoadInt32."  But there is no way to
say whether the LoadInt32 follows the StoreInt32 or not, and if it
does not come later then you know nothing about the state of memory.

storage/cloudspanner/getdb_test.go
```diff
@@ -61,14 +62,21 @@ func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
 	}
 }
 
-var dbCount uint32 // Unique per test invocation
+var dbCount struct {
+	sync.Mutex
```
AlCutter, line 65: nice use of embedding :)

storage/cloudspanner/getdb_test.go
```diff
@@ -61,15 +61,22 @@ func GetTestDB(ctx context.Context, t *testing.T) *spanner.Client {
...
 func uniqueDBName(project, instance string) string {
 	// Unique per test binary invocation
 	timestamp := time.Now().UTC().Format("jan-02-15-04-05")
 	testBinary := strings.ToLower(strings.Replace(path.Base(os.Args[0]), ".test", "", 1))
 	invocationID := fmt.Sprintf("%s-%s", timestamp, testBinary)
 
-	database := fmt.Sprintf("%s-%d", invocationID, atomic.AddUint32(&dbCount, 1))
+	dbCount.Lock()
+	dbCount.count++
+	defer dbCount.Unlock()
```
AlCutter, line 77: nit: typically you'd swap this line and the one above - keep the `Lock()` and `defer Unlock()` right next to each other.

gdbelvin: done

AlCutter:
LGTM, just looks like the base64 stuff is tripping over a difference in behaviour between linux and MacOS?

Could try `base64 | tr -d '\n'` so that both GNU (which wraps by default at 76 cols) and MacOS (which doesn't) end up outputting everything on a single line (unfortunately, GNU and MacOS `base64` commands use different flags to control wrapping column behaviour).

Issue #1740, opened 2019-06-27
Create Database Integration Test Suite


gdbelvin:
Each database implementation has its own set of "unit tests" 
We should really be testing these implementations at the interface layer so we can have some confidence that the assumptions we have about one storage implementation carry to the other implementations

gdbelvin:
The storage integration test has revealed an interesting edge case. 

Here's the cloud spanner [documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.Commit):

> Commit might return an ABORTED error. This can occur at any time; commonly, the cause is conflicts with concurrent transactions. However, it can also happen for a variety of other reasons. If Commit returns ABORTED, the caller should re-attempt the transaction from the beginning, re-using the same session.

There are a new instances of 
```
	err := admin.ReadWriteTransaction(ctx, func(ctx context.Context, tx AdminTX) error {
```
That do not retry on `ABORTED`. 
These should probably be moved to `RunInAdminSnapshot` or `RunInTransaction`. 
`RunInX` should also be modified to retry on `ABORTED` errors.

gdbelvin:
@AlCutter for cloud spanner API comments

gdbelvin:
Implementation ongoing:
#2134 
#2133 
#2132

gdbelvin:
Working cloudspanner :-D 
cc: @AlCutter @rmhrisk

rmhrisk:
Woot woot ;)

mhutchinson:
@pavelkalinnikov do you have any context on what work is remaining here?

pavelkalinnikov:
`integration/storagetest` contains generic tests that can be used to test any storage implementation. If I remember correctly, it works for the storages in this repos, and what was left is making it work for our internal storage implementation. We can track it internally though.

PR #2134, opened 2020-06-09
Move Test{Snapshot,ReadWriteTransaction} to log integration test

gdbelvin:
Bug fixes in cloudspanner


storage/cloudspanner/tree_storage.go
```diff
@@ -208,12 +209,8 @@ func (t *treeTX) currentSTH(ctx context.Context) (*spannerpb.TreeHead, error) {
 }
 
 func (t *treeTX) writeRev(ctx context.Context) (int64, error) {
-	if err := t.getLatestRoot(ctx); err == storage.ErrTreeNeedsInit {
-		return 0, nil
-	} else if err != nil {
-		return -1, fmt.Errorf("writeRev(): %v", err)
-	}
-	return t._writeRev, nil
+	err := t.getLatestRoot(ctx)
```
AlCutter, line 212: This looks as though it may break tree creation due to removing the special case when a tree exists, but has no root?

gdbelvin: This is where the new integration tests help us :-)
The `ErrTreeNeedsInit` error was being special cased a few *too* many times. It's special cased both here and in `getLatestRoot`. The result is that the second special case overrides the first, causing the first revision to be 0 rather than -1.

gdbelvin: I take this back -- another nice reason to have integration tests :-)

gdbelvin:
Ready for a second look. These integration tests are quickly making a difference!

storage/cloudspanner/log_storage.go
```diff
@@ -379,7 +376,7 @@ func (tx *logTX) StoreSignedLogRoot(ctx context.Context, root *trillian.SignedLo
 			int64(logRoot.TreeSize),
 			logRoot.RootHash,
 			root.LogRootSignature,
-			writeRev,
+			int64(logRoot.Revision),
```
AlCutter, line 379: Can a caller now write a log root of revision N+M leaving a gap?
Also, would any tiles added in this same transaction be using `writeRev` for _their_ revision, but no be longer guaranteed to match the "corresponding" logRoot revision?

storage/cloudspanner/tree_storage.go
```diff
@@ -209,7 +210,7 @@ func (t *treeTX) currentSTH(ctx context.Context) (*spannerpb.TreeHead, error) {
 
 func (t *treeTX) writeRev(ctx context.Context) (int64, error) {
 	if err := t.getLatestRoot(ctx); err == storage.ErrTreeNeedsInit {
-		return 0, nil
+		return t._writeRev, nil
```
AlCutter, line 213: Is it ok to return -1 rather than 0 here now?

I think the sequencer / admin stuff is where the most tricksy `ErrTreeNeedsInit` support is located, BTW.

storage/cloudspanner/log_storage.go
```diff
@@ -141,16 +141,20 @@ func (ls *logStorage) begin(ctx context.Context, tree *trillian.Tree, readonly b
...
-	return &logTX{
+	ltx := &logTX{
 		ls:       ls,
 		dequeued: make(map[string]*QueuedEntry),
 		treeTX:   tx,
-	}, nil
+	}
+
+	// Needed to generate ErrTreeNeedsInit in SnapshotForTree and other methods.
+	if err := ltx.getLatestRoot(ctx); err == storage.ErrTreeNeedsInit {
```
AlCutter, line 151: This might be forcing an extra DB scan for an STH when one wouldn't have been done previously (e.g. if the LogServer was requesting just an inclusion proof for a given leaf, perhaps?)

storage/cloudspanner/log_storage.go
gdbelvin: Yes, but this is controlled by the sequencer -- as is also the case in the mysql and prod implementations.  Checking this is a business layer concern, not a storage layer one. 

I view this change as correcting what could be a very confusing bug -- roots being stored at a different revision than the one requested.

storage/cloudspanner/tree_storage.go
gdbelvin: Yes, but rather than having this be another edgecase, I prefer this style.
This is also the way that the mysql impl does it.

storage/cloudspanner/log_storage.go
gdbelvin: It's wasteful, yes, but our current integration tests enforce this property, and I don't know how much of the tower of logic on top relies on this. 
It should be a separate action item to change the API contract of the storage layer.

gdbelvin: The mysql impl also has this wasteful operation.

storage/cloudspanner/log_storage.go
AlCutter: Hmm, I'm not sure I agree about the responsibility - the stacked "swiss cheese" revisions is a storage implementation detail, but with this change I think a caller can corrupt the stored tree by doing:

1. adding some leaves at N, but write a log root at rev N+3
2. add a leaf (which, due to writeRev logic) will be at N+3+1, and store a log root at N+1

etc.

gdbelvin: Perhaps, but for the sake of consistency I suggest we follow what the [mysql implementation does](https://github.com/google/trillian/blob/master/storage/mysql/log_storage.go#L890).  It's also worth noting that changing the log root will invalidate any log root signatures. See https://github.com/google/trillian/pull/1052

Changing this is an API contract change and should be done consistency across storage implementations and with a corresponding unit test.

AlCutter: I take your point on signatures, although that might be a happy side-effect of invalidating an STH which is clearly produced by a logic error :).

I'd argue there has been no explicit API contract, **but** the use of `writeRev` throughout the implementations was intended to enforce the implicit and "reasonably expected" contract that all writes in a given TX would be written at the same revision returned by `WriteRevision()` on the same TX.
FWIW the currently committed implementation here is consistent with the InfraSpanner implementation, BTW.
So I'd say the MySQL behaviour is the odd one out and it's a bug.

As a path forward, how about this for a compromise: this method (and indeed the corresponding other implementations) should return an error if `logRoot.Revision != writeRevision`?

We can add that to the comment for this method on the storage API interface too.

gdbelvin:
#1740

PR #2135, opened 2020-06-11
Make the leaf counter metric more usable

pavelkalinnikov:
- Extend the existing leaf counter metric to include pre-ordered logs.
- Add log ID label to distinguish different logs.
- Rename the metric path to not conflict with the existing one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ attempting to exit has been addressed.
 A count of the total number of individual leaves the logserver attempts to
 fetch via the GetEntries.\* API methods has been added.
 
+The `queued_leaves` metric is removed, and replaced by `added_leaves` which
+covers both `QueueLeaves` and `AddSequencedLeaves`, and is labeled by log ID.
+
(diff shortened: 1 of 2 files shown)


Martin2112:
What's the conflicting metric? Renaming and changing the dimensions of it could break existing monitoring. Might be more compatible to add a new metric and deprecate the old one.

pavelkalinnikov:
This PR effectively adds a new metric, it doesn't change the dimensions of the old one (which indeed could be dangerous, because monitoring system might be confused by mismatching schemas).

But it does remove the old metric. I could leave it there and continue to populate it to be 100% safe, but it feels like a pretty useless metric, and on our end we don't use it for alerting etc. I think that removing a metric is equivalent to deprecating it: the jobs just stop reporting it.

Martin2112:
We don't know what monitoring third parties have but maybe it's unlikely to include the old metric. 

Can you update the changelog so if it does break people will know what needs updating.

pavelkalinnikov:
Done.

PR #2136, opened 2020-06-11
Check for writeRev in StoreSignedLogRoot

gdbelvin:
Enforce that each write of rev N+1 follows rev N.

Fix tests that don't abide by this.

### DIFF
--- a/storage/mysql/log_storage_test.go
+++ b/storage/mysql/log_storage_test.go
@@ -231,7 +231,7 @@ func TestReadWriteTransaction(t *testing.T) {
 			desc:          "unknownBegin",
 			tree:          logTree(-1),
 			wantNeedsInit: true,
-			wantTXRev:     -1,
+			wantTXRev:     0,
 		},
 		{
 			desc: "activeLogBegin",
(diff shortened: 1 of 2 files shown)


storage/mysql/log_storage_test.go
```diff
@@ -231,7 +231,7 @@ func TestReadWriteTransaction(t *testing.T) {
 			desc:          "unknownBegin",
 			tree:          logTree(-1),
 			wantNeedsInit: true,
-			wantTXRev:     -1,
+			wantTXRev:     0,
```
AlCutter, line 234: This is surprising, is it hitting `ErrTreeNeedsInit` somehow?

gdbelvin: Rev = -1 + 1 = 0

gdbelvin: I had to add a special case for it. Here's the line https://github.com/google/trillian/blob/aa1a68cda895fdf4cf1792f98dbe32f2cccddb23/storage/mysql/log_storage.go#L257

The reason is that, in an uninitialized tree, the first thing we write is an empty signed log root at ... wait for it ...  revision 0. Therefore `writeRevision` for a `ErrTreeNeedsInit` needs to be 0 in order to be consistent.

AlCutter: Hmm, I suspect this test has a wrong name - given it's looking for `ErrTreeNeedsInit`, and the comment above talks about unknown trees being handled elsewhere, this test is really testing `uninitializedBegin`, in which case `writeRev==0` is correct :)

Shall we rename the test case?

gdbelvin: sounds good

PR #2140, opened 2020-06-17
merkle: Inline the used-once skipMissingLevels

pavelkalinnikov:
This is a mechanical change that merges two functions together, as
the second function merely calls the first one and converts the result.

### DIFF
--- a/merkle/log_proofs.go
+++ b/merkle/log_proofs.go
@@ -384,12 +384,14 @@ func lastNodePresent(level, ts int64) bool {
 	return b&mask != 0
 }
 
-// skipMissingLevels moves down the tree a level towards the leaves until the node exists. This
-// must terminate successfully as we will eventually reach the leaves, which are always written
-// and are at level 0. Missing nodes are intermediate nodes with one child, hence their value
-// is the same as the node lower down the tree as there is nothing to hash it with.
-func skipMissingLevels(snapshot, lastNode int64, level int, node int64) (int, int64) {
+// siblingIDSkipLevels returns the ID of the sibling to the passed-in node,
+// taking into account the skipping of levels in a non-perfect tree.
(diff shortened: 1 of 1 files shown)


merkle/log_proofs.go
```diff
@@ -384,12 +384,15 @@ func lastNodePresent(level, ts int64) bool {
 	return b&mask != 0
 }
 
-// skipMissingLevels moves down the tree a level towards the leaves until the node exists. This
-// must terminate successfully as we will eventually reach the leaves, which are always written
-// and are at level 0. Missing nodes are intermediate nodes with one child, hence their value
-// is the same as the node lower down the tree as there is nothing to hash it with.
-func skipMissingLevels(snapshot, lastNode int64, level int, node int64) (int, int64) {
+// siblingIDSkipLevels creates a new NodeID for the supplied node, accounting for levels skipped
```
AlCutter, line 387: Maybe consider flipping the info around here; the primary job is to return sibling id , an implementation detail/tweaks is that it takes into account the floating node approach of the tree. E.g:

Func returns the id of the sibling to the passed in node, taking into account the skipping of levels in a non perfect tree.

pavelkalinnikov: Done.

merkle/log_proofs.go
```diff
@@ -398,8 +400,7 @@ func skipMissingLevels(snapshot, lastNode int64, level int, node int64) (int, in
 			glog.Infof("Move down: S:%d L:%d LN:%d", sibling, level, lastNode)
```
Martin2112, line 400: While you're changing this can you make this a verbose only log message, seems it was overlooked before.

pavelkalinnikov: It checks `vvLevel` in the line above.

PR #2147, opened 2020-06-19
Client side INNER JOIN

gdbelvin:
`spannertest` does not support JOIN at the moment.

### DIFF
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -52,6 +52,22 @@ const (
 													AND t.Deleted=false`
 )
 
+type leafDataCols struct {
+	TreeID              int64
+	LeafIdentityHash    []byte
+	LeafValue           []byte
+	ExtraData           []byte
+	QueueTimestampNanos int64
+}
+
+type sequencedLeafDataCols struct {
+	TreeID                  int64
+	SequenceNumber          int64
+	LeafIdentityHash        []byte
+	MerkleLeafHash          []byte
+	IntegrateTimestampNanos int64
+}
+
 // LogStorageOptions are tuning, experiments and workarounds that can be used.
 type LogStorageOptions struct {
 	TreeStorageOptions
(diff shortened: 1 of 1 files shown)


storage/cloudspanner/log_storage.go
```diff
@@ -753,12 +810,43 @@ WHERE sd.TreeID = @tree_id AND sd.SequenceNumber >= @start AND sd.SequenceNumber
...
+
+	// Results need to be returned in order [start, end), all of which
+	// should be available (as we restricted xend/count to TreeSize).
 	leaves := make(leafmap)
-	rows := tx.stx.Query(ctx, stmt)
-	if err := rows.Do(leaves.addFullRow); err != nil {
+	if err := tx.stx.Query(ctx, stmt).
+		Do(leaves.addFullRow(seqLeaves)); err != nil {
 		return nil, err
 	}
```
pavelkalinnikov, line 851: This seems identical to lines 725-745. Maybe factor out a function?

Also, lines 813-823 are repeated as well, so could be included in that function.

storage/cloudspanner/log_storage.go
```diff
@@ -681,18 +696,53 @@ func (tx *logTX) GetLeavesByIndex(ctx context.Context, indices []int64) ([]*tril
...
 
-	leaves := make(leafmap)
 	stmt := spanner.NewStatement(
-		`SELECT sd.MerkleLeafHash, ld.LeafValue, ld.ExtraData, sd.SequenceNumber, ld.LeafIdentityHash, ld.QueueTimestampNanos, sd.IntegrateTimestampNanos
-FROM SequencedLeafData as sd
-INNER JOIN LeafData as ld
-ON sd.TreeID = ld.TreeID AND sd.LeafIdentityHash = ld.LeafIdentityHash
-WHERE sd.TreeID = @tree_id and sd.SequenceNumber IN UNNEST(@seq_nums)`)
+		`SELECT
```
pavelkalinnikov, line 700: Could you add a comment somewhere elaborating why you don't use JOIN? Seems like the reason is that the cloudspanner testing tool does not support JOIN, while the production cloudspanner does. Would we want to switch back to JOIN when the testing tool supports it? In this case I would also add a TODO.

pavelkalinnikov: In a TODO please add the link where the testing tool status lists the unsupported features.

gdbelvin: done

pavelkalinnikov: Not done?

gdbelvin: Missed the TODO, added it to the next PR

PR #2155, opened 2020-06-22
TestDequeueLeaves

gdbelvin:
Promote the `TestDequeueLeaves` test to the Log Storage integration test suite. 

#1740

### DIFF
--- a/integration/storagetest/logtests.go
+++ b/integration/storagetest/logtests.go
@@ -30,8 +30,6 @@ import (
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/testonly"
 	"github.com/google/trillian/types"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	tcrypto "github.com/google/trillian/crypto"
 	storageto "github.com/google/trillian/storage/testonly"
@@ -465,47 +463,59 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
 	}
 
 	// Now try to dequeue them
-	cctx, cancel := context.WithTimeout(ctx, 3*time.Second) // Retry until timeout
-	defer cancel()
-	if err := s.ReadWriteTransaction(cctx, tree,
-		func(ctx context.Context, tx2 storage.LogTreeTX) error {
-			leaves2, err := tx2.DequeueLeaves(ctx, 99, fakeDequeueCutoffTime)
-			if err != nil {
-				t.Fatalf("Failed to dequeue leaves: %v", err)
-			}
-			if len(leaves2) != leavesToInsert {
-				t.Logf("Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
-				return status.Errorf(codes.Aborted, "retry")
-			}
-			ensureAllLeavesDistinct(leaves2, t)
-			iTimestamp := ptypes.TimestampNow()
-			for i, l := range leaves2 {
-				l.IntegrateTimestamp = iTimestamp
-				l.LeafIndex = int64(i)
-			}
-			if err := tx2.UpdateSequencedLeaves(ctx, leaves2); err != nil {
-				t.Fatalf("UpdateSequencedLeaves(): %v", err)
-			}
-			return nil
-		}); err != nil {
-		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
+	// Some dequeue implementations probabalistically dequeue and require retrying until timeout.
+	leaves2 := dequeueLeaves(ctx, t, s, tree, fakeDequeueCutoffTime, 5, 0)
+	if len(leaves2) != leavesToInsert {
+		t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
 	}
 
 	// If we dequeue again then we should now get nothing
-	if err := s.ReadWriteTransaction(cctx, tree,
+	if err := s.ReadWriteTransaction(ctx, tree,
 		func(ctx context.Context, tx3 storage.LogTreeTX) error {
 			leaves3, err := tx3.DequeueLeaves(ctx, 99, fakeDequeueCutoffTime)
 			if err != nil {
 				t.Fatalf("Failed to dequeue leaves (second time): %v", err)
 			}
 			if len(leaves3) != 0 {
-				t.Logf("Dequeued %d leaves but expected to get %d", len(leaves3), leavesToInsert)
-				return status.Errorf(codes.Aborted, "retry")
+				t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves3), leavesToInsert)
 			}
 			return nil
 		}); err != nil {
 		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
 	}
+}
+
+// dequeueLeaves starts a transaction and uses it to dequeue leaves.
+func dequeueLeaves(ctx context.Context, t *testing.T, ls storage.LogStorage, tree *trillian.Tree, ts time.Time, limit int, startIndex int64) []*trillian.LogLeaf {
+	// We'll retry a few times if we get nothing back since we're now dependent
+	// on the underlying queue delivering unsequenced entries.
+	var ret []*trillian.LogLeaf
+	cctx, cancel := context.WithTimeout(ctx, 5*time.Second) // Retry until timeout
+	defer cancel()
+	err := ls.ReadWriteTransaction(cctx, tree, func(ctx context.Context, tx storage.LogTreeTX) error {
+		ret = make([]*trillian.LogLeaf, 0, limit)
+		for len(ret) < limit {
+			got, err := tx.DequeueLeaves(ctx, limit, ts)
+			if err != nil {
+				return err
+			}
+			ret = append(ret, got...)
+		}
+		ensureAllLeavesDistinct(ret, t)
+		iTimestamp := ptypes.TimestampNow()
+		for i, l := range ret {
+			l.IntegrateTimestamp = iTimestamp
+			l.LeafIndex = int64(i) + startIndex
+		}
+		if err := tx.UpdateSequencedLeaves(ctx, ret); err != nil {
+			return fmt.Errorf("UpdateSequencedLeaves(): %v", err)
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("dequeueLeaves(limit: %v): %v", limit, err)
+	}
+	return ret
 }
 
 func ensureAllLeavesDistinct(leaves []*trillian.LogLeaf, t *testing.T) {
--- a/integration/storagetest/logtests.go
+++ b/integration/storagetest/logtests.go
@@ -474,7 +474,8 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
 				t.Fatalf("Failed to dequeue leaves: %v", err)
 			}
 			if len(leaves2) != leavesToInsert {
-				return status.Errorf(codes.Aborted, "Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
+				t.Logf("Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
+				return status.Errorf(codes.Aborted, "retry")
 			}
 			ensureAllLeavesDistinct(leaves2, t)
 			iTimestamp := ptypes.TimestampNow()
@@ -498,7 +499,8 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
 				t.Fatalf("Failed to dequeue leaves (second time): %v", err)
 			}
 			if len(leaves3) != 0 {
-				return status.Errorf(codes.Aborted, "Dequeued %d leaves but expected to get none", len(leaves3))
+				t.Logf("Dequeued %d leaves but expected to get %d", len(leaves3), leavesToInsert)
+				return status.Errorf(codes.Aborted, "retry")
 			}
 			return nil
 		}); err != nil {
--- a/integration/storagetest/logtests.go
+++ b/integration/storagetest/logtests.go
@@ -25,6 +25,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/protobuf/ptypes"
 	"github.com/google/trillian"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/testonly"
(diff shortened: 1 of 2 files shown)


integration/storagetest/logtests.go
```diff
@@ -485,21 +489,23 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
...
 				t.Fatalf("Failed to dequeue leaves (second time): %v", err)
 			}
 			if len(leaves3) != 0 {
-				t.Fatalf("Dequeued %d leaves but expected to get none", len(leaves3))
+				return status.Errorf(codes.Aborted, "Dequeued %d leaves but expected to get none", len(leaves3))
 			}
 			return nil
-		})
+		}); err != nil {
+		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
```
AlCutter, line 505: Maybe add a (got, want) for the number of leaves?

integration/storagetest/logtests.go
```diff
@@ -465,15 +467,17 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
 		t.Fatalf("Failed to queue leaves: %v", err)
 	}
 
-	{
-		// Now try to dequeue them
-		runLogTX(s, tree, t, func(ctx context.Context, tx2 storage.LogTreeTX) error {
+	// Now try to dequeue them
+	cctx, cancel := context.WithTimeout(ctx, 3*time.Second) // Retry until timeout
```
AlCutter, line 468: Could you talk a bit about why we need this timeout?

gdbelvin: ReadWriteTransaction will retry `codes.Aborted` forever unless there is a timeout :-)

integration/storagetest/logtests.go
gdbelvin: Unfortunately, this information is not available in this block. I'll add the err details via `t.Logf`

storage/cloudspanner/log_storage.go
```diff
@@ -104,13 +106,16 @@ type unsequencedCols struct {
 
 // NewLogStorage initialises and returns a new LogStorage.
 func NewLogStorage(client *spanner.Client) storage.LogStorage {
-	return NewLogStorageWithOpts(client, LogStorageOptions{})
+	return NewLogStorageWithOpts(client, LogStorageOptions{
+		DequeueAcrossMerkleBuckets:              true,
```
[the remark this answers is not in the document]
gdbelvin: How strongly do you feel about this?  
Having tested this, I consider the default behavior badly broken and very unexpected. 
The probability of dequing leaves in the default behavior is actually very low. 
1/4 time bucket * 1/16 merkle bucket ~= 64 seconds to dequeue a single leaf. 

Using dequeue across multiple buckets, this drops to 2 seconds, which is the behavior provided by default from mysql. The client libraries depending on the default behavior trigger timeouts all over the place.

storage/cloudspanner/log_storage.go
```diff
@@ -575,46 +580,46 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
-			  TreeID = @tree_id AND 
-			  Bucket >= @start_bucket AND 
-			  Bucket <= @limit_bucket
-			LIMIT 
-			  @max_num`)
-	stmt.Params["tree_id"] = tx.treeID
-	stmt.Params["start_bucket"] = startBucket
-	stmt.Params["limit_bucket"] = limitBucket
-	stmt.Params["max_num"] = limit
+	now := time.Now().UTC().Unix()
```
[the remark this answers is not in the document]
gdbelvin: Would that help the review process?
This commit is the only change to this file so there aren't stacked diffs to review on this file. 
This commit fixes issues that are validated by the tests. I thought there might be some value in bundling them.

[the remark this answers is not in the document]
gdbelvin: I take that back - adding the limit support does change this file - I'd be happy to remove that commit from this PR.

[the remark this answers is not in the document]
gdbelvin: done

integration/storagetest/logtests.go
```diff
@@ -453,3 +454,66 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+			}
+			return nil
+		})
+	}
+}
+
+func ensureAllLeavesDistinct(leaves []*trillian.LogLeaf, t *testing.T) {
+	t.Helper()
+	// All the leaf value hashes should be distinct because the leaves were created with distinct
+	// leaf data. If only we had maps with slices as keys or sets or pretty much any kind of usable
```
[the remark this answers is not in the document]
gdbelvin: done

integration/storagetest/logtests.go
gdbelvin: done

storage/cloudspanner/log_storage.go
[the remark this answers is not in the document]
AlCutter: The problem with this as a default is that the performance will drop dramatically with time - I think what's happening is that Spanner marks "deleted" rows with tombstones and needs a compaction cycle to remove them, so when you're scanning the whole queue to find leaves you also need to scan over all the tombstoned rows too which can be a very large number.  The other "statistical" approach takes longer to integrate, but it stays ~ constant - I'm not totally sure why that is the case, I suspect that it might be something to do with the finer grained keys allow jumping directly into starts of ranges, perhaps avoiding tombstones at the front(?), but the difference in performance is borne out from empirical measurements.

This behaviour is mentioned in some [comments in the code](https://github.com/google/trillian/blob/master/storage/cloudspanner/log_storage.go#L573-L586), but could do with being more clearly documented in a README in this directory.

integration/storagetest/logtests.go
AlCutter: aha, great - thanks, possibly worth popping this in a comment there for others who pass by later on?

storage/cloudspanner/log_storage.go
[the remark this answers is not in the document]
pavelkalinnikov: +1 on leaving the old default behaviour intact, as we don't know what performance impact it may bring on MySQL log operators. The O(minutes) integration time might look unexpected, but it's something at least I was aware, and by now the operators probably are as well (I think we have an integration time metric somewhere?).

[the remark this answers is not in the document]
pavelkalinnikov: Ah wait, this is`cloudspanner`. For this one, I don't have strong preference :)

integration/storagetest/logtests.go
gdbelvin: done

storage/cloudspanner/log_storage.go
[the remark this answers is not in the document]
gdbelvin: Got the tests working with the default settings :-)

gdbelvin:
Got the tests working without any code changes :-)

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("dequeueLeaves(limit: %v): %v", limit, err)
+	}
+	return ret
+}
+
+func ensureAllLeavesDistinct(leaves []*trillian.LogLeaf, t *testing.T) {
```
pavelkalinnikov, line 525:
```suggestion
func ensureAllLeavesDistinct(t *testing.T, leaves []*trillian.LogLeaf) {
```

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+			ret = append(ret, got...)
+		}
+		t.Logf("DequeueLeaves took %d tries and %v to dequeue %d leaves", i, time.Since(start), len(ret))
+		ensureAllLeavesDistinct(ret, t)
+		iTimestamp := ptypes.TimestampNow()
+		for i, l := range ret {
+			l.IntegrateTimestamp = iTimestamp
+			l.LeafIndex = int64(i) + startIndex
+		}
+		if err := tx.UpdateSequencedLeaves(ctx, ret); err != nil {
```
pavelkalinnikov, line 514: Given that this does sequencing in addition to dequeueing, I'd rename the function to something like `dequeueAndSequence`?

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+	mustSignAndStoreLogRoot(ctx, t, s, tree, 0)
+
+	leaves := createTestLeaves(leavesToInsert, 20)
+	if _, err := s.QueueLeaves(ctx, tree, leaves, fakeDequeueCutoffTime); err != nil {
+		t.Fatalf("Failed to queue leaves: %v", err)
+	}
+
+	// Now try to dequeue them
+	// Some dequeue implementations probabalistically dequeue and require retrying until timeout.
+	leaves2 := dequeueLeaves(ctx, t, s, tree, fakeDequeueCutoffTime, 5, 0)
```
pavelkalinnikov, line 467:
```suggestion
	leaves2 := dequeueLeaves(ctx, t, s, tree, fakeDequeueCutoffTime, leavesToInsert, 0)
```

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+	if len(leaves2) != leavesToInsert {
+		t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves2), leavesToInsert)
+	}
+
+	// If we dequeue again then we should now get nothing
+	if err := s.ReadWriteTransaction(ctx, tree,
+		func(ctx context.Context, tx3 storage.LogTreeTX) error {
+			leaves3, err := tx3.DequeueLeaves(ctx, 99, fakeDequeueCutoffTime)
```
pavelkalinnikov, line 475:
```suggestion
		func(ctx context.Context, tx storage.LogTreeTX) error {
			leaves, err := tx.DequeueLeaves(ctx, 99, fakeDequeueCutoffTime)
```

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+			if err != nil {
+				t.Fatalf("Failed to dequeue leaves (second time): %v", err)
+			}
+			if len(leaves3) != 0 {
+				t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves3), leavesToInsert)
```
pavelkalinnikov, line 480:
```suggestion
			if len(leaves) != 0 {
				t.Fatalf("Dequeued %d leaves but expected to get 0", len(leaves))
```

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,85 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+			}
+			return nil
+		}); err != nil {
+		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
```
pavelkalinnikov, line 484: The indentation is confusing here. I'd store `err` without `if`, and `if err != nil` in the next line.

integration/storagetest/logtests.go
gdbelvin: done

integration/storagetest/logtests.go
gdbelvin: done

integration/storagetest/logtests.go
gdbelvin: done

integration/storagetest/logtests.go
gdbelvin: done

integration/storagetest/logtests.go
gdbelvin: Agree, fixed it.

integration/storagetest/logtests.go
```diff
@@ -465,49 +463,61 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
...
+				t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves3), leavesToInsert)
 			}
 			return nil
 		}); err != nil {
 		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
 	}
 }
 
+// dequeueLeaves starts a transaction and uses it to dequeue leaves.
+func dequeueLeaves(ctx context.Context, t *testing.T, ls storage.LogStorage, tree *trillian.Tree, ts time.Time, limit int, startIndex int64) []*trillian.LogLeaf {
```
AlCutter: Should this be renamed to `dequeueAndSequenceLeaves()` or similar?

gdbelvin: yes. done

integration/storagetest/logtests.go
```diff
@@ -465,49 +463,61 @@ func (*logTests) TestDequeueLeaves(ctx context.Context, t *testing.T, s storage.
...
+	// We'll retry a few times if we get nothing back since we're now dependent
+	// on the underlying queue delivering unsequenced entries.
+	var ret []*trillian.LogLeaf
+	cctx, cancel := context.WithTimeout(ctx, 5*time.Second) // Retry until timeout
```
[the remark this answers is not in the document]
gdbelvin: Doing it here makes it easier to not share the cctx between multiple invocations.

integration/storagetest/logtests.go
```diff
@@ -453,3 +451,87 @@ func createFakeLeaf(ctx context.Context, s storage.LogStorage, tree *trillian.Tr
...
+				t.Fatalf("Dequeued %d leaves but expected to get %d", len(leaves), leavesToInsert)
+			}
+			return nil
+		},
+	); err != nil {
+		t.Fatalf("Could not dequeue the expected number of leaves: %v", err)
+	}
+}
+
+// dequeueAndSequence repeatedly dequques in a single transaction until limit is reached or a timeout occurs.
```
pavelkalinnikov, line 489:
```suggestion
// dequeueAndSequence repeatedly dequeues in a single transaction until limit is reached or a timeout occurs.
```

integration/storagetest/logtests.go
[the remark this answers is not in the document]
pavelkalinnikov: You do share it though, right?
I generally agree that `ctx` timeout should be set as high up the stack as possible.

integration/storagetest/logtests.go
gdbelvin: done

integration/storagetest/logtests.go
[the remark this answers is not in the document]
pavelkalinnikov: Btw, do you need this timeout? It completes within a couple of seconds, so maybe the test's timeout should be enough?

integration/storagetest/logtests.go
gdbelvin: done

PR #2156, opened 2020-06-22
Remove tx.QueueLeaves interface

gdbelvin:
Rewrite tests in terms of the public `storage.QueueLeaves` API rather than the private `tx.QueueLeaves` interface.

Now we can delete `NotImplemented` functions in spanner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 Not yet released; provisionally v1.4.0 (may change).
 
+### Storage TX Interfaces
+- `QueueLeaves` has been removed from the `LogTreeTX` interface because
+  `QueueLeaves` is not transactionaal.  All callers use the
+  `QueueLeaves` function in the `LogStorage` interface.
+
 ### HTTP APIs
 
 The HTTP/JSON APIs have been removed in favor of a pure gRPC intereface.
--- a/storage/log_storage.go
+++ b/storage/log_storage.go
@@ -76,16 +76,6 @@ type LogTreeTX interface {
 	// StoreSignedLogRoot stores a freshly created SignedLogRoot.
 	StoreSignedLogRoot(ctx context.Context, root *trillian.SignedLogRoot) error
 
-	// QueueLeaves enqueues leaves for later integration into the tree.
-	// If error is nil, the returned slice of leaves will be the same size as the
-	// input, and each entry will hold:
-	//  - the existing leaf data (the LogLeaf fields not marked output-only) if
-	//    a duplicate has been submitted
-	//  - nil otherwise.
-	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
-	// considered duplicate if their leaf.LeafIdentityHash matches.
-	QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.LogLeaf, error)
-
 	// DequeueLeaves returns between [0, limit] leaves to be integrated to the
 	// tree.
 	//
@@ -159,9 +149,6 @@ type LogStorage interface {
 	//
 	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
 	// considered duplicate if their leaf.LeafIdentityHash matches.
-	//
-	// Note that in contrast to LogTX.QueueLeaves, implementations of this func must not return
-	// slices with nil values.
 	QueueLeaves(ctx context.Context, tree *trillian.Tree, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.QueuedLogLeaf, error)
 
 	// AddSequencedLeaves stores the `leaves` and associates them with the log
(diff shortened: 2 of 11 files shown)


pphaneuf:
How about this here:

https://github.com/google/trillian/blob/f77465b0b600f3b2946d72d170a4bd81e200422c/storage/tools/dump_tree/dumplib/dumplib.go#L397

gdbelvin:
Fixed up the missing tests, sorry about that

storage/log_storage.go
```diff
@@ -76,16 +76,6 @@ type LogTreeTX interface {
...
 
-	// QueueLeaves enqueues leaves for later integration into the tree.
-	// If error is nil, the returned slice of leaves will be the same size as the
-	// input, and each entry will hold:
-	//  - the existing leaf data (the LogLeaf fields not marked output-only) if
-	//    a duplicate has been submitted
-	//  - nil otherwise.
-	// Duplicates are only reported if the underlying tree does not permit duplicates, and are
-	// considered duplicate if their leaf.LeafIdentityHash matches.
-	QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.LogLeaf, error)
```
pavelkalinnikov, line 87: There is a mention of `tx.QueueLeaves` below in line 155, please remove it as well.

pavelkalinnikov:
Please update mocks.

pavelkalinnikov:
Technically, this is a breaking change. `QueueLeaves` is not a private method. Should we mark it as deprecated and leave empty bodies in all the implementations for now?

pavelkalinnikov:
After another though, this is not a breaking change, as it **removes** an interface method. So, any existing implementations would still be in "implements" relation with the interface.

storage/log_storage.go
gdbelvin: done

storage/mysql/log_storage_test.go
```diff
@@ -176,32 +178,28 @@ func TestQueueDuplicateLeaf(t *testing.T) {
...
-						t.Fatalf("QueueLeaves()[%d]=nil; want non-nil", i)
-					} else if !bytes.Equal(got.LeafIdentityHash, want.LeafIdentityHash) {
-						t.Fatalf("QueueLeaves()[%d].LeafIdentityHash=%x; want %x", i, got.LeafIdentityHash, want.LeafIdentityHash)
+			if len(existing) != len(test.want) {
+				t.Fatalf("|QueueLeaves()|=%d; want %d", len(existing), len(test.want))
+			}
+			for i, want := range test.want {
+				got := existing[i]
+				if want == nil {
+					if got, want := status.FromProto(got.GetStatus()).Code(), codes.OK; got != want {
```
AlCutter, line 192: Would `gotStatus`, `wantStatus` to avoid shadowing variables in the enclosing scope be more readable?

storage/tools/dump_tree/dumplib/dumplib.go
```diff
@@ -388,19 +388,13 @@ func sequenceLeaves(ls storage.LogStorage, seq *log.Sequencer, tree *trillian.Tr
...
-
-			if _, err := tx.QueueLeaves(context.TODO(), leaves, time.Now()); err != nil {
-				glog.Fatalf("QueueLeaves got: %v, want: no err", err)
-			}
-			return nil
-		})
-		if err != nil {
-			glog.Fatalf("ReadWriteTransaction: %v", err)
+		hash := sha256.Sum256(leafData)
+		lh := []byte(hash[:])
```
AlCutter, line 392: can this just be `lh := hash[:]`?

AlCutter:
Could you pop a note in the CHANGELOG too?

storage/tools/dump_tree/dumplib/dumplib.go
gdbelvin: yep

storage/mysql/log_storage_test.go
gdbelvin: Discovered today that this can be simplified to `got.Status != nil`

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 Not yet released; provisionally v1.4.0 (may change).
 
+### Storage TX Interfaces
```
pavelkalinnikov, line 7: Move this down to the "Storage" section?

gdbelvin: Is time ordered change easier? The change log hasn't been updated since the last release

pavelkalinnikov: I don't know for sure, but looking at `CHANGELOG.md` as it is now, it feels like it's split by category rather than time, and the time is factored into versions.

gdbelvin: The topical ordering of the CHANGELOG file caused me to make [broken releases](https://github.com/google/trillian/releases/tag/v1.3.4) in the past, so I'm nervous about continuing the topical ordering scheme.

pavelkalinnikov: I might be lacking some context (e.g. not sure how ordering in changelog can cause a broken release), but here is my mental model:

Topics in the changelog don't cover the entire history (it seems like your concern regards that). Topics are only for the stuff between the last release and `HEAD`. When a new release is created, this text is copied from changelog to the release notes, and in the changelog it is replaced by a very short version of it. The new topical list of changes is reset to empty.

PR #2159, opened 2020-06-23
Delete leaves from queue in UpdateLeaves

gdbelvin:
This change brings the mysql implementation into line with what spanner does so we can have a consistent behavior between implementations. 

Previous behavior:
- mysql impl deleted leaves from the queue inside `Dequeue`
- spanner does not delete inside `Dequeue`, it does it in `UpdateLeaves`

New / spanner behavior:
- mysql deletes leaves inside of `UpdateLeaves`
- updated mysql tests

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -250,8 +250,9 @@ func (m *mySQLLogStorage) beginInternal(ctx context.Context, tree *trillian.Tree
 	}
 
 	ltx := &logTreeTX{
-		treeTX: ttx,
-		ls:     m,
+		treeTX:   ttx,
+		ls:       m,
+		dequeued: make(map[string]dequeuedLeaf),
 	}
 	ltx.slr, err = ltx.fetchLatestRoot(ctx)
 	if err == storage.ErrTreeNeedsInit {
@@ -348,9 +349,10 @@ func (m *mySQLLogStorage) QueueLeaves(ctx context.Context, tree *trillian.Tree, 
 
 type logTreeTX struct {
 	treeTX
-	ls   *mySQLLogStorage
-	root types.LogRootV1
-	slr  *trillian.SignedLogRoot
+	ls       *mySQLLogStorage
+	root     types.LogRootV1
+	slr      *trillian.SignedLogRoot
+	dequeued map[string]dequeuedLeaf
 }
 
 func (t *logTreeTX) ReadRevision(ctx context.Context) (int64, error) {
@@ -389,7 +391,6 @@ func (t *logTreeTX) DequeueLeaves(ctx context.Context, limit int, cutoffTime tim
 	defer stx.Close()
 
 	leaves := make([]*trillian.LogLeaf, 0, limit)
-	dq := make([]dequeuedLeaf, 0, limit)
 	rows, err := stx.QueryContext(ctx, t.treeID, cutoffTime.UnixNano(), limit)
 	if err != nil {
 		glog.Warningf("Failed to select rows for work: %s", err)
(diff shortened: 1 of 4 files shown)


Martin2112:
Looks like the batched queue version of the code will need to be adjusted as well.

gdbelvin:
After taking a closer look at the Batched Queue implementation, I'm not sure if I can make this change.  The batched queue version takes `QueueID`, a value not passed in `LogLeaf`. 

I'm unsure how to proceed now, since there's not really a behavior we can standardize and test against between the different storage implementations.

gdbelvin:
@Martin2112 @AlCutter -- would love your thoughts on this.  I was going to abandon this, but the cloudspanner dequeue operations are currently broken, and this is a prerequisite to having dequeue tests.

gdbelvin:
@Martin2112 the build is [passing](https://travis-ci.org/github/google/trillian/builds/701611375) now. Not quite sure why the github / travis hook isn't picking it up.

storage/mysql/queue.go
```diff
@@ -103,14 +104,21 @@ func (t *logTreeTX) UpdateSequencedLeaves(ctx context.Context, leaves []*trillia
 			glog.Warningf("Failed to update sequenced leaves: %s", err)
 			return err
 		}
+
+		qe, ok := t.dequeued[string(leaf.LeafIdentityHash)]
+		if !ok {
+			return fmt.Errorf("attempting to assign unknown merkleleafhash %x", leaf.MerkleLeafHash)
+		}
+		dequeuedLeaves = append(dequeuedLeaves, qe)
```
Martin2112, line 112: Do we expect that after this loop finishes all leaves in `t.dequeued` will have been copied to `dequeuedLeaves`?

If so could enforce this by removing them from the map and checking it's empty after.

storage/mysql/queue.go
```diff
@@ -103,14 +104,21 @@ func (t *logTreeTX) UpdateSequencedLeaves(ctx context.Context, leaves []*trillia
 			glog.Warningf("Failed to update sequenced leaves: %s", err)
 			return err
 		}
+
+		qe, ok := t.dequeued[string(leaf.LeafIdentityHash)]
+		if !ok {
+			return fmt.Errorf("attempting to assign unknown merkleleafhash %x", leaf.MerkleLeafHash)
```
Martin2112, line 110: This error message could be made clearer. How about something like "attempting to update leaf that wasn't dequeued?" + same in batched version.

storage/mysql/log_storage.go
```diff
@@ -348,9 +349,10 @@ func (m *mySQLLogStorage) QueueLeaves(ctx context.Context, tree *trillian.Tree,
 
 type logTreeTX struct {
 	treeTX
-	ls   *mySQLLogStorage
-	root types.LogRootV1
-	slr  *trillian.SignedLogRoot
+	ls       *mySQLLogStorage
+	root     types.LogRootV1
+	slr      *trillian.SignedLogRoot
+	dequeued map[string]dequeuedLeaf
```
Martin2112, line 355: Accumulating leaves in the tx seems a big strange. Is this the only option that works?

storage/mysql/queue.go
gdbelvin: I don't think so? Leaves only get properly sequenced upon `UpdateLeaves`

storage/mysql/log_storage.go
gdbelvin: Following the spanner impl... 

I think the API is a bit strange, but at least it's consistent.  

One we have good test coverage across everything, we could consider changing the API.

storage/mysql/queue.go
Martin2112: I'm trying to decide what the invariants should be and if it could go wrong in new ways. If we update a subset of the leaves and commit then presumably the others will just get returned next time so that should be safe.

Martin2112: Given we have another version that does the same thing this is probably ok. The batched queue goes through integration as well as unit tests?

gdbelvin: yes, integration and unit test

storage/mysql/queue.go
gdbelvin: Updated

gdbelvin:
Tweeked the error message.  Anything else needed?

Martin2112:
As long as we're happy that this is being properly tested and won't cause issues for users.

PR #2162, opened 2020-06-23
Remove Unused AddSequencedLeaves from LogTX interface

gdbelvin:

### DIFF
--- a/storage/log_storage.go
+++ b/storage/log_storage.go
@@ -100,12 +100,6 @@ type LogTreeTX interface {
 	// employing this property. Consult the call sites of this method to be sure.
 	DequeueLeaves(ctx context.Context, limit int, cutoff time.Time) ([]*trillian.LogLeaf, error)
 
-	// AddSequencedLeaves stores the passed in leaves at the log positions
-	// specified in the `LeafIndex` field. The indices must be contiguous.
-	//
-	// See LogStorage.AddSequencedLeaves comment for more details.
-	AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf, timestamp time.Time) ([]*trillian.QueuedLogLeaf, error)
-
 	// UpdateSequencedLeaves associates the leaves with the sequence numbers
 	// assigned to them.
 	UpdateSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf) error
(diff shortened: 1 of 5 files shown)


storage/log_storage.go
```diff
@@ -100,12 +100,6 @@ type LogTreeTX interface {
 	// employing this property. Consult the call sites of this method to be sure.
 	DequeueLeaves(ctx context.Context, limit int, cutoff time.Time) ([]*trillian.LogLeaf, error)
 
-	// AddSequencedLeaves stores the passed in leaves at the log positions
-	// specified in the `LeafIndex` field. The indices must be contiguous.
-	//
-	// See LogStorage.AddSequencedLeaves comment for more details.
-	AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogLeaf, timestamp time.Time) ([]*trillian.QueuedLogLeaf, error)
```
Martin2112, line 107: All the implementations have been cleaned up?

storage/postgres/log_storage_test.go
```diff
@@ -373,13 +373,12 @@ func initAddSequencedLeavesTest(t *testing.T) addSequencedLeavesTest {
...
-			t.t.Fatalf("Failed to add sequenced leaves: %v", err)
-		}
-		// TODO(pavelkalinnikov): Verify returned status for each leaf.
-		return nil
-	})
+	ctx := context.Background()
+	if _, err := t.s.AddSequencedLeaves(ctx, t.tree, leaves, fakeQueueTime); err != nil {
+		t.t.Fatalf("Failed to add sequenced leaves: %v", err)
+	}
+
```
pavelkalinnikov, line 380: Could you remove this blank line?

storage/log_storage.go
pavelkalinnikov: MySQL?
In-memory storage impl? (not sure it still exists)
Mocks?

pavelkalinnikov: The MySQL one is still used but only internally. Could you unexport it as `addSequencedLeaves`?

gdbelvin: No implementions use this now

storage/postgres/log_storage_test.go
gdbelvin: done

storage/log_storage.go
Martin2112: It looks like the MySQL one might still exist? Anyway it can be sorted out later.

gdbelvin:
It only exists as a private method

pavelkalinnikov:
It's still [exported](https://github.com/google/trillian/blob/e66c3f708182ff36c01864e45677058b7bf025ea/storage/mysql/log_storage.go#L564), so not really a "private" one. I suggested unexporting it, i.e. renaming to `addSequecedLeaves`.

gdbelvin:
True, but because it's on a private struct, and because it only gets
returned through the storage.ReadOnlyLogTX interface, I'm not sure an
external caller can get to it.

Issue #2143, opened 2020-06-18
Simplify log proofs generation


pavelkalinnikov:
The proofs generation code needs a refreshment. It is old, bulky, and has references to C++ implementation that is obsolete. For example, the [merkle/log_proofs.go](https://github.com/google/trillian/blob/bc537e07cd165bc877c05412174eba700f6981e6/merkle/log_proofs.go) file can be [reduced](https://github.com/google/trillian/pull/2142) by >200 LoC if it uses helpers from the compact ranges package.

Algorithmic components of Merkle trees are scattered across the codebase. For instance, it is clear that `Rehash` parameter is meant to be consumed by the [rehasher](https://github.com/google/trillian/blob/2123c407a3646b7544a218fa266b4cd87a2885f6/server/proof_fetcher.go#L52) type. This type should probably migrate to `merkle` folder and come in pair with `log_proofs.go`.

Plan:

- [x] Refactor tests (e.g. [merkle/log_proofs_test.go](https://github.com/google/trillian/blob/bc537e07cd165bc877c05412174eba700f6981e6/merkle/log_proofs_test.go)).
- [x] Increase coverage to >100% to prevent regressions.
- [x] Convert to using compact ranges in backwards-compatible way [[prototype PR](https://github.com/google/trillian/pull/2142)].
- [x] Move purely algorithmic components to `merkle`.

pavelkalinnikov:
@AlCutter @Martin2112

pavelkalinnikov:
The bigger goal is that `merkle` package is reuse-friendly.

Martin2112:
Not sure what input you're looking for? It sounds like an improvement worth having. Al might have a view on whether to do it now or later.

pavelkalinnikov:
This is just an FYI, to explain the bigger picture. I will be sending some PRs to you soon.
Re now/later, in my mind reusability of `merkle` package fits into the current goals as we'll be using it internally for some experiments.

Martin2112:
OK if it helps current work that is an argument for doing it now.

AlCutter:
👍  Sounds great!

PR #2163, opened 2020-06-24
merkle: Refactor snapshotConsistency function

pavelkalinnikov:
This change cleans up `snapshotConsistency` function:
 - Don't allocate memory for the empty case.
 - Replace the loop of integer shifts by a single line that counts bits in one go.
 - Convert to `uint/uint64` asap.

Part of #2143

### DIFF
--- a/merkle/log_proofs.go
+++ b/merkle/log_proofs.go
@@ -90,37 +90,29 @@ func CalcConsistencyProofNodeAddresses(snapshot1, snapshot2, treeSize int64) ([]
 	return snapshotConsistency(snapshot1, snapshot2, treeSize)
 }
 
-// snapshotConsistency does the calculation of consistency proof node addresses between
-// two snapshots. Based on the C++ code used by CT but adjusted to fit our situation.
+// snapshotConsistency does the calculation of consistency proof node addresses
+// between two snapshots in a bigger tree of the given size.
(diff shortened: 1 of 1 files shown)


merkle/log_proofs.go
```diff
@@ -90,37 +90,29 @@ func CalcConsistencyProofNodeAddresses(snapshot1, snapshot2, treeSize int64) ([]
 	return snapshotConsistency(snapshot1, snapshot2, treeSize)
 }
 
-// snapshotConsistency does the calculation of consistency proof node addresses between
-// two snapshots. Based on the C++ code used by CT but adjusted to fit our situation.
+// snapshotConsistency does the calculation of consistency proof node addresses
+// between two snapshots in a bigger tree of the given size.
```
Martin2112, line 94: We might want to remove 'snapshot' terminology at some point as it was carried over from C++.

What about just saying "between two tree sizes" and noting that snapshot2 must be >= snapshot1.

pavelkalinnikov: Good idea. Will remove this terminology in a follow-up PR.

pavelkalinnikov:
For extra assurance, I ran a test that ensures the output is identical between the old/new function for all tree/snapshot sizes between -1 and 500.

PR #2166, opened 2020-06-26
Implement Mutli-MerkleBucketDequeue

gdbelvin:
Implement MultiBucketDequeue with Wraparound

- Implement wrap-around logic for both default and multi-bucket dequeue
- Implement multi-bucket dequeue

Performance:
- MultiBucketDequeue: false : 
`DequeueLeaves took map[1593164584:3478 1593164585:6219 1593164586:21] tries and 1.574777097s to dequeue 5 leaves`
- MultiBucketDequeue: true: 
`DequeueLeaves took map[1593164747:2578 1593164748:6143 1593164749:1] tries and 1.484381074s to dequeue 5 leaves`
- Original Impl: 
`DequeueLeaves took map[1593164498:3613 1593164499:4202 1593164500:11] tries and 1.921977141s to dequeue 5 leaves`

It's also worth noting that the original implementation, as I read it used a merkle bucket fraction of 1.0, but did not implement wraparound, meaning that merkel bucket 0xff would always be fetched, while bucket 0x00 would rarely be fetched. 

#1740

### Checklist

### DIFF
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"sort"
 	"sync"
@@ -592,7 +593,7 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 	if tx.ls.opts.DequeueAcrossMerkleBuckets {
 		suffixFraction = tx.ls.opts.DequeueAcrossMerkleBucketsRangeFraction
 	}
-	suffixEnd := suffixStart + int64(suffixBuckets*suffixFraction)
+	suffixEnd := suffixStart + int64(math.Ceil(suffixBuckets*suffixFraction))
 
 	keysets := []spanner.KeySet{}
 	if suffixEnd < suffixBuckets {
@@ -614,8 +615,9 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
 			},
 			spanner.KeyRange{
 				Start: spanner.Key{tx.treeID, prefix},
-				End:   spanner.Key{tx.treeID, prefix | suffixEnd},
-				Kind:  spanner.ClosedClosed,
+				// XXX: When suffixFraction = 1, this produces an overlapping range at suffixStart
+				End:  spanner.Key{tx.treeID, prefix | suffixEnd},
+				Kind: spanner.ClosedClosed,
 			})
 	}
 
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -52,6 +52,8 @@ const (
 WHERE (t.TreeType = 1 OR t.TreeType = 3)
 AND (t.TreeState = 1 OR t.TreeState = 5)
 AND t.Deleted=false`
+
+	suffixBuckets = 256
 )
 
 // LogStorageOptions are tuning, experiments and workarounds that can be used.
@@ -104,13 +106,16 @@ type unsequencedCols struct {
 
 // NewLogStorage initialises and returns a new LogStorage.
 func NewLogStorage(client *spanner.Client) storage.LogStorage {
-	return NewLogStorageWithOpts(client, LogStorageOptions{})
+	return NewLogStorageWithOpts(client, LogStorageOptions{
+		DequeueAcrossMerkleBuckets:              true,
+		DequeueAcrossMerkleBucketsRangeFraction: 1.0,
+	})
 }
 
 // NewLogStorageWithOpts initialises and returns a new LogStorage.
 // The opts parameter can be used to enable custom workarounds.
 func NewLogStorageWithOpts(client *spanner.Client, opts LogStorageOptions) storage.LogStorage {
-	if opts.DequeueAcrossMerkleBucketsRangeFraction <= 0 || opts.DequeueAcrossMerkleBucketsRangeFraction > 1.0 {
+	if got := opts.DequeueAcrossMerkleBucketsRangeFraction; got <= 0 || got > 1.0 {
 		opts.DequeueAcrossMerkleBucketsRangeFraction = 1.0
 	}
 	return &logStorage{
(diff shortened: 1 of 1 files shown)


storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
-	if err := rows.Do(func(r *spanner.Row) error {
-		var l trillian.LogLeaf
-		var qe QueuedEntry
-		if err := r.Columns(&qe.bucket, &qe.timestamp, &l.MerkleLeafHash, &l.LeafIdentityHash); err != nil {
-			return err
-		}
+	rows := tx.stx.Read(ctx, unseqTable, spanner.KeySets(keysets...),
+		[]string{"Bucket", colQueueTimestampNanos, colMerkleLeafHash, colLeafIdentityHash})
+	defer rows.Stop()
+	for {
```
pavelkalinnikov, line 627: Would `rows.Do(...)` make the `switch` unnecessary? It would require a special `errors.New("break")` for stopping after the limit though.

storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
-	stmt.Params["max_num"] = limit
+	suffixStart := rand.Int63n(suffixBuckets)
+	suffixFraction := float64(cfg.NumMerkleBuckets) / float64(suffixBuckets)
+	if tx.ls.opts.DequeueAcrossMerkleBuckets {
+		suffixFraction = tx.ls.opts.DequeueAcrossMerkleBucketsRangeFraction
+	}
+	suffixEnd := suffixStart + int64(suffixBuckets*suffixFraction)
+
+	keysets := []spanner.KeySet{}
+	if suffixEnd <= 0xff {
```
pavelkalinnikov, line 599:
```suggestion
	if suffixEnd < suffixBuckets {
```

storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
+		keysets = append(keysets,
+			spanner.KeyRange{
+				Start: spanner.Key{tx.treeID, prefix | suffixStart},
+				End:   spanner.Key{tx.treeID, prefix | suffixEnd},
+				Kind:  spanner.ClosedClosed,
+			})
+	} else {
+		// The range is too big and wraps around, overflowing a byte value, so we'll
+		// start the second range at 0x00 and end at the upper limit modulo 256:
+		suffixEnd %= 0x100
```
pavelkalinnikov, line 609:
```suggestion
		suffixEnd %= suffixBuckets
```

storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
+	if suffixEnd <= 0xff {
+		keysets = append(keysets,
+			spanner.KeyRange{
+				Start: spanner.Key{tx.treeID, prefix | suffixStart},
+				End:   spanner.Key{tx.treeID, prefix | suffixEnd},
+				Kind:  spanner.ClosedClosed,
+			})
+	} else {
+		// The range is too big and wraps around, overflowing a byte value, so we'll
+		// start the second range at 0x00 and end at the upper limit modulo 256:
```
pavelkalinnikov, line 608:
```suggestion
		// start the second range at 0 and end at the upper limit modulo suffixBuckets:
```

storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
+		suffixEnd %= 0x100
+		keysets = append(keysets,
+			spanner.KeyRange{
+				Start: spanner.Key{tx.treeID, prefix | suffixStart},
+				End:   spanner.Key{tx.treeID, prefix | 0xff},
```
pavelkalinnikov, line 613:
```suggestion
				End:   spanner.Key{tx.treeID, prefix | (suffixBuckets-1)},
```

storage/cloudspanner/log_storage.go
```diff
@@ -52,6 +53,8 @@ const (
 WHERE (t.TreeType = 1 OR t.TreeType = 3)
 AND (t.TreeState = 1 OR t.TreeState = 5)
 AND t.Deleted=false`
+
+	suffixBuckets = 256
```
pavelkalinnikov, line 57: This is used only in one function? I'd inline closer to where it's used, esp. given that the code relies on this being <= 256 (there is a shift left by 8 in one place).

storage/cloudspanner/log_storage.go
```diff
@@ -577,70 +580,85 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
-			  Bucket <= @limit_bucket
-			LIMIT 
-			  @max_num`)
-	stmt.Params["tree_id"] = tx.treeID
-	stmt.Params["start_bucket"] = startBucket
-	stmt.Params["limit_bucket"] = limitBucket
-	stmt.Params["max_num"] = limit
+	suffixStart := rand.Int63n(suffixBuckets)
+	suffixFraction := float64(cfg.NumMerkleBuckets) / float64(suffixBuckets)
+	if tx.ls.opts.DequeueAcrossMerkleBuckets {
```
pavelkalinnikov, line 593: I don't get the difference of behaviour with/without this flag. In both cases, a range of buckets is queried. I thought that the difference is 1 vs many buckets, but it doesn't appear so?

storage/cloudspanner/log_storage.go
gdbelvin: Could do that as wel

storage/cloudspanner/log_storage.go
gdbelvin: done

storage/cloudspanner/log_storage.go
gdbelvin: done

storage/cloudspanner/log_storage.go
gdbelvin: done

storage/cloudspanner/log_storage.go
gdbelvin: done

storage/cloudspanner/log_storage.go
gdbelvin: Without the flag, it's not 1 bucket it's `cfg.NumMerkleBuckets = 16`

storage/cloudspanner/log_storage.go
```diff
@@ -555,6 +556,8 @@ func (tx *logTX) AddSequencedLeaves(ctx context.Context, leaves []*trillian.LogL
 	return nil, ErrNotImplemented
 }
 
+const suffixBuckets = 0x100
+
```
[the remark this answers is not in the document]
gdbelvin: done

storage/cloudspanner/log_storage.go
```diff
@@ -577,44 +580,51 @@ func (tx *logTX) DequeueLeaves(ctx context.Context, limit int, cutoff time.Time)
...
-	stmt.Params["start_bucket"] = startBucket
-	stmt.Params["limit_bucket"] = limitBucket
-	stmt.Params["max_num"] = limit
+	suffixStart := rand.Int63n(suffixBuckets)
+	suffixFraction := float64(cfg.NumMerkleBuckets) / float64(suffixBuckets)
+	if tx.ls.opts.DequeueAcrossMerkleBuckets {
+		suffixFraction = tx.ls.opts.DequeueAcrossMerkleBucketsRangeFraction
+	}
+	suffixEnd := suffixStart + int64(suffixBuckets*suffixFraction)
+
```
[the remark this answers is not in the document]
gdbelvin: I don't follow what this block would do or how it would be triggered?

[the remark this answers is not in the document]
pavelkalinnikov: If `suffixFraction` is low enough, `suffixEnd` will be equal to `suffixStart`. Since below I suggest excluding `End`, it would result in an empty request. So +1 fixes that. As another alternative, you round `suffixEnd` up?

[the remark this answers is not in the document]
gdbelvin: Used `math.Ceil`

PR #2169, opened 2020-06-27
TestAddSequencedLeavesAndDequeueLeaves

gdbelvin:
- Adds special case for preordered logs to Dequeue

#1740 

### Checklist

### DIFF
--- a/integration/storagetest/logtests.go
+++ b/integration/storagetest/logtests.go
@@ -24,13 +24,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/trillian"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/types"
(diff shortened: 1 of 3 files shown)


integration/storagetest/logtests.go
```diff
@@ -535,3 +518,68 @@ func ensureAllLeavesDistinct(t *testing.T, leaves []*trillian.LogLeaf) {
...
+			IntegrateTimestamp: its,
+		})
+	}
+
+	// Check that the first sequenced entry is returned.
+	dequeued, err := dequeueLeavesInTx(ctx, s, tree, now, 1)
+	if err != nil {
+		t.Fatalf("dequeueLeaves(): %v", err)
+	}
+	if want := partial[:1]; !cmp.Equal(dequeued, want, cmp.Comparer(proto.Equal)) {
```
pavelkalinnikov, line 565: I wonder if you could reduce this to 1 call:
```golang
if diff := cmp.Diff(...); diff != "" {
	t.Errorf(..., diff)
}
```
Same 15 lines below.

gdbelvin: done

PR #2171, opened 2020-06-29
Map TestSetGetRoundTrip

gdbelvin:
#1740

### DIFF
--- a/integration/storagetest/maptests.go
+++ b/integration/storagetest/maptests.go
@@ -197,15 +197,14 @@ func (*mapTests) TestTileRoundTrip(ctx context.Context, t *testing.T, ms storage
 	}
 }
 
-func (*mapTests) TestGetSetRoundTrip(ctx context.Context, t *testing.T, ms storage.MapStorage, as storage.AdminStorage) {
+func (*mapTests) TestSetGetRoundTrip(ctx context.Context, t *testing.T, ms storage.MapStorage, as storage.AdminStorage) {
 	tree := mustCreateTree(ctx, t, as, storageto.MapTree)
 	mustSignAndStoreMapRoot(ctx, t, ms, tree, &types.MapRootV1{Revision: uint64(0)})
 
-	num := int64(10)
 	leafRevs := make(map[int64]*trillian.MapLeaf)
 
 	// First create several revisions of a given map leaf and store them:
-	for rev := int64(1); rev <= num; rev++ {
+	for rev := int64(1); rev <= 10; rev++ {
 		index := sha256.Sum256([]byte("Key"))
 		value := []byte{byte(rev)}
 		leafHash := sha256.Sum256(value)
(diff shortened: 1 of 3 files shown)


gdbelvin:
Famous last words -- but the last PR of this batch?

integration/storagetest/maptests.go
```diff
@@ -193,3 +196,64 @@ func (*mapTests) TestTileRoundTrip(ctx context.Context, t *testing.T, ms storage
 		t.Fatalf("ReadWriteTransaction(): %v", err)
 	}
 }
+
+func (*mapTests) TestGetSetRoundTrip(ctx context.Context, t *testing.T, ms storage.MapStorage, as storage.AdminStorage) {
```
mhutchinson, line 200:
```suggestion
func (*mapTests) TestSetGetRoundTrip(ctx context.Context, t *testing.T, ms storage.MapStorage, as storage.AdminStorage) {
```

You need to set before you can get :-)

integration/storagetest/maptests.go
```diff
@@ -193,3 +196,64 @@ func (*mapTests) TestTileRoundTrip(ctx context.Context, t *testing.T, ms storage
...
+	tree := mustCreateTree(ctx, t, as, storageto.MapTree)
+	mustSignAndStoreMapRoot(ctx, t, ms, tree, &types.MapRootV1{Revision: uint64(0)})
+
+	num := int64(10)
```
mhutchinson, line 204: Just inline this? I'm not sure the documentation provided by `num` materially decreases the magic number factor here.

integration/storagetest/maptests.go
```diff
@@ -193,3 +196,64 @@ func (*mapTests) TestTileRoundTrip(ctx context.Context, t *testing.T, ms storage
...
+		}
+		if got, want := len(leaves), 1; got != want {
+			t.Errorf("Get(rev: %d): %d leaves, want %d", rev, got, want)
+			continue
+		}
+		if got := leaves[0]; !cmp.Equal(got, want, protocmp.Transform()) {
+			t.Errorf("Get(rev: %d) = diff:\n%v", rev, cmp.Diff(got, want, protocmp.Transform()))
+			continue
+		}
+		if err := tx.Commit(ctx); err != nil {
```
mhutchinson, line 255: The commit here was surprising in a read-only test, but worth testing it doesn't blow  up I guess.

integration/storagetest/maptests.go
gdbelvin: :-)

integration/storagetest/maptests.go
gdbelvin: done

integration/storagetest/maptests.go
gdbelvin: Hmm, good point. At the very least it should be outside the loop

PR #2179, opened 2020-06-30
cloudspanner: Simplify parallelism with errgroup

gdbelvin:
Get seems to have a deadlock in it somewhere.
This should simplify processing and ensure that when an error condition
occurs, other threads are canceled.

### Checklist

### DIFF
--- a/storage/cloudspanner/map_storage.go
+++ b/storage/cloudspanner/map_storage.go
@@ -340,7 +340,9 @@ func (t *treeTX) parallelGetMerkleNodes(ctx context.Context, rev int64) func([]s
 		close(c)
 		ret := make([]*storagepb.SubtreeProto, 0, len(ids))
 		for st := range c {
-			ret = append(ret, st)
+			if st != nil {
+				ret = append(ret, st)
+			}
 		}
 		return ret, nil
 	}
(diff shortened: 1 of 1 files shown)


storage/cloudspanner/map_storage.go
```diff
@@ -340,9 +340,7 @@ func (t *treeTX) parallelGetMerkleNodes(ctx context.Context, rev int64) func([]s
 		close(c)
 		ret := make([]*storagepb.SubtreeProto, 0, len(ids))
 		for st := range c {
-			if st != nil {
-				ret = append(ret, st)
-			}
+			ret = append(ret, st)
```
pavelkalinnikov, line 343: Does this fix some bug? I can see how the upstream code could fail because of this (see line 298 getting a nil `subtree`).

gdbelvin: I'm getting hung requests in practice.  
I can't say for sure that this resolves the bug, but I'm happier with this style of of parallelism than the previous one. Seems like less can go wrong.

gdbelvin: Wrong comment -- this if just seemed unnecessary to me. All code paths that result in a nil response return an error.

pavelkalinnikov: It feels like `getSubtree` call 13 lines above can return nil. This is a valid case, because some empty tiles can be missing in the storage (e.g. when the tree is initially empty).

gdbelvin: Fixed.

PR #2180, opened 2020-06-30
Refactor consistency proof tests

pavelkalinnikov:
This is a purely mechanical change that cleans up consistency
proof tests in `log_proofs_test.go` file.

Part of #2143

### DIFF
--- a/merkle/log_proofs_test.go
+++ b/merkle/log_proofs_test.go
@@ -224,7 +224,7 @@ func TestCalcInclusionProofNodeAddresses(t *testing.T) {
 			if err != nil {
 				t.Fatalf("CalcInclusionProofNodeAddresses: %v", err)
 			}
-			comparePaths(t, "", proof, tc.want)
+			comparePaths(t, proof, tc.want)
 		})
 	}
 }
@@ -280,7 +280,7 @@ func TestCalcConsistencyProofNodeAddresses(t *testing.T) {
 			if err != nil {
 				t.Fatalf("CalcConsistencyProofNodeAddresses: %v", err)
 			}
-			comparePaths(t, "", proof, tc.want)
+			comparePaths(t, proof, tc.want)
 		})
 	}
 }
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
PTAL. Recommend commit by commit.

merkle/log_proofs_test.go
```diff
@@ -297,11 +297,12 @@ func TestCalcConsistencyProofNodeAddressesBadInputs(t *testing.T) {
 		{size1: 0, size2: 0},
 		{size1: 9, size2: 8},
 	} {
-		_, err := CalcConsistencyProofNodeAddresses(tc.size1, tc.size2, tc.size2)
-
-		if err == nil {
-			t.Fatalf("consistency path calculation accepted bad input: %v", tc)
-		}
+		t.Run(fmt.Sprintf("%d:%d", tc.size1, tc.size2), func(t *testing.T) {
```
Martin2112, line 256: Could combine these two tests into one.

merkle/log_proofs_test.go
```diff
@@ -150,108 +150,47 @@ func TestCalcInclusionProofNodeAddressesBadRanges(t *testing.T) {
 func TestCalcConsistencyProofNodeAddresses(t *testing.T) {
 	// Expected consistency proofs built from the examples in RFC 6962. Again, in
 	// our implementation node layers are filled from the bottom upwards.
+	//
```
Martin2112, line 153: If you put this outside the func it will be less indented.

merkle/log_proofs_test.go
pavelkalinnikov: Yeah, this also applies to similar inclusion proof tests. I will do both in a follow-up PR.

merkle/log_proofs_test.go
```diff
@@ -169,46 +169,38 @@ func TestCalcConsistencyProofNodeAddresses(t *testing.T) {
...
-		}
-		expectedConsistencyProofFromSize1To4 = []NodeFetch{
+		}},
+		{size1: 1, size2: 4, want: []NodeFetch{
 			newNodeFetch(0, 1, false), // b
 			newNodeFetch(1, 1, false), // h
-		}
-		expectedConsistencyProofFromSize3To7 = []NodeFetch{
+		}},
+		{size1: 6, size2: 7, want: []NodeFetch{
```
Martin2112: Looks like there was a typo in the old name.

Martin2112: Is there a reason this was moved up?

merkle/log_proofs_test.go
```diff
@@ -151,7 +151,7 @@ func TestCalcConsistencyProofNodeAddresses(t *testing.T) {
 	// Expected consistency proofs built from the examples in RFC 6962. Again, in
 	// our implementation node layers are filled from the bottom upwards.
 	//
-	//                hash6                         hash7
+	//                hash5                         hash7
```
Martin2112: It's not wrong but it probably invalidates the comment about the examples being taken from the RFC.

merkle/log_proofs_test.go
pavelkalinnikov: The order of constants did not match the order in tests, so this is why the diff looks odd. I picked one of the orders. In a follow-up commit it was reordered lexicographically anyway.

merkle/log_proofs_test.go
pavelkalinnikov: True. In fact, this tests has/will have more tests than in RFC, so I'll just remove the reference to RFC.

pavelkalinnikov: Done.

PR #2182, opened 2020-06-30
log_proofs_test: Increase coverage to 100%

pavelkalinnikov:
This change increases coverage of consistency proof generation to 100%,
and adds tests for the rehash scenarios.

Part of #2143


pavelkalinnikov:
PTAL. Might be convenient to review commit by commit.

PR #2185, opened 2020-07-01
Track session handles

gdbelvin:
Determines whether the session pool will keep track of the stacktrace of the goroutines that take sessions from the pool.
This can be used to track session leaks

### Checklist


gdbelvin:
Yes, the current `cloudspanner` implementation freezes after a while with the following error:

```
level=info grpc.start_time=2020-06-30T17:59:06Z system=grpc span.kind=server grpc.service=trillian.TrillianMap grpc.method=GetSignedMapRoot msg="finished unary call with code Canceled" error="spanner: code = \"Canceled\", desc = \"timeout / context canceled during getting session.\\nEnable SessionPoolConfig.TrackSessionHandles if you suspect a session leak to get more information about the checked out sessions.\"" grpc.code=Canceled grpc.time_ms=71440
```

This PR will give a stack trace to figure out where the session is leaking from.

PR #2186, opened 2020-07-01
merkle: Move rehashing algorithm in

pavelkalinnikov:
This change moves the `rehasher` algorithm to `merkle` package, and refactors it
to be a single function.

Part of #2143

### DIFF
--- a/merkle/log_proofs.go
+++ b/merkle/log_proofs.go
@@ -179,7 +179,6 @@ type Rehasher struct {
 	rehashing  bool
 	rehashHash []byte
 	proof      [][]byte
-	proofError error
 }
 
 func (r *Rehasher) Process(hash []byte, rehash bool) {
@@ -220,9 +219,9 @@ func (r *Rehasher) endRehashing() {
 	}
 }
 
-func (r *Rehasher) RehashedProof() ([][]byte, error) {
+func (r *Rehasher) RehashedProof() [][]byte {
 	r.endRehashing()
-	return r.proof, r.proofError
+	return r.proof
 }
 
 func reverse(ids []compact.NodeID) []compact.NodeID {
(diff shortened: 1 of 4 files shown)


pavelkalinnikov:
Might be convenient or not to review commit by commit.

merkle/log_proofs.go
```diff
@@ -177,7 +176,7 @@ func proofNodes(index uint64, level uint, size uint64, rehash bool) []NodeFetch
 
 // Rehasher bundles the rehashing logic into a simple state machine
 type Rehasher struct {
-	Hasher     hashers.LogHasher
+	Hash       func(left, right []byte) []byte
```
Martin2112: Using the same name for functions and values. HashFn?

merkle/log_proofs.go
```diff
@@ -178,7 +177,7 @@ func proofNodes(index uint64, level uint, size uint64, rehash bool) []NodeFetch
 type Rehasher struct {
 	Hash       func(left, right []byte) []byte
 	rehashing  bool
-	rehashNode tree.Node
+	rehashHash []byte
```
Martin2112: If other field is renamed this could just be 'hash'.

merkle/log_proofs.go
```diff
@@ -173,55 +174,27 @@ func proofNodes(index uint64, level uint, size uint64, rehash bool) []NodeFetch
...
-	r.rehashHash = hash
-	r.rehashing = true
-}
-
-func (r *Rehasher) endRehashing() {
-	if r.rehashing {
-		r.proof = append(r.proof, r.rehashHash)
-		r.rehashing = false
+	cursor := 0
+	for i, ln := 0, len(h); i < ln; i, cursor = i+1, cursor+1 {
```
Martin2112, line 191: I think this readers of this block could benefit from a couple of brief comments or an example. e.g. at the top what is cursor and inside what's the terminating condition on the loop.

pavelkalinnikov: Done.

PR #2192, opened 2020-07-06
Travis: go mod tidy allowed to fail

AlCutter:
Make the Travis `go mod tidy` check advisory.
Bumps the test to the top of the list too, so that it's in the same place on all our repos.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-version: ~> 1.0
 os: linux
 dist: xenial
 language: go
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,11 @@ cache:
 jobs:
   fast_finish: true
   include:
+    - name: "gomod"
```
pavelkalinnikov, line 16: How about "gomodtidy"?

AlCutter: Oke doke, done

PR #2197, opened 2020-07-17
Try to dump innodb status after the log test runs. 

Martin2112:
This might have info on the deadlock.

### Checklist


Martin2112:
This doesn't work as the script exits before it gets to run. Will have another think.

PR #2200, opened 2020-07-24
Count fetched leaves only after they are fetched

pavelkalinnikov:
do not count your chickens before they are hatched


pavelkalinnikov:
Done.

PR #2240, opened 2020-10-22
remove PeekTokens from quota.Manager

pphaneuf:
This remove the `PeekTokens` (which is only used for testing) method from the `quota.Manager` interface.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## HEAD
 
+### Misc improvements
+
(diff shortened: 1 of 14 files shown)


AlCutter (commented):
Worth a note in the CHANGELOG in case it lets anyone with an alternative implementation clean that up too?

pphaneuf:
Added an entry to the change log.

AlCutter (approved):

PR #2256, opened 2020-11-10
Demo that creates a map from generated key/values

mhutchinson:
The output map tiles are simply summarized into a single local file

### DIFF
--- a/experimental/batchmap/cmd/build/mapdemo.go
+++ b/experimental/batchmap/cmd/build/mapdemo.go
@@ -21,7 +21,6 @@ import (
 	"crypto"
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -30,6 +29,7 @@ import (
 	"github.com/apache/beam/sdks/go/pkg/beam/io/filesystem/local"
 	"github.com/apache/beam/sdks/go/pkg/beam/x/beamx"
 
+	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 
 	"github.com/google/trillian/experimental/batchmap"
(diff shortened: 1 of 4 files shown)


mhutchinson:
This now generates one file per tile for simplicity, while preserving all of the information (the initial commit in this PR threw away lots of information to summarize the tiles into a single file).

There is a verifier provided which can confirm inclusion of values by verifying the merkle structure.

I think this is a minimal enough demo that shows how this could be worked into a solution and the value it provides.

experimental/batchmap/cmd/build/mapdemo.go
```diff
@@ -0,0 +1,152 @@
...
+	beam.RegisterType(reflect.TypeOf((*writeTileFn)(nil)).Elem())
+}
+
+func main() {
+	flag.Parse()
+	beam.Init()
+
+	output := filepath.Clean(*output)
+	if output == "" {
+		log.Fatal("No output provided")
```
AlCutter, line 62: Could use `glog` for consistency with the verifier below, and then switch to `Exitf` in here?

experimental/batchmap/cmd/verify/verify.go
```diff
@@ -0,0 +1,157 @@
...
+	key          = flag.Int64("key", 0, "This is the seed for the key that will be looked up.")
+	prefixStrata = flag.Int("prefix_strata", 1, "The number of strata of 8-bit strata before the final strata.")
+)
+
+func main() {
+	flag.Parse()
+
+	mapDir := filepath.Clean(*mapDir)
+	if mapDir == "" {
+		glog.Fatal("No output provided")
```
AlCutter, line 50: Can use `Exitf`

AlCutter (approved):
nice!

Couple nitty suggestions inline, but LGTM

experimental/batchmap/cmd/build/mapdemo.go
mhutchinson: Cool, done.

experimental/batchmap/cmd/verify/verify.go
mhutchinson: Done.

PR #2272, opened 2021-01-08
cache: Ensure that goroutines never block

### LINKED ISSUE #2271: Potential goroutine leak in storage/cache/subtree_cache.go (method SubtreeCache.preload())
I think there is a potential goroutine leak in method preload() at lines 108-187 of [storage/cache/subtree_cache.go](https://github.com/google/trillian/blob/master/storage/cache/subtree_cache.go#L108).

For instance, assume:

- len(want) = 1
- s.populateConcurrency = 3 
- len(subtrees) = 3

If err != nil at [line 170](https://github.com/google/trillian/blob/master/storage/cache/subtree_cache.go#L170) in the first iteration of the range, then method preload() will return while three goroutines are stuck:

- Two goroutines are stuck at [line 159](https://github.com/google/trillian/blob/master/storage/cache/subtree_cache.go#L159) waiting to send on channel ch which is full.
- One goroutine is stuck at [line 164](https://github.com/google/trillian/blob/master/storage/cache/subtree_cache.go#L164) waiting on wg (which needs the two goroutines above to finish).

A potential fix is to change [line 141](https://github.com/google/trillian/blob/master/storage/cache/subtree_cache.go#L141)  
`ch := make(chan *storagepb.SubtreeProto, len(want))` to 
`ch := make(chan *storagepb.SubtreeProto, int(math.Max(len(want), len(subtrees)))`

FYI, I have found this bug with my tool [Gomela](https://github.com/nicolasdilley/gomela) which is a static analyser for concurrent Go programs. Let me know if you find this useful!

pavelkalinnikov:
Fixes #2271

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -138,7 +138,7 @@ func (s *SubtreeCache) preload(ids []tree.NodeID, getSubtrees GetSubtreesFunc) e
 		return err
 	}
 
-	ch := make(chan *storagepb.SubtreeProto, len(want))
+	ch := make(chan *storagepb.SubtreeProto, len(subtrees))
 	workTokens := make(chan bool, s.populateConcurrency)
 	for i := 0; i < s.populateConcurrency; i++ {
 		workTokens <- true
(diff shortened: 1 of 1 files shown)


Martin2112:
The good news is that it looks like the condition that could have triggered this is very unlikely, internal errors resulting from corruption or the wrong proto type.

pavelkalinnikov:
Yeah, this fix is mainly to make the static analysis tool happy.

AlCutter (approved):

pavelkalinnikov:
We can also assert that `len(subtrees) <= len(want)`, otherwise it can allocate an arbitrarily large channel.

PR #2278, opened 2021-01-12
Split/move hasher/registry packages to break inadvertent deps

AlCutter:
Break some more inadvertent deps.

The result of this PR & the earlier verifier-splitting one, is an 8MB reduction in client binary size - mostly gained since these changes allow us to avoid importing protobuf etc. in unnecessarily.

### Checklist

### DIFF
--- a/storage/cloudspanner/log_storage.go
+++ b/storage/cloudspanner/log_storage.go
@@ -29,7 +29,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/google/trillian"
-	"github.com/google/trillian/merkle/hashers"
+	"github.com/google/trillian/merkle/hashers/registry"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/cache"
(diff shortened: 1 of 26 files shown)


storage/cloudspanner/log_storage.go
```diff
@@ -839,16 +839,16 @@ func (tx *logTX) GetLeavesByIndex(ctx context.Context, indices []int64) ([]*tril
 	// TODO: replace with INNER JOIN when spannertest supports JOINs
 	// https://github.com/googleapis/google-cloud-go/tree/master/spanner/spannertest
 	stmt := spanner.NewStatement(
-		`SELECT 
+		`SELECT
```
mhutchinson, line 842: This might be harder than it's worth the effort, but I would like to see these trailing whitespace changes in a separate change. I'd be very sad if this changed any functionality (!), but my cautious side needs to speak out.

mhutchinson (approved):

storage/cloudspanner/log_storage.go
AlCutter: done (I think/hope)

PR #2283, opened 2021-01-18
Add CloudBuild CI configs

AlCutter:
Adds CloudBuild configs for doing CI tests for PRs.

#2298

### Checklist

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -15,7 +15,7 @@ substitutions:
   _MYSQL_ROOT_PASSWORD: ""
   _MYSQL_PASSWORD: ""
 options:
-  machineType: N1_HIGHCPU_32
+  machineType: E2_HIGHCPU_32
   env:
     - GO111MODULE=on
     - GOPROXY=https://proxy.golang.org
@@ -101,7 +101,7 @@ steps:
     - GOPATH=/throwaway
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmit
 
 # Presubmit (PKCS11)
 - id: presubmit_pkcs11
@@ -114,7 +114,19 @@ steps:
     - GOFLAGS=-race --tags=pkcs11
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmit
+
+# Try to spread the load a bit, we'll wait for all the presubmit.* steps
+# to finish before starting the integration.* ones.
+# Having too many "big" things running concurrently leads to problems
+# with timeouts and mysql issues.
+- id: presubmits_done
+  name: 'gcr.io/${PROJECT_ID}/trillian_testbase'
+  entrypoint: /bin/true
+  waitFor:
+    - presubmit
+    - presubmit_batched
+    - presubmit_pkcs11
 
 # Integration
 - id: integration
@@ -125,7 +137,7 @@ steps:
     - HAMMER_OPTS=--operations=150
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmits_done
 
 # Integration (etcd)
 - id: integration_etcd
@@ -138,7 +150,7 @@ steps:
     - HAMMER_OPTS=--operations=50
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmits_done
 
 # Integration (Batched queue)
 - id: integration_batched
@@ -150,7 +162,7 @@ steps:
     - HAMMER_OPTS=--operations=50
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmits_done
 
 # Integration (PKCS11)
 - id: integration_pkcs11
@@ -162,7 +174,7 @@ steps:
     - HAMMER_OPTS=--operations=50
     - GO_TEST_TIMEOUT=20m
   waitFor:
-    - prepare
+    - presubmits_done
 
 ############################################################################
 ## End of replicated section.
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -13,12 +13,7 @@ substitutions:
   _MASTER_ZONE: us-central1-a
   _MYSQL_TAG: "5.7"
   _MYSQL_ROOT_PASSWORD: ""
-  _MYSQL_HOST: "cloudbuild_db_1"
-  _MYSQL_PORT: "3306"
-  _MYSQL_USER: "test"
-  _MYSQL_PASSWORD: "zaphod"
-  _MYSQL_ROOT_PASSWORD: "bananas"
-  _MYSQL_USER_HOST: '%'
+  _MYSQL_PASSWORD: ""
 options:
   machineType: N1_HIGHCPU_32
   env:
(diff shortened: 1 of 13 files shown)


AlCutter:
Docker tests still need looking at, but I think I'll take that on in a follow-up, and there are issues with etcd which we can look at afterwards too.

storage/testdb/testdb.go
```diff
@@ -34,14 +35,40 @@ import (
...
 
+// mysqlURI returns the connection URI to use for tests.
+// It will prefer the value explicitly set by the --test_mysql_uri flag, if
+// that is unset it will fall back to the value in the MYSQL_URI env var, and
+// if that is unset, then it'll try the DefaultMySQLURI value.
+//
+// Only a subset of the suite of tests in this repo require a database it's
+// not possible to blanket apply the test_mysql_uri flag, and maintaining a separate
+// list of tests which needs DB connectivity is brittle, so this fallback approach
+// means we can provide the connection info in a more ambient fashion.
```
mhutchinson, line 58: This really makes me sad and a little frightened. Can we just remove the flag and say that it always uses the env variable to at least remove one layer of magic from this?

It really feels like the root problem is that we're creating multiple different mysql databases, yet wanting to use `testdb` like there is a global static singleton. Trying to solve this is going to be a whole new adventure though.

mhutchinson (commented):
There's a lot going on here. I don't think we can submit this while the docker tests and etcd tests are broken. Can we update the PR description to have checkboxes for doing this?

Otherwise I think it's looking reasonable.

cloudbuild.yaml
```diff
@@ -6,8 +15,171 @@ substitutions:
...
+  name: 'gcr.io/${PROJECT_ID}/trillian_testbase'
+  entrypoint: ./integration/cloudbuild/run_presubmit.sh
+  args:
+    - --no-linters
+    - --no-generate
+  env:
+    - GOFLAGS=-race --tags=pkcs11
+    - GO_TEST_TIMEOUT=20m
+  waitFor:
+    - presubmit
```
pavelkalinnikov, line 117: There is a lot going on here and below. Should we attempt to add one build at a time, for easier reviewability?

cloudbuild.yaml
```diff
@@ -1,3 +1,12 @@
+#############################################################################
```
pavelkalinnikov, line 1: The other 2 files ("cloudbuild_master.yaml" and "cloudbuild_tag.yaml") have almost zero comments, and don't explain why they are needed. Is there a doc somewhere?

integration/cloudbuild/run_integration.sh
```diff
@@ -0,0 +1,29 @@
...
+# Wait for mysql instance to be ready
+while ! mysql --protocol=TCP --host=${MYSQL_HOST} --port=${MYSQL_PORT} --user=root -pbananas -e quit ; do
+ sleep 5
+done
+
+export MYSQL_URI="${MYSQL_USER}:${MYSQL_PASSWORD}@tcp(${MYSQL_HOST}:${MYSQL_PORT})/${MYSQL_DATABASE}"
+
+# If the test will use etcd, then install etcd + tools.
+if [ "${ETCD_DIR}" != "" ]; then
+    go install go.etcd.io/etcd go.etcd.io/etcd/etcdctl github.com/fullstorydev/grpcurl/cmd/grpcurl
```
pavelkalinnikov, line 25: This installs the "latest" version of etcd, doesn't it? Should it fetch the version compatible with Trillian?

integration/functions.sh
```diff
@@ -128,6 +129,11 @@ log_prep_test() {
   local logsigner_opts=''
   local has_etcd=0
 
+  if [[ "${MYSQL_URI}" != "" ]]; then
+    logserver_opts+=" --mysql_uri=${MYSQL_URI}"
+    logsigner_opts+=" --mysql_uri=${MYSQL_URI}"
+  fi
+
   # Start a local etcd instance (if configured).
   if [[ -x "${ETCD_DIR}/etcd" ]]; then
```
pavelkalinnikov, line 138: Obsolete after rebasing on https://github.com/google/trillian/pull/2295?

cloudbuild.yaml
AlCutter: They manage/test kube configs for the CI/CD env, I don't know if there are docs but we should make some if not.

integration/cloudbuild/prepare.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -x
+set -e
+
+git clone --depth=1 https://github.com/googleapis/googleapis.git "$GOPATH/src/github.com/googleapis/googleapis"
+go install \
+    github.com/golang/protobuf/proto \
```
pavelkalinnikov, line 7: Again, this installs some "latest" version (master?), which might be why there is an etcd/protobuf panic.

integration/cloudbuild/testbase/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN curl -sfL https://golang.org/dl/go1.15.7.linux-amd64.tar.gz | tar -xzf - -C /usr/local
+ENV PATH /usr/local/go/bin:$PATH
+
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.18.0
+RUN mkdir protoc && \
+    (cd protoc && \
+    PROTOC_VERSION="3.11.4" && \
+    PROTOC_ZIP="protoc-${PROTOC_VERSION}-linux-x86_64.zip" && \
+    wget "https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}/${PROTOC_ZIP}" && \
```
pavelkalinnikov, line 21: This corresponds to this repo: https://github.com/protocolbuffers/protobuf

The "go import" in the other file imports from a different repo: https://github.com/golang/protobuf.

I guess, we need to make sure they are compatible. I.e. we need to "go import" from a version not far off in time from protoc v3.11.4.

integration/cloudbuild/run_integration.sh
pavelkalinnikov: I think we should probably use one of releases: https://github.com/etcd-io/etcd/releases

integration/cloudbuild/testbase/Dockerfile
AlCutter: Hmm, I think the first repo gives us the `protoc` command, but the second is the golang specific bits (including `protoc-gen-go`).  You may be onto something with the golang/protobuf version but the `pb/protobuf` version is as it was in the [travis config](https://github.com/google/trillian/blob/master/.travis.yml#L40)

integration/cloudbuild/run_presubmit.sh
```diff
@@ -0,0 +1,20 @@
...
+trap "docker-compose -p ${HOSTNAME} -f ./integration/cloudbuild/docker-compose-mysql.yaml down" EXIT
+
+# Wait for mysql instance to be ready
+while ! mysql --protocol=TCP --host=${MYSQL_HOST} --port=${MYSQL_PORT} --user=root -pbananas -e quit ; do
+ sleep 5
+done
+
+# Presumbits need a user with CREATE DATABASE grants since they create temporary databases.
+# For the same reason, this is a URI prefix - tests will add DB names to the end.
+export TEST_MYSQL_URI="root:bananas@tcp(${MYSQL_HOST}:${MYSQL_PORT})/"
```
pavelkalinnikov, line 18: Remove the trailing slash consistently? So that "mysqlURI()+"/"+name" remains valid.

AlCutter:
Subsumed by #2296

PR #2286, opened 2021-01-21
quota: Move flag out of the package to main.go

### LINKED ISSUE #2026: --quota_system flag does not list registered quota systems
The --quota_system flag's description is of the form "Quota system to use. One of: []". Within the brackets, it is expected that the registered quota systems are listed. However, because the flag description is generated during package init, it can occur before the quota systems are registered, resulting in an unexpectedly empty list.

https://github.com/google/trillian/blob/bd4c7e90dad839982cac456cfcb78d83129b5925/server/quota.go#L37

pavelkalinnikov:
Move the flag so that it is always initialised after quota providers are imported.

```
$ go run main.go --help
...
  -quota_system string
        Quota system to use. One of: [noop etcd mysql] (default "mysql")
...
```

The `systems`->`providers` rename is for consistency with `storage.Providers()`.

Fixes #2026.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
     interested in and avoid the Trillian hasher registry+protobuf deps.
  * Moved some packages intended for internal-only use into `internal` packages:
    * InMemoryMerkleTree (indended to only be used by Trillian tests)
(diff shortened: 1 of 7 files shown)


pavelkalinnikov:
PTAL

AlCutter (approved):

PR #2288, opened 2021-01-21
etcd: Remove the unnecessary util

### LINKED ISSUE #2021: Consider deleting util/etcd/client.go
The documentation in this file says:
```
// TODO(pavelkalinnikov): Remove this when there is a way to compatibly import  
// the same version of etcd in external codebases. Could Go modules help? 
```
We are now using Go Modules, which suggests that we may now be able to delete the file. The documentation also refers to the "Trillian vendor/ directory" that no longer exists, which further suggests that the code may be obsolete.

pavelkalinnikov:
With Go modules this util became unnecessary.

Fixes #2021.


pphaneuf (approved):

Martin2112:
LGTM

Issue #1150, opened 2018-05-24
If WaitForMastership fails log_signer sits around doing nothing


rolandshoemaker:
If the call to [`WaitForMastership` fails in `electionRunner.Run`](https://github.com/google/trillian/blob/a58c1a0253e92d7496de46a0950bd9b1ce128339/server/log_operation_manager.go#L150-L153) it returns from the function and sits around doing nothing for that log (i.e. `masterFor` calls will always return empty and nothing can be done). This can be triggered by recoverable intermittent etcd issues.

The `return` should either be replaced with `continue` (perhaps with a backoff?) when hitting errors from `WaitforMastership` or the `electionRunner` should be removed from `l.electionRunner[logID]` when it exits so a new election goroutine can be created the next time `masterFor` is called. I'd probably go for the first of the two options.

pavelkalinnikov:
Hi Roland, thanks for filing this. I am working right now on factoring out the election runner, and also noticed this issue. I have a work-in-progress PR that largely interferes with this code. Would you like to review that?

rolandshoemaker:
Ah cool, I'll take a look, thanks!

Martin2112:
Did #1144 fix your issue?

pavelkalinnikov:
No, that was a refactoring, and it ended up postponed.

pavelkalinnikov:
This is effectively fixed now. Election code has migrated to [util/election](/util/election/) (types `Runner` and `MasterTracker`). If election fails at any step, it will be attempted again.

pavelkalinnikov:
No, actually the problem seems to be still there: https://github.com/google/trillian/blob/85d73fd1e11bcca6a0bd2f2c6ef8409d5289dcd3/log/operation_manager.go#L242

PR #2290, opened 2021-01-22
log: Clean up election in the operation loop

pavelkalinnikov:
Comments, variable renames, and factoring out some code.
Review commit by commit.

This is in preparation for fixing #1150.

### DIFF
--- a/log/operation_manager.go
+++ b/log/operation_manager.go
@@ -114,12 +114,15 @@ type OperationManager struct {
 	// logOperation is the task that gets run across active logs in the scheduling loop
 	logOperation Operation
 
+	// runnerWG groups all goroutines with election Runners.
+	runnerWG sync.WaitGroup
 	// runnerCancels contains cancel function for each logID election Runner.
-	runnerCancels       map[string]context.CancelFunc
+	runnerCancels map[string]context.CancelFunc
+	// pendingResignations delivers resignation requests from election Runners.
 	pendingResignations chan election.Resignation
-	runnerWG            sync.WaitGroup
-	tracker             *election.MasterTracker
-	lastHeld            []int64
+
+	tracker  *election.MasterTracker
+	lastHeld []int64
 	// Cache of logID => name; assumed not to change during runtime
 	logNamesMutex sync.Mutex
 	logNames      map[int64]string
--- a/log/operation_manager.go
+++ b/log/operation_manager.go
@@ -93,7 +93,7 @@ type OperationInfo struct {
 	// The following parameters govern the overall scheduling of Operations
 	// by a OperationManager.
 
-	// Election-related configuration.
+	// Election-related configuration. Copied for each log.
 	ElectionConfig election.RunnerConfig
 
 	// RunInterval is the time between starting batches of processing.  If a
--- a/log/operation_manager.go
+++ b/log/operation_manager.go
@@ -264,13 +264,13 @@ func (o *OperationManager) runElection(ctx context.Context, logID string) (conte
 		cancel()
 		return nil, fmt.Errorf("failed to create election for %v: %v", logID, err)
 	}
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Note: This is a cosmetic PR. It's convenient to review commit-by-commit.

log/operation_manager.go
```diff
@@ -225,22 +228,14 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
 	// Synchronize the set of log IDs with those we are tracking mastership for.
 	for _, logID := range allStringIDs {
 		knownLogs.Set(1, logID)
-		if o.electionRunner[logID] != nil {
+		if o.runnerCancels[logID] != nil {
```
Martin2112, line 231: Not sure whether this or `if _, ok := ` is preferred style, or maybe doesn't matter.

log/operation_manager.go
```diff
@@ -225,22 +228,14 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
...
-			return nil, fmt.Errorf("failed to create election for %v: %v", logID, err)
+			return nil, err
 		}
-		o.electionRunner[logID] = election.NewRunner(logID, &o.info.ElectionConfig, o.tracker, cancel, el)
-		o.runnerWG.Add(1)
-		go func(r *election.Runner) {
-			defer o.runnerWG.Done()
-			r.Run(innerCtx, o.pendingResignations)
-		}(o.electionRunner[logID])
+		o.runnerCancels[logID] = cancel
```
Martin2112, line 238: This also happens at line 267?

log/operation_manager.go
pavelkalinnikov: In this case it is equivalent.

log/operation_manager.go
pavelkalinnikov: Yes. Removed here.

log/operation_manager.go
```diff
@@ -261,6 +256,26 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
...
 
+func (o *OperationManager) runElection(ctx context.Context, logID string) (context.CancelFunc, error) {
+	glog.Infof("create master election goroutine for %v", logID)
+	cctx, cancel := context.WithCancel(ctx)
+	e, err := o.info.Registry.ElectionFactory.NewElection(cctx, logID)
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("failed to create election for %v: %v", logID, err)
+	}
+	// TODO(pavelkalinnikov): Passing the cancel function is not needed here.
```
pavelkalinnikov, line 267: Something went wrong during rebase, I'll fix it.

pavelkalinnikov:
All comments are addressed.

Martin2112 (approved):

pavelkalinnikov:
Interesting: https://travis-ci.org/github/google/trillian/jobs/755698264
`WARNING: DATA RACE`.
I'm taking a look at it.

pavelkalinnikov:
Got it: `NewRunner` takes `ElectionConfig` by pointer, and tries to modify it in `Run` method. The config is also used in other places. It probably shouldn't be used by pointer.

Martin2112:
Are the modifications in `NewRunner` necessary for it to work? It might not be entirely straightforward to fix.

Martin2112:
OK that sounds like it should work.

pavelkalinnikov:
There is an unrelated test failure `TestLogSuite/DequeueLeavesTwoBatches` for the in-memory CloudSpanner instance. It's flaky because the in-memory CloudSpanner doesn't actually implement transactions.

Martin2112:
OK.

pavelkalinnikov:
Added a commit with a minor comment to trigger builds again.

Issue #1640, opened 2019-05-28
Refactor logs operation


pavelkalinnikov:
`LogOperation` and its co-types are poorly designed:
- [ ] They are meant to be stateless, but in reality there is a cache in `SequencerManager`. We should rather make an explicit per-tree cache that can be extended (e.g. with compact ranges cached between sequencing runs as in #1598).
- [ ] Interfaces and method signatures are redundant. For example, `log.NewSequencer` takes a `Signer` created from a `Tree`, but then the same `Tree` is passed into `sequencer.IntegrateBatch` when in fact there is only one `Tree` that can be accepted.
- [ ] TODO: Keep listing changes.

pavelkalinnikov:
The current design has 3 types (`OperationManager`, `SequencerManager` and `Sequencer`), which all have their own caches for individual tree IDs.

I propose to shift "caching" one level up the stack, so that `OperationManager` merely does couple of things: tracks log IDs and the corresponding masterships, and for each active log ID has an `Operation` object. `Operation` object encapsulates all "cached" items that were previously scattered across different types (log names, signers, compact ranges, etc), and contains the code for handling only one log (effectively some blend of former `SequencerManager` and `Sequencer`).

pavelkalinnikov:
@Martin2112 @AlCutter WDYT?

Martin2112:
Yes the structure could be improved. If you're going to cache tree related things then it must be done in a way that's completely safe. You might want to make that a separate work item.

pavelkalinnikov:
Yep. I will probably start from just restructuring the code with equivalent safety guarantees. For example, consider `SequencerManager` and its Signers cache: it is never cleared, and its items are never updated. Same with `OperationManager` and log names.

Once it's restructured, we could improve guarantees. E.g. if an `Operation` fails, we can delete it altogether, so on the next run time it will be re-created with fresh signer, log name, etc.

PR #2291, opened 2021-01-22
log: Restart election Runner in case of an error

### LINKED ISSUE #1150: If WaitForMastership fails log_signer sits around doing nothing

pavelkalinnikov:
This change makes the election/resignation loop restart indefinitely (with a reasonable
delay) in case it fails temporarily due to etcd issues.

Fixes #1150.

### Checklist

### DIFF
--- a/log/operation_manager.go
+++ b/log/operation_manager.go
@@ -228,14 +228,9 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
 	// Synchronize the set of log IDs with those we are tracking mastership for.
 	for _, logID := range allStringIDs {
 		knownLogs.Set(1, logID)
-		if o.runnerCancels[logID] != nil {
-			continue
+		if o.runnerCancels[logID] == nil {
+			o.runnerCancels[logID] = o.runElectionWithRestarts(ctx, logID)
 		}
-		cancel, err := o.runElection(ctx, logID)
-		if err != nil {
-			return nil, err
-		}
-		o.runnerCancels[logID] = cancel
 	}
 
 	held := o.tracker.Held()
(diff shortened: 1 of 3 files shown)


log/operation_manager.go
```diff
@@ -256,25 +251,42 @@ func (o *OperationManager) masterFor(ctx context.Context, allIDs []int64) ([]int
...
-	e, err := o.info.Registry.ElectionFactory.NewElection(cctx, logID)
-	if err != nil {
-		cancel()
-		return nil, fmt.Errorf("failed to create election for %v: %v", logID, err)
-	}
-	o.runnerWG.Add(1)
-	go func() {
-		defer o.runnerWG.Done()
+	run := func(ctx context.Context) {
+		e, err := o.info.Registry.ElectionFactory.NewElection(ctx, logID)
```
Martin2112, line 261: Previously if the election was fundamentally broken these goroutines would return an error, now they'll loop.

We might need to look at adding metrics for election restart rate.

Martin2112 (approved):

log/operation_manager.go
pavelkalinnikov: Agree.

I mitigated this in Migrillian by having a metric of restarts as you suggest, but instead of restarting only the election loop, I restarted the whole "operation" for this log. I think it will be a better granularity.

This will be easy to do if we restructure log operation as suggested in #1640, so that all the logs are nicely isolated (and only share some semaphores limiting the number of simultaneously running operations).

I'll add a TODO here.

Issue #2298, opened 2021-01-27
Migrate CI from Travis to Cloud Build


pavelkalinnikov:
Travis [migrates](https://docs.travis-ci.com/user/migrate/open-source-repository-migration) to travis-ci.com from travis-ci.org. The migration requires new repository permissions that go against our security policies.

As a result, we will be migrating to Google Cloud Build. The plan is to reproduce all the build steps from the [.travis.yml](https://github.com/google/trillian/blob/593325ffaeb18c18ee8eb714dbb9eeb137713996/.travis.yml) config.

pavelkalinnikov:
So far I have seen 2 flakes.

The first is this (which means someone is running the executable while someone else writes to it):
```
2021-01-27 14:15:05.698 GMTStep #9 - "integration": Starting Map RPC server on localhost:6962, HTTP on localhost:6963
2021-01-27 14:15:05.698 GMTStep #9 - "integration": /workspace/integration/functions.sh: line 307: ./trillian_map_server: Text file busy
2021-01-27 14:15:16.711 GMTStep #9 - "integration": Failed to get response on localhost:6962
```

The other one is map integration tests failing in "batch" storage mode. Possibly related to #1845.

pavelkalinnikov:
Some Travis builders are still missing in Cloud Build:

- [ ] Codecov
- [ ] go mod tidy + check that the diff is empty
- [ ] go generate + check that the diff is empty

pavelkalinnikov:
There are more flows to extend to a full functionality:

- [ ] Committing to master.
- [ ] Creating a release.

pavelkalinnikov:
Codecov is now working, but it needs to be activated on commits to `master` as well, in order to correctly find the base commit when generating a report for a PR.

pavelkalinnikov:
We should have a way to GC the numerous Docker images generated by each PR/commit.
- [ ] Done.

pavelkalinnikov:
Tests in `storage/cloudspanner` are flaky.

pavelkalinnikov:
- [ ] MariaDB 10.3 tests are still missing.

pavelkalinnikov:
This can be considered done.

PR #2301, opened 2021-01-27
Upgrade etcd to v3.4.12

pavelkalinnikov:
Integration tests fail with the following errors:

1. `panic: invalid Go type int for field
   go_etcd_io.etcd.etcdserver.etcdserverpb.loggablePutRequest.value_size`
2. `panic: field etcdserverpb.Request.ID has invalid type: got uint64,
   want pointer`

The first one is [fixed](https://github.com/etcd-io/etcd/blob/master/CHANGELOG-3.4.md) in etcd @ `v3.4.10`, the second - in `v3.4.12`.

Currently we are pinned to `v0.0.0-20200513171258-e048e166ab9c`, which is
in between `v3.4.8` and `v3.4.9`. This change updates the pin to `v3.4.12`, to
include both fixes.

Part of #2298

### Checklist


Martin2112 (approved):
The new version includes "alpha" but I suspect it's not any less stable than what we've been using.

AlCutter (approved):

pavelkalinnikov:
The "alpha" bit is irrelevant, it's because of the way etcd migrated to Go modules. I ran `go get go.etcd.io/etcd@17cef6e3e9d5`, and Go tooling automatically updated it to this "alpha", but you can [see](https://github.com/etcd-io/etcd/tree/v3.4.12) that this commit hash corresponds to v3.4.12.

Martin2112:
OK.

PR #2303, opened 2021-01-28
Fix MySQL URI env definition, clarify comments

pavelkalinnikov:


storage/testdb/testdb.go
```diff
@@ -114,7 +115,7 @@ func newEmptyDB(ctx context.Context) (*sql.DB, func(context.Context), error) {
 	}
 
 	db.Close()
-	db, err = sql.Open("mysql", mysqlURI()+"/"+name)
+	db, err = sql.Open("mysql", mysqlURI()+name)
```
NatalieDoduc, line 118: how about using https://golang.org/pkg/path/filepath/#Join?

storage/testdb/testdb.go
```diff
@@ -39,21 +39,22 @@ const (
...
 
 var (
 	trillianSQL = testonly.RelativeToPackage("../mysql/schema/storage.sql")
 )
 
-// mysqlURI returns the connection URI to use for tests.
-// It will prefer the value in the MYSQL_URI env var, and if empty/unset, it'll
-// return the defaultTestMySQLURI value.
+// mysqlURI returns the MySQL connection URI to use for tests. It returns the
+// value in the TEST_MYSQL_URI env var, if not empty, otherwise falls back to
```
NatalieDoduc, line 51: How about mentioning the env var as defined by `MySQLURIEnv`? If the assignment on L40 changes, the mention of `TEST_MYSQL_URI` no longer holds.

storage/testdb/testdb.go
```diff
@@ -39,21 +39,22 @@ const (
 	// instance URI to use.
 	MySQLURIEnv = "TEST_MYSQL_URI"
 
-	defaultTestMySQLURI = "root@tcp(127.0.0.1)"
+	// Note: sql.Open requires the URI to end with a slash.
```
NatalieDoduc, line 42: Is this true? looks like its rather that the way you use this value within a construction of a path that is then passed to sql.Open that requires a `/`? (I'm referring to L118 where I don't think mysqlURI()+name ends with a `/`, but there may be other uses.)

NatalieDoduc (commented):
I really like the new comments, much clearer.

Does this file have tests?

storage/testdb/testdb.go
pavelkalinnikov: This is true. At the moment some integration tests fail because of this bug (but CI is temporarily unavailable, so we don't see it).

See also line 67, where `mysqlURI()` is passed to `sql.Open` as is. This is the line where currently the fail happens.

storage/testdb/testdb.go
pavelkalinnikov: Excellent suggestion, done.

storage/testdb/testdb.go
pavelkalinnikov: Seems good, but I don't know how compatible the "file paths" are with MySQL URIs. Leaving as is for now.

AlCutter (approved):

storage/testdb/testdb.go
```diff
@@ -35,25 +35,26 @@ import (
...
 //
-// Only a subset of the suite of tests in this repo require a database it's
-// not possible to blanket apply the test_mysql_uri flag, and maintaining a separate
-// list of tests which needs DB connectivity is brittle, so this fallback approach
-// means we can provide the connection info in a more ambient fashion.
+// We use an ENV variable, rather than a flag, for flexibility. Only a subset
+// of the tests in this repo require a database and import this package. With a
+// flag, it would be necessary to distinguish "go test" invocations that need a
+// database, and those that don't. ENV allows to "blanket apply" this setting.
 func mysqlURI() string {
```
NatalieDoduc, line 58: as per the other comments in the file - perhaps mysqlURI() should check and guarantee that what it returns ends with a `/` so that all sql.Open(mysqlURI()) calls are covered?

storage/testdb/testdb.go
NatalieDoduc: ok, maybe add a TODO() for checking usage of `filepath.Join`?

and here, you may want to add a trailing `/` after the `name` as well, if `sql.Open` requires it.

pavelkalinnikov: `sql.Open` requires it only for when the database name is not included into the URI (see line 104, for instance).

storage/testdb/testdb.go
pavelkalinnikov: If it doesn't end with a slash, `sql.Open` will return an error, the tests will fail, and the caller will add a slash. IMO, there is no much need in trying to prevent this. It would be a problem if `sql.Open` didn't fail in this case, but silently do some evil.

PR #2304, opened 2021-01-28
Make CB go faster

AlCutter:
Use a shared GOPATH volume to cache module downloads between steps.
Results in ~2m30 speedup.

#2298

### Checklist

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -16,6 +16,10 @@ substitutions:
   _MYSQL_PASSWORD: ""
 options:
   machineType: E2_HIGHCPU_32
+  volumes:
+  # A shared volume for caching Go modules between steps.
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -16,6 +16,9 @@ substitutions:
   _MYSQL_PASSWORD: ""
 options:
   machineType: E2_HIGHCPU_32
+  volumes:
+  - name: go-modules
```
pavelkalinnikov, line 20: Add comment?
`# A shared volume for caching Go modules between steps.`

pavelkalinnikov (approved):

cloudbuild.yaml
AlCutter: done

PR #2308, opened 2021-01-29
Fix the missing cached image, add comments

pavelkalinnikov:
This change reduces CI running time from 9 to 7 minutes.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -31,6 +31,8 @@ options:
 # Cache the testbase image in Container Regisrty, to be reused by subsequent
 # builds. The technique is described here:
 # https://cloud.google.com/cloud-build/docs/speeding-up-builds#using_a_cached_docker_image
+#
+# TODO(pavelkalinnikov): Consider pushing this image only on commits to master.
 images: ['gcr.io/$PROJECT_ID/trillian_testbase:latest']
 
 steps:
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -28,11 +28,18 @@ options:
     - DOCKER_CLIENT_TIMEOUT=120
     - COMPOSE_HTTP_TIMEOUT=120
 
+# Cache the testbase image in Container Regisrty, to be reused by subsequent
+# builds. The technique is described here:
+# https://cloud.google.com/cloud-build/docs/speeding-up-builds#using_a_cached_docker_image
+images: ['gcr.io/$PROJECT_ID/trillian_testbase:latest']
```
AlCutter, line 34: Is there any potential for unfortunate interactions if there are multiple PRs being worked on at the same time?
I wasn't sure if this sort of thing was a problem or not, so my original plan was to just push the images on the merge-to-master trigger.

pavelkalinnikov: I see. I'm not sure how "thread-safe" this is. I think, as long as one push wins another, it should be fine. This is just for caching.

pavelkalinnikov: Pushing only when updating master seems fine as well.

pavelkalinnikov: How about I add a TODO here?

AlCutter: oke doke, sgtm!

AlCutter (approved):

PR #2311, opened 2021-01-30
cloudbuild: Add Codecov reporting

pavelkalinnikov:
This change adds a step to Cloud Build config, which generates tests
coverage reports and uploads to codecov.io. The step has a secret token
`_CODECOV_TOKEN` which should be configured via substitution variables.

Part of #2298

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -14,6 +14,7 @@ substitutions:
   _MYSQL_TAG: "5.7"
   _MYSQL_ROOT_PASSWORD: ""
   _MYSQL_PASSWORD: ""
+  _CODECOV_TOKEN: "" # The auth token for uploading coverage to Codecov.
 options:
   machineType: E2_HIGHCPU_32
   volumes:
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -14,6 +14,7 @@ substitutions:
   _MYSQL_TAG: "5.7"
   _MYSQL_ROOT_PASSWORD: ""
   _MYSQL_PASSWORD: ""
+  _CODECOV_TOKEN: "" # The auth token for uploading coverage to Codecov.
```
AlCutter, line 17: You probably know this already, but the `sekret` value can be added in the cloudbuild trigger configuration in the GCP console.

pavelkalinnikov: Yep, that's what I did.

pavelkalinnikov:
The codecov error is because it can't find a preceding report. It will go away after merging.

pphaneuf (approved):

PR #2318, opened 2021-02-03
run `go generate`

pphaneuf:


pavelkalinnikov:
The CI setup uses:
- protoc-gen-go [v1.25.0](https://github.com/google/trillian/blob/f44024791677ffa9d8be23a3bcc495d4d80e6ad4/go.mod#L63)
- protoc [v3.11.4](https://github.com/google/trillian/blob/f44024791677ffa9d8be23a3bcc495d4d80e6ad4/integration/cloudbuild/testbase/Dockerfile#L29)

We should either upgrade CI, or downgrade the generated files?

One of the action items in #2298 is to do a CI check that the `go generate`-d files are of the correct version.

pphaneuf:
Is there some recommended trick to make sure that I've got the correct version? I'm not really *intending* to update things, but I want to make other changes that will need re-generating those files, and I want to make sure that the diff is small and relevant.

pavelkalinnikov (approved):

PR #2310, opened 2021-01-30
Upgrade golangci-lint

pavelkalinnikov:
This change upgrades the linter suite, and fixes the corresponding errors.

The tool in CI is too outdated, and doesn't work:

```
golangci-lint run --deadline=8m
Step #3 - "lint": level=warning msg="[runner] Can't run linter unused: <lots of errors> (unknown iexport format version 1), export data is newer version - update tool))]"
```

### Checklist

- [x] I have updated the [CHANGELOG](CHANGELOG.md).
  - Adjust the draft version number according to [semantic versioning](https://semver.org/) rules.
- [ ] I have updated [documentation](docs/) accordingly (including the [feature implementation matrix](docs/Feature_Implementation_Matrix.md)).

PR #2320, opened 2021-02-03
Stop using the deprecated hasher symbols

pavelkalinnikov:
This change removes direct referencing the symbols that were moved
and deprecated in #2278.

Factored out from #2310.


pavelkalinnikov:
Might be more convenient to review commit-by-commit.

Martin2112 (approved):

PR #2323, opened 2021-02-04
Run presubmit.sh with more linter checks

pavelkalinnikov:
This change replaces the inline script in the "linter" Cloud Build step with the pre-existing
`scripts/presubmit.sh`. The replacement script does more useful checks.

Part of #2298


pavelkalinnikov:
PTAL. Might be convenient commit-by-commit.

pphaneuf (approved):

PR #2325, opened 2021-02-04
Don't run time-sensitive tests in parallel

pavelkalinnikov:
Many parallel goroutines can possibly cause delayed time events. Because of this,
some tests relying on physical time, are flaky in CI.


Martin2112 (approved):
Yep, the intention was good but safer without it.

PR #2327, opened 2021-02-04
client: Fix flaky WaitForInclusion test

pavelkalinnikov:
Sometimes the test fails with this error:
```
WaitForInclusion before sequencing: rpc error: code = Unknown desc = context deadline exceeded, want: not-nil
```
in the place where it expects `status.DeadlineExceeded`.

This is because not all Trillian gRPC methods return errors wrapped into
status. This change adds a workaround to unflake the test.

### DIFF
--- a/client/log_client_test.go
+++ b/client/log_client_test.go
@@ -18,9 +18,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto" //nolint:staticcheck
 	"github.com/google/trillian"
 	rfc6962 "github.com/google/trillian/merkle/rfc6962/hasher"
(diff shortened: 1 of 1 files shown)


client/log_client_test.go
```diff
@@ -227,6 +229,23 @@ func TestWaitForInclusion(t *testing.T) {
...
 }
 
+func isDeadlineExceeded(err error) bool {
+	if err == nil {
+		return false
+	}
+	if status.Code(err) == codes.DeadlineExceeded {
+		return true
+	}
+	// TODO(pavelkalinnikov): Sometimes Trillian gRPC returns codes.Unknown for a
```
Martin2112, line 239: Is there an issue to fix this? Can you create one if not.

Martin2112 (approved):

client/log_client_test.go
pavelkalinnikov: Done: #2328

PR #2330, opened 2021-02-05
use QueueLeaf instead of QueueLeaves in integration test

pphaneuf:
QueueLeaves isn't used, as it has problems with number of leaves greater than 1.


mhutchinson (approved):
This is the only place this method is invoked, so LGTM to change to the standard one.

pavelkalinnikov:
This might help resolving the second flake in https://github.com/google/trillian/issues/2298#issuecomment-768325606. The thing causing it was a race between `QueueLeaves` and sequencer transactions.

PR #2334, opened 2021-02-05
Run CI tests against multiple versions of golang

pgporada:
### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 os: linux
 dist: xenial
 language: go
-go: "1.14.x"
+go:
+  - "1.14.x"
+  - "1.15.x"
 go_import_path: github.com/google/trillian
 
 cache:


pavelkalinnikov:
@pgporada We no longer run CI on Travis: #2298.

pavelkalinnikov:
Are you using Go 1.14? Maybe we could just bump the version in CI. Our README anyway says "To build and test Trillian you need: Go 1.14 or later", so these should be consistent.

pgporada:
We're currently using Go 1.14.x to build our copies of trillian and ctgo. I had wanted to rebuild our containers on Go 1.15.x and thought that having CI run tests parallelly would be good for all.

I now see that cloudbuild is going to use golang 1.15.7 per https://github.com/google/trillian/blob/master/cloudbuild.yaml#L50-L53 and https://github.com/google/trillian/blob/master/integration/cloudbuild/testbase/Dockerfile#L23.

Sorry for the time waste.

pavelkalinnikov:
No problem at all. Yes, I think you are pretty safe to try Go 1.15. We should probably indicate it consistently across README/Dockerfiles/etc. The CI configuration is a bit messy now, because we had to do the migration quickly.

PR #2340, opened 2021-02-08
client: Fix flaky WaitForInclusion test attempt 2

pavelkalinnikov:
Sometimes the test fails with this error:

```
WaitForInclusion before sequencing: rpc error: code = Unknown desc =
sql: transaction has already been committed or rolled back, want:
not-nil
```

This is because MySQL storage layer can return a different error when the
context is canceled (ErrTxDone: https://golang.org/pkg/database/sql/#Tx).

For the purposes of the test it is sufficient to check that *some* error
is returned rather than a concrete cancelation error.

### DIFF
--- a/client/log_client_test.go
+++ b/client/log_client_test.go
@@ -18,18 +18,14 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto" //nolint:staticcheck
 	"github.com/google/trillian"
 	rfc6962 "github.com/google/trillian/merkle/rfc6962/hasher"
 	"github.com/google/trillian/testonly/integration"
 	"github.com/google/trillian/types"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/google/trillian/storage/testdb"
 	stestonly "github.com/google/trillian/storage/testonly"
@@ -211,8 +207,8 @@ func TestWaitForInclusion(t *testing.T) {
 
 			if !test.skipPreCheck {
 				cctx, cancel := context.WithTimeout(ctx, 50*time.Millisecond)
-				if err := client.WaitForInclusion(cctx, test.leaf); !isDeadlineExceeded(err) {
-					t.Errorf("WaitForInclusion before sequencing: %v, want: not-nil", err)
+				if err := client.WaitForInclusion(cctx, test.leaf); err == nil {
+					t.Error("WaitForInclusion before sequencing succeeded, want error")
 				}
 				cancel()
 			}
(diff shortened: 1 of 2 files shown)


pphaneuf:
Shouldn't that be fixed in the MySQL storage instead? Those kind of highly MySQL-specific errors "escaping" from the storage implementation seem like a bug waiting to happen...

pavelkalinnikov:
Investigating why MySQL returns this double-Close error can be an asynchronous TODO. I tried to reproduce it, but so far didn't succeed.
My suspicion though is that when context is canceled during a MySQL query, it implicitly rolls back the transaction. The Trillian storage, however, assumes that it controls all the rollback/commit paths, so it also tries to Close the transaction one last time.

For the purposes of this test/API, I think it's incorrect to assume that Context cancel should result in a very specific error returned, because there are many layers underneath this call, and each layer may react differently. It's impractical to force all layers and code paths to return the same error.

Martin2112:
This code was originally written when there was no context passed to the
SQL layer so it would have been a reasonable assumption that our code
controlled all the paths. There was no way to get into this situation. I
think this feature got added in Go 1.8.

Martin

pavelkalinnikov:
I managed to reproduce the problem. Indeed, the issue is that the transaction lifetime also depends on the context: https://pkg.go.dev/database/sql#Conn.BeginTx.

pphaneuf:
This here seems pretty clear:

https://golang.org/pkg/context/#Canceled

pavelkalinnikov:
@pphaneuf Not sure what you mean. What is your suggestion?

pavelkalinnikov:
Basically, right now any operation of the MySQL layer [can return](https://golang.org/pkg/database/sql/#Tx) `database/sql.ErrTxDone` after an explicit or implicit (via context) transaction rollback or commit. In order to fix it properly, we need to instrument all MySQL storage implementation methods with an `if err == sql.ErrTxDone { return <non-MySQL error> }`. This is out of scope of this PR, but I can add a TODO somewhere.

pphaneuf:
I'm alright with adding a TODO, although not entirely sure where... The bug is in the MySQL storage, breaking the API contract of the storage interface, and leaking it all the way back to the client library!

Anywhere would do, your call.

pavelkalinnikov:
@pphaneuf @Martin2112 Added TODO. PTAL.

Martin2112:
Not entirely our fault that the Go API contract changed under us.

Martin2112 (approved):
OK, let's consider the other implications separately.

pphaneuf:
No, not saying that its our fault, that's life...

But just making sure we're clear here: this test is correctly reporting that there is an (unfortunately intermittent) legitimate problem, and we're effectively disabling it until we have a proper fix for it.

pphaneuf (approved):

pavelkalinnikov:
Correct, we are disabling a thing that reminds there is a problem. I will add this to an open issue, to not lose this reminder.

This test wasn't the right place to observe this error: it expects that the whole stack of client-grpcClient-grpcServer-mysqlStorage-MySQL will return a specific error for the specific case of a canceled context. It's too strong of a requirement.

PR #2343, opened 2021-02-08
cloudbuild: Slightly improve the configs

pavelkalinnikov:


pavelkalinnikov:
@AlCutter Review commit-by-commit?

cloudbuild.yaml
```diff
@@ -1,26 +1,19 @@
...
-  - name: go-modules
+  # A shared volume for caching Go modules, binaries and builds between steps.
+  - name: go-workspace
     path: /go
   env:
     - GO111MODULE=on
-    - GOPROXY=https://proxy.golang.org
-    - PROJECT_ROOT=github.com/google/trillian
     - GOPATH=/go
+    - GOCACHE=/go/.cache/go-build
```
pavelkalinnikov, line 16: Not sure this should be added. It doesn't seem to bring a big speed-up. But there is a risk that it caches some succeeded tests, and prevent running them in another step. The doc says it shouldn't happen if build flags differ, but it doesn't cover all cases (e.g. the `cgo` changes).

pavelkalinnikov: Removed.

pphaneuf (approved):

PR #2344, opened 2021-02-08
Enable docker-compose test

pavelkalinnikov:
Part of #2298.

### DIFF
--- a/examples/deployment/docker-compose.yml
+++ b/examples/deployment/docker-compose.yml
@@ -68,6 +68,9 @@ services:
     depends_on:
       - mysql
 
+# TODO(pavelkalinnikov): Find a way to remove this network declaration. It is
+# needed for Cloud Build integration tests, but makes local deployment less
+# convenient.
 networks:
   default:
     external:
--- a/examples/deployment/docker-compose.yml
+++ b/examples/deployment/docker-compose.yml
@@ -68,3 +68,7 @@ services:
     depends_on:
       - mysql
 
+networks:
+  default:
+    external:
+      name: cloudbuild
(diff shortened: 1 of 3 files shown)


examples/deployment/docker-compose.yml
```diff
@@ -68,3 +68,10 @@ services:
     depends_on:
       - mysql
 
+# TODO(pavelkalinnikov): Find a way to remove this network declaration. It is
```
AlCutter, line 71: I wonder if putting the network specific bits in a separate `yml` file and using something like below might work:
```
docker-compose -d examples/deployment/docker-compose.yml -f integration/cloudbuild/network.override.yml up
```
?

AlCutter (approved):

examples/deployment/docker-compose.yml
pavelkalinnikov: What is the "-d" flag? Can't find it.

pavelkalinnikov: Oh wait, that's the detach. But I still don't get how the 2 files a fed to the docker-compose.

pavelkalinnikov: It seems like you can't combine 2 docker-compose files like that? The "free" parameters are for the service names to start from the the config file, not other docker-compose files.

pavelkalinnikov: Ah, but you can supply multiple `-f` parameters apparently: https://docs.docker.com/compose/extends/

AlCutter: Sorry, typo - that was meant to be `-f` twice >_<

pavelkalinnikov: Ok, I'll merge this PR for now and try to play with this suggestion.

pavelkalinnikov: #2350.

PR #2348, opened 2021-02-10
Travis no more

pavelkalinnikov:
This change removes remnants of Travis in the repository. All the build steps now
have equivalents in Cloud Build.

Part of #2298.

### Checklist


integration/run_integration_test.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/bash
 #
-# run_integration_test.sh is a wrapper around the log_prep and run_test functions.
-# It's intended to be used by the Travis CT/Trillian integration test only,
-# and will go away when we migrate off of Travis.
+# run_integration_test.sh is a wrapper around the log_prep and run_test
+# functions. It's intended to be used by the Travis CT/Trillian integration test
+# only, and will go away when we migrate off of Travis.
+#
+# TODO(al): Make the script go away.
```
pavelkalinnikov, line 7: @AlCutter Based on "blame" for this file, you wrote this. I am not sure how this file is used.

Martin2112: I think it was a workaround for something. Al will know. It's referenced from travis.yaml so you might be removing the only usage.

docs/CloudSpanner.md
```diff
@@ -46,6 +46,8 @@ and running on Google Cloud.
 Setting up continuous integration with Travis
 =============================================
 
+(Note: We no longer use Travis)
```
Martin2112, line 49: In this file we could remove mentions of Travis and update the text in step 8 with something more generic that won't get out of date. Most CI environments will need the service key.

Martin2112 (approved):

integration/run_integration_test.sh
pavelkalinnikov: Ah, I think there was some test using this: It cloned CT repo and ran this script from within CT.

We don't have such test in Cloud Build. Should we replicate it? It seems that such a thing should be tested in CT repo rather than Trillian?

docs/CloudSpanner.md
pavelkalinnikov: Made it slightly more generic.

Issue #2284, opened 2021-01-21
Tombstone and remove current map server & storage code


AlCutter:
The newer approach is already present in the experimental directory, and early examples of how to use it is in the [experimental sumdb](https://github.com/google/trillian-examples/tree/master/experimental/batchmap/sumdb) project in trillian-examples.

pavelkalinnikov:
KT depends on Maps as currently implemented. We would need to pin it to the latest release.

pphaneuf:
Is KT in use? I think their repository is currently broken anyway?

mhutchinson:
I propose that we create a Trillian release in the next week or so that still includes the map in its current state. The next release of Trillian will cut the map out. This seems like a reasonable compromise so that anyone still wanting to play with the old map can pin to this release, and it prevents the release being blocked on PRs which are deleting the map (which I imagine will take a while to see through).

pphaneuf:
SGTM

pavelkalinnikov:
Also, maybe we can time this with "finalising" v1 branch, and setting foundation for v2:

- Create a "permanent" v1 branch, with maps still in it. The last v1 release will be tagged on this v1 branch.
- Commit to v1 only to fix issues and support prod (including our internal).
- Start breaking things in master. Delete maps, minimise the rest of the code.
- Whenever ready, create v2 branch/label/release from master.

pavelkalinnikov:
Deleting maps to some non-breaking degree could still be done in v1.

pphaneuf:
I missed a detail previously: the KT repo (which I believe has been broken for other reasons for a good while) is pinning two Trillian versions behind.

So there's nothing we can do that will break KT any further, and this issue is essentially done?

pavelkalinnikov:
Yeah, the KT aspect is no longer a factor/blocker here, I think.

The only remaining factor is that the last release was a few months ago, so we might want: a) fresh "final" map release, b) declaring in release notes that Map is being removed in the next release.

pphaneuf:
If the "final" map-inclusive release is for the benefit of map users, my hypothesis is that there are none, and I would think that having it "one last time" can only encourage new potential users to appear (and soon be disappointed), even with an announcement?

pphaneuf:
The backing for my hypothesis is that the map API as it is at the moment is so incredibly difficult to use "correctly" that I would consider it broken, and code using it likely to be broken as well (so removing it breaks nothing new).

mhutchinson:
[...]

PR #2361, opened 2021-02-18
Removed some map stragglers from schemas and protos

mhutchinson:
This is a continuation of #2284


Martin2112:
I can't see why the build failed because of permissions. Was that legit?

pphaneuf:
This will need some tweaks to `storage/mysql/log_storage_test.go`, I believe (and maybe elsewhere?).

mhutchinson:
The tests are passing now.

Martin2112 (approved):

pphaneuf (approved):

Issue #2364, opened 2021-02-23
Tests / presubmit fails under go 1.16


mhutchinson:
`go test ./...` fails under go 1.16, but works under go 1.15. The two tests that are failing are:

`github.com/google/trillian/cmd/updatetree`
`github.com/google/trillian/cmd/createtree`

Both fail with `panic: unexpected call to os.Exit(0) during test`, which looks to be a direct consequence of https://golang.org/doc/go1.16#go-test.

mhutchinson:
For the time being (#2365) this is hacked out by making the presubmit scripts flag off the panic on Exit(0).

Now I'm starting to look into the root cause over at https://github.com/mhutchinson/trillian/tree/debug. I've done a hack to get a stacktrace on the panic, and it appears at the end of the test run. This rules out it being something deeper in the actual code which is exiting early. At first it seemed that my addition of the `TestMain` was the problem, but I've added an identical `TestMain` to a test which passes and even though `TestMain` calls `os.Exit`, this one works. My hunch now is that the tests are not being seen as being "finished" properly. Looking into the [change that introduced this behaviour in go1.16](https://github.git.corp.google.com/golang/go/+/4f76fe86756841befb6574ce4bf04113d14389d4), and more specifically into [testing.go#after](https://github.git.corp.google.com/golang/go/+/4f76fe86756841befb6574ce4bf04113d14389d4%5E%21/#F6) then we see:

```
 // after runs after all testing.
 func (m *M) after() {
 	m.afterOnce.Do(func() {
+		if *panicOnExit0 {
+			m.deps.SetPanicOnExit0(false)
+		}
 		m.writeProfiles()
 	})
 }
```

This mechanism looks like it is intended to clear the panic condition so that the generated test's `main` can `Exit` when it's finished without triggering the check.

My guess is that this `after` call is not being made in the failing tests for some reason. I suspect some teardown / timeout is pending. The options available in my "choose your own adventure" here seem to be:
 1. Look at what mechanism makes `after` be invoked
 2. Look through the failing tests for anything suspicious / hack in a timeout in `TestMain`

mhutchinson:
Haha, priceless. OK, so these tests both have the following called in one of their test cases:

```
// resetFlags sets all flags to their default values.
func resetFlags() {
	flag.Visit(func(f *flag.Flag) {
		f.Value.Set(f.DefValue)
	})
}
```

[...]

PR #2365, opened 2021-02-23
Temporary fix for go1.16 tests

mhutchinson:
This is a workaround to make presubmit work while we work out and address the root cause for #2364.


Martin2112 (approved):

Martin2112:
Breaks on 1.15 so the easy option doesn't work.

Martin2112:
Looks promising.

PR #2366, opened 2021-02-23
Fixed some lint errors that come up locally for me

mhutchinson:
* Unused field
* Declaration/assignment merging
* Duplicate nil check

### DIFF
--- a/quota/etcd/quota_provider.go
+++ b/quota/etcd/quota_provider.go
@@ -58,8 +58,7 @@ func newEtcdQuotaManager() (quota.Manager, error) {
 		return nil, fmt.Errorf("failed to connect to etcd at %v: %v", *Servers, err)
 	}
 
-	var qm quota.Manager
-	qm = etcdqm.New(client)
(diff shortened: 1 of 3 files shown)


quota/etcd/quota_provider.go
```diff
@@ -58,8 +58,7 @@ func newEtcdQuotaManager() (quota.Manager, error) {
 		return nil, fmt.Errorf("failed to connect to etcd at %v: %v", *Servers, err)
 	}
 
-	var qm quota.Manager
-	qm = etcdqm.New(client)
+	var qm quota.Manager = etcdqm.New(client)
```
Martin2112, line 61: I assume we're deliberately dropping the specific type here. Is that necessary?

Martin2112 (approved):

quota/etcd/quota_provider.go
mhutchinson: It needs to be this way to support the `qm = cachedQM` line below.

PR #2369, opened 2021-02-24
Reset flags in a cleaner way

### LINKED ISSUE #2364: Tests / presubmit fails under go 1.16

mhutchinson:
This avoids fiddling with the flag used by go's test framework, which fixes #2364 and means that the hack from #2365 can be removed.

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -102,14 +102,13 @@ main() {
     go build ./...
 
     export TEST_FLAGS="-timeout=${GO_TEST_TIMEOUT:-5m}"
-    export TEST_FLAGS_SUFFIX="-alsologtostderr"
 
     if [[ ${coverage} -eq 1 ]]; then
       TEST_FLAGS+=" -covermode=atomic -coverprofile=coverage.txt"
     fi
 
     echo "running go test ${TEST_FLAGS} ./..."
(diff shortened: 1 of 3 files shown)


scripts/presubmit.sh
```diff
@@ -110,11 +108,6 @@ main() {
       TEST_FLAGS+=" -covermode=atomic -coverprofile=coverage.txt"
     fi
 
-    if [[ ${GOVER} -ge 16 ]]; then
-      # https://github.com/google/trillian/issues/2364
-      TEST_FLAGS_SUFFIX+=" -test.paniconexit0=false"
-    fi
-
     echo "running go test ${TEST_FLAGS} ./..."
     go test ${TEST_FLAGS} ./... ${TEST_FLAGS_SUFFIX}
```
pavelkalinnikov, line 112: Consider rolling this back to pre-#2365. The `TEST_FLAGS_SUFFIX` equals `-alsologtostderr` unconditionally.

pavelkalinnikov (approved):

PR #2375, opened 2021-03-01
Remove HStar2

pavelkalinnikov:
This change removes `HStar2` algorithm in favour of `HStar3`. Both were used only in the
`batchmap` experimental code (in the builder and verifier components), so this change also
makes both components use the same algorithm.

#2284


pavelkalinnikov:
This PR depends on #2370. The build failure will go away after that one.

mhutchinson (approved):

PR #2381, opened 2021-03-03
Make better use of etcd services & endpoints

mhutchinson:
This affects the usage of `trillian_log_server` and `trillian_log_signer`.

Previously the log server had two services, each with a single endpoint (for http and rpc). etcd now supports (nay, forces) that service endpoints can't be announced raw and need to be in a service namespace. Thus we now announce endpoints at `trillian-logserver/http` and `trillian-logserver/rpc`. The `--etcd_http_service` flag has been removed and only `--etcd_service` is available to override the service name (`trillian-logserver` by default).

This supercedes the old behaviour of announcing at, e.g. `trillian-logserver-http/localhost:5055` and `trillian-logserver/localhost:5054`. Note that the endpoint address was included in the key. That seems risky as multiple keys could co-exist and a client querying with `--prefix` won't know which is canonical.

This change also affects the log signer, and changes its key from `trillian-logsigner-http` to `trillian-logsigner/http`, and the flag for changing the service name is renamed from `--etcd_http_service` to just `--etcd_service`.

Keys announced as of v1.3.13 release:

```
etcdctl get trillian --prefix
trillian-logserver-http/localhost:50055
{"Op":0,"Addr":"localhost:50055","Metadata":null}
trillian-logserver/localhost:50054
{"Op":0,"Addr":"localhost:50054","Metadata":null}
```

Keys announced after this change:

```
etcdctl get trillian --prefix
trillian-logserver/http
{"Op":0,"Addr":"localhost:50055","Metadata":null}
trillian-logserver/rpc
{"Op":0,"Addr":"localhost:50054","Metadata":null}
```

### Checklist

### DIFF
--- a/cmd/internal/serverutil/main.go
+++ b/cmd/internal/serverutil/main.go
@@ -210,7 +210,7 @@ func (m *Main) newGRPCServer() (*grpc.Server, error) {
 // AnnounceSelf announces this binary's presence to etcd.  Returns a function that
 // should be called on process exit.
 // AnnounceSelf does nothing if client is nil.
-func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, endpoint string) func() {
+func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, endpointName, endpointAddr string) func() {
 	if client == nil {
 		return func() {}
 	}
@@ -226,13 +226,13 @@ func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, end
 	if err != nil {
 		glog.Exitf("Failed to create etcd manager: %v", err)
 	}
-	fullEndpoint := fmt.Sprintf("%s/%s", etcdService, endpoint)
-	em.AddEndpoint(ctx, fullEndpoint, endpoints.Endpoint{Addr: endpoint})
+	fullEndpoint := fmt.Sprintf("%s/%s", etcdService, endpointName)
+	em.AddEndpoint(ctx, fullEndpoint, endpoints.Endpoint{Addr: endpointAddr})
 	glog.Infof("Announcing our presence in %v", etcdService)
 
 	return func() {
 		// Use a background context because the original context may have been cancelled.
-		glog.Infof("Removing our presence in %v", etcdService)
+		glog.Infof("Removing our presence at %v", fullEndpoint)
 		ctx := context.Background()
 		em.DeleteEndpoint(ctx, fullEndpoint)
 		client.Revoke(ctx, leaseRsp.ID)
--- a/cmd/trillian_log_server/main.go
+++ b/cmd/trillian_log_server/main.go
@@ -64,13 +64,12 @@ import (
 )
 
 var (
-	rpcEndpoint     = flag.String("rpc_endpoint", "localhost:8090", "Endpoint for RPC requests (host:port)")
-	httpEndpoint    = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP metrics (host:port, empty means disabled)")
-	healthzTimeout  = flag.Duration("healthz_timeout", time.Second*5, "Timeout used during healthz checks")
-	tlsCertFile     = flag.String("tls_cert_file", "", "Path to the TLS server certificate. If unset, the server will use unsecured connections.")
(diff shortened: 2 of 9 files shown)


mhutchinson:
After proposing this, I'm starting to suspect that this isn't what we want to do. There's probably a good reason to allow multiple "grouped" keys so that all of the frontends can be found.

cmd/trillian_log_server/main.go
```diff
@@ -64,13 +64,12 @@ import (
...
-	tlsKeyFile      = flag.String("tls_key_file", "", "Path to the TLS server key. If unset, the server will use unsecured connections.")
-	etcdService     = flag.String("etcd_service", "trillian-logserver", "Service name to announce ourselves under")
-	etcdHTTPService = flag.String("etcd_http_service", "trillian-logserver-http", "Service name to announce our HTTP endpoint under")
+	rpcEndpoint    = flag.String("rpc_endpoint", "localhost:8090", "Endpoint for RPC requests (host:port)")
+	httpEndpoint   = flag.String("http_endpoint", "localhost:8091", "Endpoint for HTTP metrics (host:port, empty means disabled)")
+	healthzTimeout = flag.Duration("healthz_timeout", time.Second*5, "Timeout used during healthz checks")
+	tlsCertFile    = flag.String("tls_cert_file", "", "Path to the TLS server certificate. If unset, the server will use unsecured connections.")
+	tlsKeyFile     = flag.String("tls_key_file", "", "Path to the TLS server key. If unset, the server will use unsecured connections.")
+	etcdService    = flag.String("etcd_service", "trillian-logserver", "Service name to announce ourselves under")
```
pavelkalinnikov, line 72: The "rpc"/"http" suffix is hard-coded which means the user needs to know it (from help messages, README, etc). Alternatively, we could have flags that set the whole path:

```
flag.String("etcd_service", "trillian-logserver/rpc", "... mention the <namespace>/<thing> format")
flag.String("etcd_http_service", "trillian-logserver/http", "...")
```

pavelkalinnikov:
> After proposing this, I'm starting to suspect that this isn't what we want to do. There's probably a good reason to allow multiple "grouped" keys so that all of the frontends can be found.

Are you saying that with this PR only one frontend (the latest starting one) will "win" the key/slot, and others won't be discoverable? I agree this would be undesirable.

cmd/internal/serverutil/main.go
```diff
@@ -226,13 +226,13 @@ func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, end
 	if err != nil {
 		glog.Exitf("Failed to create etcd manager: %v", err)
 	}
-	fullEndpoint := fmt.Sprintf("%s/%s", etcdService, endpoint)
-	em.AddEndpoint(ctx, fullEndpoint, endpoints.Endpoint{Addr: endpoint})
+	fullEndpoint := fmt.Sprintf("%s/%s", etcdService, endpointName)
```
pavelkalinnikov, line 229: It looks like if you put `endpointAddr` instead of `endpointName` here, it will work as previously.

pavelkalinnikov: I'm judging by this guide: https://etcd.io/docs/v3.4.0/dev-guide/grpc_naming/

Writing `("my-service", naming.Update{Op: naming.Add, Addr: "1.2.3.4", Metadata: "..."})` resulted in a `"my-service/1.2.3.4"` key.

pavelkalinnikov: But the new guide has a misleading "/e1" suffix which I'm not sure how to interpret: https://etcd.io/docs/current/dev-guide/grpc_naming/

mhutchinson: This documentation is what led me down this adventure, thinking I was being really smart in making the key unique. I'm now less sure, and think that this is only useful for master-election type of stuff. For service discovery, having a common prefix may actually be more useful. i.e. this PR should just be reverted. I still need to step back and consider this when the world gives me an hour without a high priority interrupt :-)

pavelkalinnikov:
Is this still WIP, or obsolete?

mhutchinson:
I'll close this. I'm not in the market for functional changes here, and I'm now pretty convinced that the behaviour has stayed the same even through the etcd api upgrade.

PR #2382, opened 2021-03-05
Bumped cloud.google.com/go/spanner and mod tidy

mhutchinson:
This supersedes #2374. We still need to fix the presubmit to not die on a diff in the sum file...


mhutchinson:
I've created #2383 which should avoid the need to do this manually on each dependabot change.

pavelkalinnikov (approved):

PR #2383, opened 2021-03-05
Support skipping processing go mod tidy

mhutchinson:
This allows the cloudbuild to accept diffs from dependabot without more manual intervention. It does allow humans to also submit changes without running mod tidy, but hopefully we're all better disciplined than that. By default, ./scripts/presubmit will still fix up the go.sum file.

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -26,12 +26,13 @@ check_cmd() {
 }
 
 usage() {
-  echo "$0 [--coverage] [--fix] [--no-build] [--no-linters] [--no-generate] [--empty-diff]"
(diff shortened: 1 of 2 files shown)


scripts/presubmit.sh
```diff
@@ -26,12 +26,13 @@ check_cmd() {
 }
 
 usage() {
-  echo "$0 [--coverage] [--fix] [--no-build] [--no-linters] [--no-generate] [--empty-diff]"
+  echo "$0 [--coverage] [--fix] [--skip-gomod] [--no-build] [--no-linters] [--no-generate] [--empty-diff]"
```
pavelkalinnikov, line 29: How about `--no-mod-tidy`?

pavelkalinnikov (approved):

scripts/presubmit.sh
mhutchinson: Yeah, good call. Done.

Issue #2378, opened 2021-03-01
Reduce code complexity related to maps


pavelkalinnikov:
Some of the code is unnecessarily complex because previously it tried to accommodate both `Log` and `Map` trees. After maps removal [#2284] it makes sense to minimize this code, so that it's log-specific. If anything related to maps reappears in the repo, it still makes sense to keep the code decoupled.

Examples:

- [storage/cache](https://github.com/google/trillian/tree/3615e67ca1e376d40e92178d524a5499ba485545/storage/cache) can be removed completely, or significantly simplified.
- The [NodeID&co](https://github.com/google/trillian/blob/3615e67ca1e376d40e92178d524a5499ba485545/storage/tree/node.go) types should be made log-specific. It likely needs to be replaced by a simple [compact.NodeID](https://github.com/google/trillian/blob/3bc7219b903515f5a24d5646c8c3679d017cda15/merkle/compact/nodes.go#L30), whereas for maps the [NodeID2](https://github.com/google/trillian/blob/3615e67ca1e376d40e92178d524a5499ba485545/storage/tree/node_id2.go) type is more suitable.
- The LogStorage/TreeStorage/MapStorage interfaces can be reduced.

mhutchinson:
@pavelkalinnikov is the checklist in the original comment still accurately tracking the remaining work on this?

pavelkalinnikov:
Yes, it still looks relevant.

PR #2391, opened 2021-03-10
storage: Remove revision from GetMerkleNodes

pavelkalinnikov:
Log trees always read from the revision corresponding to the root which
is read in the beginning of the transaction. There is no need for it to
make a roundtrip to the application layer.

Part of a larger refactoring [#2378]. More follow-up changes are WIP.

### Checklist

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/trillian/monitoring"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/cache"
+	"github.com/google/trillian/storage/tree"
 	"github.com/google/trillian/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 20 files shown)


storage/mysql/log_storage.go
```diff
@@ -372,6 +373,14 @@ func (t *logTreeTX) WriteRevision(ctx context.Context) (int64, error) {
 	return t.treeTX.writeRevision, nil
 }
 
+// GetMerkleNodes returns the requested nodes at the read revision.
+func (t *logTreeTX) GetMerkleNodes(ctx context.Context, nodeIDs []tree.NodeID) ([]tree.Node, error) {
```
pavelkalinnikov, line 377: This bit might look confusing: The `GetMerkleNodes` method just moved from tree storage to log storage.
Reason: tree storage (for this particular implementation) doesn't know `root.revision`, whereas the log storage does. In contrast, in other implementation (e.g. `cloudspanner` and `spanner`) tree storage does know it.

This indicates that tree vs log storage abstractions are sometimes violated / not defined. Given that we only have logs now, the bigger plan is to combine the two into one interface [#2378].

pphaneuf (approved):

PR #2394, opened 2021-03-11
Use 64 bits for node IDs and SubtreeCache

pavelkalinnikov:


pphaneuf (approved):
This is very nice, thanks!

PR #2398, opened 2021-03-11
SubtreeCache: Use log tile funcs directly

pavelkalinnikov:
This change makes the `SubtreeCache` use log tile population functions directly,
as it no longer needs to be log/map polymorphic.

#2378


pavelkalinnikov:
I recommend reviewing commit-by-commit, with "Hide whitespace changes" enabled.

storage/aliases.go
```diff
@@ -14,10 +14,7 @@
 
 package storage
 
-import (
-	"github.com/google/trillian/storage/storagepb"
-	"github.com/google/trillian/storage/tree"
```
pphaneuf, line 19: This builds without `tree` being imported?

pphaneuf (approved):

storage/aliases.go
pavelkalinnikov: Look 2 lines below :)

PR #2401, opened 2021-03-12
Use NodeID2 in map hasher

pavelkalinnikov:
`MapHasher` is the penultimate user of the `NodeID` type, which is being
removed soon. Since the maps are deprecated/experimental, we are free to
change the hasher interface to use a better `NodeID2` type.

Part of #2378.


mhutchinson (approved):
Cool. This does shine a light on `github.com/google/trillian/storage/tree` being a weird package; `storage` doesn't feel like something that should be on the API for clients to  be leaning on.

My other concern with this is that this will need importing into `trillian-examples` and seeing what delta it causes for the maps there. However, that isn't going to happen until the pending update (https://github.com/google/trillian-examples/pull/257) is applied, so I guess this will fall on me to maintain.

pavelkalinnikov:
`storage/tree` needs decomposing into `merkle` (the core/algorithm bits), and `storage`-specific things. I will do this as part of this same refactoring. Previously some bits were moved here to avoid circular dependencies.

pavelkalinnikov:
MapHasher is the penultimate user of the NodeID type, which is being
removed soon. Since the maps are deprecated/experimental, we are free to
change the hasher interface to use a better NodeID2 type.

PR #2404, opened 2021-03-17
Remove legacy NodeID

pavelkalinnikov:
This change eliminates the last cases of using `NodeID`, most notably in `SubtreeCache`,
and removes this type. Now `compact.NodeID` is used all the way down to storage, but
storage implementations still use `[]byte` slices to address tiles, for backward compatibility
with the existing databases.

Review commit by commit is recommended.

Part of #2378

### Checklist


storage/cache/gen.go
```diff
@@ -21,11 +21,10 @@ import (
 	"context"
 
 	"github.com/google/trillian/storage/storagepb"
-	"github.com/google/trillian/storage/tree"
 )
 
 // NodeStorage provides an interface for storing and retrieving subtrees.
 type NodeStorage interface {
-	GetSubtree(n tree.NodeID) (*storagepb.SubtreeProto, error)
+	GetSubtree(prefix []byte) (*storagepb.SubtreeProto, error)
```
Martin2112, line 28: Unfortunately golang has no immutable byte string type. There could be value in using string instead. OTOH the original NodeID type was not immutable and I suspect it's not worth doing.

storage/cache/subtree_cache.go
```diff
@@ -174,8 +172,8 @@ func (s *SubtreeCache) preload(ids []tree.NodeID, getSubtrees GetSubtreesFunc) e
 	// Create empty subtrees for anything left over. Note that multiple parallel readers
 	// may be be running this code and touch the same keys, although this doesn't happen
 	// normally.
-	for _, id := range want {
-		if err := s.cacheSubtree(s.newEmptySubtree(id)); err != nil {
+	for id := range want {
+		if err := s.cacheSubtree(s.newEmptySubtree([]byte(id))); err != nil {
```
Martin2112, line 176: May have been true before but there's no test covering this line.

storage/cache/subtree_cache.go
```diff
@@ -197,36 +195,26 @@ func (s *SubtreeCache) cacheSubtree(t *storagepb.SubtreeProto) error {
...
-		}
-		ids[i] = id
-	}
-
+func (s *SubtreeCache) GetNodes(ids []compact.NodeID, getSubtrees GetSubtreesFunc) ([]tree.Node, error) {
 	glog.V(2).Infof("cache: GetNodes(count=%d)", len(ids))
 	if glog.V(3) {
-		for _, n := range ids {
-			glog.Infof("  cache: GetNodes(path=%x, prefixLen=%d)", n.Path, n.PrefixLenBits)
+		for _, id := range ids {
```
Martin2112, line 201: Might be time to remove logging the IDs. Not sure we'll ever use it.

storage/cache/subtree_cache.go
```diff
@@ -197,36 +195,26 @@ func (s *SubtreeCache) cacheSubtree(t *storagepb.SubtreeProto) error {
...
+	for _, id := range ids {
 		h, err := s.getNodeHash(
 			id,
-			func(n tree.NodeID) (*storagepb.SubtreeProto, error) {
+			func(id []byte) (*storagepb.SubtreeProto, error) {
 				// This should never happen - we should've already read all the data we
 				// need above, in Preload()
-				glog.Warningf("Unexpectedly reading from within getNodeHash(): %s", n.String())
-				ret, err := getSubtrees([]tree.NodeID{n})
+				glog.Warningf("Unexpectedly reading from within getNodeHash(): %x", id)
```
Martin2112, line 216: I think we previously found the above comment to be incorrect. Might demote the log level. This condition is also not tested. Again, I suspect that was true before.

storage/cache/subtree_cache.go
```diff
@@ -271,19 +259,18 @@ func (s *SubtreeCache) prefixIsDirty(prefixKey string) bool {
 }
 
 // getNodeHash returns a single node hash from the cache.
-func (s *SubtreeCache) getNodeHash(id tree.NodeID, getSubtree GetSubtreeFunc) ([]byte, error) {
+func (s *SubtreeCache) getNodeHash(id compact.NodeID, getSubtree GetSubtreeFunc) ([]byte, error) {
 	if glog.V(3) {
-		glog.Infof("cache: getNodeHash(path=%x, prefixLen=%d) {", id.Path, id.PrefixLenBits)
+		glog.Infof("cache: getNodeHash(%+v) {", id)
```
Martin2112, line 264: Remove this logging as well?

storage/cache/subtree_cache.go
```diff
@@ -294,10 +281,10 @@ func (s *SubtreeCache) getNodeHash(id tree.NodeID, getSubtree GetSubtreeFunc) ([
 			}
 		}
 		if c.Prefix == nil {
-			panic(fmt.Errorf("getNodeHash nil prefix on %v for id %v with px %x", c, id.String(), subKey))
+			panic(fmt.Errorf("getNodeHash nil prefix on %v for id %+v with px %x", c, id, subID))
```
Martin2112, line 284: Can we propagate an error instead of the panic? It's a long standing issue that would be good to fix.

storage/cache/subtree_cache.go
```diff
@@ -321,44 +308,39 @@ func (s *SubtreeCache) getNodeHash(id tree.NodeID, getSubtree GetSubtreeFunc) ([
 		if err != nil {
 			glog.Errorf("base64.DecodeString(%v): %v", sfxKey, err)
 		}
-		glog.Infof("getNodeHash(%x | %x): %x", subKey, b, nh)
+		glog.Infof("getNodeHash(%x | %x): %x", subID, b, nh)
```
Martin2112, line 311: Too detailed?

storage/cache/subtree_cache.go
```diff
@@ -321,44 +308,39 @@ func (s *SubtreeCache) getNodeHash(id tree.NodeID, getSubtree GetSubtreeFunc) ([
...
 	return nh, nil
 }
 
 // SetNodeHash sets a node hash in the cache.
-func (s *SubtreeCache) SetNodeHash(nID compact.NodeID, h []byte, getSubtree GetSubtreeFunc) error {
-	id, err := tree.NewNodeIDForTreeCoords(int64(nID.Level), int64(nID.Index), maxSupportedTreeDepth)
-	if err != nil {
-		return fmt.Errorf("failed to create nodeID %v: %v", nID, err)
-	}
+func (s *SubtreeCache) SetNodeHash(id compact.NodeID, h []byte, getSubtree GetSubtreeFunc) error {
```
Martin2112, line 321: The tests for this function don't cover several of the branches. I suggest we look at that separately.

storage/cache/subtree_cache.go
pavelkalinnikov: Agree. Doing this separately in #2405, and will rebase this PR.

storage/cache/subtree_cache.go
pavelkalinnikov: #2405

storage/cache/subtree_cache.go
pavelkalinnikov: #2405

storage/cache/subtree_cache.go
pavelkalinnikov: Yes, it [wasn't](https://codecov.io/gh/google/trillian/src/master/storage/cache/subtree_cache.go) covered previously (same for other lines, this PR seemingly shouldn't change coverage of `subtree_cache.go`).

Unlike for maps, this line won't be executed for logs, because in this spot logs read only existing tiles on the right border, so `len(want)` will end up zero (cleared by `delete` statement a few lines above). In fact, it would probably be safer to return an error here instead of making an empty tile (which in fact might be not empty but lost due to a bug).

#2406

storage/cache/subtree_cache.go
pavelkalinnikov: #2406

storage/cache/subtree_cache.go
pavelkalinnikov: I believe it's correct (or don't see why it's not). Also, I found 0 occurrences of this warning in our logs. I think we should rather return an error here, or don't have this callback in the first place. To be addressed in follow-ups.

Martin2112: Maybe it was maps again? I remember we tried to remove it before and it did happen.

storage/cache/gen.go
pavelkalinnikov: Within the scope of this PR, the safety remains the same as it was. I might improve it in follow-ups.

pavelkalinnikov:
Left pre-existing coverage/safety things for future addressing. For now I can see that `SubtreeCache` will almost go away, so fixing small aspects of it seems more work than a straight replacement with 100% test coverage.

storage/cache/subtree_cache.go
pavelkalinnikov: Do you suggest a follow-up, or as a prereq for this PR?

pavelkalinnikov: I found a simple way to add coverage here, will send a PR shortly.

Martin2112: Followup is ok.

Martin2112 (approved):

storage/cache/subtree_cache.go
pavelkalinnikov: Checked #2407 on this PR, tests pass. So submitting it optimistically.

PR #2406, opened 2021-03-18
cache: Handle missing tiles stricter

pavelkalinnikov:
`GetMerkleNodes` is always called with existing nodes, therefore `SubtreeCache.preload`
will always try to read existing tiles. We should return an error if some tile is not found,
instead of silently creating an empty tile.

Addresses some comments in #2404.
Part of #2378.

### DIFF
--- a/storage/cache/subtree_cache.go
+++ b/storage/cache/subtree_cache.go
@@ -168,17 +168,9 @@ func (s *SubtreeCache) preload(ids []tree.NodeID, getSubtrees GetSubtreesFunc) e
 		}
 		delete(want, string(t.Prefix))
 	}
-
-	// We might not have got all the subtrees we requested, if they don't already exist.
-	// Create empty subtrees for anything left over. Note that multiple parallel readers
-	// may be be running this code and touch the same keys, although this doesn't happen
-	// normally.
-	for _, id := range want {
-		if err := s.cacheSubtree(s.newEmptySubtree(id)); err != nil {
-			return err
-		}
+	if r := len(want); r != 0 {
+		return fmt.Errorf("preload did not get all tiles: %d not found", r)
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pavelkalinnikov:
Interesting, there is some failing integration test. Looking.

storage/cache/subtree_cache.go
```diff
@@ -168,17 +168,9 @@ func (s *SubtreeCache) preload(ids []tree.NodeID, getSubtrees GetSubtreesFunc) e
...
-
-	// We might not have got all the subtrees we requested, if they don't already exist.
-	// Create empty subtrees for anything left over. Note that multiple parallel readers
-	// may be be running this code and touch the same keys, although this doesn't happen
-	// normally.
-	for _, id := range want {
-		if err := s.cacheSubtree(s.newEmptySubtree(id)); err != nil {
-			return err
-		}
+	if r := len(want); r != 0 {
```
Martin2112, line 171: I'm not 100% sure if there's an edge case where this happens. It might possibly be where the tree grows enough to require a new tile that wasn't previously written.

pavelkalinnikov: The key point is in the PR description: we will not try to read this new tile with `GetMerkleNodes` (and hence `SubtreeCache.preload`). The latter method is used in 2 scenarios:
1. Sequencer's transaction: read compact range of the current tree. These nodes (and hence tiles) always exist, unless someone removed them from the DB.
2. Reading proofs. These "read" transactions also know the current tree size, and request the existing nodes.

pavelkalinnikov: This is a leftover from maps, I believe. There we could read a leaf which previously did not exist, so we would need empty tiles silently added.

pavelkalinnikov:
Fixed the tests.

storage/cache/subtree_cache.go
Martin2112: Maybe it was added for the map. It's a long time ago now.

Martin2112 (approved):

PR #2408, opened 2021-03-19
Cleanup map hasher

pavelkalinnikov:
This change removes a few unused `MapHasher` bits, and reduces dependencies.

Review commit by commit is recommended.


pavelkalinnikov:
Some unit tests try to create `Map`, will fix them shortly.

pphaneuf (approved):
Great stuff, just one meta-nit: the "Move coniks package one level down" commit is actually moving it one level *up*?

pavelkalinnikov:
Haha, yes, 1 level up. More generally, 1 level in the **better** direction.

pphaneuf:
Haha! Yes, this is definitely in the better direction, for sure!

pavelkalinnikov:
@mhutchinson If you use `coniks` in trillian-examples, might need to adjust the import path. Also, consider using it directly, instead of referencing `MapHasher` interface, or redefine the interface locally. I tried moving some stuff to `smt`, and it turns out there are some circular dependencies because of those interfaces. I then remembered Go "best practice" is defining interfaces where they are used. Essentially, the `merkle/hashers.MapHasher` interface is not needed.

Issue #2245, opened 2020-10-23
strip out unused and unneeded code


pphaneuf:
A number of APIs (both external and internal) add burden to storage implementers, without bringing much benefits.

For example, `QueueLeaves` is complex to implement correctly, due to being a batching API with partial failure, but in practice, there are possibly no need for it, and due to the implementation complexity, is very unlikely to provide any performance benefit over using `QueueLeaf` (the non-batch version of this API) multiple times (potentially concurrently), say, and this both leaves the user in better control over what happens in the partial failure cases, and removes complexity of implementation (and the inevitable bugs that it brings).

PR #2423, opened 2021-03-22
Remove Cloud Spanner support.

pphaneuf:
Part of #2245.


pavelkalinnikov:
I remember from @AlCutter that we run some internal CloudSpanner instance. Defer the review to Al.

pphaneuf:
Sure thing. I *think* we've stopped using that instance, but should get confirmed, @AlCutter should know.

AlCutter:
The CI/CD instance uses CloudSpanner

pphaneuf:
Seems that we're using it for some post-submit testing, so we'll keep it.

PR #2438, opened 2021-03-30
Update CI golang version latest

AlCutter:
### Checklist

### DIFF
--- a/examples/deployment/docker/log_signer/Dockerfile
+++ b/examples/deployment/docker/log_signer/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13 as build
+FROM golang:buster as build
 
 WORKDIR /trillian
 
--- a/integration/cloudbuild/testbase/Dockerfile
+++ b/integration/cloudbuild/testbase/Dockerfile
@@ -1,5 +1,5 @@
 # This Dockerfile builds a base image for Trillan integration tests.
-FROM ubuntu
+FROM golang:buster
 
 WORKDIR /testbase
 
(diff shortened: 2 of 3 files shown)


integration/cloudbuild/testbase/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This Dockerfile builds a base image for Trillan integration tests.
-FROM ubuntu
+FROM golang:buster
```
pavelkalinnikov, line 2: What does "buster" mean? Is it based on ubuntu too?

AlCutter: Buster is the latest stable Debian release

pavelkalinnikov: Does it change often? Given the idea of "testbase" image being cacheable, should this be a more static version?

pavelkalinnikov: Although it seems like the OS version is outside of these labels' control. We can only change Go version.

examples/deployment/docker/log_signer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13 as build
+FROM golang:buster as build
```
pavelkalinnikov, line 1: I would slightly prefer a static Go version, for better reproduceability.

Also, what is the point of using the latest one? Are we targeting any specific feature?

pphaneuf:
Get this merged!

examples/deployment/docker/log_signer/Dockerfile
AlCutter: It's nice to find out that upstream golang releases break our build though...

pphaneuf (approved):

examples/deployment/docker/log_signer/Dockerfile
pavelkalinnikov: We'll catch it locally anyway if it happens, and also when we upgrade it next time. Not knowing which Go version is run is CI (correspondingly, not being able to tell in `README` which version to use for dev), seems worse to me.

AlCutter: I'm not sure having pins in has worked very well though; we had bits pinned to 1.11, others on 1.12, 1.13, and 1.15 - and we (at least I) only found that out when I tried to bump deps which depended on parts of the golang stdlib which didn't exist...

pphaneuf: It'd be nice if this was defined in one place, not sure if that's even possible, just wishful thinking...

PR #2441, opened 2021-04-01
Clean up a few more bits of verifiable maps.

pphaneuf:
### Checklist


trillian.proto
```diff
@@ -99,15 +99,15 @@ enum TreeType {
   // Tree represents a verifiable log.
   LOG = 1;
 
-  // Tree represents a verifiable map.
-  MAP = 2;
+  // Deprecated: verifiable maps are no longer supported.
+  DEPRECATED_MAP = 2 [deprecated = true];
```
Martin2112, line 103: Not sure there's any point deprecating it as it already no longer works. Just reserve the field?

pphaneuf: Good call, done.

trillian.proto
```diff
@@ -99,15 +99,15 @@ enum TreeType {
...
 
-  // Tree represents a verifiable map.
-  MAP = 2;
-
   // Tree represents a verifiable pre-ordered log, i.e., a log whose entries are
   // placed according to sequence numbers assigned outside of Trillian.
   PREORDERED_LOG = 3;
+
+  reserved 2;
+  reserved "MAP";
```
pavelkalinnikov, line 107: I wonder whether reserving these has any effect in Go. The diff in ".pb.go" file doesn't seem different from just deleting these lines?

pavelkalinnikov (approved):
LGTM. One question inlined.

trillian.proto
pavelkalinnikov: Ah, this looks like not a compatibility mechanism, but it just prevents future edits reusing this field number/name?

pphaneuf: Yeah, they're the same, their effect is in `protoc`, where if you try to reuse the name or value, it'll throw an error.

This used to be done by just leaving a comment, traditionally, but sometimes, you have these messages that are very long, and people miss the comment...

Martin2112 (approved):

PR #2442, opened 2021-04-01
Remove the unused KeyHint field.

pphaneuf:
Reviewing commit by commit might be easier.

### Checklist


pphaneuf:
This PR might be easiest to review commit-by-commit.

pavelkalinnikov (approved):

Martin2112 (approved):

trillian.proto
```diff
@@ -211,15 +211,8 @@ message SignedLogRoot {
...
-  // key_hint is a hint to identify the public key for signature verification.
-  // key_hint is not authenticated and may be incorrect or missing, in which
-  // case all known public keys may be used to verify the signature.
-  // When directly communicating with a Trillian gRPC server, the key_hint will
-  // typically contain the LogID encoded as a big-endian 64-bit integer;
-  // however, in other contexts the key_hint is likely to have different
-  // contents (e.g. it could be a GUID, a URL + TreeID, or it could be
-  // derived from the public key itself).
-  bytes key_hint = 7;
+  reserved "key_hint";
```
pavelkalinnikov, line 214: Can we have compatibility issues if some historical `SignedLogRoot` has this field?

pphaneuf: If we deserialise a stored proto that has this field set, for example, it will be interpreted as an unknown field (the `reserved` doesn't affect that behaviour, it only causes errors during compilation), which will be to just ignore it.

I believe there should be some effort to keep its content, but that might depend on the specific runtime library (that was the case with the C++ library for example, not sure with the Go one).

PR #2448, opened 2021-04-07
Tidy up protobuf reserved fields.

pphaneuf:


trillian.proto
```diff
@@ -251,6 +235,15 @@ message SignedLogRoot {
 
   // log_root_signature is the raw signature over log_root.
   bytes log_root_signature = 9;
+
+  reserved 1, 2, 3, 4, 5, 6, 7;
```
pavelkalinnikov, line 239: Could use the `reserved 1 to 7;` syntax?

trillian.proto
```diff
@@ -251,6 +235,15 @@ message SignedLogRoot {
...
+  reserved "key_hint";
```
pavelkalinnikov, line 240: Optionally:
Add comments with int tags? E.g.:
```
  reserved "log_id"; // 5
```
Or sort by tag instead of lexicographically?

pavelkalinnikov (approved):
LGTM % optional nit

trillian.proto
pphaneuf: Ah yes, I always forget about that syntax! Done.

trillian.proto
pphaneuf: I don't think this really brings any useful information? The important thing is that these should no longer be used, and keeping them in lexicographical order is easiest to check and maintain.

pavelkalinnikov: Fair enough.

PR #2449, opened 2021-04-07
Unexport internal members of client.LogVerifier.

pphaneuf:


client/log_verifier.go
```diff
@@ -34,22 +34,22 @@ import (
...
-	// SigHash computes the digest of LogRoot for signing.
-	SigHash crypto.Hash
-	v       logverifier.LogVerifier
+	// hasher is the hash strategy used to compute nodes in the Merkle tree.
+	hasher hashers.LogHasher
+	// pubKey verifies the signature on the digest of LogRoot.
+	pubKey crypto.PublicKey
+	// sigHash computes the digest of LogRoot for signing.
+	sigHash crypto.Hash
+	verifier       logverifier.LogVerifier
```
pavelkalinnikov, line 43: Personally, I would leave it as `v`. **And** maybe even made other private names 1-2 letters long. Feels like Go style to me.

pavelkalinnikov (approved):
LGTM % optional nit.

client/log_verifier.go
pphaneuf: I find the single-letter names difficult to search for when using more "basic" editors like vi, but done (the important part was unexporting!).

pavelkalinnikov: Good point about vi, didn't look at it from this angle. My approach is usually typing something like `c\.v` to search, but it requires knowing the receiver name (which is hopefully the same across the entire file).

Issue #2450, opened 2021-04-07
remove signing and signatures


pphaneuf:
Early on, we had hope that Trillian could be run in a separate trust domain, and that Trillian data structure might become commonly exchanged in distributed systems. In practice, there seems to be too much variation in the needs of different ecosystems. In addition to that, Trillian is not currently suitable to be run in its own trust domain, and providing signatures can provide a false sense of security. They also bring a lot of complexity, in the form of key management.

AlCutter:
Is there more to do on the signatures or can we close this?

pphaneuf:
There is a bit more, I was just letting the previous changes settle in, but I'll get back on it soon to finish it up.

garantir-km:
I am not fully following the reasoning for removing signatures. I see the reasons listed as:

1. They provide a false sense of security
2. Key management is difficult

Number 1 can be resolved with documentation on how to **fully** verify a log entry. In particular, verifying a signature should be a necessary but not sufficient condition to verifying the log entry. As for number 2, Trillian doesn't need to solve the key management issue. It can just support a standard crypto interface (e.g., PKCS11) as well as software-based keys (e.g., PKCS12) for those looking for easier deployments. How those keys are managed is up to the end users. At the very least, they are no worse off (security-wise) for using signatures as compared to not using signatures.

pphaneuf:
The "false sense of security" is not related to the verification of a given log entry, but rather, to the effective isolation of different users of a shared Trillian instance. In effect, this demonstrates my argument, as signatures would be required for such a deployment, but would be in no way sufficient, and their presence seem to make people think that the other aspects would also be covered.

These other aspects are not covered, and we have no intention of covering them (they are not a "work in progress"), so this is actively misleading.

The key management *code* in Trillian is not that complex (there was in fact PKCS11 support), but it complicates deployment beyond what is reasonable, especially in consideration of the signatures not actually permitting the benefit that they might.

The only reasonable deployment environment for Trillian at this time and for the expected future is where the Trillian RPC client (such as CTFE, for example) is in the same trust domain.

PR #2451, opened 2021-04-07
Do not verify the signature in SignedLogRoot.

pphaneuf:
Part of #2450.

### Checklist

### DIFF
--- a/crypto/signer_test.go
+++ b/crypto/signer_test.go
@@ -107,7 +107,7 @@ func TestSignLogRoot(t *testing.T) {
 			t.Errorf("len(sig): %v, want > 0", got)
 		}
 		// Check that the signature is correct
-		if _, err := VerifySignedLogRoot(key.Public(), crypto.SHA256, slr); err != nil {
+		if err := Verify(key.Public(), crypto.SHA256, slr.LogRoot, slr.LogRootSignature); err != nil {
 			t.Errorf("Verify(%v) failed: %v", test.root, err)
 		}
 	}
--- a/client/log_verifier.go
+++ b/client/log_verifier.go
@@ -15,19 +15,14 @@
 package client
 
 import (
-	"crypto"
 	"errors"
 	"fmt"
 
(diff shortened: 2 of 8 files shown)


pavelkalinnikov:
Hmm, not sure about this one. What's the bigger plan?

Was the signature verification also sort of checking the "safety" of the TLS-encoded blob? I think it came in pair with the TLS encoding. In this regard, maybe the TLS encoding is also not needed on its own?

pphaneuf:
The TLS encoding was needed for stability of the signature (but all along, since it was being kept in a `bytes` field, it could also have been a serialised protobuf). You are correct that without signature validation, TLS encoding is not necessary, but I'm going one step at a time... 😄

pphaneuf:
FYI, the plan is the same as usual, removing code "from the front": remove the client code checking the signatures, remove the signature fields in the RPCs, remove the infrastructure supporting filling out those fields, etc...

crypto/signer_test.go
```diff
@@ -107,7 +107,7 @@ func TestSignLogRoot(t *testing.T) {
 			t.Errorf("len(sig): %v, want > 0", got)
 		}
 		// Check that the signature is correct
-		if _, err := VerifySignedLogRoot(key.Public(), crypto.SHA256, slr); err != nil {
+		if err := Verify(key.Public(), crypto.SHA256, slr.LogRoot, slr.LogRootSignature); err != nil {
```
pavelkalinnikov, line 110: Is this signature/key also to be removed (separately)?

pavelkalinnikov (approved):

crypto/signer_test.go
pphaneuf: Yes, I wanted to just remove the signature verification in the PR, but without affecting the signing itself (including the code testing the signing).

Later PRs will remove the signing itself, as well as the key management infrastructure (which is unfortunately quite plentiful and deeply rooted!).

PR #2454, opened 2021-04-08
Remove signer from sequencer.

pphaneuf:
Part of #2450.

### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -33,8 +33,6 @@ import (
 	"github.com/google/trillian/types"
 	"github.com/google/trillian/util/clock"
 	"google.golang.org/protobuf/types/known/timestamppb"
-
-	tcrypto "github.com/google/trillian/crypto"
 )
 
 const logIDLabel = "logid"
@@ -103,7 +101,6 @@ type Sequencer struct {
 	hasher     hashers.LogHasher
 	timeSource clock.TimeSource
 	logStorage storage.LogStorage
-	signer     *tcrypto.Signer
 	qm         quota.Manager
 }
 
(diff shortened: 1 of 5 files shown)


log/sequencer_test.go
```diff
@@ -389,7 +385,6 @@ func TestIntegrateBatch(t *testing.T) {
 				merkleNodesGet:   &compactTree16,
 				updatedLeaves:    &noLeaves,
 				merkleNodesSet:   &noNodes,
-				signer:           fixedGoSigner,
```
pavelkalinnikov, line 392: Is `fixedGoSigner` used anywhere else? Delete it?

log/sequencer.go
```diff
@@ -103,7 +101,6 @@ type Sequencer struct {
 	hasher     hashers.LogHasher
 	timeSource clock.TimeSource
 	logStorage storage.LogStorage
-	signer     *tcrypto.Signer
```
pavelkalinnikov, line 106: Hm, was this field unused?

pavelkalinnikov (approved):

log/sequencer_test.go
pphaneuf: Yes, still used in `sequencer_manager_test.go`:

https://github.com/google/trillian/blob/e3eff67ad434aacbc281c39d841dcf25528d1cb6/log/sequencer_manager_test.go#L104

I thought of moving it, to keep it together, but it will be disappearing entirely soon, so I figured keeping the diff smaller was better.

log/sequencer.go
pphaneuf: Yes, the use was removed in this commit:

https://github.com/google/trillian/commit/884ef511ce166d1cb1ffa1ae2cb4aefb3018591d#diff-6074443f242fbef0f5f94ab179ab499f8ee83fedf0a36f79fbe52c637ca14cdd

pphaneuf: Also, it still builds! 😃

PR #2456, opened 2021-04-08
Remove trees.Hash.

pphaneuf:
Part of #2245.

### Checklist

### DIFF
--- a/integration/quota/quota_test.go
+++ b/integration/quota/quota_test.go
@@ -16,6 +16,7 @@ package quota
 
 import (
 	"context"
+	"crypto"
 	"errors"
 	"fmt"
 	"hash"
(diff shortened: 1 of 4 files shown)


integration/quota/quota_test.go
```diff
@@ -136,11 +136,7 @@ func runRateLimitingTest(ctx context.Context, s *testServer, numTokens int) erro
 	if err != nil {
 		return fmt.Errorf("InitLog() returned err = %v", err)
 	}
-	hasherFn, err := trees.Hash(tree)
-	if err != nil {
-		return fmt.Errorf("trees.Hash()=%v, want: nil", err)
-	}
-	hasher := hasherFn.New()
+	hasher := crypto.SHA256.New()
```
pavelkalinnikov, line 139: Should we still verify that `tree.HashAlgorithm == sigpb.DigitallySigned_SHA256`, or it's done elsewhere?

pavelkalinnikov: Ah, no need to do in the in the tests. What about the main code?

pavelkalinnikov (approved):

integration/quota/quota_test.go
pphaneuf: This is what I figured (that it was ok in the tests)... At the one "real" call site (`Signer`, in `trees.go`), I replaced the call with a check that it is the expected algorithm.

PR #2458, opened 2021-04-09
Strip down crypto.Signer.

pphaneuf:
### Checklist


crypto/signer.go
```diff
@@ -17,47 +17,20 @@ package crypto
 
 import (
 	"crypto"
-	"crypto/rand"
-
-	"golang.org/x/crypto/ed25519"
 )
```
pavelkalinnikov, line 20: Can squash this into 1 line. (my favourite nit :D)

pavelkalinnikov (approved):

crypto/signer.go
pphaneuf: Done. 😃

PR #2459, opened 2021-04-09
Remove unused parameter from crypto.NewSigner.

pphaneuf:
Part of #2245.

### Checklist

### DIFF
--- a/crypto/signer.go
+++ b/crypto/signer.go
@@ -24,7 +24,7 @@ type Signer struct {
 }
 
 // NewSigner returns a new signer.
-func NewSigner(signer crypto.Signer, _ crypto.Hash) *Signer {
+func NewSigner(signer crypto.Signer) *Signer {
 	return &Signer{Signer: signer}
 }
 
--- a/trees/trees.go
+++ b/trees/trees.go
@@ -18,7 +18,6 @@ package trees
 
 import (
 	"context"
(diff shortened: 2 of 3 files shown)


Martin2112 (approved):

crypto/signer.go
```diff
@@ -24,7 +24,7 @@ type Signer struct {
 }
 
 // NewSigner returns a new signer.
-func NewSigner(signer crypto.Signer, _ crypto.Hash) *Signer {
+func NewSigner(signer crypto.Signer) *Signer {
```
pavelkalinnikov, line 27: It seems like this whole struct is now removable.

pavelkalinnikov (approved):

crypto/signer.go
pphaneuf: Yes, it could be just replaced, but keeping the type allows for controlling what methods are available (only `Public()` now!), which helps reasoning about it for later PRs.

The last usage of it (and there's only *one*!) is intertwined with a lot of other things, more PRs are coming to untangle this in a way that can be understood and reviewed more comfortably.

pavelkalinnikov: SGTM. Btw, if the purpose is to reduce the number of available methods, a shorter approach might be introducing an interface with just those methods.

PR #2460, opened 2021-04-09
Migrate protobuf from v1 to v2.

pphaneuf:
### Checklist


gen.go
```diff
@@ -14,7 +14,7 @@
 
 package trillian
 
-//go:generate protoc -I=. -I=$GOPATH/src -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=plugins=grpc,paths=source_relative:. trillian_log_api.proto trillian_admin_api.proto trillian.proto --doc_out=markdown,api.md:./docs/
+//go:generate protoc -I=. -I=$GOPATH/src -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false trillian_log_api.proto trillian_admin_api.proto trillian.proto --doc_out=markdown,api.md:./docs/
```
mhutchinson, line 17: I'll bet that this one line here took about as much time to work out as the rest of the PR.

go.mod
```diff
@@ -44,5 +44,6 @@ require (
 	google.golang.org/api v0.43.0
 	google.golang.org/genproto v0.0.0-20210331142528-b7513248f0ba
 	google.golang.org/grpc v1.36.1
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0
```
mhutchinson, line 47: It won't let me leave comments on the rest of this file, but I'm interested in why the now-removed stuff (`github.com/golang/protobuf `) is still in the mod file?

mhutchinson (approved):
Thanks! I had resigned myself to doing this today or tomorrow, so this is a great PR to start the week :-)

gen.go
pphaneuf: Very good guess! 😆

go.mod
pphaneuf: It's in the generated protobufs? Which is a bit of a mystery, but as an experiment, I tried changing them by hand, and that still left the `github.com/golang/protobuf` dependency, it only got an `// indirect`, so I figured I'd keep this bit for a second pass, as it didn't make it fully go away even if I did it...

PR #2463, opened 2021-04-12
Remove `hash_strategy` from `Tree`.

pphaneuf:
Part of #2450.

### Checklist


storage/cloudspanner/log_storage.go
```diff
@@ -158,9 +158,6 @@ func (ls *logStorage) Snapshot(ctx context.Context) (storage.ReadOnlyLogTX, erro
 }
 
 func newLogCache(tree *trillian.Tree) (*cache.SubtreeCache, error) {
-	if s := tree.HashStrategy; s != trillian.HashStrategy_RFC6962_SHA256 {
-		return nil, fmt.Errorf("unknown hash strategy: %s", s)
-	}
 	return cache.NewLogSubtreeCache(defLogStrata, rfc6962.DefaultHasher), nil
 }
```
pavelkalinnikov, line 162: Inline this?

storage/sql.go
```diff
@@ -96,11 +96,6 @@ func ReadTree(row Row) (*trillian.Tree, error) {
 	} else {
 		return nil, fmt.Errorf("unknown TreeType: %v", treeType)
 	}
-	if hs, ok := trillian.HashStrategy_value[hashStrategy]; ok {
-		tree.HashStrategy = trillian.HashStrategy(hs)
-	} else {
-		return nil, fmt.Errorf("unknown HashStrategy: %v", hashStrategy)
-	}
```
pavelkalinnikov, line 103: After deleting this code, `hashStrategy` seems not used. Would here be a good place to verify that `hashStrategy == "RFC6962_SHA256"` for `[PREORDERED_]LOG` trees? As a safe-guard for the (unlikely) case that a log with a different hasher exists somewhere.

pavelkalinnikov:
The "dumplib" test failure seems non-trivial. I wonder if it used a different hasher.

pavelkalinnikov:
Or it looks like a formatting issue.

pphaneuf:
Yeah, it seems like a formatting issue, which I find somewhat baffling?

Taking a look at your other comments now...

storage/sql.go
pphaneuf: `hashStrategy` is still somewhat necessary for the SQL query (I didn't want to fiddle with those, seemed easier and safer to still read/write them the same way), but good call on having a check to see that it matches what we expect it to be.

I'm not sure I get why the "for `[PREORDERED_]LOG` trees" bit? The only "known" `HashStrategy` is `RFC6962_SHA256`, so just generally check for that?

pavelkalinnikov: Yeah, can make the check unconditional, since all trees are expected to be logs.

pphaneuf: Done, PTAL

pphaneuf:
Increased overall coverage from 65.94% to 67.09%, but getting told off about not having enough incremental coverage? Deleting code is so thankless... 🤣

pavelkalinnikov (approved):

PR #2466, opened 2021-04-12
Remove `signature_algorithm` from `Tree`.

pphaneuf:
In a few places where signature algorithms are still mentioned, forcing ECDSA as a default (but upcoming PRs will remove those mentions).

Part of #2450.

### Checklist


pphaneuf:
A few issues with GCB, sorted out now, and ready for review!

storage/mysql/admin_storage.go
```diff
@@ -295,7 +295,7 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
 		newTree.TreeType.String(),
 		"RFC6962_SHA256",
 		newTree.HashAlgorithm.String(),
-		newTree.SignatureAlgorithm.String(),
+		"ECDSA",
```
NatalieDoduc, line 298: Do you need to do something about L48 and L272?

Additionally, how do current users deal with the schema change?

pphaneuf: This is why put the string "ECDSA" here, to avoid a schema change, but still putting in something that's valid (previous PRs have been narrowing down the selection of signature algorithms to just this one, so that this is likely to be effectively what would have been written before).

In the query, I could drop it, but I didn't want to mess about with this too much...

cmd/createtree/main.go
```diff
@@ -132,19 +125,10 @@ func newRequest() (*trillian.CreateTreeRequest, error) {
...
 		ctr.Tree.PrivateKey = pk
 	} else {
-		ctr.KeySpec = &keyspb.Specification{}
-
-		switch sigpb.DigitallySigned_SignatureAlgorithm(sa) {
-		case sigpb.DigitallySigned_ECDSA:
-			ctr.KeySpec.Params = &keyspb.Specification_EcdsaParams{
+		ctr.KeySpec = &keyspb.Specification{
+			Params: &keyspb.Specification_EcdsaParams{
 				EcdsaParams: &keyspb.Specification_ECDSA{},
```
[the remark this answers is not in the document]
pphaneuf: Done.

storage/mysql/admin_storage.go
NatalieDoduc: Perhaps leave a TODO() or at least a note on the field references, noting it's deprecated and unused for future readers. Thanks.

NatalieDoduc (commented):
LGTM - will leave to other reviewers more familiar with Trillian for Approval.

Martin2112 (approved):

storage/mysql/admin_storage.go
pphaneuf: Done.

PR #2467, opened 2021-04-12
Remove `hash_algorithm` from `Tree`.

pphaneuf:
Part of #2450.

### Checklist


Martin2112:
Wait, this means it can only be used for CT? I didn't think that was part of the scope of this.

pphaneuf:
It can be used for many things, but it can only use the RFC6962 hashing algorithm (which everything I could lay my eyes on uses).

pphaneuf:
For what it's worth, this is already true: all the non-RFC6962 hashing algorithm have been removed, and any other value will return an error anyway.

Martin2112:
@AlCutter can you comment on what the scope of the changes is meant to be?

AlCutter:
This `hash_algorithm` is just describing the hash scheme for (now removed) signatures - logically it's part of #2466 which removed the corresponding encryption scheme for the sigs.

There's the tree `hash_strategy` which defines the Merkle node hash function, but that's separate (and now that the map is gone there is currently only the the one `RFC6962_SHA256` strategy)

AlCutter (approved):

Martin2112:
OK I misunderstood then. It looked from the comments above like it was affecting tree hashing.

pphaneuf:
Ah, yes, I was confusing the two as well, in that chain of PRs! 😄 

This here is the hashing *algorithm* "used" in the signature, and the hash *strategy* was removed in other PRs.

Martin2112:
So we have removed the means to implement non-CT hashing? I don't think that was necessary.

PR #2468, opened 2021-04-12
Remove `public_key` from `Tree`.

pphaneuf:
Part of #2450.

### Checklist


storage/mysql/admin_storage.go
```diff
@@ -301,7 +301,7 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
 		nowMillis,
 		nowMillis,
 		privateKey,
-		newTree.PublicKey.GetDer(),
+		[]byte{}, // Unused, filling in for backward compatibility.
```
pavelkalinnikov, line 304: Should this and other constants be moved directly to the SQL statement, instead of being added as parameters?

pavelkalinnikov (approved):

storage/mysql/admin_storage.go
pphaneuf: I did something similar to this in the previous PRs, where I'm trying to touch this as little as possible, and I'd try a separate PR to tweak the SQL statements?

pavelkalinnikov: Yep, separate PR sounds good.

PR #2469, opened 2021-04-12
Remove `private_key` from `Tree`, and `key_spec` from `CreateTreeRequest`.

pphaneuf:
Part of #2450.

### Checklist


cmd/createtree/pkcs11.go
```diff
@@ -1,64 +0,0 @@
-// +build pkcs11
```
pavelkalinnikov, line 1: Are there any more files left with this build tag?

There is a CI testing target using it. Should it be removed too?

pphaneuf: There's very little: there's the packages in `crypto/keys/pkcs11`, which are still imported by the mains.

I intend on cleaning that up separately, especially as this will impact the CI (and that can have its own complications!)...

pphaneuf: Oh, and there's a possibility that this is still used by CTFE!

storage/tree_validation_test.go
```diff
@@ -26,29 +26,12 @@ import (
...
 	"google.golang.org/protobuf/types/known/timestamppb"
 
-	ktestonly "github.com/google/trillian/crypto/keys/testonly"
-
 	_ "github.com/google/trillian/crypto/keys/der/proto"
 	_ "github.com/google/trillian/crypto/keys/pem/proto"
 )
 
-const (
-	privateKeyPath = "../testdata/log-rpc-server.privkey.pem"
```
pavelkalinnikov, line 39: Is this file still used? Remove it too?

pavelkalinnikov (approved):

storage/tree_validation_test.go
pphaneuf: Yes, still used in some `crypto/keys/pem` tests.

PR #2478, opened 2021-05-10
adding VDS paper as markdown

smeiklej:
### Checklist

### DIFF
--- a/docs/VerifiableDataStructures.md
+++ b/docs/VerifiableDataStructures.md
@@ -14,7 +14,9 @@ site­ owners trust Certificate Authorities to issue certificates, mail clients
 may trust a key server to return the right public key for a recipient they wish
 to mail to, server administrators trust their package distributors to send them
 binaries for their software. In the real world, citizens trust land registries
-to accurately record who owns which land. This paper describes a number of data
+to accurately record who owns which land.  
+
+This paper describes a number of data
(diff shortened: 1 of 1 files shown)


AlCutter (approved):

getagit:
Do you want to delete the PDF version in the papers/ subdir (or maybe move the two versions closer, or something)?

pphaneuf:
Since git is itself a "blockchain", the PDF will not disappear with anything less than a hard fork, even if you "delete" it. 😃 

What you'll get is potentially less confusion, though...

smeiklej:
Sure, but the link to it will go away/change and there are numerous places on the internet that use the current one (and will not change it), so I have a strong preference for keeping it where it is and just making sure we update all our own pointers.

getagit:
Agree that the old version is linked all over the place, so break without forwarding is sub-optimal.  But for people coming to the Trillian repo now, they see two versions of "VerifiableDataStructures", in different formats, in different directories, and (soon) with different content and no obvious indicator of what the difference is or which one to use, so it looks more a bug.  

Maybe the simplest is to rename the new version to "-2021"  (or "latest" or something to indicate that its the "living" version and not just the old paper in a new format)?

smeiklej:
Okay, makes sense to rename the Markdown one - can do that once I start committing the changes we plan to make.

PR #2481, opened 2021-05-12
Remove some references to the `crypto` directory.

pphaneuf:
Part of #2450.

### Checklist

### DIFF
--- a/storage/mysql/admin_storage_test.go
+++ b/storage/mysql/admin_storage_test.go
@@ -21,7 +21,6 @@ import (
 	"testing"
 
 	"github.com/google/trillian"
-	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/storage"
 	"github.com/google/trillian/storage/testonly"
 	"google.golang.org/protobuf/proto"
@@ -151,7 +150,7 @@ func TestAdminTX_StorageSettingsNotSupported(t *testing.T) {
 	s := NewAdminStorage(DB)
 	ctx := context.Background()
 
-	settings, err := anypb.New(&keyspb.PEMKeyFile{})
+	settings, err := anypb.New(&trillian.Tree{})
 	if err != nil {
 		t.Fatalf("Error marshaling proto: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


storage/mysql/admin_storage_test.go
```diff
@@ -151,7 +150,7 @@ func TestAdminTX_StorageSettingsNotSupported(t *testing.T) {
 	s := NewAdminStorage(DB)
 	ctx := context.Background()
 
-	settings, err := anypb.New(&keyspb.PEMKeyFile{})
+	settings, err := anypb.New(&trillian.Tree{})
```
NatalieDoduc, line 153: this is an interesting substitution. have not looked at the detailed implementations, though.

NatalieDoduc (commented):
LGTM, a few educational comments below.

I'm curious what is left of the `crypto` references. Could you detail your commit description to include which areas `crypto` has been removed from?

storage/mysql/admin_storage_test.go
pphaneuf: All that this wants is a valid non-nil "any" protobuf, the content is not actually looked at. The name of the field is `storage_settings`, so putting in a `Tree` is no less "reasonable" than putting in a `PEMKeyFile` (they're both equally nonsensical!). 😃

pphaneuf:
I think that the references left within Trillian are only within `crypto` itself:

```
$ grep -rlI trillian/crypto .
./crypto/sigpb/sigpb.proto
./crypto/keyspb/keyspb.proto
./crypto/keys/handlers_test.go
./crypto/keys/generate.go
./crypto/keys/pem/proto/register_test.go
./crypto/keys/pem/proto/register.go
./crypto/keys/pem/pem.go
./crypto/keys/pem/pem_test.go
./crypto/keys/der/proto/register_test.go
./crypto/keys/der/proto/register.go
./crypto/keys/der/der_test.go
./crypto/keys/der/der.go
./crypto/keys/testonly/keys.go
./crypto/keys/generate_test.go
./crypto/keys/pkcs11/proto/register.go
./crypto/keys/pkcs11/pkcs11.go
```

AlCutter (approved):

storage/mysql/admin_storage_test.go
NatalieDoduc: Yeah, I was wondering. The comment in tree_validation_test.go made me 71% sure this could be anything. Also the `anypb.New()`.

NatalieDoduc (approved):

PR #2487, opened 2021-05-14
Remove generate.go and put the rest of it in better places.

pphaneuf:
### Checklist


crypto/keys/pem/pem_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	. "github.com/google/trillian/crypto/keys/pem"
 	ktestonly "github.com/google/trillian/crypto/keys/testonly"
 	"github.com/google/trillian/testonly"
+
+	_ "github.com/golang/glog"
```
NatalieDoduc, line 25: why this change?

NatalieDoduc (commented):
Could you please amend the commit description to say "keys/generate.go"?

I forget what it is at the moment (proto generation maybe?), but i think we also have more generic "generate.go" files and at first I was thinking you were removing that.

crypto/keys/pem/pem_test.go
pphaneuf: Some of those tests seem to be run by hand by the GCB scripts (in addition to being run via `go test`), and get passed some glog flags, which are not recognised without this.

This was previously pulled indirectly, somehow (not obvious to me in this PR, but there's been much code deleted), it was found via GCB failures and going through the error logs.

NatalieDoduc (approved):
thanks for the clarifications!

PR #2492, opened 2021-05-17
Initialise sequencer metrics in a separate function.

pphaneuf:
### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -72,25 +72,29 @@ func quotaIncreaseFactor() float64 {
 	return QuotaIncreaseFactor
 }
 
+// InitMetrics sets up some metrics for this package. Must be called before calling IntegrateBatch.
+// Can be called more than once, but only the first call has any effect.
 // TODO(pavelkalinnikov): Create all metrics in this package together.
-func createSequencerMetrics(mf monitoring.MetricFactory) {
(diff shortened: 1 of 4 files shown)


log/sequencer.go
```diff
@@ -72,25 +72,29 @@ func quotaIncreaseFactor() float64 {
 	return QuotaIncreaseFactor
 }
 
+// InitMetrics sets up the metrics for this package. Must be called before calling IntegrateBatch.
+// Can be called more than once, but only the first call has any effect.
```
NatalieDoduc, line 76: does something still need to be done for the TODO()? 

You're calling your function InitMetrics ), just wanted to check it's not InitJustSomeMetricsButNotAll?

NatalieDoduc (approved):

log/sequencer.go
pphaneuf: I'm not sure if it is, to be honest? I've done nothing to fix that TODO, and it seems like a reasonable TODO to me.

This is likely to require more PRs to fully address...

NatalieDoduc: maybe s/the metrics/some metrics/ to avoid confusion.

pphaneuf: Done

PR #2500, opened 2021-05-19
Simplify sequencer

pphaneuf:
### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -97,19 +97,6 @@ func InitMetrics(mf monitoring.MetricFactory) {
 	})
 }
 
-// Sequencer instances are responsible for integrating new leaves into a single log.
-// Leaves will be assigned unique sequence numbers when they are processed.
-// There is no strong ordering guarantee but in general entries will be processed
-// in order of submission to the log.
-type Sequencer struct {
-	qm quota.Manager
-}
-
-// NewSequencer creates a new Sequencer instance for the specified inputs.
-func NewSequencer(qm quota.Manager) *Sequencer {
-	return &Sequencer{qm: qm}
-}
-
 // initCompactRangeFromStorage builds a compact range that matches the latest
 // data in the database. Ensures that the root hash matches the passed in root.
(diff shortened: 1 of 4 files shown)


log/sequencer.go
```diff
@@ -292,17 +270,17 @@ func (s *preorderedLogSequencingTask) update(ctx context.Context, leaves []*tril
 
 // IntegrateBatch wraps up all the operations needed to take a batch of queued
 // or sequenced leaves and integrate them into the tree.
-func (s Sequencer) IntegrateBatch(ctx context.Context, tree *trillian.Tree, limit int, guardWindow, maxRootDurationInterval time.Duration) (int, error) {
-	start := s.timeSource.Now()
+func IntegrateBatch(ctx context.Context, tree *trillian.Tree, limit int, guardWindow, maxRootDurationInterval time.Duration, timeSource clock.TimeSource, logStorage storage.LogStorage, qm quota.Manager) (int, error) {
```
pavelkalinnikov, line 273: Avoid name repetitions? Suggestion: `ts clock.TimeSource, ls storage.LogStorage`.

pavelkalinnikov (commented):
Is this just a drive-by, or you have some bigger refactoring going? Honestly, I don't think small improvements can make this code significantly better, it needs a complete redesign.

This particular PR might be good, but feels like switching from C++ to C. I don't particularly like functions with many arguments of all kinds.

pavelkalinnikov (approved):
OTOH, currently this object is used "stateless", i.e. it's created, then the method is called once, and the object is forgotten. It makes some sense to just squash this whole thing to one function.

Approved modulo naming suggestion.

log/sequencer.go
pphaneuf: Done.

pphaneuf:
Exactly, and having it as a struct with a method misleads someone reading the code into thinking that there might be cases where different code paths can happen.

This is some of the most mission-critical code in Trillian, "misleading" is the last thing we want!

PR #2501, opened 2021-05-19
Avoid and remove the pem.UnmarshalPublicKey function.

pphaneuf:
Part of #2450.

### Checklist

### DIFF
--- a/crypto/keys/pkcs11/pkcs11.go
+++ b/crypto/keys/pkcs11/pkcs11.go
@@ -18,10 +18,11 @@ package pkcs11
 
 import (
 	"crypto"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 
-	"github.com/google/trillian/crypto/keys/pem"
 	"github.com/google/trillian/crypto/keyspb"
 
 	pkcs11key "github.com/letsencrypt/pkcs11key/v4"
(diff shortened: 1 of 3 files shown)


crypto/keys/pkcs11/pkcs11.go
```diff
@@ -34,7 +35,8 @@ func FromConfig(modulePath string, config *keyspb.PKCS11Config) (crypto.Signer,
 	}
 
 	pubKeyPEM := config.GetPublicKey()
-	pubKey, err := pem.UnmarshalPublicKey(pubKeyPEM)
+	block, _ := pem.Decode([]byte(pubKeyPEM))
```
pavelkalinnikov, line 38: Is it okay that the new code doesn't check `len(rest) == 0`, whereas the old one did?

pphaneuf: Any input file that worked with the old code will also work with the new code, but the old code would give an error if there was extra data after the PEM block (say, a description, or potentially even another PEM block with another key/certificate, like what is done in a roots file).

I can imagine that checking for that might help with some very specific incorrect usage, like passing a roots file instead of a private key? But it doesn't seem like a problem to ignore it, what do you think?

pavelkalinnikov: Seems fine to ignore, but I'm slightly in favour of preserving the old behaviour in this line. It's fine to panic in the `testonly` thing below.

pphaneuf: How does that look to you?

pavelkalinnikov (approved):

crypto/keys/pkcs11/pkcs11.go
pavelkalinnikov: Beautiful.

PR #2511, opened 2021-05-24
Split out the function to create a crypto.Signer from a PrivateKey proto.

pphaneuf:
This allows it to be reused, and also tests it more directly.

### Checklist

### DIFF
--- a/crypto/keys/der/proto/register.go
+++ b/crypto/keys/der/proto/register.go
@@ -17,21 +17,11 @@
 package proto
 
 import (
-	"context"
-	"crypto"
-	"fmt"
-
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/crypto/keys/der"
 	"github.com/google/trillian/crypto/keyspb"
-	"google.golang.org/protobuf/proto"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


crypto/keys/der/proto/register.go
```diff
@@ -17,21 +17,11 @@
...
 )
 
 func init() {
-	keys.RegisterHandler(&keyspb.PrivateKey{}, func(ctx context.Context, pb proto.Message) (crypto.Signer, error) {
-		if pb, ok := pb.(*keyspb.PrivateKey); ok {
-			return der.FromProto(pb)
-		}
-		return nil, fmt.Errorf("der: got %T, want *keyspb.PrivateKey", pb)
-	})
+	keys.RegisterHandler(&keyspb.PrivateKey{}, der.FromProto)
```
cindylindeed, line 26: Is the unused ctx passed in FromProto forced by RegisterHandler here?

cindylindeed (approved):

crypto/keys/der/proto/register.go
pphaneuf: That's right, it has to match that function signature.

PR #2512, opened 2021-05-26
Always provide a pkcs11.FromConfig function.

pphaneuf:
Instead of having it only exist when PKCS11 support is present, have this function always available, but return an error when PKCS11 support is not compiled in.

### Checklist

### DIFF
--- a/crypto/keys/pkcs11/no_pkcs11.go
+++ b/crypto/keys/pkcs11/no_pkcs11.go
@@ -1,3 +1,5 @@
+// +build !pkcs11
+
 // Copyright 2017 Google LLC. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


crypto/keys/pkcs11/no_pkcs11.go
```diff
@@ -0,0 +1,29 @@
+// +build !pkcs11
```
pavelkalinnikov, line 1: Do we need this build flag at all? Maybe make it unconditional?

What is the cost? Extra package import?

pavelkalinnikov (approved):
Feels ok. Does this not bring any heavy dependencies in places where you removed pkcs11 build flag?

crypto/keys/pkcs11/no_pkcs11.go
pphaneuf: There's *two* `pkcs11.FromProto` defined, one in `pkcs11.go`, the other in `no_pkcs11.go`, and they should never "meet". 😃

pphaneuf:
The `register.go` is now compiled unconditionally, but it doesn't really bring anything, the heavy dependency is in the `pkcs11` package.

pavelkalinnikov:
The `register.go` file imports `github.com/google/trillian/crypto/keys/pkcs11`, which imports `github.com/letsencrypt/pkcs11key/v4`. Is the latter thing heavy? I see that it's the same package that `pkcs11` imports.

pphaneuf:
That's right, but it only imports it in the file marked with `// +build pkcs11`.

PR #2525, opened 2021-06-03
Move and rename map node ID type

pavelkalinnikov:
The node ID type logically belongs to the `merkle` subfolder which holds the algorithmic
aspect of Merkle trees. This change moves the `NodeID2` type, which is now known simply
as `node.ID`. It also untangles some dependencies that blocked this change.

#1855
#2378


pavelkalinnikov:
Review commit-by-commit might be convenient.

pavelkalinnikov:
It's better to review #2526 first. This PR will become much simpler after rebasing.

pavelkalinnikov:
@pphaneuf This is ready for review now. It's a big but simple one: in a nutshell, just a move + rename + update usages.

pavelkalinnikov:
@mhutchinson PTAL too, it affects `batchmap`.

merkle/coniks/coniks.go
```diff
@@ -107,27 +107,27 @@ func (m *Hasher) BitLen() int {
...
-// writeMaskedNodeID writes the node ID bits to the buffer, padded with zero
-// bits to the byte Size() of the hashes in use by this hasher.
+// writeMaskedID writes the node ID bits to the buffer, padded with zero bits
+// to the byte Size() of the hashes in use by this hasher.
 //
 // TODO(pavelkalinnikov): We must not use BitLen() and Size() interchangeably.
 // The tree height and hash size could be different.
 // TODO(pavelkalinnikov): Padding with zeroes doesn't buy us anything, as the
 // depth is also written to the Buffer.
-func (m *Hasher) writeMaskedNodeID(b *bytes.Buffer, node tree.NodeID2) {
```
mhutchinson, line 117: opinion: I preferred the previous method name. In normal conversation one would still refered to `node.ID` as a "node ID".

mhutchinson (approved):

merkle/coniks/coniks.go
pavelkalinnikov: Fair point, restored the original name.

PR #2526, opened 2021-06-03
Factor out Layout for sparse Merkle trees

pavelkalinnikov:
The old `Layout` type was made to cover both logs and maps. This change factors out the
`smt`-specific part of `Layout` into a separate type.

#2378

### DIFF
--- a/merkle/smt/layout.go
+++ b/merkle/smt/layout.go
@@ -27,13 +27,13 @@ type Layout struct {
 }
 
 // NewLayout creates a tree layout based on the passed-in tier heights.
-func NewLayout(heights []int) Layout {
+func NewLayout(heights []uint) Layout {
 	depths := make([]int, len(heights)+1)
 	for i, h := range heights {
 		if h <= 0 {
 			panic("tile heights must be positive")
 		}
-		depths[i+1] = depths[i] + h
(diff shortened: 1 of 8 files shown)


merkle/smt/layout.go
```diff
@@ -0,0 +1,56 @@
...
+	depths []int
+}
+
+// NewLayout creates a tree layout based on the passed-in tier heights.
+func NewLayout(heights []int) Layout {
+	depths := make([]int, len(heights)+1)
+	for i, h := range heights {
+		if h <= 0 {
+			panic("tile heights must be positive")
+		}
```
mhutchinson, line 35: We use `uint` below, so I'd just suck this up and make the inputs to this `uint` too so that it's properly compile-time checked and thus avoid this panic.

mhutchinson (approved):

merkle/smt/layout.go
pavelkalinnikov: Yeah, I did this first, but note that we should still check != 0. Also, there is no `SearchInts` for uints, so I kinda chose int to avoid another cast. But that is an implementation details, and from the "API" point of view `uint` indeed makes more sense - fixed.

pphaneuf (approved):

PR #2532, opened 2021-06-09
Simplify log Layout type

pavelkalinnikov:
This change removes the unused logic for dealing with flexible log tree layouts. Currently,
all trees have 8 strata of height 8.


storage/cache/subtree_cache.go
```diff
@@ -373,7 +355,7 @@ func (s *SubtreeCache) Flush(ctx context.Context, setSubtrees SetSubtreesFunc) e
 
 // newEmptySubtree creates an empty subtree for the passed-in ID.
 func (s *SubtreeCache) newEmptySubtree(id []byte) *storagepb.SubtreeProto {
-	height := s.layout.TileHeight(len(id) * 8)
+	const height = 8
```
pphaneuf, line 358: I'd be inclined to skip this here, not mention "with height %d" in the log message, and just inline it when setting the field below?

pphaneuf (approved):

storage/cache/subtree_cache.go

PR #2546, opened 2021-06-17
Comment LogRootV1 fields properly

pavelkalinnikov:
Deprecation syntax: https://github.com/golang/go/wiki/Deprecated

### DIFF
--- a/types/logroot.go
+++ b/types/logroot.go
@@ -38,8 +38,9 @@ type LogRootV1 struct {
 	TreeSize       uint64
 	RootHash       []byte `tls:"minlen:0,maxlen:128"`
 	TimestampNanos uint64
-	Revision       uint64 // Deprecated.
-	Metadata       []byte `tls:"minlen:0,maxlen:65535"`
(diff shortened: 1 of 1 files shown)


types/logroot.go
```diff
@@ -35,11 +35,21 @@ import (
...
 	TimestampNanos uint64
-	Revision       uint64 // Deprecated.
-	Metadata       []byte `tls:"minlen:0,maxlen:65535"`
+
+	// Revision is the Merkle tree revision associated with this root.
+	//
+	// Deprecated: Revision is a concept internal to the storage layer.
+	Revision uint64
+
+	// Metadata holds additional data attached to this root.
```
AlCutter, line 51: "associated with"?

AlCutter (approved):

types/logroot.go
pavelkalinnikov: Done.

NatalieDoduc (approved):
nice, thanks for this!

PR #2551, opened 2021-06-18
Roll back #2536, to fix generated code.

pphaneuf:
### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	bitbucket.org/creachadair/shell v0.0.6
 	cloud.google.com/go/spanner v1.20.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.13.8
-	github.com/apache/beam v2.30.0+incompatible
+	github.com/apache/beam v2.29.0+incompatible
 	github.com/fullstorydev/grpcurl v1.8.1
 	github.com/go-redis/redis v6.15.9+incompatible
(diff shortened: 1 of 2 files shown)


mhutchinson:
Any context on the breakage? I'm checking this locally now, but it's surprising the presubmit tests passed on the original PR if this doesn't work.

mhutchinson:
Ah, I'm seeing it now:

```
experimental/batchmap/batchmap.shims.go:28:2: no required module provides package github.com/apache/beam/sdks/go/pkg/beam/core/graphx/schema; to add it:
	go get github.com/apache/beam/sdks/go/pkg/beam/core/graphx/schema
```

mhutchinson (approved):

pavelkalinnikov (approved):
If you run `go get` from the error message, does it not fix the problem?

mhutchinson:
It looks like `runtime` is missing from that generated package import path:

https://github.com/apache/beam/tree/master/sdks/go/pkg/beam/core/runtime/graphx

I've done this locally with the latest tip of tree and the generated code resolves correctly. go-beam isn't releasing with semver, though git tags are cut for java-beam and python-beam which is why there are "releases". I think we should merge this PR to roll back, and then either wait for the next minor version, or perform ad-hoc updates. I'd prefer the dependabot updates, as it's at least regular (even if the versions it proposes are basically arbitrary for the go code).

PR #2567, opened 2021-07-05
Remove RootHash field of SubtreeProto

pavelkalinnikov:
This field is unused.


storage/storagepb/storage.proto
```diff
@@ -27,6 +27,7 @@ message SubtreeProto {
   // subtree's depth
   int32 depth = 2;
 
+  // Deprecated: Stored the root hash of the tile, but was not used.
```
pphaneuf, line 30: Is there a reason not to make this disappear (and use `reserved`)?

pavelkalinnikov: This field was stored, but its value wasn't used. I'm not sure what parser will do when it encounters this field, will it ignore it if I replace this with `reserved`?

Martin2112:
This was put in to detect bit flips etc. I doubt we'll use it now though.

Martin2112 (approved):

storage/storagepb/storage.proto
pavelkalinnikov: I switched to what appears a proper deprecation mechanism for protos. PTAL.

pphaneuf: For binary protos, unknown fields are simply ignored. The parser does put them aside, so that if the proto is marshaled again, the unknown fields will be maintained, and there's some API to access the unknown fields.

Using `reserved` has no effect on the generated code, it is only to avoid re-using the field name and number in the future (it will cause a compile-time error). Reusing a field name in the future is likely to be with some different semantics, and reusing a field number is almost assuredly incompatible, and using `reserved` gives some protection against a new binary encountering old protos with the previous meaning (in a database, say) and misinterpreting them.

Using the `deprecated` attribute is helpful if there is still code reading from the field, so that no further uses get added while that code is being fixed/migrated.

But if there are no usage left, why not just delete the field?

pavelkalinnikov: Deleting SGTM. Done.

PR #305, opened 2017-01-19
Add rehashing support for inclusion + consistency proofs.

Martin2112:
This allows us to serve proofs at arbitrary tree sizes. Snapshot recomputation annotates portions of the path that require rehashing and this is then used to recalculate a subtree node when storage returns the nodes. Add some proof tests at the node / path calculation level to improve coverage. If proofs are at STH corresponding sizes the new code paths are not taken so probably safe! 

Sorry this PR is a bit big but it was proving tricky to separate out.

PR #2568, opened 2021-07-05
Allow not storing ephemeral node hashes

pavelkalinnikov:
This change adds `--tree_ids_with_no_ephemeral_nodes` flag to the sequencer,
which allows disabling the writes of ephemeral nodes for a subset of Trillian
trees.  The plan is to disable it unconditionally in the next release, but the
flag helps to test the new behaviour safely on a subset of trees.

The stored ephemeral nodes are not used because all the higher level logic uses
compact ranges, operating only with the "perfect" nodes. Since commit df47465 on
Jul 13, 2021 ephemeral nodes are never read from the tree storage.

Not storing ephemeral nodes also has a positive performance effect. Previously,
every sequencing transaction for a typical tree of hundreds of millions entries
would write at least 4 tiles at the tree edge, because it "touched" all the
ephemeral nodes.  Now, it will only touch perfect nodes, which for a typical
transaction are in the bottom 1-2 tiles.

Relatedly, now that only non-ephemeral node updates trigger a tile write, every
tile can be written at most 256 times, whereas previously tiles could be written
millions of times (e.g. the root tile of the tree). It led to issues like #1188
which is the result of needing to scan / query millions of rows.

### Checklist


pavelkalinnikov:
Checking whether it's legit.

Martin2112:
This didn't work before compact ranges and the problems only showed up at particular tree sizes. Need to be sure it's well tested.

Martin2112:
Also when the issue occurred it wasn't always using the latest version, among other changes.

pavelkalinnikov:
It's covered by tests already (and they pass). I factored out #2579 which only removes the reading of ephemeral nodes, and the tests pass. Since not reading them is ok, there is no point in writing them, hence this PR.

AlCutter (approved):
Just wondering about release strategy for this set of changes - perhaps this is large enough to merit a bump to `1.4`, and maybe there should even be a `1.4.0` with the _"don't read ephemeral nodes"_ change and then a `1.4.1` with this _"don't write ephemeral nodes"_ included so that folks can be happy that they're good on `1.4.0` before trying `1.4.1`, because there'll be no roll-back past the _"don't read ephemeral nodes"_ version in case of trouble?

pavelkalinnikov:
Yes, there can be special effects depending on how this is rolled out. I'll leave this unmerged for now and see how far I can get without it. The 2 separate releases SGTM.

pavelkalinnikov:
I made more research on the effect of this change, and it looks like submitting this PR is backwards compatible up to #305 [Jan 2017, `v0.6`], when there was an introduction of rehashing for fetching proofs at arbitrary tree sizes. This is when ephemeral nodes stopped being read from tree storage, and writing them became unnecessary. It is safe to remove the writing now.

pavelkalinnikov:
I made more research :) It looks like the last commit that stopped reading ephemeral nodes was df47465, which was only released in `v1.4.0`. Still, @AlCutter's comment applies: there is a release that stops reading, the next release can stop writing.

pavelkalinnikov:
If this update is rolled out after df47465 (or `v1.4.0`) has been rolled out, there won't be any issues.

If this update is rolled out with skipping df47465 (or `v1.4.0`), the worst case effect would be a temporary inability to serve some proofs correctly (while the rollout is ongoing). Specifically, the conditions that would trigger it:

1. A new (rolled out) node completed a sequencing run.
2. An old node receives an (inclusion|consistency) proof request.
3. The requested tree size equals the current tree size stored in Trillian.

The amount of time in which this can happen is limited by:

- The rollout time.
- The period of time while the requested tree size equals the current tree size. Trillian updates the tree frequently (multiple times per second), so it's likely that the request will return the correct result after a couple of retries.

If this effect is completely undesirable, there is a way to avoid it:

1. Roll out `logserver` first, so that all the reading of ephemeral nodes disappears.
2. Roll out `logsigner` second, so that all the writes are compatible with all the reads.

pavelkalinnikov:
@AlCutter @mhutchinson I ended up with a safer approach. This PR now introduces a flag that can enable the new behaviour only for certain trees. We can roll this out for a few trees first, and once made sure it works, fully enable it.

So, I propose to cut a release `v1.4.2` with this flag now, and later cut `v1.5.0` with this flag removed, and the new behaviour enabled by default.

WDYT?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## HEAD
 
+* #2568: Allow disabling the writes of ephemeral nodes to storage.
```
mhutchinson, line 5:
```suggestion
* #2568: Allow disabling the writes of ephemeral nodes to storage via `tree_ids_with_no_ephemeral_nodes` flag to the sequencer.
```

mhutchinson (approved):

CHANGELOG.md
pavelkalinnikov: Done.

PR #2572, opened 2021-07-06
merkle: Use only complete subtree nodes for proofs

pavelkalinnikov:
This change makes proof construction functions always fetch non-ephemeral
nodes unconditionally. Previously it sometimes could fetch ephemeral nodes
(when the requested tree size is equal to the stored size). Not fetching them
allows not storing them, which enables storage savings and the removal of the
revisions concept.

### Checklist


pavelkalinnikov:
I recommend commit-by-commit review.

AlCutter (approved):

PR #2576, opened 2021-07-14
merkle: Refactor proof generation functions

pavelkalinnikov:
This change introduces the `merkle/proof` package which helps to generate
inclusion and consistency proofs. The new `proof.Nodes` type replaces the
old `[]merkle.NodeFetch` and the `Rehash` function.

Key improvements:

- Uses `[]compact.NodeID` natively, no need to convert from `[]merkle.NodeFetch`
   up the stack.
- Uses only compact range nodes natively.
- Stricter handling of ephemeral nodes: `proof.Nodes` allows only one whereas a
   `[]merkle.NodeFetch` could have arbitrary runs of rehashes.
- Concise naming: `proof.Inclusion` and `proof.Consistency` return `proof.Nodes`.
- Using `uint64` type for leaf index and tree size throughout the package results in
   a simpler error handling.
- Shorter code.

### Checklist


pavelkalinnikov:
Review commit-by-commit might be the easiest. Note that some commits override previous ones though, but might be easier to follow the evolution of the code this way.

pavelkalinnikov:
PTAL

pavelkalinnikov:
@AlCutter @Martin2112 @pphaneuf PTAL. Let me know if this is difficult to review / there is a good way to split it.

pavelkalinnikov:
Factored out a couple of minor commits to #2578.

merkle/proof/proof.go
```diff
@@ -0,0 +1,175 @@
...
+		return Nodes{}, fmt.Errorf("index %d out of bounds for tree size %d", index, size)
+	}
+	return nodes(index, 0, size), nil
+}
+
+// Consistency returns the information on how to fetch and construct a
+// consistency proof between the two given tree sizes of a log Merkle tree.
+func Consistency(size1, size2 uint64) (Nodes, error) {
+	if size1 > size2 {
+		return Nodes{}, fmt.Errorf("tree sizes %d and %d are mis-ordered", size1, size2)
```
AlCutter, line 57: maybe just say `tree size1 (%d) >  size2 (%d)` ?

AlCutter:
Is it worth leaving some deprecated forwarders in `merkle/` or perhaps at least a `README` with how to update old code to use the new `proof` package?

I think there are a number of folks using this package directly (https://github.com/search?q=trillian%2Fmerkle&type=code)

pavelkalinnikov:
This specific bit of the package is not used widely:
https://github.com/search?q=CalcInclusionProofNodeAddresses&type=code (unused; ccccccccjl/google seems to be a copy/fork)
https://github.com/search?q=CalcConsistencyProofNodeAddresses&type=code (an outdated version is used only by letsencrypt/ct-woodpecker)

At this point in time these funcs are not intended for public use. Although I can add a note to the changelog on how to "migrate".

merkle/proof/proof.go
pavelkalinnikov: Addressed in the new `merkle` repository.

pavelkalinnikov:
This change will be made in the new https://github.com/transparency-dev/merkle repo.

PR #2608, opened 2021-08-31
Cloudbuild: Lock the golang tooling to a minor version

mhutchinson:
Larger deltas in go tooling and expectations occur on minor version changes, which can cause problems for presubmits. Specifically, 1.16 to 1.17 has a change to how go:build lines are preferred, which is introducing diffs into PRs that are orthogonal. Locking this down means we now have an explicit decision about when we jump and address these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## HEAD
 
+* Recommended go version for development: 1.16x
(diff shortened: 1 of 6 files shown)


pavelkalinnikov:
Please add a note to CHANGELOG, and potentially to main README (now it says that one needs Go 1.14 or above).

pavelkalinnikov (approved):

PR #2609, opened 2021-08-31
Added the go:build lines that go1.17 prefers

mhutchinson:
This is related to #2608. After this change we should be able to bump the tooling to prefer go1.17 without diffs being generated in cloudbuild.

### DIFF
--- a/crypto/keys/pkcs11/no_pkcs11.go
+++ b/crypto/keys/pkcs11/no_pkcs11.go
@@ -1,3 +1,4 @@
+//go:build !pkcs11
 // +build !pkcs11
 
 // Copyright 2017 Google LLC. All Rights Reserved.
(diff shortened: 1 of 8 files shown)


crypto/keys/pkcs11/no_pkcs11.go
```diff
@@ -1,3 +1,4 @@
+//go:build !pkcs11
```
pavelkalinnikov, line 1: Are we keeping both formats only for compatibility reasons, or the new tooling wants both lines?

pavelkalinnikov (approved):

crypto/keys/pkcs11/no_pkcs11.go
mhutchinson: This comment didn't appear until I pressed merge. What fun. Luckily, the answer is that the new tooling wants both lines. The old one will be deprecated in some future version of go.

pavelkalinnikov: SGTM.

PR #2636, opened 2021-11-09
Use the extracted merkle library

mhutchinson:
The merkle library in this repo can be marked as deprecated after this. New clients would be better served by using the smaller module, especially for client-side code which will not need all the server-side dependencies that this repository has collected.

### Checklist


AlCutter:
Looks like a `goimports` might be needed

AlCutter (commented):
You still have references to `github.com/google/trillian/merkle/hashers` in here - the `LogHasher` i/f in that package is now/also in `github.com/transparency-dev/merkle`

AlCutter:
`github.com/google/trillian/merkle/logverifier` structs & funcs are now/also in the `github.com/transparency-dev/merkle` package too

AlCutter:
That last commit seems to have reverted a bunch of references back to the trillian `merkle`, was that intentional?

mhutchinson:
Yep, I figured the various merkle packages here (`merkle`, `internal/merkle` etc) should stay self-contained. If we break that self-containment we should probably do it with aliases to create forward references to the new locations, but that's a big job.

AlCutter:
> Yep, I figured the various merkle packages here (`merkle`, `internal/merkle` etc) should stay self-contained. If we break that self-containment we should probably do it with aliases to create forward references to the new locations, but that's a big job.

Ah, yup, makes sense, ta

AlCutter (approved):

PR #2651, opened 2022-01-11
Make countFromInformationSchema function compatible with MySQL 8

px3303:
Turn off statistics caching for MySQL 8.
To always retrieve the latest statistics directly from the storage engine and bypass cached values, set information_schema_stats_expiry to 0.
See https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry

### Checklist


px3303:
Moved to #2652

PR #2652, opened 2022-01-11
Query information_schema compatible with MySQL 8

### LINKED ISSUE #2653: Not compatible with MySQL 8.0
MySQL 8.0 has introduced cache to improve the efficiency of information_schema queries, so it may lead to a consequence that users could not get the latest and accurate statistics when querying information_schema.

If using MySQL 8.0 as storage layer, trillian codes, without setting the session variable `information_schema_stats_expiry` (the default is 86400 seconds), could not pass the `TestMySQLRateLimiting` and `TestQuotaManager_GetTokens_InformationSchema`. 
```
--- FAIL: TestMySQLRateLimiting (1.13s)
    quota_test.go:100: Test MySQL available at "root@tcp(127.0.0.1)/"
    quota_test.go:124: timed out before rate limiting kicked in
FAIL
FAIL    github.com/google/trillian/integration/quota    8.509s
```

```
--- FAIL: TestQuotaManager_GetTokens_InformationSchema (1.44s)
    mysql_quota_test.go:118: Test MySQL available at "root@tcp(127.0.0.1)/"
    --- FAIL: TestQuotaManager_GetTokens_InformationSchema/useSelectCount_=_false (1.34s)
        mysql_quota_test.go:170: timed out
FAIL
FAIL    github.com/google/trillian/quota/mysqlqm        4.407s
```

The countFromInformationSchema function in the codes depends on information_schema to return the latest statistics,
```golang
	countFromInformationSchemaQuery = `
		SELECT table_rows
		FROM information_schema.tables
		WHERE table_schema = schema()
			AND table_name = ?
			AND table_type = ?`
```

```golang
func countFromInformationSchema(ctx context.Context, db *sql.DB) (int, error) {
	// information_schema.tables doesn't have an explicit PK, so let's play it safe and ensure
	// the cursor returns a single row.
	rows, err := db.QueryContext(ctx, countFromInformationSchemaQuery, "Unsequenced", "BASE TABLE")
	if err != nil {
		return 0, err
	}
	defer rows.Close()
	if !rows.Next() {
		return 0, errors.New("cursor has no rows after information_schema query")
	}
	var count int
	if err := rows.Scan(&count); err != nil {
		return 0, err
	}
	if rows.Next() {
		return 0, errors.New("too many rows returned from information_schema query")
	}
	return count, nil
}
```

so anything that relies on this function may fail to work as normal.

px3303:
Turn off statistics caching for MySQL 8.
To always retrieve the latest statistics directly from the storage engine and bypass cached values, set information_schema_stats_expiry to 0.
See https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry

Fixes #2653

### Checklist

### DIFF
--- a/quota/mysqlqm/mysql_quota.go
+++ b/quota/mysqlqm/mysql_quota.go
@@ -19,6 +19,8 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
+	"strconv"
 
 	"github.com/google/trillian/quota"
 )
@@ -97,6 +99,10 @@ func (m *QuotaManager) countUnsequenced(ctx context.Context) (int, error) {
 }
 
 func countFromInformationSchema(ctx context.Context, db *sql.DB) (int, error) {
+	// turn off statistics caching for MySQL 8
+	if err := turnOffInformationSchemaCache(ctx, db); err != nil {
+		return 0, err
+	}
 	// information_schema.tables doesn't have an explicit PK, so let's play it safe and ensure
 	// the cursor returns a single row.
 	rows, err := db.QueryContext(ctx, countFromInformationSchemaQuery, "Unsequenced", "BASE TABLE")
@@ -123,4 +129,30 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
 		return 0, err
 	}
 	return count, nil
+}
+
+// turnOffInformationSchemaCache turn off statistics caching for MySQL 8
+// To always retrieve the latest statistics directly from the storage engine and bypass cached values, set information_schema_stats_expiry to 0.
+// See https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry
+func turnOffInformationSchemaCache(ctx context.Context, db *sql.DB) error {
+	opt := "information_schema_stats_expiry"
+	res := db.QueryRowContext(ctx, "SHOW VARIABLES LIKE '%"+opt+"%'")
+	var none, expiry string
+	err := res.Scan(&none, &expiry)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil
+		}
+
+		return fmt.Errorf("Failed to get variable %q: %v", opt, err)
+	}
+
+	exp, err := strconv.Atoi(expiry)
+	if err != nil || exp != 0 {
+		if _, err := db.ExecContext(ctx, "SET SESSION "+opt+"=0"); err != nil {
+			return fmt.Errorf("Failed to set variable %q: %v", opt, err)
+		}
+	}
+
+	return nil
 }
--- a/quota/mysqlqm/mysql_quota.go
+++ b/quota/mysqlqm/mysql_quota.go
@@ -20,7 +20,6 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"strconv"
 
 	"github.com/google/trillian/quota"
 )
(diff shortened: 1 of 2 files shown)


AlCutter:
Hi px3303, many thanks for the PR!

To help understanding the context for it, would you mind updating the description with a bit of background on why this change is necessary/what [potential] problem it fixes?

Thanks in advance.

px3303:
@AlCutter 
This PR can fix this problem without relying on MySQL configurations and versions, and it is compatible with earlier versions previous to MySQL 8.0.

Part of #2653

AlCutter:
Ah, that's great - thank you!
It might be worth adding a line about adding preliminary support for MySQL 8 into the `CHANGELOG` file?

quota/mysqlqm/mysql_quota.go
```diff
@@ -124,3 +130,29 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
...
+	opt := "information_schema_stats_expiry"
+	res := db.QueryRowContext(ctx, "SHOW VARIABLES LIKE '%"+opt+"%'")
+	var none, expiry string
+	err := res.Scan(&none, &expiry)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil
+		}
+
+		return fmt.Errorf("Failed to get variable %q: %v", opt, err)
```
AlCutter, line 147: nit: "Failed" -> "failed"

quota/mysqlqm/mysql_quota.go
```diff
@@ -124,3 +130,29 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
...
+	}
+
+	exp, err := strconv.Atoi(expiry)
+	if err != nil || exp != 0 {
+		if _, err := db.ExecContext(ctx, "SET SESSION "+opt+"=0"); err != nil {
+			return fmt.Errorf("Failed to set variable %q: %v", opt, err)
```
AlCutter, line 153: nit: "Failed" -> "failed"

quota/mysqlqm/mysql_quota.go
```diff
@@ -124,3 +130,29 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
...
 }
+
+// turnOffInformationSchemaCache turn off statistics caching for MySQL 8
+// To always retrieve the latest statistics directly from the storage engine and bypass cached values, set information_schema_stats_expiry to 0.
+// See https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry
+func turnOffInformationSchemaCache(ctx context.Context, db *sql.DB) error {
+	opt := "information_schema_stats_expiry"
+	res := db.QueryRowContext(ctx, "SHOW VARIABLES LIKE '%"+opt+"%'")
+	var none, expiry string
+	err := res.Scan(&none, &expiry)
```
AlCutter, line 141: nit: Can you inline this with the check on `err` like this: `if err := res.Scan...; err != nil { ...`?

quota/mysqlqm/mysql_quota.go
```diff
@@ -124,3 +130,29 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
...
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil
+		}
+
+		return fmt.Errorf("Failed to get variable %q: %v", opt, err)
+	}
+
+	exp, err := strconv.Atoi(expiry)
```
AlCutter, line 150: Can inline here too?
`if exp, err := strconv...; err != nil || exp != 0 { ...`

quota/mysqlqm/mysql_quota.go
```diff
@@ -124,3 +130,29 @@ func countFromTable(ctx context.Context, db *sql.DB) (int, error) {
 	}
 	return count, nil
 }
+
+// turnOffInformationSchemaCache turn off statistics caching for MySQL 8
+// To always retrieve the latest statistics directly from the storage engine and bypass cached values, set information_schema_stats_expiry to 0.
+// See https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry
+func turnOffInformationSchemaCache(ctx context.Context, db *sql.DB) error {
```
AlCutter, line 137: Will this fail safely for all versions of MySQL prior to 8 or should we check the version and only execute on version 8+?

quota/mysqlqm/mysql_quota.go
AlCutter, line 141: Is it possible to `Scan` expiry directly into an `int` or will that fail if it's unset?
(Would changing the query to `SELECT FROM @@SESSION.information_schema_stats_expiry` help if so?)

AlCutter (commented):
Just a few minor nits and a couple of questions on the code itself inline.

quota/mysqlqm/mysql_quota.go
px3303: ok

quota/mysqlqm/mysql_quota.go
px3303: ok

quota/mysqlqm/mysql_quota.go
px3303: ok

quota/mysqlqm/mysql_quota.go
px3303: removed

quota/mysqlqm/mysql_quota.go
px3303: Thanks very much for your reminding. Indeed, “expiry” should be scanned into `int` type.

When this variable does not exist (versions before MySQL 8.0), `SELECT @@SESSION.information_schema_stats_expiry` statement will return error instead of empty resultset.

quota/mysqlqm/mysql_quota.go
px3303: Yes, it will fail safely for versions before MySQL 8.0. This is because this function uses `SHOW VARIABLES LIKE` statement. Executing this statement in versions before MySQL 8 is equivalent to searching a system variable that does not exist, and it will only return empty result rather than error.

Furthermore, tests of 5.6.51, 5.7.36 and 8.0.27 have already passed locally.

px3303:
@AlCutter 

 I have updated the `CHANGELOG` file.

AlCutter (approved):

AlCutter:
@px3303 - Nice one, many thanks!

Issue #2249, opened 2020-10-28
etcdv3 client warn message: lease keepalive response queue is full; dropping response send


mingjunyang:
env 
```text
etcd server: 3.3.25

ct_server: 1.1.1
log_signer: 1.3.11
log_server: 1.3.11

server info : Linux ip-xx.xx.xx.xx.ap-mmmm-1.compute.internal 4.14.198-152.320.amzn2.x86_64 #1 SMP Wed Sep 23 23:57:28 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

when server start running a few moment, the console print warning message.
```text
log_signer_1_d621790f498d | {"level":"warn","ts":"2020-10-28T02:52:39.049Z","caller":"clientv3/lease.go:524","msg":"lease keepalive response queue is full; dropping response send","queue-size":16,"queue-capacity":16}
log_server_1_891d40c2df74 | {"level":"warn","ts":"2020-10-28T02:52:44.006Z","caller":"clientv3/lease.go:524","msg":"lease keepalive response queue is full; dropping response send","queue-size":16,"queue-capacity":16}
```

I don't know this warning should or not affect long-term operation.

I think it might be related to this issues https://github.com/etcd-io/etcd/issues/9911 . the man report *server memory highly load due to too many connections*.

AlCutter:
Hi,

This is not an error we've come across before, but it sounds like it might be an issue with etcd/etcd client internals.

I found this blog post which seems to be written by someone who had a similar issue and looked into what was causing it, perhaps it may give some insight:
https://blog.csdn.net/puss0/article/details/96481856

px3303:
@AlCutter 

Fixed by #2655

PR #2655, opened 2022-01-12
Gracefully exit the program when the lease expired

### LINKED ISSUE #2654: KeepAlive in etcd client cannot ensure the permanent validity of lease
`KeepAlive` in etcd client cannot ensure the permanent validity of lease. If automatic renewal of `trillian` lease is interrupted due to unexpected circumstances, trillian will still continue operating. However, the program will never be detected by ETCD server, and as a result it can no longer provide services. In addition, external process health check program cannot discover abnormalities and warnings cannot be sent to system administrators in time.

```golang
// AnnounceSelf announces this binary's presence to etcd.
func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, endpoint string) func() {
	if client == nil {
		return func() {}
	}

	// Get a lease so our entry self-destructs.
	leaseRsp, err := client.Grant(ctx, 30)
	if err != nil {
		glog.Exitf("Failed to get lease from etcd: %v", err)
	}
	client.KeepAlive(ctx, leaseRsp.ID)
...
```

```golang
// recvKeepAlive updates a lease based on its LeaseKeepAliveResponse
func (l *lessor) recvKeepAlive(resp *pb.LeaseKeepAliveResponse) {
	karesp := &LeaseKeepAliveResponse{
		ResponseHeader: resp.GetHeader(),
		ID:             LeaseID(resp.ID),
		TTL:            resp.TTL,
	}

	l.mu.Lock()
	defer l.mu.Unlock()

	ka, ok := l.keepAlives[karesp.ID]
	if !ok {
		return
	}

	if karesp.TTL <= 0 {
		// lease expired; close all keep alive channels
		delete(l.keepAlives, karesp.ID)
		ka.close()
		return
	}
...
```

Meanwhile, because `trillian` did not consume `LeaseKeepAliveResponse` channel immediately, ETCD client would print a large number of warnings #2249 .

px3303:
This PR can let `trillian` proactively Listen "LeaseKeepAliveResponse" channel returned by `KeepAlive` in ETCD client. When automatic renewal interruption is detected,  Exit the program by canceling the context.

Fixes #2654,#2249

### Checklist

### DIFF
--- a/cmd/internal/serverutil/main.go
+++ b/cmd/internal/serverutil/main.go
@@ -230,7 +230,7 @@ func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, end
 	if err != nil {
 		glog.Exitf("Failed to keep lease alive from etcd: %v", err)
 	}
-	listenKeepAliveRsp(ctx, keepAliveRspCh, cancel)
+	go listenKeepAliveRsp(ctx, keepAliveRspCh, cancel)
 
 	em, err := endpoints.NewManager(client, etcdService)
 	if err != nil {
@@ -252,19 +252,17 @@ func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, end
 // listenKeepAliveRsp listens to `keepAliveRspCh` channel, and calls the cancel function
 // to notify the lease expired.
 func listenKeepAliveRsp(ctx context.Context, keepAliveRspCh <-chan *clientv3.LeaseKeepAliveResponse, cancel func()) {
-	go func() {
-		for {
-			select {
-			case <-ctx.Done():
-				glog.Infof("listenKeepAliveRsp canceled: %v", ctx.Err())
+	for {
+		select {
+		case <-ctx.Done():
+			glog.Infof("listenKeepAliveRsp canceled: %v", ctx.Err())
+			return
+		case _, ok := <-keepAliveRspCh:
+			if !ok {
+				glog.Errorf("listenKeepAliveRsp canceled: unexpected lease expired")
+				cancel()
 				return
-			case _, ok := <-keepAliveRspCh:
-				if !ok {
-					glog.Errorf("listenKeepAliveRsp canceled: unexpected lease expired")
-					cancel()
-					return
-				}
 			}
 		}
-	}()
+	}
 }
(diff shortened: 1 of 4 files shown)


mhutchinson:
Sorry for the slow reply, I've been at work trying to understand this change. The pattern you're using here feels a bit opaque:

1) Code calls a sub-method, which returns a child context
2) Code then awaits cancellation of child context
3) On cancellation of the child context, the parent context is cancelled

I've created a change that I think is equivalent with respect to propagating etcd cancellation, but does it as:

1) Code passes the cancellation function into the sub-method
2) If the lease is cancelled, the cancel function is called

This change is on my fork at https://github.com/google/trillian/compare/master...mhutchinson:etcd_keepalive?expand=1. If you like this, then please do integrate this into the PR. If this change breaks something, then let me know!

px3303:
@mhutchinson 

Sorry for my bad english. Thanks a lot for your suggestion, I decided to go with it. But you removed the `AwaitContext` function, it is needed to exit the rpc server block, So i added it again.

cmd/internal/serverutil/main.go
```diff
@@ -238,3 +248,23 @@ func AnnounceSelf(ctx context.Context, client *clientv3.Client, etcdService, end
 		client.Revoke(ctx, leaseRsp.ID)
 	}
 }
+
+// listenKeepAliveRsp listens to `keepAliveRspCh` channel, and calls the cancel function
+// to notify the lease expired.
+func listenKeepAliveRsp(ctx context.Context, keepAliveRspCh <-chan *clientv3.LeaseKeepAliveResponse, cancel func()) {
```
AlCutter, line 254: Do you think it'd be slightly easier to grok if this method blocked and the caller is responsible for doing `go listenKeepAliveRsp(...)`?

util/process.go
```diff
@@ -42,3 +42,21 @@ func AwaitSignal(ctx context.Context, doneFn func()) {
 		glog.Infof("AwaitSignal canceled: %v", ctx.Err())
 	}
 }
+
+// AwaitContext waits for context done, then runs the given function.
```
AlCutter, line 46: I'd be tempted to keep this as an unexported function in the `serverutil` package it's called from...

px3303: fixed

cmd/internal/serverutil/main.go
px3303: fixed

AlCutter (approved):
@mhutchinson for a 2nd pair of eyes

mhutchinson (approved):

PR #2685, opened 2022-03-28
Add unique index on sequenced leaf identity hashes

aarongable:
Both the mysql and cloudspanner storage backends currently enforce that
all rows in the LeafData table have unique LeafIdentityHash values, by
virtue of that column being part of the table's primary key. This
restriction is used by Trillian to detect duplicate submissions and
return already-sequenced data rather than integrating the submitted leaf
entry a second time.

However, the same uniqueness constraint is not enforced on the
SequencedLeafData table. This means that, in exceptional circumstances
such as those documented in the Oak 2022 CT Log[1], multiple
SequencedLeafData rows can have the same foreign key, and end up
referencing the same underlying LeafData row. This results in corruption
of the log, and violation of its cryptographic guarantees.

Add a UNIQUE KEY constraint on the (TreeId, LeafIdentityHash) tuple in
the SequencedLeafData table to ensure that this situation cannot arise
again.

[1] https://groups.google.com/a/chromium.org/g/ct-policy/c/sdPvvZSp7Rw/m/6UqU1MN8AQAJ

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
   we recommend clients also migrate over to this library at the earliest
   convenient time; the long term plan is to remove `merkle` from this repo.
 * `countFromInformationSchema` function to add support for MySQL 8.
+* The mysql and cloudspanner storage schemas now require that the
+  LeafIdentityHash column is unique. This prevents multiple SequencedLeafData
+  rows from referencing the same LeafData row via foreign key.
 
 ## v1.4.0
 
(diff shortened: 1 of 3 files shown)


pavelkalinnikov:
Does this new constraint affect performance in any way?

I would be concerned if this adds an extra set of participants to each transaction. It would be trading off performance of critical paths (every sequencing transaction) for a bit of consistency during the rare migration scenarios (and those problems could probably be avoided in other ways).

aarongable:
Almost certainly. It's difficult to simulate full load, but in general yes this would be expected to affect performance. We can take other approaches.

PR #2688, opened 2022-03-31
Update the dependency on merkle

pavelkalinnikov:


mhutchinson:
@pavelkalinnikov can you rebase and `push -f` so that this includes the fixed cloudbuild config?

pavelkalinnikov:
@mhutchinson Works now, thanks for fixing.

PR #2689, opened 2022-03-31
Emit a better error message on verify fail

jsha:
Old message renders like:

calculated root:
[255 199 ...  70 147 149 140 31 227 198 230 121 101 84]
does not match expected root:
[220 94 ...  71 196 132 63 30 31 192 68 94 193 54 205]

This renders the hashes using base64 and puts them on a single line.

### DIFF
--- a/merkle/logverifier/log_verifier.go
+++ b/merkle/logverifier/log_verifier.go
@@ -16,6 +16,7 @@ package logverifier
 
 import (
 	"bytes"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"math/bits"
@@ -30,7 +31,9 @@ type RootMismatchError struct {
 }
 
 func (e RootMismatchError) Error() string {
-	return fmt.Sprintf("calculated root:\n%v\n does not match expected root:\n%v", e.CalculatedRoot, e.ExpectedRoot)
+	return fmt.Sprintf("calculated root %s does not match expected root %s",
+		base64.StdEncoding.EncodeToString(e.CalculatedRoot),
+		base64.StdEncoding.EncodeToString(e.ExpectedRoot))
 }
 
 // LogVerifier verifies inclusion and consistency proofs for append only logs.


merkle/logverifier/log_verifier.go
```diff
@@ -30,7 +31,9 @@ type RootMismatchError struct {
 }
 
 func (e RootMismatchError) Error() string {
-	return fmt.Sprintf("calculated root:\n%v\n does not match expected root:\n%v", e.CalculatedRoot, e.ExpectedRoot)
+	return fmt.Sprintf("calculated root %s does not match expected root %s",
+		base64.StdEncoding.EncodeToString(e.CalculatedRoot),
+		base64.StdEncoding.EncodeToString(e.ExpectedRoot))
```
mhutchinson, line 36:
```suggestion
	return fmt.Sprintf("calculated root %x does not match expected root %x", e.CalculatedRoot, e.ExpectedRoot)
```

I agree that the message as it was is too verbose. I suggest this approach which uses the hex encoding of the bytes instead of base64. It's just a little more compact in Go, and aligns with how we usually (though alas not exclusively) encode bytes in errors.

jsha: I generally like hex encoding for this sort of output, too. But for the Certificate Transparency use case, all hashes are expressed in the protocol as base64. That makes base64 output much easier to compare against other strings (like the output of get-sth, or get-proof-by-hash). I think CT is a big enough use case that it's worth orienting hash output around base64.

mhutchinson: I think this change is strictly better than the old message, so let's go with this for now. The proper answer if we need it would be to have personalities catch and format this error messages in a way that's suitable for them. That's a good reason for having this custom error that makes the two root hashes available.

It's worth considering that we haven't considered error message strings as part of versioning considerations. Which means that there's always a chance that this message could change in the future. If you're going to rely on this, it might be worth adding a unit test downstream that asserts the message rendering so there's at least an explicit notification.

And a final consideration: this merkle code has been forked over to https://github.com/transparency-dev/merkle. I'm _fairly_ sure that I migrated all of the code our team owns over to this. It might be worth checking that this new location isn't what you want to be updating. At some future point I want to delete this stuff here, but I've left it for now as it's likely that projects are depending on it.

(who would have thought that changing an error message could prompt such discussion :-) )

mhutchinson (approved):

mhutchinson:
Cloudbuild is broken because of #2690 which I'm going to try to address to unblock CI.

merkle/logverifier/log_verifier.go
jsha: No worries, I'm not planning to parse this error message. I just saw it during testing of https://github.com/aarongable/ctaudit/pull/1 and thought "huh, that's not very human-useful." :-)

And I'll take a look at the new merkle location and see about porting to that. Thanks for the pointer.

mhutchinson:
Can you rebase on top of master? I think that this needs the changes from #2691 before the presubmits will pass.

pavelkalinnikov:
Btw, this code is no longer used in Trillian, and is marked deprecated. As Martin pointed out, we use the code from another repository now, so I propose to either open a new PR on the other repository, or I will just incorporate these changes since I'm doing the refactoring anyway.

Issue #2690, opened 2022-04-04
Bazel not building


mhutchinson:
```
DEBUG: Rule 'com_google_googleapis' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "91220997bb6bce38c979f4980da1103e9bc23abc81a09c3045459fbc89c4d1eb"
DEBUG: Call stack for the definition of repository 'com_google_googleapis' which is a http_archive (rule definition at /builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/bazel_tools/tools/build_defs/repo/http.bzl:292:16):
 - /workspace/WORKSPACE:26:1
Analyzing: 8 targets (1 packages loaded, 0 targets configured)
INFO: SHA256 (https://github.com/bazelbuild/rules_go/archive/7d17d496a6b32f6a573c6c22e29c58204eddf3d4.zip) = bf483b3be44fbe1dc798f18470cadf67541f5d890cad691ef335546a88f5ba0f
DEBUG: Rule 'io_bazel_rules_go' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "bf483b3be44fbe1dc798f18470cadf67541f5d890cad691ef335546a88f5ba0f"
DEBUG: Call stack for the definition of repository 'io_bazel_rules_go' which is a http_archive (rule definition at /builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/bazel_tools/tools/build_defs/repo/http.bzl:292:16):
 - /workspace/WORKSPACE:41:1
Analyzing: 8 targets (14 packages loaded, 56 targets configured)
INFO: Call stack for the definition of repository 'zlib' which is a http_archive (rule definition at /builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/bazel_tools/tools/build_defs/repo/http.bzl:292:16):
 - /builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/com_google_protobuf/protobuf_deps.bzl:9:9
 - /workspace/WORKSPACE:24:1
WARNING: Download from https://zlib.net/zlib-1.2.11.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
ERROR: An error occurred during the fetch of repository 'zlib':
[...]

PR #2691, opened 2022-04-04
Remove bazel from GCB to work around #2690

mhutchinson:


AlCutter (approved):
Worth leaving a comment breadcrumb pointing to the issue until it's resolved and we can reinstate?

mhutchinson:
The plan is to leave #2690 open. We're more likely to see that than a comment in this file.

PR #2693, opened 2022-04-05
Bump Bazel version

pavelkalinnikov:
The Cloudbuild is failing because it can't download a missing zlib
archive.

### Checklist

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -64,7 +64,7 @@ steps:
 
 # Run Bazel check.
 - id: bazel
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Ah, I see the #2690 workaround. Closing this.

mhutchinson:
@pavelkalinnikov did you have good reason to believe this would work? My understanding from following the breadcrumbs was that it was the protobuf dependency causing this rather than bazel, but perhaps you found something I didn't?

pavelkalinnikov:
I was poking finger in the air. It looks like it wouldn't help, since the GCB failed on this PR too.

PR #2701, opened 2022-04-13
Make in-memory Merkle tree immutable

pavelkalinnikov:


pavelkalinnikov:
Please review #2706 instead.

internal/merkle/inmemory/tree_test.go
```diff
@@ -124,16 +124,16 @@ var testProofs = []proofTestVector{
 	}},
 }
 
-func decodeHexStringOrPanic(hs string) []byte {
+// hx decodes a hex string or panics.
+func hx(hs string) []byte {
```
phbnf, line 128: This change is unrelated to this CL, right?
Although I like the shortness, after this PR, it is unclear from looking at the name of this function wether is decodes or encodes a string in hex

pavelkalinnikov: Yeah, this could be separated, I just did it as a drive-by because otherwise the lists of constants would look too verbose.

This file was old and ugly anyway, so #2707 refactors it. After the refactoring, `hx` is used only with constants that are inlined, e.g. `hx("6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d")`, and never used with variables like `hx(someVariable)`. So it's obvious from context that it decodes, and the pros is that the name is short.

PR #2707, opened 2022-04-20
Rewrite in-memory Merkle tree tests

pavelkalinnikov:
This change completes the TODO of rewriting the unit tests for the in-memory `Tree` type.
The new tests are mostly a rework of the previous tests.

The methodology of testing remains the same:
1) There is a reference implementation of root hash and proofs computation, translated from
definitions in RFC 6962. It is tested using "golden" tests and sanity checks.
2) The `Tree` type is tested exhaustively, to make sure it produces the same result as the
reference implementation.

The goal is to move the `Tree` type and tests to https://github.com/transparency-dev/merkle
`testonly` package, as it can be widely reused for tests.

### DIFF
--- a/internal/merkle/inmemory/reference_test.go
+++ b/internal/merkle/inmemory/reference_test.go
@@ -1,0 +1,135 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package inmemory
+
+import (
+	"bytes"
+	"encoding/hex"
+	"math/bits"
+	"testing"
+
+	"github.com/transparency-dev/merkle"
+	"github.com/transparency-dev/merkle/rfc6962"
+	to "github.com/transparency-dev/merkle/testonly"
+)
+
+// refRootHash returns the root hash of a Merkle tree with the given entries.
+// This is a reference implementation for cross-checking.
+func refRootHash(entries [][]byte, hasher merkle.LogHasher) []byte {
+	if len(entries) == 0 {
+		return hasher.EmptyRoot()
+	}
+	if len(entries) == 1 {
+		return hasher.HashLeaf(entries[0])
+	}
+	split := downToPowerOfTwo(uint64(len(entries)))
+	return hasher.HashChildren(
+		refRootHash(entries[:split], hasher),
+		refRootHash(entries[split:], hasher))
+}
+
+// refInclusionProof returns the inclusion proof for the given leaf index in a
+// Merkle tree with the given entries. This is a reference implementation for
+// cross-checking.
+func refInclusionProof(entries [][]byte, index uint64, hasher merkle.LogHasher) [][]byte {
+	size := uint64(len(entries))
+	if size == 1 || index >= size {
+		return nil
+	}
+	split := downToPowerOfTwo(size)
+	if index < split {
+		return append(
+			refInclusionProof(entries[:split], index, hasher),
+			refRootHash(entries[split:], hasher))
+	}
+	return append(
+		refInclusionProof(entries[split:], index-split, hasher),
+		refRootHash(entries[:split], hasher))
+}
+
+// refConsistencyProof returns the consistency proof for the two tree sizes, in
+// a Merkle tree with the given entries. This is a reference implementation for
+// cross-checking.
+func refConsistencyProof(entries [][]byte, size2, size1 uint64, hasher merkle.LogHasher, haveRoot1 bool) [][]byte {
+	if size1 == 0 || size1 > size2 {
+		return nil
+	}
+	// Consistency proof for two equal sizes is empty.
+	if size1 == size2 {
+		// Record the hash of this subtree if it's not the root for which the proof
+		// was originally requested (which happens when size1 is a power of 2).
+		if !haveRoot1 {
+			return [][]byte{refRootHash(entries[:size1], hasher)}
+		}
+		return nil
+	}
+
+	// At this point: 0 < size1 < size2.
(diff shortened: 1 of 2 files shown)


pavelkalinnikov:
Recommendation for reviewers: Review the latest state of this PR. The chain of commits reflects the iterative process of this refactoring, and shows that all the transitions are equivalent, but you may choose to trust me on that :)

internal/merkle/inmemory/reference_test.go
```diff
@@ -0,0 +1,206 @@
...
+	// subtrees are consistent. The right subtree doesn't contain the root of
+	// size1, so set haveRoot1 = false. Record the hash of the left subtree
+	// (equal in both trees).
+	return append(
+		refConsistencyProof(entries[split:], size2-split, size1-split, hasher, false),
+		refRootHash(entries[:split], hasher))
+}
+
+// downToPowerOfTwo returns the largest power of two smaller than x.
+func downToPowerOfTwo(x uint64) uint64 {
```
jiggoha, line 108: May I suggest sqrtFloor as an alternative name?

internal/merkle/inmemory/tree_test.go
```diff
@@ -16,634 +16,190 @@ package inmemory
...
-
-// Some paths for the reference tree.
-type pathTestVector struct {
-	leaf       uint64
-	snapshot   uint64
-	testVector []string
-}
-
-// Generated from C++ ReferenceMerklePath, not the Go one so we can verify
-// that they are both producing the same paths in a sanity test.
```
jiggoha, line 64: Can we move this comment and the other similar ones to the `testonly` package? It would helpful there for understanding where these constants were coming from.

internal/merkle/inmemory/tree_test.go
```diff
@@ -16,634 +16,190 @@ package inmemory
...
-// This is the hash of an empty string
-var emptyTreeHashValue = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
-
-// Inputs to the reference tree, which has eight leaves.
-var leafInputs = [][]byte{
-	hx(""), hx("00"), hx("10"), hx("2021"), hx("3031"), hx("40414243"),
-	hx("5051525354555657"), hx("606162636465666768696a6b6c6d6e6f"),
-}
-
-// Incremental roots from building the reference tree from inputs leaf-by-leaf.
```
jiggoha, line 43: Same comment

internal/merkle/inmemory/tree_test.go
```diff
@@ -16,634 +16,190 @@ package inmemory
...
-	},
-}
-
-type proofTestVector struct {
-	snapshot1 uint64
-	snapshot2 uint64
-	proof     []string
-}
-
-// Generated from ReferenceSnapshotConsistency in C++ version.
```
jiggoha, line 108: Same comment

internal/merkle/inmemory/reference_test.go
```diff
@@ -0,0 +1,206 @@
...
+// The reference Merkle tree hashing and proof algorithms in this file directly
+// implement the definitions from RFC 6962 [1]. We use this implementation only
+// for testing correctness of other more flexible and performant algorithms,
+// such as the in-memory Tree type and compact ranges.
+//
+// [1] https://datatracker.ietf.org/doc/html/rfc6962#section-2
+
+// refRootHash returns the root hash of a Merkle tree with the given entries.
+// This is a reference implementation for cross-checking.
+func refRootHash(entries [][]byte, hasher merkle.LogHasher) []byte {
```
jiggoha, line 38: Do the functions here need to be prepended with "ref"? IIUC, "reference" here refers to the fact that this is returns an example/specific case for a root hash. However, there is nothing specific about this code, since it takes `entries` as a param. Thus, is it inaccurate to just call it `rootHash`?

internal/merkle/inmemory/reference_test.go
pavelkalinnikov: I see where your idea comes from, but it's not quite a square root, it returns the highest power of 2 which is strictly less than x.

E.g., for 9 it returns 8, whereas the square root of 9 is 3.

This name is how this func was called before the refactoring.

internal/merkle/inmemory/tree_test.go
pavelkalinnikov: My plan is:
- Complete this PR.
- Move this code to `testonly` in https://transparency-dev/merkle.
- Use these and other constants as "golden" tests there.

Looks inline with your suggestion?

internal/merkle/inmemory/reference_test.go
pavelkalinnikov: "reference" here mean that the implementation is reference (not specific values/hashes). See the comment in lines 29-34.

pavelkalinnikov: To test various Merkle tree implementation, there are a couple of techniques:
- "golden" tests: use specific inputs and outputs to check that it matches exactly. Pros: no way to corrupt the "expected" result, cons: need to consider lots of corner cases manually.
- "reference" implementation: compare our implementation to ones that are most simple/intuitive (but maybe less performanct) and follow definitions closely; we know that these implementations probably work, so we can run an exhaustive test that verifies new algorithm on all possible inputs.

pavelkalinnikov: So, the methodology is:
- Make reference implementations (this file).
- Test them against golden tests (this file).
- Test any other implementation exhaustively against the reference implementation (other files).
  - Note: Other tests will indirectly "inherit" all the golden tests, so there is no need to repeat the golden testing there.

pavelkalinnikov: This way, we get Pros of both approaches, and minimise Cons.

pavelkalinnikov: This also partly addresses you suggestion below of exposing the "known good" values. It can be done, but is not strictly necessary in the described methodology.

jiggoha: Gotcha, thanks. Re-reading your comments in the code now makes a lot more sense to me.

internal/merkle/inmemory/reference_test.go
jiggoha: Ah, I did not look at this carefully. Fair enough.

internal/merkle/inmemory/tree_test.go
jiggoha: SGTM

jiggoha (approved):

PR #2708, opened 2022-04-20
Simplify log integration test use of inmemory.Tree

pavelkalinnikov:


integration/log.go
```diff
@@ -291,7 +290,11 @@ func waitForSequencing(treeID int64, client trillian.TrillianLogClient, params T
 	return errors.New("wait time expired")
 }
 
-func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params TestParameters, expect map[string]int) (map[int64]*trillian.LogLeaf, error) {
+func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params TestParameters, expect map[string]int) ([]*trillian.LogLeaf, error) {
+	if start := params.StartLeaf; start != 0 {
+		return nil, fmt.Errorf("non-zero StartLeaf is not supported: %d", start)
```
pavelkalinnikov, line 295: This is true pre this PR, but it wasn't explicitly checked.

integration/log.go
```diff
@@ -504,11 +500,10 @@ func checkConsistencyProof(consistParams consistencyProofParams, treeID int64, t
 }
 
 // buildMerkleTree returns an in-memory Merkle tree built on the given leaves.
-func buildMerkleTree(leafMap map[int64]*trillian.LogLeaf, params TestParameters) *inmemory.Tree {
+func buildMerkleTree(leaves []*trillian.LogLeaf, params TestParameters) *inmemory.Tree {
```
pavelkalinnikov, line 503: This is why `StartLeaf` that is not 0 was not supported: note that this function builds an in-memory Merkle tree that starts from empty.

pavelkalinnikov:
Commit-by-commit review might be convenient.

integration/log.go
```diff
@@ -304,70 +303,63 @@ func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params T
...
-	leafMap := make(map[int64]*trillian.LogLeaf)
 	glog.Infof("Expecting %d unique leaves", len(expect))
 
-	for currentLeaf < params.LeafCount {
-		// We have to allow for the last batch potentially being a short one
-		numLeaves := params.LeafCount - currentLeaf
-
-		if numLeaves > params.ReadBatchSize {
-			numLeaves = params.ReadBatchSize
+	var leaves []*trillian.LogLeaf
```
mhutchinson, line 308: It would be reasonable to pre-allocate this array based on the `params.LeafCount`.

integration/log.go
```diff
@@ -304,70 +303,63 @@ func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params T
...
+	for index, end := params.StartLeaf, params.StartLeaf+params.LeafCount; index < end; {
+		// We have to allow for the last batch potentially being a short one.
```
mhutchinson, line 310: IMO this (pre-existing) comment doesn't add a lot, but perhaps I've read/written a lot of batched retrieval stuff and this is useful to someone that hasn't encountered this pattern before?

integration/log.go
```diff
@@ -304,70 +303,63 @@ func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params T
...
 			}
 
-			// Ensure that the ExtraData in the leaf made it through the roundtrip. This was set up when
-			// we queued the leaves.
-			if got, want := hex.EncodeToString(leaf.ExtraData), hex.EncodeToString([]byte(strings.Replace(string(leaf.LeafValue), "Leaf", "Extra", 1))); got != want {
-				return nil, fmt.Errorf("leaf %d extra data got: %s, want:%s (%v)", leaf.LeafIndex, got, want, leaf)
+			// Ensure that the ExtraData in the leaf made it through the roundtrip.
+			// This was set up when we queued the leaves.
+			if got, want := leaf.ExtraData, []byte(strings.Replace(value, "Leaf", "Extra", 1)); !bytes.Equal(got, want) {
+				return nil, fmt.Errorf("leaf %d ExtraData: got %x, want %x: %v", leaf.LeafIndex, got, want, leaf)
```
mhutchinson, line 350: This is also pretty weird that this function that otherwise is pretty generic Trillian -> []Leaf has prior knowledge of how the leaves were constructed. This would be better in the outer function, alongside the `expect` checks about what the contents of the leaves are.

integration/log.go
```diff
@@ -304,70 +303,63 @@ func readbackLogEntries(logID int64, client trillian.TrillianLogClient, params T
...
+			leaves = append(leaves, leaf)
 
-			if expect[lk] == 0 {
-				delete(expect, lk)
+			// Check that the leaf values are returned in the same counts. Can't rely
+			// on the ordering because they can be inserted in a different order.
+			value := string(leaf.LeafValue)
+			expect[value]--
+			if expect[value] == 0 {
+				delete(expect, value)
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

integration/log.go
pavelkalinnikov: Done.

pavelkalinnikov (commented):
@mhutchinson Addressed all your suggestions.

pavelkalinnikov:
There can be further improvements, e.g. the `map[string]int` is unnecessary. The test can go like this (items correspond to separate functions, and the order of invocation):

1. Generate `entries`.
2. Write `entries`.
3. Wait for integration of size `len(entries)`.
4. Read all -> returns `storedEntries`.
5. Compare(`entries`, `storedEntries`), which internally would just sort things using a common key (or again have a map), and compare.

pavelkalinnikov:
How about I do that too, and rename/repurpose this PR to be a more general refactoring?

mhutchinson (approved):
I'm happy with this as it is. I think we need to balance how likely this code is to be changed with how much more work we put into it. I think the current state of this strikes a reasonable balance, even if this isn't how we'd write the code from scratch today.

pavelkalinnikov:
Couldn't resist one more commit.

PR #2710, opened 2022-04-21
Remake dumplib into integration test

pavelkalinnikov:
`dumplib` was a tool which created an in-memory log storage of a particular size,
and printed its contents with various flags. The useful bit of this tool was a "golden"
unit test which checks regressions on the `SubtreeProto` format in storage.

This change removes the tool, refactors the test, and places it in `integration/format`.

### Checklist


pavelkalinnikov:
There is no need to look at all commits, can just review the last state. Commits in this PR were iteratively deleting bit by bit, so that I didn't miss anything important.

integration/format/format.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return nil
+}
+
+type treeAndRev struct {
+	subtree  *storagepb.SubtreeProto
+	revision int
+}
+
+func latestRevisions(ls storage.LogStorage, treeID int64, hasher merkle.LogHasher) (string, error) {
```
jiggoha, line 136: What do the revisions here refer to?

pavelkalinnikov: In the original Trillian storage design the Merkle tree is split into tiles (which are called `SubtreeProto` there), and each tile will have a new revision every time it's modified (when new entries are appended to the log, and some nodes of a tile get new hashes, or modified ephemeral hashes).

Trillian also had maps, which were sparse Merkle trees. Maps had different API: it was possible to modify an arbitrary location in the Merkle tree (as opposed to logs, where only appends to the end are allowed). Trillian tried to have a unified Merkle storage (the `SubtreeProto`s with revisions, etc). Since then maps were deleted and the design of logs converges to simpler version which does not store ephemeral nodes, and does not need tile revisions. But for backwards compatibility the revisions still exist.

pavelkalinnikov: A revision is a "snapshot" when a new tree head is created. For logs, a new revision is when new batch of entries are appended and committed to storage.

pavelkalinnikov: So, the way to see the state of Merkle tree at revision `x`, one should scan all tiles/`SubtreeProto`s, and take the latest revision <= `x` for each.

jiggoha: Thank you, that elucidated my understanding a lot!

integration/format/format.go
```diff
@@ -0,0 +1,175 @@
...
+	// vMap maps subtree prefixes (as strings) to the corresponding subtree proto and its revision
+	vMap := make(map[string]treeAndRev)
+	memory.DumpSubtrees(ls, treeID, func(k string, v *storagepb.SubtreeProto) {
+		// Relies on the btree key space for subtrees being /tree_id/subtree/id/revision.
```
jiggoha, line 140: If this is the convention for the layout of trees, it feels like we should be abstracting it away in a library rather than re-implementing it here separately from its definition. Where is the "other side" of this implemented? Not blocking for this PR, but if it's easy to use call a library instead, I feel that it would better to do that.

pavelkalinnikov: The details of this are in `storage/memory`. 100% agree that it can be incapsulated, and it can be done in another PR.

jiggoha (approved):

PR #2726, opened 2022-04-28
cache: Set a fixed tile height of 8

pavelkalinnikov:
This change reflects the status quo.


mhutchinson (approved):
Please add this to the changelog that this check is now more strict. It may well be enforced elsewhere, but I think no harm in having a line in the release notes clearly calling this type of additional check out.

PR #2747, opened 2022-05-26
Add restricted owners for the DB schemas

### LINKED ISSUE #1297: Need to automate database schema updates

mhutchinson:
This isn't a perfect solution, but it's a pragmatic solution that fixes #1297 enough for how much we are making changes

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,7 +8,14 @@
 # the repo. Unless a later match takes precedence,
 # @google/trillian-team will be requested for
 # review when someone opens a pull request.
-*                                @google/trillian-team
+*                                 @google/trillian-team
 
-/*.proto                         @Martin2112 @AlCutter @pphaneuf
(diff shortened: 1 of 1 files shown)


AlCutter:
hehe, *snap* #2746

PR #2755, opened 2022-06-07
Fix flag initialization

pavelkalinnikov:
Flag parsing happens in main, after all init functions have been called.
This change moves the parsing code, incorrectly placed in init, to a
custom flag Set method.

#1188

### Checklist

### DIFF
--- a/log/sequencer.go
+++ b/log/sequencer.go
@@ -66,12 +66,10 @@ var (
 	QuotaIncreaseFactor = 1.1
 )
 
+// stringSet is a set of strings that can be parsed by flag package.
 type stringSet map[string]bool
 
 func (s *stringSet) String() string {
-	if (*s)["*"] {
-		return "*"
-	}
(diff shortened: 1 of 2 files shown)


log/sequencer.go
```diff
@@ -66,20 +67,41 @@ var (
 	QuotaIncreaseFactor = 1.1
 )
 
+type stringSet map[string]bool
```
AlCutter, line 70: I know it's only "temporary", but might be worth a comment on this - it's mostly a string set except when there's a magical string present in the set in which case the rest of the set is dropped.

Just something to reduce surprise if someone inadvertently tried to use this stringSet for something else in here :)

AlCutter (approved):

AlCutter:
Seems like there might be a bad interaction with `flagsaver` in the tests?

pphaneuf (approved):
Oh, yeah, using flags before `flag.Parse()` is called is Very Forbidden!

Issue #2763, opened 2022-06-21
Scripts and documentation still refer to pre-Go-module practices


mhutchinson:
Almost anything that references `GOPATH/src` is from prior to go modules being fully adopted. It used to be that having all go code in that directory was expected (required?). Now it is only there if users go out of their way to check out code there.

I'm working on cleaning up the easy references to this, but there will be a long tail that might take some time to tidy up as they are deep in scripts that haven't been run in some time (e.g. scripts for running on public GCP).

mhutchinson:
The remaining work here is in the `example/deployment` directories. `git grep src/github.com/google/trillian` seems a pretty good way to catch these:

```
examples/deployment/aws/terraform.tf:cd /go/src/github.com/google/trillian
examples/deployment/docker/db_client/Dockerfile:ADD . /go/src/github.com/google/trillian
examples/deployment/docker/db_client/Dockerfile:WORKDIR /go/src/github.com/google/trillian
examples/deployment/docker/db_server/Dockerfile:# expects the build context to be: $GOPATH/src/github.com/google/trillian
examples/deployment/kubernetes/deploy.sh:cd $GOPATH/src/github.com/google/trillian
examples/deployment/kubernetes/mysql/README.md:kubectl apply -f $GOPATH/src/github.com/google/trillian/storage/mysql/kubernetes
examples/deployment/kubernetes/mysql/README.md:$GOPATH/src/github.com/google/trillian/storage/mysql/kubernetes/resetdb.sh
examples/deployment/kubernetes/undeploy.sh:cd $GOPATH/src/github.com/google/trillian
```

Any changes to these files should be done with a working public environment to deploy them to, in order to confirm the new steps succeed.

PR #2765, opened 2022-06-21
Manually vendor the required .proto files from googleapis

mhutchinson:
This resolves some weird references to GOPATH, that while it worked, was weird (#2763). It also simplifies new user onboarding (one less instruction to follow), makes CI more lightweight (one less big repo to clone), and avoids mismatched versions of these protos from causing havoc (because they were less hermetic before).


gen.go
```diff
@@ -14,7 +14,7 @@
 
 package trillian
 
-//go:generate protoc -I=. -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false trillian_log_api.proto trillian_admin_api.proto trillian.proto --doc_out=markdown,api.md:./docs/
+//go:generate protoc -I=. -I=third_party/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false trillian_log_api.proto trillian_admin_api.proto trillian.proto --doc_out=markdown,api.md:./docs/
```
pavelkalinnikov, line 17: Is there any prescribed naming convention for this? I'm thinking between `vendor` and `third_party`? Do you have any distinction in mind?

mhutchinson: I think it's arbitrary, but I'm happy to change it if the strong preference goes the other way. I'm perfectly ambivalent in this moment.

third_party/googleapis/README.md
```diff
@@ -0,0 +1,2 @@
+This directory is manually vendored in order to take the required proto definitions from googleapis.
+The files here were cloned at @c81bb70.
```
pavelkalinnikov, line 2: Are there stable version tags to use here instead?

gen.go
pavelkalinnikov: This reading might help to decide between the 2: https://github.com/golang-standards/project-layout
(haven't read it yet myself)

pavelkalinnikov: After a brief glance, `third_party` seems more relevant, esp. that it's a subset of the files (i.e. a special kind of "fork"). WDYT?

pavelkalinnikov:
What was the sequence of actions you did to create this. Was it a copy-paste, or some "submodule" magic? It would be nice to expand the README a bit, to know how to update this dependency in the future.

quota/etcd/quotapb/gen.go
```diff
@@ -14,4 +14,4 @@
 
 package quotapb
 
-//go:generate protoc -I=. -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false quotapb.proto
+//go:generate protoc -I=. -I=../../../third_party/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false quotapb.proto
```
AlCutter, line 17: You might be able to do something like `... -I=$(go list -f '{{.Root}}' .)/third_party/googleapis ...` to avoid ever having to update the relative path.

mhutchinson:
> What was the sequence of actions you did to create this? Was it a copy-paste, or some "submodule" magic? It would be nice to expand the README a bit, to know how to update this dependency in the future.

Yeah that's a good question. I've documented it in the README, and confirmed that the distilled steps I've put in there replace the `google` directory if it is deleted.

quota/etcd/quotapb/gen.go
mhutchinson: It's a nice idea, but failing:
```
protoc -I=. -I=$(go list -f '{{.Root}}' .)/third_party/googleapis --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --go-grpc_opt=require_unimplemented_servers=false quotapb.proto
$(go: warning: directory does not exist.
Unknown flag: -f
gen.go:17: running "protoc": exit status 1
```

Now if I hadn't spent a huge amount of time trying to work out the magic of these go:generate protoc invocations before I'd spend the "couple of minutes" looking at this. But I know it's a trap :-)

And yeah, I confirmed that running `go list -f '{{.Root}}' .` naked in a shell does the right thing.

AlCutter: I guess you'd have to explicitly invoke a shell along the lines of `go:generate bash -c "protoc -I=$(go list ...) ...` to make it work. I don't know whether that's a net win or not...

pavelkalinnikov: There is a precedent here: https://github.com/google/certificate-transparency-go/blob/1a8898f9bd49798a7a442d3691fdc79c6c31807f/proto_gen.go#L23

AlCutter: hehe, turns out it was me :)

third_party/googleapis/README.md
```diff
@@ -0,0 +1,26 @@
...
+a known location.
+
+## Updating
+
+The required files were determined by manual inspection of the codebase and closing the
+transitive dependencies by checking the imports of the files.
+Should the upstream repository be significantly refactored then it is possible that further files
+may need to be imported if an update is performed.
+
+The files here were cloned at @c81bb70.
```
pavelkalinnikov, line 16:
```suggestion
The files here were cloned at commit `c81bb70`.
```

third_party/googleapis/README.md
```diff
@@ -0,0 +1,26 @@
+## Purpose
+
+This directory is manually vendored in order to take the required proto definitions from googleapis.
```
pavelkalinnikov, line 3:
```suggestion
This directory is manually vendored in order to take the required proto definitions from `googleapis`.
```

pavelkalinnikov (approved):
LGTM in principle, see some more nits.

quota/etcd/quotapb/gen.go
mhutchinson: Oh wow, this works by invoking a shell to invoke protoc. It's cool that this is possible, in a weird way. But are we really sure that the relative paths are _that bad_ ? If we were moving this around a lot, I can see the argument. If this wasn't covered by presubmits then that would add further weight to it.

I think I'm particularly resistant because I've spent a lot of time in the last week debugging shell scripts and it has been whatever the opposite of fun is.

third_party/googleapis/README.md
mhutchinson: Done.

third_party/googleapis/README.md
mhutchinson: Done.

quota/etcd/quotapb/gen.go
pavelkalinnikov: I am 80/20 in favor of the simpler command, at the cost of a relative path.

AlCutter (approved):

pavelkalinnikov:
Still LGTM. Maybe worth mentioning in CHANGELOG?

quota/etcd/quotapb/gen.go
mhutchinson: I'll keep this simpler invocation with the relative paths.

PR #2766, opened 2022-06-24
Change broken installation instructions to just run the binary directly

mhutchinson:
The  command given didn't work. We could change this to specify the install with  as a suffic, but skipping installation and path setup seems even easier and more direct.

This fixes another case of #2763.


AlCutter (approved):
LGTM, but I think the PR description text might need a tweak?

PR #2779, opened 2022-07-19
Update contrib.go.opencensus.io/exporter/stackdriver to `v0.13.13`

saschagrunert:
This also requries a bump to the latest etcd alpha and helps to fix the
usage of the retracted module github.com/prometheus/prometheus@v2.5.0+incompatible.

How to reproduce:

```
> go list --mod=mod -u -m --json all
go: github.com/prometheus/prometheus@v2.5.0+incompatible: retracted by module author
```

### Checklist


roger2hk (changes requested):
Thank you very much for your contribution. There are two issues in this PR.

The PR title and description do not match because there are more than one changes. I would recommend you to bump a dependency version in each PR.

For the go.etcd.io/etcd dependency version updates, they are alpha versions. Unless there is any security alert, I would prefer to upgrade when the stable release is ready.

saschagrunert:
> Thank you very much for your contribution. There are two issues in this PR.
> 
> The PR title and description do not match because there are more than one changes. I would recommend you to bump a dependency version in each PR.
> 
> For the go.etcd.io/etcd dependency version updates, they are alpha versions. Unless there is any security alert, I would prefer to upgrade when the stable release is ready.

Alright, let's wait for etcd to be out of alpha and give the update another try when the time is right.

PR #2787, opened 2022-07-29
Switch from glog to klog

jdolitsky:
Please see the following explanation of the reasons for klog vs. glog: https://github.com/kubernetes/klog#why-was-klog-created

### DIFF
--- a/client/admin.go
+++ b/client/admin.go
@@ -19,11 +19,11 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/google/trillian"
 	"github.com/google/trillian/client/backoff"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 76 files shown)


jdolitsky:
This should pull in latest go-licenses once this is merged: https://github.com/google/go-licenses/pull/138

mhutchinson:
/gcbrun

I've just rebased this to fix conflicts in the go.mod file

mhutchinson:
I've got #2791 to update to the latest linter, but it doesn't fix this problem. https://github.com/dominikh/go-tools/blob/v0.3.3/go/ir/exits.go#L100 - looks like more work is required upstream in the staticcheck library to detect the `v2` version of `klog`.

AlCutter:
I've rebased/jimmied onto `HEAD` and fixed a few follow-up bits, @hickford would you mind casting an eye over the tweaks so I'm not approving my own code?

AlCutter (approved):

AlCutter:
Added #2801 to track removing the `//nolint` comments once @mhutchinson's fix to `golangci-lint` is present in a released version.

PR #2799, opened 2022-08-17
Update CHANGELOG.md

jiggoha:
Prepare CHANGELOG.md for v.1.5.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,29 @@
 
 ## HEAD
 
+## v.1.5.0
+
 * Deprecate `--tree_ids_with_no_ephemeral_nodes` flag by setting it to all trees
-  (`*`) by default.
-* Updated golangci-lint to v1.47.3 (developers should update to this version)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,29 @@
 
 ## HEAD
 
+## v.1.5.0
+
```
AlCutter, line 6: I wonder if it's worth emphasising this one a bit?

How about something like this:
> 
> ### Storage
> 
> Ephemeral nodes are no-longer written for any tree by default (and have not been read since the `v1.4.0` release), the corresponding `---tree_ids_with_no_ephemeral_nodes` flag is now deprecated and will be removed in a future release).
>
?

AlCutter (approved):
Small suggestion inline, but LGTM

CHANGELOG.md
jiggoha: That is more clear, thanks!

PR #2805, opened 2022-09-02
Fix log server not exiting properly on SIGINT

gregoire-mullvad:
Make the tree garbage collection coroutine exit immediately when the context is cancelled.

This fixes an issue that appeared in 1.4.1: after receiving SIGINT `trillian_log_server` starts shutting down but seems to get stuck and needs to be killed.

### Checklist

### DIFF
--- a/server/admin/tree_gc.go
+++ b/server/admin/tree_gc.go
@@ -98,7 +98,6 @@ func (gc *DeletedTreeGC) Run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-timeAfter(d):
-		default:
 		}
 
 	}
--- a/server/admin/tree_gc.go
+++ b/server/admin/tree_gc.go
@@ -35,7 +35,7 @@ const (
 
 var (
 	timeNow   = time.Now
-	timeSleep = time.Sleep
+	timeAfter = time.After
 
 	hardDeleteCounter monitoring.Counter
 	metricsOnce       sync.Once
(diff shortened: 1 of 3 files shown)


server/admin/tree_gc.go
```diff
@@ -100,7 +94,13 @@ func (gc *DeletedTreeGC) Run(ctx context.Context) {
 		}
 
 		d := gc.minRunInterval + time.Duration(rand.Int63n(gc.minRunInterval.Nanoseconds()))
-		timeSleep(d)
+		select {
+		case <-ctx.Done():
+			return
+		case <-timeAfter(d):
+		default:
```
AlCutter, line 101: I suspect you probably don't want the `default:` case in there? I think having it will defeat the new `timeAfter()` sleep and cause the loop to "busy spin" until `ctx` is done (i.e. I think you want this `select` to block until either `ctx` is done or `d` time has passed?)

AlCutter (changes requested):

server/admin/tree_gc.go
gregoire-mullvad: Yes, of course.  I still find `select` a bit tricky :sweat_smile:

gregoire-mullvad: Fixed

AlCutter: > Yes, of course. I still find `select` a bit tricky 😅

No worries, it happens to us all sometime or other :)

AlCutter (approved):

AlCutter:
Many thanks for the fix!

PR #2833, opened 2022-11-08
WIP: Add initial schema and storage provider for CockroachDB

JAORMX:
This creates an initial schema for CRDB by mirroring a lot of the
existing constructs from the MySQL storage provider. It also adds
scripts and test code to run unit and integration tests with CockRoachDB

While this lead to a lot of code duplication, the intention is to
address that in a separate PR in order to keep the work separated into
logical batches. Doing all the refactoring to create common constructs
for SQL backends would lead to a patch too big for folks to review in a
reasonable time and would create more risk.

The plan, thus, is as follows:

- [x] Create CRDB storage provider (this PR).
- [ ] Abstract SQL admin storage to its own package.
- [ ] Abstract Statement cache into its own package.
- [ ] Abstract (Log)Tree Transaction code into its own package.

This way, we could have re-usable parts for other SQL providers.

### Checklist


roger2hk:
You can make use of the draft pull request if it is not yet ready for review.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

JAORMX:
superceded by https://github.com/google/trillian/pull/2834

PR #2834, opened 2022-11-10
Add initial schema, storage provider and quota manager for CockroachDB

JAORMX:
This creates an initial schema for CRDB by mirroring a lot of the existing constructs from the MySQL storage provider. It also adds scripts and test code to run unit and integration tests with CockroachDB

While this lead to a lot of code duplication, the intention is to address that in a separate PR in order to keep the work separated into logical batches. Doing all the refactoring to create common constructs for SQL backends would lead to a patch too big for folks to review in a reasonable time and would create more risk.

### Checklist


JAORMX:
Current status: The driver spawns a single node crdb instance when the tests start, and these are passing. The tests and some basic linting is running on GH actions. I'm currently looking into getting the integration tests to run on GH actions too.

One caveat is that in order to easily parse database URI's, I did use a golang 1.19 function: https://github.com/equinixmetal-security/trillian/blob/crdb/storage/testdb/testdb.go#L191

Don't know how much of an issue this would be 😕 , but it's sure a convenient function! 😄 

Looking forward for initial feedback.

JAORMX:
@AlCutter this is ready for review.

storage/crdb/log_storage.go
```diff
@@ -0,0 +1,848 @@
+// Copyright 2022 <TBD>
```
AlCutter, line 1: I think these lines should retain the the date from the original mysql files they derive from, but could be "Copyright XXXX The Trillian Authors" (and you can add an entry in the AUTHORS/CONTRIBUTORS files).

.github/workflows/test.yaml
```diff
@@ -0,0 +1,71 @@
+---
```
AlCutter, line 1: Should this filename (and the `name:` values inside) specifically call out that it's focused on the cockroachDB tests?

JAORMX: Done, the workflow itself is now `Test CRDB`, and this name will be prepended on all jobs in it.

JAORMX:
@AlCutter seems I can't read the results of the unit test failure:
```
You do not have sufficient permissions to view this page
```

roger2hk:
> @AlCutter seems I can't read the results of the unit test failure:
> 
> ```
> You do not have sufficient permissions to view this page
> ```

There are errors when executing the presubmit script (`./scripts/presubmit.sh --no-build --fix --no-mod-tidy --empty-diff`).

```
storage/testdb/testdb.go:191:12: uri.JoinPath undefined (type *url.URL has no field or method JoinPath) (typecheck)
	uri = uri.JoinPath(name)
```

storage/crdb/log_storage.go
```diff
@@ -0,0 +1,848 @@
...
+func sortLeavesForInsert(leaves []*trillian.LogLeaf) []leafAndPosition {
+	ordLeaves := make([]leafAndPosition, len(leaves))
+	for i, leaf := range leaves {
+		ordLeaves[i] = leafAndPosition{leaf: leaf, idx: i}
+	}
+	sort.Sort(byLeafIdentityHashWithPosition(ordLeaves))
+	return ordLeaves
+}
+
+func (t *logTreeTX) QueueLeaves(ctx context.Context, leaves []*trillian.LogLeaf, queueTimestamp time.Time) ([]*trillian.LogLeaf, error) {
```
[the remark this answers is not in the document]
AlCutter: Just to capture the offline discussion: Os and I chatted about this, we can do this in a follow-up PR once we have something we know is working with the minimal number changes from the MySQL code.

JAORMX:
@roger2hk ah! I used a golang 1.19 construct. I'll refactoring so it works in 1.17.

AlCutter:
Can you add a note in `CHANGELOG` about adding an initial/alpha storage driver for crdb?

JAORMX:
> Can you add a note in `CHANGELOG` about adding an initial/alpha storage driver for crdb?

done.

AlCutter (approved):

PR #2836, opened 2022-11-14
Add initial schema, storage provider and quota manager for CockroachDB

JAORMX:
This supercedes #2834 while preserving history for the mysql driver inside the crdb storage driver.

This creates an initial schema for CRDB by mirroring a lot of the existing constructs from the MySQL storage provider. It also adds scripts and test code to run unit and integration tests with CockroachDB

While this lead to a lot of code duplication, the intention is to address that in a separate PR in order to keep the work separated into logical batches. Doing all the refactoring to create common constructs for SQL backends would lead to a patch too big for folks to review in a reasonable time and would create more risk.

### Checklist


JAORMX:
Getting rid of this in favor of the original PR.

PR #2855, opened 2022-11-28
Provide cockroach in test env

AlCutter:
Install CockroachDB in the testbase image.

### Checklist


JAORMX (approved):

roger2hk (approved):
LGTM but we have to keep the tech debt (repo complexity) in mind.

JAORMX:
@roger2hk normally that binary is fetched directly at the test setup. However, it seems that the test infra no longer allows egress access. So doing this was necessary.

In GitHub actions it's possible to download the binary and cache it if needed (making it available for other test runs in a determined period of time). I'm not sure if that's easy to do with cloudbuild. One option would be for us to limit the cloudbuild tests to *not* run the crdb test cases and only run those in GH actions. That way we'd keep the test images small.

AlCutter:
> LGTM but we have to keep the tech debt (repo complexity) in mind.

Good call, what do you think would be a good way to help toward that here?

I'll submit this for now so PRs are unblocked, but more than happy to change approach in a follow-up.

PR #2876, opened 2022-12-14
Migrate spanner types off deprecated package

AlCutter:
See https://github.com/googleapis/google-cloud-go/blob/main/migration.md

### Checklist

### DIFF
--- a/storage/cloudspanner/getdb_test.go
+++ b/storage/cloudspanner/getdb_test.go
@@ -32,7 +32,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	database "cloud.google.com/go/spanner/admin/database/apiv1"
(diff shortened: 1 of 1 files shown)


zkpjedi (commented):
looks like library itself https://github.com/googleapis/google-cloud-go/blob/main/migration.md still points to the old databasepb proto?

https://github.com/googleapis/google-cloud-go/blob/main/migration.md

AlCutter:
> looks like library itself https://github.com/googleapis/google-cloud-go/blob/main/migration.md still points to the old databasepb proto?
> 
> https://github.com/googleapis/google-cloud-go/blob/main/migration.md

Yeah, I was hoping to get this in before the dep update (on the chance that it may _Just Work™️_, but that wasn't to be so I've rolled it into #2871 .

PR #2936, opened 2023-03-15
When a SQL error occurs, clean up all prepared statement.

px3303:
As described in # 420, when the prepared statement fails unexpectedly in the database, the program cannot be recovered.

This PR will clear all the prepared statements in the cache when the statement execution error occurs (this is because other prepared statements may also make errors).

### DIFF
--- a/storage/mysql/log_storage.go
+++ b/storage/mysql/log_storage.go
@@ -741,6 +741,7 @@ func (t *logTreeTX) getLeavesByHashInternal(ctx context.Context, leafHashes [][]
 	args = append(args, t.treeID)
 	rows, err := stx.QueryContext(ctx, args...)
 	if err != nil {
+		t.ls.cleanAllStmt()
 		klog.Warningf("Query() %s hash = %v", desc, err)
 		return nil, err
 	}
--- a/storage/mysql/tree_storage.go
+++ b/storage/mysql/tree_storage.go
@@ -104,6 +104,20 @@ func expandPlaceholderSQL(sql string, num int, first, rest string) string {
 	return strings.Replace(sql, placeholderSQL, parameters, 1)
 }
 
+// clearAllStmt clean up all sql.Stmt in cache
+func (m *mySQLTreeStorage) cleanAllStmt() {
+	m.statementMutex.Lock()
+	defer m.statementMutex.Unlock()
+
+	for _, ns := range m.statements {
+		for _, s := range ns {
+			s.Close()
+		}
+	}
+
+	m.statements = make(map[string]map[int]*sql.Stmt)
+}
+
 // getStmt creates and caches sql.Stmt structs based on the passed in statement
 // and number of bound arguments.
 // TODO(al,martin): consider pulling this all out as a separate unit for reuse
@@ -200,6 +214,7 @@ func (t *treeTX) getSubtrees(ctx context.Context, treeRevision int64, ids [][]by
 
 	rows, err := stx.QueryContext(ctx, args...)
 	if err != nil {
+		t.ts.cleanAllStmt()
 		klog.Warningf("Failed to get merkle subtrees: %s", err)
 		return nil, err
 	}
@@ -296,6 +311,7 @@ func (t *treeTX) storeSubtrees(ctx context.Context, subtrees []*storagepb.Subtre
 
 	r, err := stx.ExecContext(ctx, args...)
 	if err != nil {
+		t.ts.cleanAllStmt()
 		klog.Warningf("Failed to set merkle subtrees: %s", err)
 		return err
 	}


storage/mysql/tree_storage.go
```diff
@@ -104,6 +104,20 @@ func expandPlaceholderSQL(sql string, num int, first, rest string) string {
 	return strings.Replace(sql, placeholderSQL, parameters, 1)
 }
 
+// clearAllStmt clean up all sql.Stmt in cache
+func (m *mySQLTreeStorage) cleanAllStmt() {
+	m.statementMutex.Lock()
+	defer m.statementMutex.Unlock()
+
```
mhutchinson, line 111: Some kind of logging for this would be really useful. This should be a rare event in an otherwise healthy system:

`klog.Info("Clearing all prepared statements")`

Actually, for bonus points I think it's worth incrementing a counter whenever we clear statements, which should appear on monitoring dashboards and allow correlation with potential impacts such as increased latency or errors. Take a look at `queuedCounter` in this class as an example (though we won't need a `logID` for this as it's effectively global).

storage/mysql/tree_storage.go
```diff
@@ -104,6 +104,20 @@ func expandPlaceholderSQL(sql string, num int, first, rest string) string {
...
+	for _, ns := range m.statements {
+		for _, s := range ns {
+			s.Close()
```
mhutchinson, line 114: Please check the error returned from Close. Maybe not much more to do other than log the failure, but I don't like letting errors like this simply disappear.

mhutchinson (changes requested):
Can you also remove the TODO in `getStmt` to do this work:
```// TODO(al,martin): we'll possibly need to expire Stmts from the cache,`

Thanks!

File: storage/mysql/tree_storage.go (hunk 1 above)
Comment (px3303) [reply storage/mysql/tree_storage.go:111]: fixed

File: storage/mysql/tree_storage.go (hunk 2 above)
Comment (px3303) [reply storage/mysql/tree_storage.go]: fixed

**Review by mhutchinson — COMMENTED** [review]
This LGTM. Would appreciate @AlCutter giving final approval as an original author of this code.

File: storage/mysql/log_storage.go [hunk 3]
```diff
@@ -101,6 +102,7 @@ func createMetrics(mf monitoring.MetricFactory) {
 	queuedCounter = mf.NewCounter("mysql_queued_leaves", "Number of leaves queued", logIDLabel)
 	queuedDupCounter = mf.NewCounter("mysql_queued_dup_leaves", "Number of duplicate leaves queued", logIDLabel)
 	dequeuedCounter = mf.NewCounter("mysql_dequeued_leaves", "Number of leaves dequeued", logIDLabel)
+	clearedAllStmtCounter = mf.NewCounter("mysql_cleared_all_prepared_statements", "Number of all prepared statements cleared")
```
AlCutter, line 105: nit: -> "number of times the prepared-statement cache has been cleared"?

px3303: fixed

storage/mysql/tree_storage.go
```diff
@@ -104,6 +104,25 @@ func expandPlaceholderSQL(sql string, num int, first, rest string) string {
 	return strings.Replace(sql, placeholderSQL, parameters, 1)
 }
 
+// clearAllStmt clean up all sql.Stmt in cache
+func (m *mySQLTreeStorage) cleanAllStmt() {
```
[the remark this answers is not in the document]
px3303: fixed

AlCutter:
@px3303 Would you mind adding a small bit of text to the CHANGELOG to capture this? It's an important change which definitely should be highlighted in the next release's set of release notes.

storage/mysql/log_storage.go
```diff
@@ -101,6 +102,7 @@ func createMetrics(mf monitoring.MetricFactory) {
 	queuedCounter = mf.NewCounter("mysql_queued_leaves", "Number of leaves queued", logIDLabel)
 	queuedDupCounter = mf.NewCounter("mysql_queued_dup_leaves", "Number of duplicate leaves queued", logIDLabel)
 	dequeuedCounter = mf.NewCounter("mysql_dequeued_leaves", "Number of leaves dequeued", logIDLabel)
+	clearedStmtCacheCounter = mf.NewCounter("mysql_cleared_prepared-statement_caches", "Number of times the prepared-statement cache has been cleared")
```
Martin2112, line 105: Might be worth noting in the description that this is an unexpected condition / could indicate a problem.

storage/mysql/tree_storage.go
```diff
@@ -47,9 +47,9 @@ const (
 	 n.TreeId = ? AND n.SubtreeRevision <= ?
 	GROUP BY n.TreeId, n.SubtreeId
  ) AS x
- INNER JOIN Subtree 
- ON Subtree.SubtreeId = x.SubtreeId 
- AND Subtree.SubtreeRevision = x.MaxRevision 
+ INNER JOIN Subtree
```
Martin2112, line 50: I assume these are gofmt changes. Could be done separately but up to you.

Martin2112 (approved):
I feel a bit bad that we never got round to this. The way this works here means some errors will surface but it should recover. Alternatively, it could retry on error but I think this code is probably already complex enough and this isn't something that should happen often.

px3303:
@AlCutter @Martin2112 @mhutchinson 

I submitted an improved version #2937 , please review it again.

PR #2945, opened 2023-03-21
Pin all usages of golangci-lint to same rule version

mhutchinson:


hickford (approved):

mhutchinson:
@JAORMX FYI. I think it makes sense to run explicitly the same version of the linter over this code as the rest of the repo.

JAORMX:
> @JAORMX FYI. I think it makes sense to run explicitly the same version of the linter over this code as the rest of the repo.

It does!  Thanks for looking into this!

PR #2971, opened 2023-04-18
Pin the version of cockroach for tests to a specific version

mhutchinson:
./storage/crdb tests have started failing in GCP recently. Investigation shows that this is pulling the latest tagged version, which breaks hermetic builds. This PR is a test to see if using the previous version fixes the issue, and if so, I propose merging this to unblock things and then discussing with the community to look for better solutions.

### DIFF
--- a/storage/crdb/common_test.go
+++ b/storage/crdb/common_test.go
@@ -43,7 +43,7 @@ func (db *testDBHandle) GetDB() *sql.DB {
 func TestMain(m *testing.M) {
 	flag.Parse()
 
-	ts, err := testserver.NewTestServer()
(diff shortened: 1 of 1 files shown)


mhutchinson:
@JAORMX FYI

roger2hk (approved):

pavelkalinnikov:
> @pavelkalinnikov in case this is interesting to you. No expectation or pressure :-)

@mhutchinson There are a couple of CRDB issues with similar errors:
- https://github.com/cockroachdb/cockroach/issues/80683
- https://github.com/cockroachdb/cockroach/issues/90114

See if your symptoms match, or transaction does anything similar.

PR #2978, opened 2023-04-26
Fixed all lint errcheck in number of TLDs

mhutchinson:
This fixes all errors in client, cmd, experimental, integration, log, merkle, quota, and testonly. The storage top level directory still has a huge number, so this change does not modify the .golangci.yaml file to enable checks across the repo.

### DIFF
--- a/merkle/coniks/coniks.go
+++ b/merkle/coniks/coniks.go
@@ -57,9 +57,13 @@ func (m *Hasher) HashEmpty(treeID int64, root node.ID) []byte {
 	buf := bytes.NewBuffer(make([]byte, 0, 32))
 	h := m.New()
 	buf.Write(emptyIdentifier)
-	binary.Write(buf, binary.BigEndian, uint64(treeID))
+	if err := binary.Write(buf, binary.BigEndian, uint64(treeID)); err != nil {
+		klog.Errorf("binary.Write(): %v", err)
+	}
 	m.writeMaskedNodeID(buf, root)
-	binary.Write(buf, binary.BigEndian, uint32(depth))
+	if err := binary.Write(buf, binary.BigEndian, uint32(depth)); err != nil {
+		klog.Errorf("binary.Write(): %v", err)
+	}
 	h.Write(buf.Bytes())
 	r := h.Sum(nil)
 	if klog.V(5).Enabled() {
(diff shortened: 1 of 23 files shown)


merkle/coniks/coniks.go
```diff
@@ -57,9 +57,13 @@ func (m *Hasher) HashEmpty(treeID int64, root node.ID) []byte {
 	buf := bytes.NewBuffer(make([]byte, 0, 32))
 	h := m.New()
 	buf.Write(emptyIdentifier)
-	binary.Write(buf, binary.BigEndian, uint64(treeID))
+	if err := binary.Write(buf, binary.BigEndian, uint64(treeID)); err != nil {
```
AlCutter, line 60: Interesting that the `buf.Write` didn't trigger the linter, that's very clever...

Anyway, given that these `binary.Write` calls in here (and below on L82 etc.) are writing to a `bytes.Buffer`, I think any errors will `panic` rather than return (https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/bytes/buffer.go;l=164-165), so you _could_ opt to `_ = ...` these with a comment if you liked?

quota/crdbqm/common_test.go
```diff
@@ -37,7 +37,10 @@ func TestMain(m *testing.M) {
 	dburl.Path = "/"
 
 	// Set the environment variable for the test server
-	os.Setenv(testdb.CockroachDBURIEnv, dburl.String())
+	if err := os.Setenv(testdb.CockroachDBURIEnv, dburl.String()); err != nil {
+		klog.Errorf("Failed to set CockroachDBURIEnv: %v", err)
+		os.Exit(1)
```
AlCutter, line 42: If you use `klog.Exitf` you'll get an `os.Exit(1)` thrown in for free :)

AlCutter (approved):
Just a few nits/suggestions

merkle/coniks/coniks.go
mhutchinson: Ta!

testonly/mdm/mdm.go
```diff
@@ -132,7 +132,9 @@ func (m *MergeDelayMonitor) monitor(ctx context.Context, idx int) error {
 			createNew = true
 		}
 		if createNew {
-			crand.Read(data)
+			if _, err := crand.Read(data); err != nil {
+				return fmt.Errorf("rand.Read(): %v", err)
```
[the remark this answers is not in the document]
mhutchinson: Just dropped the package altogether.

quota/crdbqm/common_test.go
mhutchinson: I've done it this way for consistency with L29-30. For better or worse.

PR #2980, opened 2023-04-28
Friendlier intro/discussion for the Claimant Model

mhutchinson:
I've referred to it as a book in a tongue-in-cheek way. It isn't intended to be as long as a book, but there are similarities in that it is intended to be read in order, and should favour accessibility over being absolutely technically correct from the offset.

### DIFF
--- a/docs/claimantmodel/Book.md
+++ b/docs/claimantmodel/Book.md
@@ -1,0 +1,130 @@
+# Claimant Model In-Depth
+
+## Introduction
+
+The Claimant Model is very concisely introduced in the [Core Model](./CoreModel.md).
+While this concise documentation works well as a refresher for those that already understand the model, it has
+empirically proved to be a tough introduction to those new to the domain.
+This longer form text on the Claimant Model serves as a more gentle on-ramp to understanding this powerful tool.
+
+This text is broken into sections, but is intended to be read from start to finish.
+
+## Motivation
+
+Prior to the Claimant Model being developed, new transparency projects were designed by copying and modifying
+patterns from existing deployments, most notably [Certificate Transparency](https://certificate.transparency.dev/) (CT).
+While this allows those familiar with CT to reach consensus on a rough design quickly, it has a number of drawbacks:
+
+1. Those that don't already understand CT now have to read [RFC 6962](https://www.rfc-editor.org/rfc/rfc6962) to understand
+   the domain that the other designers are drawing from
+1. There are still terms in CT that people disagree on the precise definition of (e.g. Auditor and Monitor)
+1. CT uses mechanisms such as SCTs that transparency systems should aim to avoid unless strictly necessary
+1. Any differences in the roles, interactions, actors, or motivations between CT and the new system being designed can
+   have knock-on effects that can undermine the security properties achieved
+
+With this in mind, a clean, role-based model for describing transparency systems was required, and the Claimant Model was born.
+
+## Applicability
+
+The Claimant Model can be used to describe any system where an actor (`Believer`) performs some trusted action based on some
+information (`Claim`) provided to them by a trusted third-party (`Claimant`). There is _usually_ an asymmetry, where the
+Believer can not verify that the Claim is true, but believes it because they trust the Claimant.
+
+The situation above appears in countless settings in our everyday human lives.
+On a technical level, this also happens frequently, often invisibly to the user.
+For example, when a Chrome browser (`Believer`) receives a TLS certificate (`Claim`) for a domain, it only performs
+the connection (trusted action) if the certificate is issued by a Certificate Authority (`Claimant`) that chains to its
+trusted roots.
+As another example, a software package manager (`Believer`) will only install software (trusted action) when the package
+(`Claim`) is signed by the correct package maintainer (`Claimant`).
+
+Above we state that it must be possible to verify that the Claim is true.
+This means that all Claims must be _falsifiable_, i.e. they are able to be proven false.
+We'll discuss _who_ can falsify Claims later.
+
+### Exceptions
+
+There are some situations that the Claimant Model _could_ be used to describe, but taking the time to do so would be a poor
+return on investment, or more simply, just confuse a situation that can be easily stated in simpler terms.
+
+- Believer can trivially verify the Claim: trust in the Claimant is a non-factor in this case. For example, if a web
+  service exists that can return the prime factors of some large number, the Believer may not have been able to perform
+  this factorization itself, but it can trivially multiply the factors together to confirm the result.
+- Believer will trust Claims from anyone: if the identity of the Claimant is not a factor in the Believer's decision to
+  perform some action, then the Claimant Model provides little benefit. Such situtations are unfortunately
+  still common, e.g. installing unsigned software/firmware.
+
+## Trust
+
+Any ecosystem where the Claimant Model is [applicable](#applicability) has potential for exploit.
+Specifically, a false Claim made under the identity of the Claimant would be believed by a Believer, and they would
+perform some trusted action they would not have made given a true Claim.
+
+### Motivations for Betraying Trust
+
+There are many reasons that a Claimant may make incorrect Claims, and these are worth considering along with a threat model.
+As a non-exhaustive list of reasons that a Claimant may make a false claim:
+- Accidental:
+  - Bugs in code that the Claimant runs
+  - Human error in following process
+  - Massively unlikely, but [possible](https://www.agwa.name/blog/post/how_ct_logs_fail) random uncontrollable events
+- Greed: the Claimant decided that abusing their trust was financially more beneficial than continuing operating in
+  a trustworthy manner (e.g. exit scams in the cryptocurrency world)
+- Compromise:
+  - Insider Threat: a single agent at the company has motivation to exploit trust
+  - Leverage: a third-party could coerce the Claimant into betraying trust
+  - Compromise: actors that breach the Claimant security mechanisms could issue false claims directly by stealing signing keys,
+    or change code or playbooks in such a way that the Claimant themselves issues false claims
+
+There are also pressures in the opposite direction, i.e. motivations to *never* issue a bad claim.
+Consider the situation where each Claim is provided to a single Believer who would never realize the Claim was bad, and then
+compare this to a situation where every Claim is made in a publicly visible forum.
+In the first situation, maybe a Claimant could be coerced into making a bad Claim "just this one time", where in the second
+case it would be discovered and a public investigation launched.
+This is the motivation for [transparency logs](https://transparency.dev).
+
+## Verification
+
+Given enough time, potential motivation, and/or bad luck, any Claimant could issue a false Claim.
+The question is how to make a system with enough security given this fact.
+
+Earlier we established that Claims must be falsifiable, but that in general the Believer will not be able
+to perform this verification.
+Thus we introduce the role of the `Verifier`, which will be able to verify all Claims in a system.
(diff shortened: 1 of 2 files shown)


docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+## Motivation
+
+Prior to the Claimant Model being developed, new transparency projects were designed by copying and modifying
+patterns from existing deployments, most notably [Certificate Transparency](https://certificate.transparency.dev/) (CT).
+While this allows those familiar with CT to reach consensus on a rough design quickly, it has a number of drawbacks:
+
+1. Those that don't already understand CT now have to read [RFC 6962](https://www.rfc-editor.org/rfc/rfc6962) to understand
+   the domain that the other designers are drawing from
+1. There are still terms in CT that people disagree on the precise definition of (e.g. Auditor and Monitor)
+1. CT uses mechanisms such as SCTs that transparency systems should aim to avoid unless strictly necessary
```
haydentherapper, line 21: Would it be worthwhile to provide a brief explanation for this comment, given this is one of the primary designs we want to avoid copying over from CT?

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+1. Any differences in the roles, interactions, actors, or motivations between CT and the new system being designed can
+   have knock-on effects that can undermine the security properties achieved
+
+With this in mind, a clean, role-based model for describing transparency systems was required, and the Claimant Model was born.
+
+## Applicability
+
+The Claimant Model can be used to describe any system where an actor (`Believer`) performs some trusted action based on some
+information (`Claim`) provided to them by a trusted third-party (`Claimant`). There is _usually_ an asymmetry, where the
+Believer can not verify that the Claim is true, but believes it because they trust the Claimant.
```
haydentherapper, line 31: I see later you define the Verifier role, but would it be worth specifying what "verify the Claim" means here? This was something confusing initially, thinking that verifying = signature verification, not accuracy of the claim

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+
+The situation above appears in countless settings in our everyday human lives.
```
haydentherapper, line 33: Did you want to expand on this? Provide a non-technical example of a Believer and Claimant?

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+On a technical level, this also happens frequently, often invisibly to the user.
+For example, when a Chrome browser (`Believer`) receives a TLS certificate (`Claim`) for a domain, it only performs
+the connection (trusted action) if the certificate is issued by a Certificate Authority (`Claimant`) that chains to its
+trusted roots.
+As another example, a software package manager (`Believer`) will only install software (trusted action) when the package
+(`Claim`) is signed by the correct package maintainer (`Claimant`).
+
+Above we state that it must be possible to verify that the Claim is true.
+This means that all Claims must be _falsifiable_, i.e. they are able to be proven false.
```
haydentherapper, line 42: Maybe a concrete example here, falsifying the TLS certificate claim?

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+
+There are some situations that the Claimant Model _could_ be used to describe, but taking the time to do so would be a poor
+return on investment, or more simply, just confuse a situation that can be easily stated in simpler terms.
+
+- Believer can trivially verify the Claim: trust in the Claimant is a non-factor in this case. For example, if a web
+  service exists that can return the prime factors of some large number, the Believer may not have been able to perform
+  this factorization itself, but it can trivially multiply the factors together to confirm the result.
+- Believer will trust Claims from anyone: if the identity of the Claimant is not a factor in the Believer's decision to
+  perform some action, then the Claimant Model provides little benefit. Such situtations are unfortunately
+  still common, e.g. installing unsigned software/firmware.
```
haydentherapper, line 55: Either in this section or somewhere, is it worth explicitly stating that this expects that claims are present/discoverable in a transparency log? That's something that wasn't immediately obvious to me learning the model.

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+  - Leverage: a third-party could coerce the Claimant into betraying trust
+  - Compromise: actors that breach the Claimant security mechanisms could issue false claims directly by stealing signing keys,
+    or change code or playbooks in such a way that the Claimant themselves issues false claims
+
+There are also pressures in the opposite direction, i.e. motivations to *never* issue a bad claim.
+Consider the situation where each Claim is provided to a single Believer who would never realize the Claim was bad, and then
+compare this to a situation where every Claim is made in a publicly visible forum.
+In the first situation, maybe a Claimant could be coerced into making a bad Claim "just this one time", where in the second
+case it would be discovered and a public investigation launched.
+This is the motivation for [transparency logs](https://transparency.dev).
```
haydentherapper, line 84: This part about the motivation of transparency might be useful higher up.

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+a digital signature over the Statement.
+
+### Composite Claims
+
+Thus far we've referred to the Believer taking action based on a single Claim.
+This isn't incorrect, but at times it is useful to consider a single Claim as multiple sub-claims.
+In particular, this is useful when a claim has multiple sub-claims that can be verified by different actors.
+
+TODO: example
+
```
haydentherapper, line 126: One thing I'd like more explanation on is when actors are the same for multiple roles

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+- A falsifiable logical statement
+- Data
+
+The first of these is the correct usage of `Claim`: a falsifiable logical statement made by the Claimant.
+The second of these is often referred to as the Claim at high levels, but where precision is required we should refer
+to this as the `Statement`.
+
+A Statement must be issued by the Claimant and it must be possible to construct the Claim from it.
+In practical scalable systems, Statements are determined to have been issued by a Claimant by the presence of
+a digital signature over the Statement.
```
haydentherapper, line 117: A concrete example here, i.e TLS certificate?

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Composite Claims
+
+Thus far we've referred to the Believer taking action based on a single Claim.
+This isn't incorrect, but at times it is useful to consider a single Claim as multiple sub-claims.
+In particular, this is useful when a claim has multiple sub-claims that can be verified by different actors.
+
```
haydentherapper, line 124: +1 to expanding this

docs/claimantmodel/Book.md
```diff
@@ -0,0 +1,130 @@
...
+- Believer can trivially verify the Claim: trust in the Claimant is a non-factor in this case. For example, if a web
+  service exists that can return the prime factors of some large number, the Believer may not have been able to perform
+  this factorization itself, but it can trivially multiply the factors together to confirm the result.
+- Believer will trust Claims from anyone: if the identity of the Claimant is not a factor in the Believer's decision to
+  perform some action, then the Claimant Model provides little benefit. Such situtations are unfortunately
+  still common, e.g. installing unsigned software/firmware.
+
+## Trust
+
+Any ecosystem where the Claimant Model is [applicable](#applicability) has potential for exploit.
```
phbnf, line 59: I get what you mean, but I find this quite negative. It seems like the Claimant model is incomplete, and allows for exploit, when in really it gives you the framework to think about these exploits.

How about something like "Claims made by the Claimant can be true of false. Since they are trusted by the Believers, a false Claim made under the identity of the Claimant would be believed by a Believer, and they would
perform some trusted action based on this false Claim.
The claimant model describes clarifies the relationship and incentives between Actors such that Believers can safely perform actions based on Claims."

docs/claimantmodel/Book.md
phbnf: +1, one or more example that people can follow along in this doc, with details related to this example in each section would ease understanding (and ideally, an example without SCTs)

docs/claimantmodel/Book.md
mhutchinson: I agree we should have public documentation about why these promises are a non-trivial change to the security model and what this entails, but I will take a strong default position that this is not an appropriate location for this content to live. We should link to it from here when it exists though. Do either of you have a counter argument to this position?

Specifically I'd want to be convinced that this documentation which is about the claimant model would be improved by adding content about a) the history of CT, and/or b) design considerations of SCTs.

haydentherapper: I think it's fine for documentation about CT and SCTs to be linked from elsewhere, but I don't know if we do have those docs yet. Given that we really do want to discourage the use of SCTs, I think we either need to create a brief doc about the downsides of them and alternatives, or briefly summarize the issue here.

docs/claimantmodel/Book.md
mhutchinson: I understand why _you_ think that as the operator of a log. I have tried to stage this doc in a way that leads from the symptoms, to the underlying problem, and then on to the solution. Given that logs are the solution, it feels natural that they should appear later in the doc. Perhaps this is a fallacy though and the majority of readers will be familiar with the concept of logs and delaying introduction of them is unsatisfying.

I'll leave this comment open to reconsider the narrative structure. Perhaps instead of A-B-C-D this should be something more like A-c-B-C-D, i.e. tease the ending near the beginning but without going full https://en.wikipedia.org/wiki/Reverse_chronology.

docs/claimantmodel/Book.md
mhutchinson: Yep, added an example here.

docs/claimantmodel/Book.md
mhutchinson: Added a little stub in a new section but this can be expanded. Feel free to make any suggestions or queries on this new stub.

docs/claimantmodel/Book.md
mhutchinson: Thanks for the +1. Resolving this as it's tracked by a TODO, and leaving comments open for the remaining stuff.

mhutchinson (commented):
Great comments, thanks for the engagement. I'm feeling that this is still miles away from done, but hopefully the latest changes take it a step in the right direction.

tracymiranda:
Thanks for sharing this - I've had a read through and this is a very useful document and complementary to the existing material in github and on the transparency.dev website. I'd +1 merging as is and then iterating for improvements. 

Some suggestions (which shouldn't gate merging a first version):
I'd recommend treating this doc as an entry level document to folks looking to understand the claimant model, and in turn the Core Model doc is a reference for folks who already understand the claimant model. With that in mind:
- Introduction: remove all todos and guff here and turn it into a genuine introduction stating what the goals of the claimant model are. 
- Motivation: this is a good section that covers the motivation nicely in a way I haven't seen captured elsewhere
- Applicability - I like the examples given here
- Trust - good as is 
- Verification - add a link for the todo part 
- More precision - most of the section here could just be definitions? In particular the actors and roles sections seems overexplained. The core model doc  says it more concisely 'This model describes roles, not actors. In a concrete deployment of this model, the same actor may play multiple roles, or multiple actors may need to work together to cover a role.'
- SCT appendix - doesn't seem like it belongs here but doesn't really matter if this material isn't really covered well anywhere else.

AlCutter (approved):

PR #3058, opened 2023-08-08
Build the trillian_log_server and trillian_log_signer binary without depending on glibc

roger2hk:
This PR avoids similar issue like https://github.com/google/certificate-transparency-go/issues/1118.

Here is the distroless Dockerfile example for Go.
https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/Dockerfile

### Checklist

### DIFF
--- a/examples/deployment/docker/log_server/Dockerfile
+++ b/examples/deployment/docker/log_server/Dockerfile
@@ -13,7 +13,7 @@ RUN go mod download
 COPY . .
 
 # Build the server.
-RUN go install ./cmd/trillian_log_server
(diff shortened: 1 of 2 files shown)


roger2hk:
> Have you tried these images to confirm they work without cgo? I know that there are some libraries (e.g. sqlite, which is not relevant here) that require cgo. I'm not sure if others do?

Thanks for the reminder. I've revisited this issue and found out https://github.com/GoogleContainerTools/distroless/issues/1342, which is an open issue related to the Debian version compatibility.

PR #3083, opened 2023-09-05
Register DoFns

AlCutter:
Beam seems to now require all `DoFn`s to be registered at package init time.

This PR fixes that, including pulling out anonymous funcs used only in the tests.

### Checklist


mhutchinson (approved):

mhutchinson:
I thought it generated code to do this, but let's just get this in and we can tidy things up if we context load this code for any future work. 👍

Issue #3182, opened 2023-11-02
switch to MySQL 8.x as 5.7 is now end of life


bobcallaway:
[MySQL 5.7 is now end of life](https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-44.html) and will no longer be providing security or bug fixes. It would be good to move the integration testing away from 5.7 and to 8.x

roger2hk:
Fixed in #3183.

PR #3183, opened 2023-11-02
Updated all MySQL deps to 8.0 #3182

mhutchinson:
This PR is primarily being raised as a smoke test to see whether this is going to be simple or not

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## HEAD
 
+* Bump CI version of MySQL from 5.7 to 8.0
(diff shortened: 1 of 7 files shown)


mhutchinson:
This passes the tests. We haven't performed long-lived performance testing against this version of the DB, so this isn't yet a raving endorsement that clients should go upgrade their MySQL instances. However, any clients that have tried MySQL 8 and can report back any performance notes would be greatly appreciated.

Anyone seeing write performance issues in MySQL 8 could take a look at https://forums.mysql.com/read.php?24,668698,668704#msg-668704 and report back whether turning off binary logging fixes it.

roger2hk (approved):

PR #3201, opened 2023-11-09
Support for skipping subtree revisions to increase read performance and reduce disk usage

mhutchinson:
Just to confirm that removing this will actually work in practice.

The tests pass, so now we need to design an upgrade mechanism.
Probably forking this mysql driver / schema to be a mysql2 which is
effectively a different driver.

We'll also need to make some changes to this to support MariaDB.
The 'REPLACE INTO' syntax is, I'm pretty sure, a MySQL specfic
extension. There is a more standard format for doing this, but it
involved passing in more '?' parameters which was already getting
too wild for me with the `<placeholder>` expansion. Probably a
solveable problem, but one for when I'm a bit less ill :-)


jsha:
MariaDB also documents the `REPLACE INTO` syntax: https://mariadb.com/kb/en/replace/

jsha:
I think there's a way to get the benefits of this branch without the disruption of a table migration. Rather than remove the column, we could pick some very large revision number (larger than any currently existing one), and start writing / replacing all subtrees at that revision number. We could then modify the `selectSubtreeSQL` query to always request that specific revision (eliminating the subquery).

Doing this on an existing table would not reduce the cardinality of the existing index but would dramatically slow its growth, which could have a performance benefit, depending on whether the issue is the absolute size or the growth rate.

mhutchinson:
I have the same intuition and imagined a similar solution. Such an elegant
hack wouldn't be without risks, and would require a careful rollout. I'm
particular a staged rollout where some old code and new code were live at
the same time could really break things.

If you would be happy to work through these risks then I would be happy to
think about this approach some more.

mhutchinson:
As a data point, I've run the integration tests in [CT-go](https://github.com/google/certificate-transparency-go/) against this version of Trillian and confirmed that it works: `./trillian/integration/integration_test.sh`

mhutchinson:
For anyone following along at home, the commit history got weird here. Somehow (maybe during a rebase?) I ended up setting the branch to an intermediate version of the work that was incomplete. Luckily `git reflog` spelunking got me back to a version that works. I've force-pushed and so you should only see a single commit, which leaves the tree in a working state.

I've abandoned the `mysql2` idea that I was pursuing for the moment. It may be possible to get it to work, but it ended up with a lot of duplication and started sprawling to changes outside of the driver directory (introducing new flags, `storage/testdb.go`, etc). I'll take a look at some other options.

mhutchinson:
There are a number of approaches that could be used to transition from the current world where subtrees are revisioned, to one where they are unrevisioned. I'll list them here:

## The breaking change

This simply changes the schema to drop the revisions column and then updates the read/write queries. This is what the first commit to this PR does.

This is the simplest thing to do, but it has no backwards compatibility for logs running on the older schema. This would need to be a Trillian 2.0.

## Different storage layer

Copying the `mysql` storage layer into a `mysql2` storage directory avoids the breaking change for existing deployments. Deploying trillian with `--storage_system=mysql2` would change to the revisionless version.

This involves a lot of duplicate code for an indeterminate amount of time. It also means that new flags need to be introduced such as `mysql2_uri` to avoid conflicts with the existing flags in the `mysql` storage implementation. This approach also has a limitation that a single deployment of Trillian could only serve with either `mysql` or `mysql2`. This means that log operators that have a number of historic logs with revisioned subtrees would need to turn up a new instance of Trillian for any new logs that desired the `mysql2` revisionless feature.

## Different code paths depending on tree features

Trillian stores [tree metadata](https://github.com/google/trillian/blob/master/storage/mysql/schema/storage.sql#L9) that is read at the start of each tree operation. In this approach we store some marker in the tree metadata that indicates that this is a revisionless tree, and an absence of this marker means that it will be treated as a historic revisioned subtree log.

This approach allows for backwards compatibility with existing logs, and a single instance of Trillian can support both revisioned and unrevisioned subtrees. Further, having this data in the database means that it's more difficult for a log operator to point the wrong configuration of Trillian at a database and potentially make incompatible writes by mixing revisioned / unrevisioned approaches on the same log.

This approach has a little more complexity in terms of code, but simpler in terms of deployment. For this reason, this is the primary exploration path at this point. The current commit chain demonstrates one approach to doing this, where the marker is the presence of magic bytes '🙃norevs🙃' as the prefix of the tree description field.

A more elegant approach would be to add a new column to the `Trees` table, but this would require a DDL schema update migration for all log operators using MySQL.

pavelkalinnikov:
Just passing by here with some thoughts dump :)

Would a blatant `ALTER TABLE Trees ADD COLUMN IF NOT EXISTS SomeInfo LONGBLOB;` do the trick? Containing the `norevs` and other potential new things (bundled as a proto, for instance). If the binary is rolled back to an older code, would it choke upon seeing an extra field in this table?

To make this column creation seamless, the `ALTER TABLE` could be run upon the process startup (a bit wasteful, but perhaps fine if it's done only once by each node on each rollout). After a few releases, this step could be removed.

In general, some seamless/automatic/stateful migration mechanism (built into the concrete Trillian storage implementations) would remove a lot of friction here, and would allow Trillian schema/code to iterate. For example, have some migrations versioning (like a simple counter), a table in the schema containing the "min supported version" which is bumped after each migration and ensuring all nodes know this version, and some form of election and progress tracking to facilitate these migrations. All done from within Trillian (without reliance on operators, manual steps, and external tooling).

As an example, CockroachDB seamless migrations [mechanism](https://github.com/cockroachdb/cockroach/blob/11eb2972ba1c7edfe5adaf2bee39d1f8df61b13c/pkg/clusterversion/cockroach_versions.go#L21-L172) could be of interest, but may be a bit complex.

mhutchinson:
@pavelkalinnikov Hello, and welcome back :-)

We were at a conference last week, apologies for the slow reply. Replies are here, and I'll put them in line:

> Would a blatant `ALTER TABLE Trees ADD COLUMN IF NOT EXISTS SomeInfo LONGBLOB;` do the trick? Containing the `norevs` and other potential new things (bundled as a proto, for instance). If the binary is rolled back to an older code, would it choke upon seeing an extra field in this table?
> 
> To make this column creation seamless, the `ALTER TABLE` could be run upon the process startup (a bit wasteful, but perhaps fine if it's done only once by each node on each rollout). After a few releases, this step could be removed.

I considered this but ruled it out because I think we would still need to have manual steps for users that have deployed with a locked down configuration, i.e. where the DB user that Trillian connects as has data read/write permissions, but does not have permissions to modify the schema (i.e. no `ALTER TABLE` permission). I suspect this concern doesn't affect you as a Cockroach maintainer because your code is the database and thus has all the permissions :-)

> In general, some seamless/automatic/stateful migration mechanism (built into the concrete Trillian storage implementations) would remove a lot of friction here, and would allow Trillian schema/code to iterate. For example, have some migrations versioning (like a simple counter), a table in the schema containing the "min supported version" which is bumped after each migration and ensuring all nodes know this version, and some form of election and progress tracking to facilitate these migrations. All done from within Trillian (without reliance on operators, manual steps, and external tooling).

+1 for a metadata table. I wished that we had one when I started this PR :-) Something to add to the roadmap, along with automatic upgrades, and documentation on the permissions that the Trillian user needs. That said, I'm still very nervous about even proposing this feature because this adds complexity that is somewhat hard to test, and if it goes wrong, logs don't support recovery from a backup.

storage/mysql/tree_storage.go
```diff
@@ -34,7 +34,7 @@ import (
 
 // These statements are fixed
 const (
-	insertSubtreeMultiSQL = `INSERT INTO Subtree(TreeId, SubtreeId, Nodes, SubtreeRevision) ` + placeholderSQL
+	insertSubtreeMultiSQL = `REPLACE INTO Subtree(TreeId, SubtreeId, Nodes, SubtreeRevision) ` + placeholderSQL
```
n-canter, line 37: Should we use `INSERT INTO ... ON DUPLICATE KEY UPDATE Nodes = VALUES(Nodes)` instead?
On my benchmarks I didn't see any difference, but I think it's [safer](https://code.openark.org/blog/mysql/replace-into-think-twice/comment-page-5#comments) to use it. Wdyt?

mhutchinson: Thanks. I've switched over to this version. I have a minor concern that the `VALUES()` syntax is unfavourable in both MySQL[1] and to a lesser degree MariaDB[2] in the newest versions, but for the time being they both support it so let's roll with it.

Tested this with the following docker images:
  - mysql:5
  - mysql:8.0
  - mariadb:10.5

[1] https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html - "The use of VALUES() to refer to the new row and columns is deprecated beginning with MySQL 8.0.20, and is subject to removal in a future version of MySQL"
[2] https://mariadb.com/kb/en/values-value/ - In MariaDB 10.3.3 this function was renamed to VALUE(), because it's incompatible with the standard Table Value Constructors syntax ... The VALUES() function can still be used even from MariaDB 10.3.3 but only in INSERT ... ON DUPLICATE KEY UPDATE statements; it's a syntax error otherwise.

storage/mysql/tree_storage.go
```diff
@@ -147,6 +157,9 @@ func (m *mySQLTreeStorage) beginTreeTx(ctx context.Context, tree *trillian.Tree,
 		klog.Warningf("Could not start tree TX: %s", err)
 		return treeTX{}, err
 	}
+	// If the tree description starts with magic constant then we will only write a single
+	// revision for each subtree tile, which simplifies the reading query.
+	subtreeRevs := !strings.HasPrefix(tree.Description, "[NO_SUBTREE_REVISIONS]")
```
AlCutter, line 162: Should this magic constant be somewhere it can be referenced from here and from the `createTree` cmd? (Top level `/internal/...` pkg?)

storage/mysql/tree_storage.go
```diff
@@ -190,18 +205,28 @@ func (t *treeTX) getSubtrees(ctx context.Context, treeRevision int64, ids [][]by
 		}
 	}()
 
-	args := make([]interface{}, 0, len(ids)+3)
+	var args []interface{}
+	if t.subtreeRevs {
+		args = make([]interface{}, 0, len(ids)+3)
+		// populate args with ids.
+		for _, id := range ids {
+			klog.V(4).Infof("  id: %x", id)
```
AlCutter, line 213: nit: It might be worth changing this (and below) to be `if klog.V(4) { klog.Infof("...") }` while you're here to avoid creating and discarding the formatted log string for every subtree read?

storage/mysql/tree_storage.go
```diff
@@ -283,6 +308,12 @@ func (t *treeTX) storeSubtrees(ctx context.Context, subtrees []*storagepb.Subtre
 	// a really large number of subtrees to store.
 	args := make([]interface{}, 0, len(subtrees))
 
+	subtreeRev := t.writeRevision
```
AlCutter, line 311: Would flipping the logic here be easier for the reader?

```golang
var subtreeRev int64
if t.subtreeRevs {
  // We're using subtrees so ensure we write at the correct revision:
  subtreeRev = t.writeRev
}
```

Separately, is there anything special about the `0` value? The current comment sort of implies it's arbitrary, but e.g. does it break if the `treeTX` default of `-1` were used?

mhutchinson:
I've convinced myself that the current approach is too leaky an abstraction and I'm going to put more work into this. Specifically, changing the description in `createTree` means that other storage implementations other than mysql will be affected by this change. This is ugly and confusing (especially if these other storage implementations still use subtree revisions), and sets the wrong precedent for isolation of these different storage backends.

I have some ideas about doing this in a cleaner way, and I'll take another run at this tomorrow.

storage/mysql/tree_storage.go
phbnf: +1 that caught my attention as well

storage/mysql/tree_storage.go
mhutchinson: This constant no longer exists, which is good news all round.

storage/mysql/tree_storage.go
mhutchinson: Will look at this again tomorrow. Today has been spent doing heavy lifting. I'll come back around for a nuance pass after some time away from this code :-)

phbnf (commented):
For anyone else reviewing this, I've found it useful to break this review in two halves since some changes were done and undone by some commits: the first commit, and all the other ones together.

storage/mysql/tree_storage.go
AlCutter: Ah, nice - I didn't know `klog` did that internally - thanks!

storage/mysql/admin_storage.go
```diff
@@ -15,17 +15,21 @@
...
 	"github.com/google/trillian"
 	"github.com/google/trillian/storage"
+	"github.com/google/trillian/storage/mysql/mysqlpb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"k8s.io/klog/v2"
 )
```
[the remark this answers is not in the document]
mhutchinson: Done.

storage/mysql/admin_storage.go
```diff
@@ -223,6 +227,39 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
...
 
+	// When creating a new tree we automatically add StorageSettings to allow us to
+	// determine that this tree can support newer storage features. When reading
+	// trees that do not have this StorageSettings populated, it must be assumed that
+	// the tree was created with the oldest settings.
+	// The gist of this code is super simple: create a new StorageSettings with the most
+	// modern defaults if the created tree does not have one, and then create a struct that
+	// represents this to store in the DB. Unfortunately because this involves anypb, struct
+	// copies, marshalling, and proper error handling this turns into a scary amount of code.
+	if tree.StorageSettings != nil {
```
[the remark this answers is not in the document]
mhutchinson: Thanks :-)

storage/mysql/admin_storage.go
```diff
@@ -223,6 +227,39 @@ func (t *adminTX) CreateTree(ctx context.Context, tree *trillian.Tree) (*trillia
...
+	}
+	buff := &bytes.Buffer{}
+	enc := gob.NewEncoder(buff)
+	if err := enc.Encode(ss); err != nil {
+		return nil, fmt.Errorf("failed to encode storageSettings: %v", err)
+	}
+
 	insertTreeStmt, err := t.tx.PrepareContext(
 		ctx,
 		`INSERT INTO Trees(
```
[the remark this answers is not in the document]
mhutchinson: Done.

storage/mysql/sql.go
```diff
@@ -124,5 +128,30 @@ func readTree(r row) (*trillian.Tree, error) {
...
+	// storing settings.
+	buff := bytes.NewBuffer(publicKey)
+	dec := gob.NewDecoder(buff)
+	ss := &storageSettings{}
+	var o *mysqlpb.StorageOptions
+	if err := dec.Decode(ss); err != nil {
+		// If there are no storageSettings then this tree was created before settings
+		// were supported, and thus we have to populate the settings with the oldest
+		// settings for features.
+		o = &mysqlpb.StorageOptions{
```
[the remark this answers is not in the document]
mhutchinson: What change do you think that this introduces for old trees? The goal, which I believe I have scored, is that this non-nil value establishes the same behaviour for old trees as before this PR.

Stepping back, this is to ensure that there is always a StorageSettings available that explicitly describes how this tree should be treated at the storage layer. I _could_ set this to be nil, but then it will require code elsewhere to infer how it should be treated from a nil value. Having a central place (i.e. here) that establishes the modality makes it easier to reason about and to test.

storage/mysql/tree_storage.go
mhutchinson: I've gone with Al's suggestion here. `-1` would also work, as would anything else. We just need a non-`NULL` value to write, and then it will never be read again.

mhutchinson (commented):
Thanks for the comments. Addressed, and still eagerly courting more eyes and opinions.

mhutchinson:
I've squashed the history for this PR into 2 commits. The first of these shows what the real change is to support the new revisionless approach, but does so in a breaking way. The second change makes it so that all new trees get the new revisionless behaviour, but old trees continue to work exactly as they used to.

storage/mysql/sql.go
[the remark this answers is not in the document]
mhutchinson: Addressed in the latest commit.

roger2hk (approved):

phbnf (approved):

storage/mysql/sql.go
```diff
@@ -124,5 +128,30 @@ func readTree(r row) (*trillian.Tree, error) {
...
 
+	// We're going to try to interpret PublicKey as storageSettings, but it could be a
+	// public key from a really old tree, or an empty column from a tree created in the
+	// period between Trillian key material being removed and this column being used for
+	// storing settings.
+	buff := bytes.NewBuffer(publicKey)
+	dec := gob.NewDecoder(buff)
+	ss := &storageSettings{}
+	var o *mysqlpb.StorageOptions
+	if err := dec.Decode(ss); err != nil {
```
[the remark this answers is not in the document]
mhutchinson: As the comments above and below here say, there are the options that this column contains one of {public key, nothing, storageSettings}. An errror will occur in either of the first two cases, which maps to having a revisioned tree (which is what we want). The third case should parse properly, and then switch on the storageSettings.Revisioned field (also what we want).

There is another option which is that database corruption mangled a valid storageSettings. At this point it is somewhat normal to expect things to die. But actually I believe this will work well. I've reasoned that going from revisionless to revisioned will work (a mangled storageSettings will either fail to parse, or in an extreme case will be mangled to another valid storageSettings where the revisioned bit is true; either still leads to a working outcome).

The final option is that we make bonkers backwards incompatible changes to storageSettings, but that's just something we have to control for by proper engineering considerations in future changes. I don't see that much more that can be added here to safeguard for that outcome.

mhutchinson:
@AlCutter @jsha @n-canter this is looking final now. If you have any comments positive/neutral/negative before this gets merged, now is the time!

CHANGELOG.md
```diff
@@ -5,6 +5,42 @@
 * Bump CI version of MySQL from 5.7 to 8.0
 * Bump golangci-lint from 1.51.1 to 1.55.1 (developers should update to this version)
 
+### MySQL: Changes to Subtree Revisions
+
+Support for skipping subtree revisions to increase read performance and disk usage: added in #3201
```
AlCutter, line 10: nit: and _reduce_ disk usage? :)

mhutchinson: Done!

storage/mysql/sql.go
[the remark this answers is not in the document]
roger2hk: Gotcha, thanks!

AlCutter (approved):

PR #3236, opened 2023-12-06
Fix deadlock in log_client

n-canter:
Deadlock in WaitForRootUpdate() method when UpdateRoot() is called simultaneously.


roger2hk (commented):
I could be wrong but it looks like the ctx timeout in `WaitForRootUpdate` func will break the deadlock.

n-canter:
> I could be wrong but it looks like the ctx timeout in `WaitForRootUpdate` func will break the deadlock.

You're right, closing this pr

PR #3272, opened 2024-01-09
Switch from using unmaintained Google Cloud mysql db image to dockerhub official image

patflynn:
Alternative to https://github.com/google/trillian/pull/3270 
https://github.com/GoogleCloudPlatform/mysql-docker claims it is unmaintained and
hasn't seen an update in 5 months.  Without a specific reason for using this fork
we should try using the official image on dockerhub which is less likely to see breakages
or fall out of date.

Instead of pinning to a hash I'm following the 8.0 tag to automatically receive patch versions.
The reasoning being that the container is used during testing and is not
an included dependency in our final artifact.  Happy to consider pinning
if anybody thinks there is a call for it.

### Checklist


mhutchinson (approved):
Tests seem to pass, so LGTM. I suspect we can/should remove the steps from the cloudbuild files that create the `gcr.io/trillian-opensource-ci/mysql8` image from the unmaintained project.

PR #3318, opened 2024-01-27
Add SQL election module

### LINKED ISSUE #3219: Leader election using SQL database?
I see that Trillian currently supports leader election using etcd or Chubby.  Given that there is _also_ a requirement for an ACID database store (Spanner/MySQL/CockroachDB), would it be reasonable to add an election implementation based on rows in a SQL database with a schema like the following?

```sql
CREATE TABLE IF NOT EXISTS leader_election (
  leader_name     varchar(255) NOT NULL PRIMARY KEY,
  node_name       varchar(255) NOT NULL,
  election_expiry datetime     NOT NULL
);
```

The leader could do conditional updates of `election_expiry` based on `node_name` being the same, and non-leader nodes could do a periodic (e.g. 1s) query for expired elections they could claim.

This would have the advantage of reducing t he set of required Trillian dependencies, at a cost of somewhat more database traffic.

evankanderson:
Adds a generic SQL leader election module, as discussed in #3219; it sounds like Sigstore is looking at possibly using this for simplicity.

Fixes #3219

Note that I haven't yet wired this up into Trillian nor added database table declarations -- I'm happy to do that in this PR or a follow-up.  I'm not including this in the changelog until it's wired in, but I wanted to get feedback on how to do that first.

### Checklist

### DIFF
--- a/util/election2/sql/election.go
+++ b/util/election2/sql/election.go
@@ -1,0 +1,269 @@
+// Copyright 2023 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package etcd provides an implementation of leader election based on a SQL database.
+package sql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/google/trillian/util/election2"
+	"k8s.io/klog/v2"
+)
+
+type leaderData struct {
+	currentLeader string
+	timestamp     time.Time
+}
+
+// Election is an implementation of election2.Election based on a SQL database.
+type Election struct {
+	db         *sql.DB
+	instanceID string
+	resourceID string
+
+	currentLeader leaderData
+	leaderLock    sync.Cond
+
+	// If a channel is supplied with the cancel, it will be signalled when the election routine has exited.
+	cancel           chan *chan error
+	electionInterval time.Duration
+}
+
+var _ election2.Election = (*Election)(nil)
+
+// Await implements election2.Election
+func (e *Election) Await(ctx context.Context) error {
+	e.leaderLock.L.Lock()
+	defer e.leaderLock.L.Unlock()
+	if e.cancel == nil {
+		e.cancel = make(chan *chan error)
+		go e.becomeLeaderLoop(context.Background(), e.cancel)
+	}
+	if e.currentLeader.currentLeader == e.instanceID {
+		return nil
+	}
+	for e.currentLeader.currentLeader != e.instanceID {
+		e.leaderLock.Wait()
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			klog.Infof("Waiting for leadership, %s is the leader at %s", e.currentLeader.currentLeader, e.currentLeader.timestamp)
+		}
+	}
+	klog.Infof("%s became leader for %s at %s", e.instanceID, e.resourceID, e.currentLeader.timestamp)
+	return nil
+}
+
+// Close implements election2.Election
+func (e *Election) Close(ctx context.Context) error {
+	if err := e.Resign(ctx); err != nil {
+		klog.Errorf("Failed to resign leadership: %v", err)
+		return err
+	}
+	return nil
+}
+
+// Resign implements election2.Election
+func (e *Election) Resign(ctx context.Context) error {
+	e.leaderLock.L.Lock()
+	closer := e.cancel
+	e.cancel = nil
+	e.leaderLock.L.Unlock()
+	if closer == nil {
+		return nil
+	}
+	// Stop trying to elect ourselves
+	done := make(chan error)
+	closer <- &done
+	return <-done
+}
+
+// WithMastership implements election2.Election
+func (e *Election) WithMastership(ctx context.Context) (context.Context, error) {
+	cctx, cancel := context.WithCancel(ctx)
+	e.leaderLock.L.Lock()
+	defer e.leaderLock.L.Unlock()
+	if e.currentLeader.currentLeader != e.instanceID {
+		// Not the leader, cancel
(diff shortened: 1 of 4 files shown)


util/election2/sql/election.go
```diff
@@ -0,0 +1,269 @@
...
+			return
+		default:
+			leader, err := e.tryBecomeLeader(ctx)
+			if err != nil {
+				klog.Errorf("Failed attempt to become leader for %s, retrying: %v", e.resourceID, err)
+			} else {
+				e.leaderLock.L.Lock()
+				if leader != e.currentLeader {
+					// Note: this code does not actually care _which_ instance was
+					// elected, it sends notifications on each leadership cahnge.
```
JAORMX, line 149: s/cahnge/change/

util/election2/sql/election.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (e *Election) tryBecomeLeader(ctx context.Context) (leaderData, error) {
+	leader := leaderData{}
+	tx, err := e.db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable})
+	if err != nil {
+		return leader, fmt.Errorf("BeginTX: %w", err)
+	}
+	defer tx.Rollback()
+	row := tx.QueryRow(
+		"SELECT leader, last_update FROM leader_election WHERE resource_id = $1",
```
JAORMX, line 168: Wouldn't the `$<num>` format be problematic in mysql?

util/election2/sql/election.go
```diff
@@ -0,0 +1,269 @@
...
+		return fmt.Errorf("Update: %w", err)
+	}
+	if n, err := res.RowsAffected(); n != 1 || err != nil {
+		return fmt.Errorf("failed to resign leadership: %d, %w", n, err)
+	}
+	return nil
+}
+
+func (e *Election) initializeLock(ctx context.Context) error {
+	insert, err := e.db.Prepare("INSERT INTO leader_election (resource_id, leader, last_update) VALUES ($1, $2, $3)")
```
JAORMX, line 216: Wouldn't we need to handle the case where more than one replica executes an initialization at the same time? Or is the idea is to crash loop, restart and continue?

evankanderson:
I fixed the remaining lint issue and added what I think is the right integration hooks.  I'm still working on running the integration tests -- in particular, adding a test that uses the `--election_backend=mysql` option.

cmd/trillian_log_signer/main.go
```diff
@@ -47,13 +47,15 @@ import (
 	"github.com/google/trillian/util/election"
 	"github.com/google/trillian/util/election2"
 	etcdelect "github.com/google/trillian/util/election2/etcd"
+	"github.com/google/trillian/util/election2/sql"
```
mhutchinson, line 50: This comment relates to the package of the new code. I recommend `mysql` over `sql` to be clear at which implementation it is targeting, and for consistency with the flag names used in this file to pick it.

JAORMX:
Ah! Linter found something

cmd/trillian_log_signer/main.go
```diff
@@ -143,13 +146,22 @@ func main() {
 	instanceID := fmt.Sprintf("%s.%d", hostname, os.Getpid())
 	var electionFactory election2.Factory
 	switch {
-	case *forceMaster:
+	case *forceMaster || *electionBackend == "noop":
```
phbnf, line 148: It's not fantastic that we can now control this behavior with two different flags, but I also cannot thing of a better way forward for now. On the long run, we could / should probably deprecate forceMaster. No action required!

util/election2/mysql/election.go
```diff
@@ -0,0 +1,280 @@
...
+		default:
+			klog.Infof("Waiting for leadership, %s is the leader at %s", e.currentLeader.currentLeader, e.currentLeader.timestamp)
+		}
+	}
+	klog.Infof("%s became leader for %s at %s", e.instanceID, e.resourceID, e.currentLeader.timestamp)
+	return nil
+}
+
+// Close implements election2.Election
+func (e *Election) Close(ctx context.Context) error {
```
phbnf, line 77: Given this is essentially a wrapper around Resign, this election object could be re-used for a new election, which is slightly at odds with the interface definition. In practice, I don't think anything relies on this behaviour, and there aren't any resources that can be released, so I think it's safe. Does that match your understanding?

util/election2/mysql/election.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil
+	}
+	// Stop trying to elect ourselves
+	done := make(chan error)
+	closer <- &done
+	return <-done
+}
+
+// WithMastership implements election2.Election
+func (e *Election) WithMastership(ctx context.Context) (context.Context, error) {
```
roger2hk, line 101: Having the inclusivity and diversity in mind, may I suggest replacing `master` with another inclusive term?

https://developers.google.com/style/word-list#master

evankanderson: I can rewrite the `election2.Election` interface if you'd like.  😁 

Otherwise, I think those are the only two instances of the word "master" rather than "leader".

util/election2/mysql/election.go
evankanderson: Yes.

Do you want me to make the object unusable, to catch errors that might occur by calling other methods after `Close`?

util/election2/mysql/election.go
```diff
@@ -0,0 +1,280 @@
...
+// Election is an implementation of election2.Election based on a SQL database.
+type Election struct {
+	db         *sql.DB
+	instanceID string
+	resourceID string
+
+	currentLeader leaderData
+	leaderLock    sync.Cond
+
+	// If a channel is supplied with the cancel, it will be signalled when the election routine has exited.
```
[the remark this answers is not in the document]
evankanderson: I wrote this before realizing that I could re-use `Resign` for `Close`, and I think one of those did not wait for the error in the first implementation.  I can remove the comment, and change the type from `*chan error` to `chan error` probably...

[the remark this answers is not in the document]
phbnf: Gotcha - makes sense to me to remove the comment then yes!

util/election2/mysql/election.go
phbnf: If you can think of a way of an elegant way of doing this, sure thing. I don't think it's super super important, it's just some extra care to make sure it's aligned with the interface definition :)

vipulagarwal:
@evankanderson thanks for this PR as it simplifies Sigstore dependencies. Do we have any plans to merge this PR any soon? Thanks

evankanderson:
Sorry, I got distracted by this by life.  If this is still of interest, I can try to revive it after my camping break this week -- maybe by around 14 July or so.

PR #3584, opened 2024-08-06
Don't update to MySQL 9.0

mhutchinson:
Be more explicit about which dependency to apply the filter to. Somehow
in #3560 I thought that the dependency to ignore was implicit because
there is only one base image. However, #3581 shows that it didn't work.
Trying again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,8 +35,8 @@ updates:
   schedule:
     interval: weekly
   ignore:
-    versions:
-      - ">= 9.0"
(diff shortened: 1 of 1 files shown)


patflynn (approved):

phbnf (approved):
I don't think that the brackets are required, but it doesn't hurt either! https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore is not supper clear.

Issue #3592, opened 2024-08-13
Add TLS Support for MySQL Connections in Trillian log Server/Signer


fghanmi:
Currently, Trillian does not support TLS connections to MySQL. 
This issue aims to track the necessary changes required to enable TLS support when connecting to MySQL database.

_We'll consider potential backward compatibility for users not utilizing TLS._

PR #3593, opened 2024-08-13
Add MySQL TLS support

fghanmi:
This PR adds TLS support for MySQL connections in the Trillian server/signer. The key changes include:
- Added new flags:
  - `mysql_tls_ca`: Path to the CA certificate file for the MySQL TLS connection.
  - `mysql_server_name`: Name of the MySQL server to be used as the Server Name in the TLS configuration.

- TLS Configuration Registration:
  - Added a new function `registerTLSConfig()` to handle the registration of the custom TLS configuration.

_If no TLS configuration is provided, the connection defaults to non-TLS, ensuring backward compatibility._

Issue: https://github.com/google/trillian/issues/3592
### Checklist

### DIFF
--- a/storage/mysql/provider.go
+++ b/storage/mysql/provider.go
@@ -15,8 +15,12 @@
 package mysql
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"database/sql"
+	"errors"
 	"flag"
+	"os"
 	"sync"
 
 	"github.com/google/trillian/monitoring"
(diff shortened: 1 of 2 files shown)


storage/mysql/provider.go
```diff
@@ -81,7 +87,14 @@ func getMySQLDatabaseLocked() (*sql.DB, error) {
 	if mysqlDB != nil || mysqlErr != nil {
 		return mysqlDB, mysqlErr
 	}
-	db, err := OpenDB(*mySQLURI)
+	if err := registerTLSConfig(); err != nil {
+		return nil, err
+	}
```
AlCutter, line 92: Could this be called inside the `if *mySQLTLSCA {` check on line 93?

That would make it clear that this is related to the MySQL TLS functionality.

storage/mysql/provider.go
```diff
@@ -107,3 +120,32 @@ func (s *mysqlProvider) AdminStorage() storage.AdminStorage {
 func (s *mysqlProvider) Close() error {
 	return s.db.Close()
 }
+
+func registerTLSConfig() error {
```
AlCutter, line 124: Maybe rename this to `registerMySQLTLSConfig` to be clear about what it's doing & why.

storage/mysql/provider.go
```diff
@@ -107,3 +120,32 @@ func (s *mysqlProvider) AdminStorage() storage.AdminStorage {
...
+	}
+	if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+		return errors.New("failed to append PEM")
+	}
+	serverName := ""
+	if *mySQLServerName != "" {
+		serverName = *mySQLServerName
+	} else {
+		serverName = os.Getenv("MYSQL_HOSTNAME")
+	}
```
AlCutter, line 142: This isn't mentioned in the PR description/CHANGELOG or flags. 
Is this ENV var used anywhere else or is it a new one? If it's new then let's remove it for now and rely on just flags for configuration to avoid confusion.

storage/mysql/provider.go
```diff
@@ -107,3 +120,32 @@ func (s *mysqlProvider) AdminStorage() storage.AdminStorage {
...
+	err = mysql.RegisterTLSConfig("custom", &tls.Config{
+		RootCAs:    rootCertPool,
+		ServerName: serverName,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
AlCutter, line 150: Since you're not wrapping the error, this can just be shortened to `return mysql.RegisterTLS...`

AlCutter:
Hi @fghanmi - thanks for the PR, I've dropped a few comments on for you.
Cheers,
Al.

storage/mysql/provider.go
fghanmi: I thought it is used explicitly by trillian (as it is used by mySQL), but you're right, this is a new var (removed it)

fghanmi:
> Hi @fghanmi - thanks for the PR, I've dropped a few comments on for you. Cheers, Al.

Hello @AlCutter,
Thanks for the comments! all done ;) 
Thank you!

AlCutter (approved):

AlCutter:
Thanks again @fghanmi!

PR #3664, opened 2024-11-07
Control included quota and storage providers via build tags

robstradling:
https://github.com/google/trillian/pull/3644#discussion_r1822420304 noted that the log-signer and log-server binaries are huge, due in part to the presence of multiple quota and storage providers. This PR enables providers to be included selectively via Go build tags.

By default and for backwards compatibility, all providers are included:

> go build && ls -sh trillian_log_server
62M trillian_log_server*

Specify one build tag to include just one provider:

> go build -tags=cloudspanner && ls -sh trillian_log_server
54M trillian_log_server*
> go build -tags=crdb && ls -sh trillian_log_server
37M trillian_log_server*
> go build -tags=mysql && ls -sh trillian_log_server
34M trillian_log_server*
> go build -tags=postgresql && ls -sh trillian_log_server
40M trillian_log_server*

Specify multiple, comma-separated build tags to include multiple providers. e.g.,:

> go build -tags=mysql,cloudspanner && ls -sh trillian_log_server
55M trillian_log_server*
> go build -tags=cloudspanner,crdb,mysql,postgresql && ls -sh trillian_log_server
62M trillian_log_server*

### Checklist

- [x] I have updated the [CHANGELOG](CHANGELOG.md).
- [x] I have updated [documentation](docs/) accordingly (including the [feature implementation matrix](docs/Feature_Implementation_Matrix.md)).

PR #3679, opened 2024-11-15
Sensible default quota and storage systems when MySQL support unavailable

robstradling:
It no longer makes sense to hardcode "mysql" as the default value for `-quota_system` and `-storage_system`, because after https://github.com/google/trillian/pull/3664 it's now possible to omit MySQL support when building the log_server and log_signer.

For backwards compatibility "mysql" should remain the default when MySQL support is available, but otherwise an available storage system and an available (and if possible matching) quota system should be the default.

### Checklist

### DIFF
--- a/cmd/internal/provider/default_systems.go
+++ b/cmd/internal/provider/default_systems.go
@@ -1,0 +1,30 @@
+package provider
+
+import (
+	"slices"
+
+	"github.com/google/trillian/quota"
+	"github.com/google/trillian/storage"
+)
+
+var (
+	DefaultQuotaSystem   string
+	DefaultStorageSystem string
+)
+
+func init() {
+	defaultProvider := "mysql"
+	providers := storage.Providers()
+	if len(providers) > 0 && !slices.Contains(providers, defaultProvider) {
+		slices.Sort(providers)
+		defaultProvider = providers[0]
+	}
+	DefaultStorageSystem = defaultProvider
+
+	providers = quota.Providers()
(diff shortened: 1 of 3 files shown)


cmd/internal/provider/default_systems.go
```diff
@@ -0,0 +1,30 @@
...
+		defaultProvider = providers[0]
+	}
+	DefaultStorageSystem = defaultProvider
+
+	providers = quota.Providers()
+	if len(providers) > 0 && !slices.Contains(providers, defaultProvider) {
+		slices.Sort(providers)
+		defaultProvider = providers[0]
+	}
+	DefaultQuotaSystem = defaultProvider
```
roger2hk, line 29: This may change the backward compatibility of using `mysql` as the default provider in the following assumption. I'm not sure if we want to update the behavior.

**Input**
storage.Providers(): `postgresql`
quota.Providers(): `mysql`, `postgresql`

**Want**
DefaultStorageSystem: `postgresql`
DefaultQuotaSystem: `mysql`

**Got**
DefaultStorageSystem: `postgresql`
DefaultQuotaSystem: `postgresql`

robstradling: This scenario shouldn't occur.  The build tags, if used, keep the quota and storage provider lists synchronised.  So if `mysql` is returned by `quota.Providers()` then it will also be returned by `storage.Providers()`, and vice versa.

roger2hk: Great, thanks!

roger2hk (approved):

PR #3713, opened 2025-01-09
chore: relax go directive to permit 1.22.x

dnwe:
Setting this to 1.22.7 requires all consumers to be building with the 1.22.7 or newer release of Go 1.22 and to update their own go.mod accordingly — this seems unnecessarily restrictive for a library module, particularly as the code itself doesn't currently use any modern language constructs and builds fine even with older Go versions.

Instead set the go directive to 1.22.0 and use the toolchain directive to recommend the latest 1.22.x when building locally.

### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/google/trillian
 
-go 1.22.7
+go 1.22.0
+
+toolchain go1.22.10
 
 require (
 	bitbucket.org/creachadair/shell v0.0.8


AlCutter:
Thanks @dnwe, good call.

AlCutter (approved):

dnwe:
@AlCutter thanks Al — would you be happy to cut a v1.7.1 so I can adopt it in some other modules?

AlCutter:
> @AlCutter thanks Al — would you be happy to cut a v1.7.1 so I can adopt it in some other modules?

Sure!

AlCutter:
--> https://github.com/google/trillian/releases/tag/v1.7.1 👍

PR #3770, opened 2025-04-03
Improve PostgreSQL functions

robstradling:
In our CT logs that use the PostgreSQL quota manager, we've found that in practice `ANALYZE (SKIP_LOCKED TRUE)` in the count_estimate() function gets skipped a lot of the time.  Removing `(SKIP_LOCKED TRUE)` hurts performance, because it first has to wait for a lock and then it holds that lock until the `ANALYZE` operation completes.  For tables with a relatively small number of rows, it's actually a lot quicker to do a simple `SELECT count(1)` query (whereas for large tables, `SELECT count(1)` is very slow).  This PR implements a hybrid approach that avoids running `ANALYZE` when there are <1000 rows and avoids running `SELECT count(1)` for >1000 rows.  We've deployed this to our Elephant and Tiger CT logs, and it's definitely improved performance.

This PR also implements a couple of minor editorial changes to the queue_leaves() and add_sequenced_leaves() functions.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Storage
 
 * For PostgreSQL, explicitly create index on SequencedLeafData(TreeId, LeafIdentityHash) by @robstradling in https://github.com/google/trillian/pull/3695
+* Improve PostgreSQL functions by @robstradling in https://github.com/google/trillian/pull/3770
 
 ### Misc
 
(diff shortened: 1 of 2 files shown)


roger2hk:
@AlCutter ping :)

mhutchinson (approved):

PR #3862, opened 2026-03-11
Mitigate PostgreSQL connection pool contamination

robstradling:
When a gRPC context is cancelled (client disconnect or timeout) while a database query is in-flight, pgx closes the underlying connection or leaves it in a dirty transaction state ('T' or 'E'), but the connection pool may then recycle this connection and assign it to a subsequent query, causing "failed to deallocate cached statement(s): conn closed" errors. This pgx behaviour is described by https://github.com/jackc/pgx/issues/2100.

Attempt to work around this problem by registering PrepareConn and AfterRelease callbacks on the pgxpool configuration that check connection health before acquisition and after release. Connections that are closed or not in idle transaction state ('I') are destroyed rather than recycled.

### Checklist

### DIFF
--- a/storage/postgresql/admin_storage.go
+++ b/storage/postgresql/admin_storage.go
@@ -102,6 +102,9 @@ func (t *adminTX) Commit() error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	t.closed = true
+	if t.tx.Conn().IsClosed() {
+		return fmt.Errorf("commit aborted: connection already closed")
+	}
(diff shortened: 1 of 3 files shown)


dhggsf (approved):

robstradling:
Abandoning this PR.  We had it running on our Dumbo log for the past couple of months, but it apparently caused the log_server pods to crashloop, and it didn't do anything noticeable to improve performance or reduce failed requests.

PR #3893, opened 2026-06-03
[CI] Fix govulncheck warning

mhutchinson:
This uses the minimal bump to go that avoids the CVEs blocking the vuln
checker.


AlCutter (approved):

.github/workflows/govulncheck.yml
```diff
@@ -16,8 +16,13 @@ jobs:
     runs-on: ubuntu-latest
     name: Run govulncheck
     steps:
-      - id: govulncheck
-        uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
+      - name: Checkout code
+        uses: actions/checkout@v4
```
AlCutter: Do these need pins?

.github/workflows/govulncheck.yml
```diff
@@ -16,8 +16,13 @@ jobs:
...
+      - name: Setup Go
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-          go-package: ./...
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
AlCutter: Worth pinning to latest release/commit?

.govulncheck-ignore.json
```diff
@@ -0,0 +1,14 @@
+{
```
AlCutter: This could be a plain text file e.g. `${id} ${module} <free form text>` and then just `readline -a` and index to get the fields (or sed/awk/whatever :) ), that'd be easier for a human to maintain than JSON with special rules about commas.

AlCutter (approved):

scripts/govulncheck-wrapper.sh
```diff
@@ -0,0 +1,138 @@
...
+
+if [[ -z "$UNIQUE_VULNS_TSV" ]]; then
+  log_info "No vulnerabilities found"
+  exit 0
+fi
+
+IGNORED_LIST=""
+while read -r id module reason || [[ -n "$id" ]]; do
+  # Skip comments and empty lines
+  if [[ -z "$id" || "$id" =~ ^# ]]; then
```
AlCutter: Probably worth quoting the `^#`

AlCutter (approved):

AlCutter (approved):

.github/workflows/govulncheck.yml
mhutchinson: Good call. I got blind trying to make this work.

.github/workflows/govulncheck.yml
mhutchinson: done

.govulncheck-ignore.json
mhutchinson: done now

scripts/govulncheck-wrapper.sh
mhutchinson: I'll do this tomorrow - need to escape during this break in the rain!

PR #3903, opened 2026-07-13
Silence unactionable govulncheck findings

phbnf:
This change silences unactionable `govulncheck` reports for `docker`.


phbnf:
Unfortunately, this tool will not work (without further modification) if a fix has been publishes, which is the case for prometheus.

PR #3917, opened 2026-08-21
offer batching for mysql quota provider

bobcallaway:
Currently, the MySQL quota provider does a lookup on the row count of unsequenced rows for every single insert. By using the caching provider that already exists, grabbing tokens in batches will cause the provider to only check the row count when that batch of tokens is depleted. This mimics the setup for the etcd provider.

The batch size is set to 0, which retains the current default behavior of fetching a row count on every QueueLeaf call.

### DIFF
--- a/quota/mysqlqm/quota_provider.go
+++ b/quota/mysqlqm/quota_provider.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 
 	"github.com/google/trillian/quota"
+	"github.com/google/trillian/quota/cacheqm"
 	"github.com/google/trillian/storage/mysql"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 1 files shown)


quota/mysqlqm/quota_provider.go
```diff
@@ -43,6 +50,14 @@ func newMySQLQuotaManager() (quota.Manager, error) {
 		DB:                 db,
 		MaxUnsequencedRows: *maxUnsequencedRows,
 	}
+	if *mysqlQuotaMinBatchSize > 0 {
+		cachedqm, err := cacheqm.NewCachedManager(qm, *mysqlQuotaMinBatchSize, *mysqlQuotaMaxCacheEntries)
```
roger2hk, line 58: The following validation is necessary to prevent a permanent write lockout on misconfigured servers.

```suggestion
	if *mysqlQuotaMinBatchSize > 0 {
		if *mysqlQuotaMinBatchSize >= *maxUnsequencedRows {
			return nil, fmt.Errorf("mysql_quota_min_batch_size (%d) must be less than max_unsequenced_rows (%d)", *mysqlQuotaMinBatchSize, *maxUnsequencedRows)
		}
		cachedqm, err := cacheqm.NewCachedManager(qm, *mysqlQuotaMinBatchSize, *mysqlQuotaMaxCacheEntries)
```

roger2hk (approved):

